PR #11, opened 2022-12-02
2 have indexerpy store file metadata in the opensearch cluster

sdhoyt:
# Change Summary

## Overview
- Added OpenSearch utility classes to abstract away details of the opensearch-py client
- Added metadata ingestion to OpenSearch cluster in indexer.py
- Added new environment variables for use in lambda function
- Added unit tests (32 tests)

## New Dependencies
- [opensearch-py](https://github.com/opensearch-project/opensearch-py)

## New Files
- document.py
   - represents a document containing data to be sent to OpenSearch
- index.py
   - represents an OpenSearch index
- action.py
   - represents an action to be performed on a document sent to OpenSearch
- payload.py
   - represents a bulk payload containing multiple documents to be sent to OpenSearch
- client.py
   - wrapper for opensearch-py client that works with Index, Document, and Payload objects

## Updated Files
- sds_in_a_box_stack.py
   - added environment variables to lambda function init input for instantiating the OpenSearch client
- indexer.py
   - instantiated an OpenSearch client
   - added index, document, payload creation for data received in the lambda function.
   - added bulk upload request for received data

## Deleted Files
- testing.py
   - replaced with test_indexer.py inside tests/unit/

## TODOs Added
- payload.py line 63
   - haven't been able to find how the OpenSearch request limits work / what the actual limit is. After testing it out, the ascii encoded string size seems to be the most accurate way to check if a payload is over the request limit and after running some tests I got consistent results around a certain size and am using that as a placeholder for now. But still need to figure out what the actual limit is and how it's determined.


sds_in_a_box/SDSCode/opensearch_utils/client.py
```diff
@@ -0,0 +1,200 @@
...
+    def send_payload(self, payload):
+        """
+        Sends a bulk payload of documents to the OpenSearch cluster.
+
+        Parameters
+        ----------
+        payload: Payload
+            payload containing bulk documents to be sent to the OpenSearch cluster.
+        """
+        self.client.bulk(payload.get_contents(), params={"request_timeout":1000000})
```
sdhoyt, line 133: Need to use a smaller timeout

tests/unit/opensearch_utils/test_client.py
```diff
@@ -0,0 +1,186 @@
...
+from opensearchpy import OpenSearch, RequestsHttpConnection, AWSV4SignerAuth
+import boto3
+from botocore.exceptions import ClientError
+
+class TestClient(unittest.TestCase):
+    """tests for client.py"""
+
+    def setUp(self):
+        #Opensearch client Params
+        host = 'search-sds-metadata-uum2vnbdbqbnh7qnbde6t74xim.us-west-2.es.amazonaws.com'
```
bryan-harter, line 16: I wonder if there is a way to make this not hard-coded? This is probably best saved for another ticket, but maybe we should save the output of of the CDK stack somewhere, and then have all the code reference that "somewhere".  Like an output file, environment variables, a small dynamodb database, etc

bryan-harter (commented):
Overall it looks really really good! The opensearch_utils library is great and will be a big help for other functions in the SDC going forward!

sds_in_a_box/SDSCode/opensearch_utils/payload.py
```diff
@@ -0,0 +1,80 @@
...
+    def get_contents(self):
+        """Returns the contents of the payload as a string."""
+        full_contents = "".join(self.payload_contents)
+        return full_contents
+
+    def __repr__(self):
+        return str(self.payload_contents)
+
+    def __add_to_payload(self, document):
+        # TODO: not sure what the actual request limit is or how it's
```
bmcclellan-cu, line 63: Can you clarify what limit is being hit, how, and what errors you're receiving?

bmcclellan-cu (approved):
The code looks good to me! I'm interested in the limit your are hitting gin the payload.py file, as we'll want to be sure we are aware of it for Libera and can reach out to our AWS contacts to confirm.

sds_in_a_box/SDSCode/opensearch_utils/payload.py
sdhoyt: When I use the opensearch-py bulk() method I get a "TransportError(429, 'Too Many Requests') error when the bulk upload is too large. The purpose of the chunking is to avoid that limit, but I've been having trouble finding official documentation on what exactly that limit is. Right now I'm using a rough string size limit that I found from iteratively increasing the chunk size until I got the error. It seems to allow me to consistently avoid getting that error, but I am still trying to figure out what that limit officially is. I originally thought it was a size issue, but after looking into it more, it seems like it's more of a rate issue, so I'm not sure why lowering the chunk size helps. Let me know if you know of anything on Libera that was used to avoid this issue.

tests/unit/opensearch_utils/test_client.py
sdhoyt: I agree, that would be good to do. I can make another ticket for that.

sds_in_a_box/SDSCode/opensearch_utils/payload.py
bmcclellan-cu: From Gavin based on our testing with OpenSearch:

It sounds very similar to what we did for Libera ITDC. My understanding is that 429 means there are too many simultaneous bulk upload requests for the server to handle – i.e. the ingest queue is FULL. Basically the upshot is that data is being sent faster than it can be indexed. I solved this by adding a random 1-3s delay between bulk request retries to give opensearch an opportunity to clear the queue without constantly pelting it with HTTP requests. The other error that occurred a lot during my development is a timeout error, which just means the server doesn’t have the resources to send back a response to your HTTP request at all HOWEVER, this doesn’t necessarily mean that the data didn’t get ingested, so if you retry after that, expect to either create duplicate records or get a bunch of errors that the records all exist.
 
*As we discovered in our data lab, the hands down most reliable way to fix all of this is to just throw some more ingest nodes into the cluster.* We are using m6g.large.search

sdhoyt: Did you upgrade to m6g.large.search to avoid needing the delay or is that delay still needed?

bryan-harter (approved):

PR #13, opened 2023-01-06
Adding in Cognito to the code repository

bryan-harter:
* Added in cognito sign up service with a custom message
* Added api lambda template, including a function to verify a cognito token 
* Added a script where users can test the API with their username and password


sds_in_a_box/SDSCode/api.py
```diff
@@ -0,0 +1,68 @@
...
+def _verify_cognito_token(token):
+    region = 'us-west-2'
+    userpool_id = os.environ["COGNITO_USERPOOL_ID"]
+    app_client_id = os.environ["COGNITO_APP_ID"]
+    keys_url = f'https://cognito-idp.{region}.amazonaws.com/{userpool_id}/.well-known/jwks.json'
+    response = requests.get(keys_url)
+    keys = (response.json())['keys']
+    
+    headers=jwt.get_unverified_headers(token)
+    kid=headers['kid']
```
tech3371, line 17: Is there really a 'kid' key in the header? :) funny

tech3371: looked it up and apparently stands for 'key-id'.

sds_in_a_box/SDSCode/api.py
```diff
@@ -0,0 +1,68 @@
...
+import requests
+import time
+from jose import jwk, jwt
+from jose.utils import base64url_decode
+
+def _verify_cognito_token(token):
+    region = 'us-west-2'
+    userpool_id = os.environ["COGNITO_USERPOOL_ID"]
+    app_client_id = os.environ["COGNITO_APP_ID"]
+    keys_url = f'https://cognito-idp.{region}.amazonaws.com/{userpool_id}/.well-known/jwks.json'
```
tech3371, line 12: What are we storing there? a pub/private ssh key or something like that?

sds_in_a_box/SDSCode/cognito_signup_message.py
```diff
@@ -0,0 +1,16 @@
+import os
+import boto3 
+
+def lambda_handler(event, context):
+    client = boto3.client('cognito-idp')
+    domain_desc = client.describe_user_pool_domain(Domain=os.environ["COGNITO_DOMAIN_PREFIX"])
+    clients = client.list_user_pool_clients(UserPoolId = domain_desc["DomainDescription"]['UserPoolId'])
+    command_line_client = clients['UserPoolClients'][0]['ClientId']
+    
+    if 'triggerSource' in event:
```
tech3371, line 10: What event is triggering this lambda?

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -72,27 +83,122 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+                                                                                 user_srp=True,
+                                                                                 custom=True),
+                                                     prevent_user_existence_errors=True)
+        
+        # Add a random unique domain name where users can sign up / reset passwords
+        # Users will be able to reset their passwords at https://sds-login-{SDS_ID}.auth.us-west-2.amazoncognito.com/login?client_id={}&redirect_uri=https://example.com&response_type=code
+        userpooldomain = userpool.add_domain(id="TeamLoginCognitoDomain",
+                                             cognito_domain=cognito.CognitoDomainOptions(domain_prefix=f"sds-login-{SDS_ID}"))
+
+        # Add a lambda function that will trigger whenever an email is sent to the user (see the lambda section above)
```
tech3371, line 184: These comments are helpful

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -13,43 +15,52 @@
...
         # This is the S3 bucket where the data will be stored
         data_bucket = s3.Bucket(self, "DATA-BUCKET",
-                                #bucket_name="DataBucket",
+                                bucket_name=f"sds-data-{SDS_ID}",
                                 versioned=True,
                                 removal_policy=RemovalPolicy.DESTROY,
                                 auto_delete_objects=True
                                 )
 
+########### DATABASE
```
tech3371, line 44: nice way to group codes

tech3371 (commented):
Did you mean to leave `.idea` folder? Just checking.

And it looks good to me.

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -72,27 +83,122 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
         )
 
-        # This is the role that the lambdas will assume
-        # We'll narrow things down later
+########### LAMBDA FUNCTIONS
+
+        # This is where we install dependencies for the lambda functions
+        # We take advantage of something called "lambda layers"
+        os.system("pip install requests -t ./external/python")
+        os.system("pip install python-jose -t ./external/python")
```
sdhoyt, line 91: why are these dependencies being installed here and not part of the library requirements?

sds_in_a_box/SDSCode/api.py
bryan-harter: Yeah cognito automatically sets up these "JSON Web Key Sets" that have the public keys that you use to verify someone's JSON Web Token.  It comes automatically with a new cognito user group.  For example, here is the public key for the latest version of the code I have set up in the development environment: https://cognito-idp.us-west-2.amazonaws.com/us-west-2_rHAHpNHQi/.well-known/jwks.json

.devcontainer/Dockerfile
```diff
@@ -13,5 +13,7 @@ RUN sudo apt-get install -y python3 python3-pip
 RUN pip install aws-cdk-lib==2.48.0 
 RUN pip install "constructs>=10.0.0,<11.0.0"
 RUN pip install boto3
+RUN pip install requests
+RUN pip install python-jose
```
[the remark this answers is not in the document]
bryan-harter: I think that would be nice, but I'm not sure how to set up this Codespace environment to use the requirements.txt file.  The best I can tell, Codespaces first looks inside the ".devcontainer/devcontainer.json" file, and then that file pulls in the "Dockerfile".  I'm not sure if the Dockerfile is able to pull in more stuff from the repository while it is setting up the Codespace.  

devcontainer.json might have a command where it is able to run stuff after the Dockerfile loads.  But that said, these extra libraries should definitely be added to the current requirements.txt file

app.py
```diff
@@ -1,17 +1,20 @@
...
 import boto3 
+import string
+import random
 import aws_cdk as cdk
 import aws_cdk.assertions as assertions
 from sds_in_a_box.sds_in_a_box_stack import SdsInABoxStack
 
+### Configuration variables
+# A random unqiue ID for this particular instance of the SDS
+SDS_ID = "".join( [random.choice(string.ascii_lowercase) for i in range(8)] )
```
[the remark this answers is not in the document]
bryan-harter: Yes I would say definitely so!  I just got somewhat lazy this ticket and kind of skipped out on some of quality-of-life tools.  

One thing I was wondering about was maybe the configuration might live in another type of configuration file (or even the same config.json file as the file types), rather than asking the user for prompts? Or maybe it asks the user for prompts if it can't find an initial config file with an ID, intial email, instance types, etc.   

That said, I think maybe adding these more "user-friendly" components might be better suited to seperate ticket, especially since this one contains some pretty large changes as it is?

sds_in_a_box/SDSCode/cognito_signup_message.py
bryan-harter: Whenever a user receives an email from the cognito userpool, first it will send the event through this lambda.  This lambda's job is to take that event, change it in some way, and then return the changed event back to cognito.  

So in this lambda's case, we only want to change the message a user receives when their account is first created by the admin (i.e from the event "CustomMessage_AdminCreateUser").  What I wanted to do here was simply add a link to the signup URL, because by default congito doesn't actually tell the user where to go to verify their account and change their initial password.  

There are several other events that will cause this lambda to trigger.  For example if the user wants to reset their password, cognito sends them an email which triggers this lambda (with the tigger "CustomMessage_ForgotPassword").  However, we don't want to change those specific emails from the user, so the event in that case is just passed along untouched.  

To make this all more clear, I'll add some documentation to the top of this lambda

app.py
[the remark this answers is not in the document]
sdhoyt: I like the configuration file idea. Would a configuration file check happen in app.py or is there another place where that makes more sense? And I agree, that wouldn't be something for this ticket.

sds_in_a_box/sds_in_a_box_stack.py
bryan-harter: So these are the libraries that the lambda function needs in order to function.  Generally we handle dependencies in one of three ways in lambda functions:

1) We use Docker containers for our lambda functions, and have that dockerfile install all dependencies.  (https://docs.aws.amazon.com/lambda/latest/dg/images-create.html).  This is the route that Libera is taking when it comes to python dependencies.  *However*, what bothers me about that is we can't edit the code directly from the AWS console any more, making it much harder to develop and debug.  I really enjoy the flexibilty of making tweaks directly in the console.  

2) We can also package the dependencies into the lambda function itself.  One potentatial problem here is that the total amount of python code can't exceed 50MB, so it's possible that your code and all of the code in the libraries we need get so large that it can't fit in the lambda function any more.  

3) We use something called "lambda layers", which is what I do here.  ( https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html ) I'm installing things into a seperate directory locally, and CDK takes that directory, zips it up, and creates a "layer" object from it.  You can go here and check out the most recent layer this stack created: https://us-west-2.console.aws.amazon.com/lambda/home?region=us-west-2#/layers/SDSDependencieskugzmjqr394CC578/versions/1?tab=versions.  

Now I really have no idea if this is the best way to go about it.  I just googled how to handle python dependencies in the CDK, and it seems like its possible lambda might just automatically install and include things in a requirements.txt file.  I'll experiment a little more with this and see if there is a nicer looking way to go about this that doesn't install things into a local directory.

app.py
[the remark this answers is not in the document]
bryan-harter: We could try it in the app.py file, and see if it runs when we call "cdk synth" or "cdk deploy".  But if that doesn't work, we could make a separate script that checks for a config file and/or asks for user input, and then calls cdk synth/cdk deploy for the user.

sds_in_a_box/sds_in_a_box_stack.py
bryan-harter: So I was just playing around with lambda dependencies for awhile, and unfortunately it doesn't look like there is a good way to do this without things getting pretty complex.  

There is an experimental construct in the CDK which is supposed to be able to auto-install a the libraries in a requirements.txt file: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda_python_alpha/PythonFunction.html

But since it is experimental, it might not make sense to be using this yet.  Hopefully it'll get out of alpha soon.  

There is also a way to specify using a Docker image to help build your lambda function, as seen here: https://docs.aws.amazon.com/cdk/api/v1/docs/aws-lambda-readme.html#bundling-asset-code.  (That code is in javascript, but I sort of got it working for python.)  The problem is, this will spin up a docker container locally in order to help build the code that gets uploaded to the lambda function.  But since we're doing our development inside of a docker container already, building a Docker image inside of another Docker image looks like it requires a lot of weird parameters set, if it is even possible at all.  

So all that is to say, I don't necessarily see a better option at the moment.  I'm open to ideas though.  It might be something we have to revisit later on, maybe when development has stabilized a bit?

sdhoyt (approved):

sds_in_a_box/SDSCode/cognito_signup_message.py
tech3371: thanks!

sds_in_a_box/SDSCode/api.py
tech3371: I see.

tech3371 (approved):

.devcontainer/Dockerfile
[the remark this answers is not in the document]
greglucas: These should be a part of `pyproject.toml` if they are a part of the project requirements so I can install these outside of the GitPod structure if I want.

But, if they are a part of individual Lambdas they should be in `requirements.txt` within that directory structure and not force someone to install it globally.

README.md
```diff
@@ -1,22 +1,71 @@
...
+
+After about 20 minutes or so, you should have a brand new SDS set up in AWS.  
+
+
+### Virtual Desktop for Development
+Codespaces actually comes with a fully functional virtual desktop.  To open, click on the "ports" tab and then "open in new browser".  The default password is "vscode".  
 
 
+### Testing the APIs
+Inside of the "scripts" folder is a python script you can use to call the APIs.  It is completely independent of the rest of the project, so you should be able to pull this single file out and run it anywhere.  It only depends on basic python libraries.
```
greglucas, line 69: This should be a separate project then?

app.py
```diff
@@ -1,17 +1,20 @@
...
 
-app = cdk.App()
-SdsInABoxStack(app, "SdsInABoxStack")
-
-app.synth()
+# An initial user of the APIs
+initial_user = "harter@lasp.colorado.edu"
```
greglucas, line 15: Don't hardcode your credentials in the shared repository.

sds_in_a_box/SDSCode/api.py
```diff
@@ -0,0 +1,76 @@
+import json
+import os
+import requests
+import time
+from jose import jwk, jwt
```
greglucas, line 5: Why did you choose to use jose instead of boto3 to take care of the authentication for you?

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -13,43 +15,52 @@
...
         super().__init__(scope, construct_id, **kwargs)
+        
+########### INIT
+        # Determines the initial configuration
+        if not SDS_ID:
+            SDS_ID="".join( [random.choice(string.ascii_lowercase) for i in range(8)] )
+        
+        if not initial_user:
+            print("Please set an initial user before calling this stack")
+            return
```
greglucas, line 33: Should raise in these two situations?

sds_in_a_box/sds_in_a_box_stack.py
greglucas: Please don't system install, that is dangerous and if I'm running locally I absolutely don't want this happening for me without me knowing.

See here for an example of how we structured this for swxtrec_cdk. There are a few various lambda models over there where we put this `requirements.txt` file in and it works nicely.
https://github.com/SWxTREC/swxtrec-cdk/tree/main/swxtrec_cdk/msis/lambda

I think using the alpha Lambda version is totally fine. We may have to update syntax later, but that is easy enough to do.

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -72,27 +83,122 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+        os.system("pip install python-jose -t ./external/python")
+        layer = lambda_.LayerVersion(self, f"SDSDependencies-{SDS_ID}",
+                                    code=lambda_.Code.from_asset("./external"),
+                                    description="A layer that contains all dependencies needed for the lambda functions"
+                                )
+
+        # This is the role that the lambdas will all assume
+        # TODO: We'll narrow things down later, right now we're just giving admin access
         lambda_role = iam.Role(self, "Indexer Role", assumed_by=iam.ServicePrincipal("lambda.amazonaws.com"))
         lambda_role.add_managed_policy(iam.ManagedPolicy.from_aws_managed_policy_name("AdministratorAccess"))
```
greglucas, line 100: 👀 Please don't add Administrator access as a blanket policy!! That is not secure. We should follow the principle of least access.

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -72,27 +83,122 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+                                               )],
+                                               username="harter@lasp.colorado.edu"
+                                              )
+
+########### OUTPUTS
+        # This is a list of the major outputs of the stack
+        cdk.CfnOutput(self, "API_URL", value=api_url.url)
+        cdk.CfnOutput(self, "COGNITO_USERPOOL_ID", value=userpool.user_pool_id)
+        cdk.CfnOutput(self, "COGNITO_APP_ID", value=command_line_client.user_pool_client_id)
+        cdk.CfnOutput(self, "SIGN_IN_WEBPAGE", value=f"https://sds-login-{SDS_ID}.auth.us-west-2.amazoncognito.com/login?client_id={command_line_client.user_pool_client_id}&redirect_uri=https://example.com&response_type=code")
```
greglucas, line 204: Do we have an imap url yet? Might be nice to use `dev.imap-soc.com` `imap-soc.com` for nicer domains?

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -72,27 +83,122 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+        userpool.add_trigger(cognito.UserPoolOperation.CUSTOM_MESSAGE, signup_lambda)
+
+        # Create an initial user of the API
+        initial_user = cognito.CfnUserPoolUser(self, "MyCfnUserPoolUser",
+                                               user_pool_id=userpool.user_pool_id,
+                                               desired_delivery_mediums=["EMAIL"],
+                                               force_alias_creation=False,
+                                               user_attributes=[cognito.CfnUserPoolUser.AttributeTypeProperty(
+                                                  name="email",
+                                                  value="harter@lasp.colorado.edu"
```
greglucas, line 194: remove individual credentials here.

If you want this set automatically I suppose you could add this in as an argument into the stack itself so everyone could pass in what they want.

greglucas (dismissed):
I think this needs some work still. There are a few security implications that should be addressed I think.

When making the Cognito stack here, it might be nice to separate out the Cognito into a more "global" version across this project so we aren't creating/destroying the pool each time we update the app? So users don't have to add their credentials each time.

here are some links for adding tests as well:
https://github.com/SWxTREC/swxtrec-cdk/blob/main/swxtrec_cdk/legacy/cognito_stack.py
https://github.com/SWxTREC/swxtrec-cdk/blob/main/swxtrec_cdk/tests/test_cognito.py

sds_in_a_box/SDSCode/api.py
bryan-harter: I actually couldn't figure out how to have boto3 do it! If you do this on another project let me know.  But on this page: https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-verifying-a-jwt.html it basically tells you exactly what you need to do to verify. 

There are a bunch of AWS SDK operations here: https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_Operations.html but I couldn't see one that verified a token.  There is one at the bottom called "VerifySoftwareToken" but that seems to involve MFA one time codes.

sds_in_a_box/sds_in_a_box_stack.py
bryan-harter: Alrighty, I'll try using the alpha version.

.devcontainer/Dockerfile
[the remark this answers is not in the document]
bryan-harter: I need to check out what a pyproject.toml is, I've been using setup.py and requirements.txt forever but it seems like this is the newer way of doing things!

README.md
bryan-harter: Possibly, though "completely independent" might be the wrong term.  It still depends on the specific API endpoint and cognito userpool that the stack sets up, so in that sense it still needs to be modified before its useful.  But once those values are set, the script doesn't rely on the rest of the project.  

I was still debating on how we want to go about modifying them once the stack is created, whether it is an automated process that puts the variables into the file, or what.  I was mainly keeping this file in here as a code sample for when we started the tickets for https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box/issues/8 https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box/issues/9 and https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box/issues/10

.devcontainer/Dockerfile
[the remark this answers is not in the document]
greglucas: I added one that will come in via the other PR for reference:
https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box/pull/18/files

sds_in_a_box/SDSCode/api.py
greglucas: The web team does it on their end and uses the AWS specific frameworks on their end.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cognito-idp.html#CognitoIdentityProvider.Client.initiate_auth

https://docs.aws.amazon.com/cognito-user-identity-pools/latest/APIReference/API_InitiateAuth.html

Or actually, are you trying to add Cognito as an auth mechanism to all APIs? We can let AWS handle that by adding an authorizer in CDK.

Brian M set that up in TREC for us and had it working, but we got rid of Cognito because people didn't want to deal with a login:
https://github.com/SWxTREC/swxtrec-cdk/commit/7edb962a89868a10a545a62df30ada2b53faeec9

```python
        # Add Cognito auth requirement to API gateway
        authorizer = apigw2_auth.HttpUserPoolAuthorizer(
            "APIGWAuthorizer",
            authorizer_name="APIGWAuthorizer",
            pool=user_pool,
        )
```

sds_in_a_box/sds_in_a_box_stack.py
bryan-harter: Nope no URL yet, we should get started thinking about that

sds_in_a_box/SDSCode/api.py
bryan-harter: Yeah we're adding it as an authorization mechanism for all the APIs.  The current plan is that if they are a team user, they can get access to non-public data.  If they don't add an authorization token in their header, then they can still access the public data.  

As of right now we aren't using API Gateway, we're invoking the lambda functions directly via HTTPS calls, so the API Gateway authorizer won't work with it.  The advantage of invoking the lambda directly is 1) Less infrastructure in general to worry about and 2) The limits on API gateway don't apply to invoking the lambda directly (for example, the 29 second timeout with API gateway)

.devcontainer/Dockerfile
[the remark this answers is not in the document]
bryan-harter: Ahh gotcha, maybe we should merge yours first, then I can pull in your changes to this branch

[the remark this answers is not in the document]
greglucas: Sure, it doesn't matter either way to me, that one hasn't been reviewed at all, so I'm happy to rebase that onto this one as well.

sds_in_a_box/SDSCode/api.py
greglucas: OK, sounds good for now! I don't know anything about the Lambda URLs yet and asked about them on the AWS working group call yesterday, but no one there had experimented with them either. So it sounds like you're breaking new ground and will be the expert here :)

> 1) Less infrastructure in general to worry about 

I actually think you get a lot from APIGW for as simple as it is to implement and it can handle DDOS attack vectors and other security things a little easier.

> 2) The limits on API gateway don't apply to invoking the lambda directly (for example, the 29 second timeout with API gateway)

I think this is probably the important reason why you'd want this to be urls rather than APIGW. Although I'd also say that as a user I get annoyed if my request is taking longer than 30 seconds and not sending me any streaming data back or things to let me know it is alive still, so if we're planning on long running Lambdas we should think if there is a better way to handle that situation. (I thought you did something like this on EMM where you'd put the data into a bucket for the user to download and send them a link to that instead of sending it directly back to the user)

sds_in_a_box/sds_in_a_box_stack.py
bryan-harter: Darn, so the problem with the alpha version and a requirements.txt file is that it still ultimately uses a Docker container to install the requirements for the lambda.  And then the problem there is that we're already running our code inside of a docker container, so we get an error about not being able to find Docker.  So the issue I mentioned above still applies.  

It may be possible by specifying custom bundling options as seen here: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda_python_alpha/README.html#custom-bundling

Though if we got that working, we'd have to complicate the CDK a bit by adding some sort of user input that specifies if you're running this CDK inside a docker container or not.

bryan-harter:
Closing this to re-open

PR #18, opened 2023-01-18
TST/CI: Add automated testing for pull requests

greglucas:
Add pyproject.toml so we can install the project
Add GitHub Actions CI
Add network test markers to ignore tests that require network connectivity

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,32 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  cdk-tests:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -l {0}
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.9"
+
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -0,0 +1,52 @@
+[build-system]
```
tech3371, line 1: I haven't used toml file before. Are these value referenced by something?

.github/workflows/test.yml
```diff
@@ -0,0 +1,32 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
tech3371, line 7: I see. Just stating pull_request allows it to run test when anybody creates PR. Didn't know that before.

tech3371 (approved):
Looks good to me!

pyproject.toml
greglucas: This is the way to specify your project setup now and can reference other tools and requirements as well `[tool.xyz]`. For example, poetry would look different than this.
https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

https://peps.python.org/pep-0517/

bryan-harter (approved):

PR #21, opened 2023-02-09
4 create api for querying the data

sdhoyt:
# Change Summary

*Note: Deployments need to be run locally to avoid docker issues with lambda layers*

## Overview

- Added new class to opensearch_utils to represent an OpenSearch query
- Added search method to opensearch_utils Client class
- Added Query API lambda
- Updated config.json L0 pattern to match ICD
- Now indexing files using the full S3 bucket path as the ID
- Added new OpenSearch domain resource policy so lambda can access OpenSearch
- Reverted to python 3.7 for indexer and queries lambdas due to python lambda alpha limitation (more details under sds_in_a_box_stack bullet point below)

## New Dependencies
- aws-cdk.aws-lambda-python-alpha

## New Files

- query.py
   - opensearch_utils class that takes in a dictionary of field:value API Query parameters and construct an opensearch Query DSL. This currently ignores any query parameters sent to the API that are not expected (currently: instrument, level, start_date, end_date). The valid list may need to be expanded or we might decide to return an error if invalid search fields are sent over the API rather than just ignoring them (open for discussion). the Query object can be passes directly into the Client.search method.
- queries.py
   - lambda function that receives a query from the lambda url, constructs a query, sends the query to opensearch, and returns the search results. The search results will be returned as a list of dictionaries. Example below:

```json
{
        "_index": "metadata",
        "_type": "_doc",
        "_source": {
            "date": "20230202",
            "mission": "imap",
            "extension": "pkts",
            "level": "l0",
            "instrument": "swe",
            "type": "sci",
            "version": "v01"
        },
        "_id": "s3://sdsinaboxstack-databucketa168b937-193uladkg9d2y/imap/l0/imap_l0_sci_swe_20230202_v01.pkts",
        "_score": 0.4700036
    }
```

## Updated Files

- sds_in_a_box_stack.py
   - added Query API lambda 
   - added new resource policy for opensearch domain
   - changed indexer and queries lambda to run with python 3.7. Seems to be the most recent python version that can be used with the python lambda alpha library which is used for layers. I think this might be a result of the aws-cdk version that we're using, but not 100% sure.
- indexer.py
   - changed opensearch document ID to use full S3 bucket path instead of only file name
- client.py
   - added new search method to send queries to opensearch.
- payload.py
   - added new payload_chunk method to fix payload chunking and bulk upload metadata without limit
   - changed import paths to other opensearch_utils classes to relative to work in lambda and local tests
- config.json
   - updated L0 pattern to match ICD
- document.py
   - changed import paths to other opensearch_utils classes to relative to work in lambda and local tests
- index.py
   - changed import paths to other opensearch_utils classes to relative to work in lambda and local tests

## Testing
- Added 7 new unit tests
- All unit tests are passing, however there is an issue with import paths for unit tests vs lambdas for indexer.py and queries.py. The import paths inside these files need to be changed to relative to work when unit tests are run. I temporarily made them relative to run the tests and switched them back after. This isn't ideal, so I'm trying to find another solution.
- I uploaded 2 files empty files to the S3 bucket (imap_l0_sci_mag_20230201_v01.pkts and map_l0_sci_swe_20230201_v01.pkts) and the files were correctly indexed into opensearch. I then used the lambda url to pass search queries using Postman and the expected opensearch metadata was returned.


sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -85,20 +97,48 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
                                           handler="indexer.lambda_handler",
                                           role=lambda_role,
-                                          runtime=lambda_.Runtime.PYTHON_3_9,
+                                          runtime=lambda_.Runtime.PYTHON_3_7,
                                           timeout=cdk.Duration.minutes(15),
                                           memory_size=1000,
                                           environment={
                                             "OS_ADMIN_USERNAME": "master-user", 
-                                            "OS_ADMIN_PASSWORD_LOCATION": os_secret.secret_name,
+                                            "OS_ADMIN_PASSWORD_LOCATION": os_secret.secret_value.unsafe_unwrap(),
```
sdhoyt, line 105: Is it ok to use the unsafe_unwrap() method for the two lambdas? Or does anyone have any better suggestions?

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -72,6 +73,17 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
             )
         )
 
+        # add an access policy for opensearch
+        domain.add_access_policies(
```
sdhoyt, line 77: Is this access policy too unrestricted? There needs to be an access policy for the lambda to access opensearch, I got it working with this policy, but I'm not sure if this is too open for our purposes.

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -85,20 +97,48 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
                                             "OS_ADMIN_USERNAME": "master-user", 
-                                            "OS_ADMIN_PASSWORD_LOCATION": os_secret.secret_name,
+                                            "OS_ADMIN_PASSWORD_LOCATION": os_secret.secret_value.unsafe_unwrap(),
                                             "OS_DOMAIN": domain.domain_endpoint,
                                             "OS_PORT": "443",
-                                            "OS_INDEX": "metadata"
-                                            }
+                                            "OS_INDEX": "metadata",
+                                            "S3_BUCKET": data_bucket.s3_url_for_object()},
+                                            layers=[lambda_alpha_.PythonLayerVersion(self, "SDSIndexerCodeLayer", entry=os.path.join(os.path.dirname(os.path.realpath(__file__)), "SDSCode/"))]
```
greglucas, line 110: Rather than a layer here, can you actually make the entire Function through lambda_alpha? That should automatically create the appropriate layer for you I _think_.
https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda_python_alpha/README.html#python-function

I think that might help with the Python Version required here too.

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -85,20 +97,48 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
                                           )
 
         indexer_lambda.add_event_source(S3EventSource(data_bucket,
                                                       events=[s3.EventType.OBJECT_CREATED]
                                                       )
                                         )
         indexer_lambda.apply_removal_policy(cdk.RemovalPolicy.DESTROY)
+
+        # The purpose of this lambda function is to trigger off of a lambda URL.
+        query_api_lambda = lambda_.Function(self,
```
greglucas, line 120:
```suggestion
        query_api_lambda = lambda_alpha_.PythonFunction(self,
```

PythonFunction here which I think will add your layer automatically for you 🤞

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -85,20 +97,48 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+                                          id="QueryAPILambda",
+                                          code=lambda_.Code.from_asset(os.path.join(os.path.dirname(os.path.realpath(__file__)), "SDSCode/")),
+                                          handler="queries.lambda_handler",
+                                          role=lambda_role,
+                                          runtime=lambda_.Runtime.PYTHON_3_7,
+                                          timeout=cdk.Duration.minutes(15),
+                                          memory_size=1000,
+                                          environment={
+                                            "OS_ADMIN_USERNAME": "master-user",
```
greglucas, line 129: Maybe we can add query-only users if we need to do this? It would be great if we can do it through iam roles if possible though.

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -85,20 +97,48 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
         indexer_lambda.apply_removal_policy(cdk.RemovalPolicy.DESTROY)
+
+        # The purpose of this lambda function is to trigger off of a lambda URL.
+        query_api_lambda = lambda_.Function(self,
+                                          id="QueryAPILambda",
+                                          code=lambda_.Code.from_asset(os.path.join(os.path.dirname(os.path.realpath(__file__)), "SDSCode/")),
+                                          handler="queries.lambda_handler",
+                                          role=lambda_role,
+                                          runtime=lambda_.Runtime.PYTHON_3_7,
+                                          timeout=cdk.Duration.minutes(15),
```
greglucas, line 126: How long do queries usually take? 15 minutes seems like a really long time.

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -85,20 +97,48 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+                                            layers=[lambda_alpha_.PythonLayerVersion(self, "SDSQueryCodeLayer", entry=os.path.join(os.path.dirname(os.path.realpath(__file__)), "SDSCode/"))]
+                                          )
+        # add function url for lambda query API
+        lambda_query_api_function_url = lambda_.FunctionUrl(self,
+                                                 id="QueryAPI",
+                                                 function=query_api_lambda,
+                                                 auth_type=lambda_.FunctionUrlAuthType.NONE,
+                                                 cors=lambda_.FunctionUrlCorsOptions(
+                                                                     allowed_origins=["*"],
+                                                                     allowed_methods=[lambda_.HttpMethod.GET, lambda_.HttpMethod.POST]))
```
greglucas, line 144: If it is only a query API, do we want to allow POST or only GET?

tests/unit/opensearch_utils/test_client.py
```diff
@@ -183,6 +197,79 @@ def test_send_payload(self):
...
+        self.client.send_document(self.document2, action_override=Action.DELETE)
+        self.client.send_document(self.document3, action_override=Action.DELETE)
+
+    def test_scroll_search(self):
+        """
+        test that the search method correctly scrolls through a large set of results
+        and returns all results without error
+        """
+        ## Arrange ##
+        search_true = [{'_index': 'test_data', '_type': '_doc', '_id': '3', '_score': 1.0, '_source': {'test body': 10}},
```
greglucas, line 229: Are they always in this order? It seems like the indexes jump around.

tests/unit/test_indexer.py
```diff
@@ -82,19 +82,20 @@ def setUp(self):
...
 
         ## Act
         indexer.lambda_handler(self.sample_payload, "")
-
+        time.sleep(1)
         document_out = self.client.get_document(self.document)
 
         ## Assert
+        print("out:  " + json.dumps(document_out))
+        print("true: "  + json.dumps(document_true))
```
greglucas, line 98:
```suggestion
```

tests/unit/test_queries.py
```diff
@@ -0,0 +1,74 @@
...
+from sds_in_a_box.SDSCode.opensearch_utils.payload import Payload
+from sds_in_a_box.SDSCode.opensearch_utils.client import Client
+
+
+@pytest.mark.network
+class TestQueries(unittest.TestCase):
+
+    def setUp(self):
+        #Opensearch client Params
+        os.environ["OS_DOMAIN"] = 'search-sds-metadata-uum2vnbdbqbnh7qnbde6t74xim.us-west-2.es.amazonaws.com'
```
greglucas, line 23: Is this specific to your stack?

We might want to look at mocking opensearch in the future: https://pypi.org/project/openmock/

tests/unit/test_queries.py
```diff
@@ -0,0 +1,74 @@
...
+        os.environ["OS_PORT"] = '443'
+        os.environ["OS_INDEX"] = "test_data"
+
+        hosts = [{"host":os.environ["OS_DOMAIN"], "port":os.environ["OS_PORT"]}]
+        
+        secret_name = "OpenSearchPassword9643DC3D-uVH94BjrbF9u"
```
greglucas, line 29: We probably don't want this on the open internet or is this a test name somehow? It seems like someone could possibly piece together things here since we are using a lot of admin and hardcoded urls throughout here.

greglucas (commented):
Are you comfortable rebasing and removing content from old commits?

There is a 28.8 MB `AWSCLIV2.pkg` that should be removed and it is best if we remove it from all history so we don't have to pull it down with git when cloning. Let me know if you have questions or want to walk through that together.

sds_in_a_box/SDSCode/opensearch_utils/query.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    Query class to represent an AWS OpenSearch query domain-specific language (DSL),
+    providing a range of search options. Query objects will be used by the Client
+    search method and used in OpenSearch searches.
+
+    ...
+
+    Attributes
+    ----------
+    query: dict
```
tech3371, line 13: did you mean `query_params` here?

sds_in_a_box/SDSCode/opensearch_utils/query.py
```diff
@@ -0,0 +1,101 @@
...
+        query_match_structure = {"match": {}}
+        query_date_structure = {"range": {"date": {}}}
+        
+        # remove all params that are not valid
+        query_params = self.__filter_params(query_params)
+
+        # create the query
+        for param in query_params:
+            # create a date query using start and end date parameters
+            if param == "start_date" or param == "end_date":
```
tech3371, line 62: I remember things being tricky when you query opensearch. In this function, you are trying to construct this query logic
```
{
  "query": {
    "bool": {
      "filter": {
        "range": {
          "date": {
            "gte": start_date,
            "lte": end_date
          }
        }
      }
    }
  }
}
```
is that right?

sds_in_a_box/SDSCode/opensearch_utils/query.py
```diff
@@ -0,0 +1,101 @@
...
+                # already exist
+                if "must" not in query["query"]["bool"]:
+                    query["query"]["bool"]["must"] = []
+                
+                # add the search parameters to the must query structure
+                query_match = query_match_structure.copy()
+                query_match["match"] = {param:query_params[param]}
+                query["query"]["bool"]["must"].append(query_match)
+            
+        return query
```
tech3371, line 84: so the end query will look something like this?

```
{
  "query": {
    "bool": {
      "filter": {
        "range": {
          "date": {
            "gte": 10,
            "lte": 20
          }
        }
      }
      , "must": [
        {
          "match": {
            "level": "l0"
          }
        },
        {
          "match": {
            "instrument": "instrument"
          }
        }
      ]
    }
    
  }
}
```

sds_in_a_box/sds_in_a_box_stack.py
tech3371: This link has some good examples of IAM policy: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/ac.html

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -85,20 +97,48 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
                                           code=lambda_.Code.from_asset(os.path.join(os.path.dirname(os.path.realpath(__file__)), "SDSCode")),
                                           handler="indexer.lambda_handler",
                                           role=lambda_role,
-                                          runtime=lambda_.Runtime.PYTHON_3_9,
+                                          runtime=lambda_.Runtime.PYTHON_3_7,
```
tech3371, line 100: Is there reason why we went down to python3.7 instead of python3.9?

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -85,20 +97,48 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
                                             "OS_DOMAIN": domain.domain_endpoint,
                                             "OS_PORT": "443",
-                                            "OS_INDEX": "metadata"
-                                            }
+                                            "OS_INDEX": "metadata",
+                                            "S3_BUCKET": data_bucket.s3_url_for_object()},
+                                            layers=[lambda_alpha_.PythonLayerVersion(self, "SDSIndexerCodeLayer", entry=os.path.join(os.path.dirname(os.path.realpath(__file__)), "SDSCode/"))]
                                           )
 
         indexer_lambda.add_event_source(S3EventSource(data_bucket,
```
tech3371, line 113: Nice! I have not done s3 trigger this way.

sds_in_a_box/SDSCode/indexer.py
```diff
@@ -87,11 +87,12 @@ def lambda_handler(event, context):
             logger.info(f"Found no matching file types to index this file against.")
             return None
         
-        # Rather than returning the metadata, we should insert it into the DB
         logger.info("Found the following metadata to index: " + str(metadata))
 
+        # use the bucket path to file as the ID in opensearch
+        bucket_path = os.path.join(os.environ["S3_BUCKET"], filetypes[0]['path'], filename)
```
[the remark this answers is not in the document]
sdhoyt: - no problem. I can change the name to s3_path
- the s3:// is included in the S3_BUCKET environment variable

sds_in_a_box/SDSCode/opensearch_utils/query.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def query_dsl(self):
+        """Returns the query parameters in the AWS OpenSearch Query DSL format"""
+        return json.dumps(self.query_dsl_formatted)
+
+    def size(self):
+        """Returns the number of results the query is allowed to return in the search"""
+        return self.query_size
+
+    def __build_query_dsl(self, query_params):
```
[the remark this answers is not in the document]
sdhoyt: Yeah, not sure what I was thinking. I'll change to single underscore

sds_in_a_box/SDSCode/opensearch_utils/query.py
```diff
@@ -0,0 +1,101 @@
...
+        Builds a Query DSL using a dictionary with field:value pairings.
+
+        Parameters
+        ----------
+        query_params: dict
+            dictionary containing field:value search parameters.
+
+        """
+        # define the query structure
+        query = {"query": {"bool":{}}}
```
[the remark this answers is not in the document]
sdhoyt: This structure is to match the Query DSL format: https://opensearch.org/docs/latest/opensearch/query-dsl/bool/

I think opensearch needs to see that query key to know that a query is contained in it because you can send more complicated json requests with your search. For example, you could send a query key along with a sort key that will query and then sort the results that are returned:

{
  "query": {
    "term": {
      "play_name": {
        "value": "Henry IV"
      }
    }
  },
  "sort": [
    {
      "line_id": {
        "order": "desc"
      }
    }
  ]
}

https://opensearch.org/docs/2.4/opensearch/search/sort

It looks a little weird since we're just sending queries right now, but I think that key is necessary.

sds_in_a_box/SDSCode/opensearch_utils/client.py
```diff
@@ -130,12 +131,41 @@ def send_payload(self, payload):
...
+
+        Parameters
+        ----------
+        query: Query
+            query object instantiated with the desired query parameters.
+        index: Index
+            OpenSearch index to use for the search.
+        """
+        # search the opensearch instance with scroll to handle larger responses
+        result = self.client.search(body=query.query_dsl(), index=index.get_name(), params={"scroll": "1m"})
```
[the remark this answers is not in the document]
sdhoyt: "1m" tells opensearch to keep the search context open for 1 minute. I'm hoping that is enough for our searches, but I'm not 100% sure. I'm hoping this is something we can stress test with more test data.

sds_in_a_box/SDSCode/opensearch_utils/client.py
```diff
@@ -130,12 +131,41 @@ def send_payload(self, payload):
...
+        scroll_id = result['_scroll_id']
+        scroll_size = len(result["hits"]["hits"])
+        counter = 0
+        full_result = result["hits"]["hits"]
+
+        # scroll through the results and add results to list
+        while scroll_size > 0:
```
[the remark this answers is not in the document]
sdhoyt: Sorry, responded to the other comment before I saw this. Correct, there's a request limit that can get hit when querying for large amounts of data. The scroll features helps to avoid that limit.

sds_in_a_box/SDSCode/opensearch_utils/query.py
sdhoyt: Yes, I'll fix that

sds_in_a_box/SDSCode/opensearch_utils/query.py
sdhoyt: That's correct

sds_in_a_box/SDSCode/opensearch_utils/query.py
sdhoyt: Yeah, that looks correct

sds_in_a_box/SDSCode/queries.py
```diff
@@ -0,0 +1,32 @@
...
+def lambda_handler(event, context):
+    logger.info("Received event: " + json.dumps(event, indent=2))
+    # create the opensearch query from the API parameters
+    query = Query(event["queryStringParameters"])
+    client = _create_open_search_client()
+    logger.info("Query: " + query.query_dsl())
+    # search the opensearch instance
+    search_result = client.search(query, Index(os.environ["OS_INDEX"]))
+    logger.info("Query Search Results: " + json.dumps(search_result))
+    return search_result
```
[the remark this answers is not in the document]
sdhoyt: Interesting. I can add an issue for that

sds_in_a_box/sds_in_a_box_stack.py
sdhoyt: for some reason when I use the lambda_python_alpha library for layers, it doesn't allow me to use a python version greater than 3.7. I'm going to add an issue for this to spend more time trying to figure out why.

sds_in_a_box/sds_in_a_box_stack.py
sdhoyt: Yeah, I forgot to change that time. I can probably change it to 1 minute for now. I don't feel like I have a good idea of how long searches will take / how much data there will be to search through. Maybe this is something we could eventually stress test with some test data.

greglucas (approved):
Approving based on our conversation earlier.

TODO:
* Fix up easy things
* Make tickets for complicated things to look into later (auth, ...)
* Mark your tests as `pytest.mark.network()` so the tests aren't run on CI and create a ticket to look into mocking / fixing that discrepancy later.

sds_in_a_box/sds_in_a_box_stack.py
sdhoyt: The following issues were opened to address this comment:
https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box/issues/22
https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box/issues/23

sds_in_a_box/sds_in_a_box_stack.py
sdhoyt: The following issue was opened to address this comment
#22

pyproject.toml
```diff
@@ -28,6 +28,7 @@ classifiers = [
 requires-python = ">=3.9"
 dependencies = [
     "aws-cdk-lib==2.48.0",
```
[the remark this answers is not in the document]
sdhoyt: The PythonFunction suggestion seems to have fixed the python version issue, so this update shouldn't be necessary

.devcontainer/Dockerfile
```diff
@@ -11,6 +11,7 @@ RUN unzip awscliv2.zip
 RUN sudo ./aws/install
 RUN sudo apt-get install -y python3 python3-pip
 RUN pip install aws-cdk-lib==2.48.0 
+RUN pip install aws-cdk.aws-lambda-python-alpha
```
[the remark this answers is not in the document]
sdhoyt: Removed from docker

sds_in_a_box/SDSCode/opensearch_utils/query.py
```diff
@@ -0,0 +1,101 @@
...
+                    query["query"]["bool"]["must"] = []
+                
+                # add the search parameters to the must query structure
+                query_match = query_match_structure.copy()
+                query_match["match"] = {param:query_params[param]}
+                query["query"]["bool"]["must"].append(query_match)
+            
+        return query    
+
+    def __filter_params(self, query_params):
```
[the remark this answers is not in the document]
sdhoyt: Fixed

sds_in_a_box/SDSCode/opensearch_utils/query.py
```diff
@@ -0,0 +1,101 @@
...
+        # define the query structure
+        query = {"query": {"bool":{}}}
+        query_match_structure = {"match": {}}
+        query_date_structure = {"range": {"date": {}}}
+        
+        # remove all params that are not valid
+        query_params = self.__filter_params(query_params)
+
+        # create the query
+        for param in query_params:
```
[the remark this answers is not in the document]
sdhoyt: moved valid params to top of file and removed unnecessary method

sds_in_a_box/sds_in_a_box_stack.py
sdhoyt: This fixed the python version issue. Thanks!

sds_in_a_box/sds_in_a_box_stack.py
sdhoyt: The following issues were opened to address this comment:
https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box/issues/22
https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box/issues/23

sds_in_a_box/sds_in_a_box_stack.py
sdhoyt: removed POST

tests/unit/opensearch_utils/test_client.py
sdhoyt: The keys have always been in the same order since I've been running tests

tests/unit/test_indexer.py
sdhoyt: removed

tests/unit/test_queries.py
sdhoyt: This is a test domain that I've been using for my unit tests. 
Opened issue https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box/issues/24 as a result of this comment.

tests/unit/test_queries.py
sdhoyt: The good thing is this is the secret location, not the actual secret value. However, I agree, this isn't ideal. Opened issue #23 as a result of this comment.

sds_in_a_box/SDSCode/queries.py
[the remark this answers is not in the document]
sdhoyt: Opened issue #25

sds_in_a_box/sds_in_a_box_stack.py
sdhoyt: This issue was fixed by a suggestion in another comment. The lambdas are now using 3.9

sds_in_a_box/SDSCode/indexer.py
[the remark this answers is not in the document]
tech3371: I see. That might be confusing later.

sds_in_a_box/SDSCode/opensearch_utils/client.py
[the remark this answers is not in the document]
tech3371: Good idea. We can stress test that in the new issue you opened.

sds_in_a_box/sds_in_a_box_stack.py
tech3371: nice!

tech3371 (approved):
looks good to me

sdhoyt:
Thanks for all the help/comments with this PR!

PR #28, opened 2023-02-16
Improving the development environment on codespaces

bryan-harter:
This is just cleaning up some of the things we'll need for codespaces to function


.devcontainer/Dockerfile
```diff
@@ -13,7 +13,10 @@ RUN sudo apt-get install -y python3 python3-pip
 RUN pip install aws-cdk-lib==2.48.0 
 RUN pip install "constructs>=10.0.0,<11.0.0"
 RUN pip install boto3
+RUN pip install requests
 RUN pip install opensearch-py
+RUN pip install python-jose
+RUN pip install aws_cdk.aws_lambda_python_alpha
```
greglucas, line 19:
```suggestion
```

We should let the **package** define the dependencies rather than through the dockerfile so we are all kept in-sync even if we aren't using Docker. So you'd add these dependencies to `pyproject.toml`. Then `pip install -e .[dev]` to get the dependencies installed into this environment. These should also be the dependencies required to run the CDK/infrastructure and not the lambda's as well.

bryan-harter: I had these in here primarily because Codespaces doesn't have access to the repo while its building the docker image, but I did find out a way to automatically run a pip install command after container is built.  So that's what it'll do now.

.devcontainer/devcontainer.json
```diff
@@ -9,7 +9,10 @@
...
 			"version": "latest"
+		}, "ghcr.io/devcontainers/features/docker-in-docker:2": { 
+			"version": "latest" 
 		}
 	},
-	"forwardPorts": [6080]
-}
\ No newline at end of file
+	"forwardPorts": [6080],
+    "postCreateCommand": "pip install ."
```
tech3371, line 17: Does this mean that this command will execute when Git workspace is up?

.devcontainer/devcontainer.json
```diff
@@ -9,7 +9,10 @@
 	"features": {
 		"ghcr.io/devcontainers/features/desktop-lite:1": {
 			"version": "latest"
+		}, "ghcr.io/devcontainers/features/docker-in-docker:2": {
```
tech3371, line 12: nice work on finding this out!

tech3371 (approved):
looks good to me!

.devcontainer/devcontainer.json
bryan-harter: Yeah after the docker container is spun up and your code is put into the codespace, the final command it'll run will be a pip install.  It should grab all of the required libraries in the pyproject.toml file.

pyproject.toml
```diff
@@ -31,7 +31,9 @@ dependencies = [
     "aws-cdk.aws-lambda-python-alpha",
     "constructs>=10.0.0,<11.0.0",
     "boto3",
-    "opensearch-py"
+    "opensearch-py",
+    "requests",
+    "python-jose"
```
greglucas, line 36: These aren't CDK/infrastructure dependencies? I think they are Lambda dependencies, so should be in the `requirements.txt` file next to the lambda that needs them.

bryan-harter: True, I mainly wanted them installed so I could play around with them in the devcontainer if I was troubleshooting something for the lambdas (for example, import jose in a python terminal and make sure a certain function in that library worked).  I could either add these libraries to the devcontainer dockerfile, or just pip install them manually if I play around with them like that.

greglucas: Yeah, I'd prefer that. I like to keep **requirements** as minimal as possible. There is `optional.requires` where we added the `pip install .[dev]`, you could add some of those into another requires section and then install that way too?

.devcontainer/devcontainer.json
maxinelasp: Why not put the pip install inside the docker container?

pyproject.toml
maxinelasp: Putting the lambda dependencies in a different optional group will help keep them seperate, but still within the poetry dependency management. This can also be good for splitting up dev/prod/test dependencies. 

https://python-poetry.org/docs/master/managing-dependencies/#optional-groups

pyproject.toml
```diff
@@ -31,7 +31,9 @@ dependencies = [
     "aws-cdk.aws-lambda-python-alpha",
```
maxinelasp, line 31: This isn't part of your changeset, but why is the pyproject.toml set up like this? The documentation has it so that your dependencies are under their own header: `[tool.poetry.dependencies]`. This gives more flexibility for different groupings or optional dependencies. Also, I'm just in general kind of confused about the formatting here, it doesn't seem to align with what the docs say. 

https://python-poetry.org/docs/master/pyproject/#dependencies-and-dependency-groups

maxinelasp (changes requested):

pyproject.toml
maxinelasp: I'm happy to change the pyproject format and make the changes I suggested btw, especially since this isn't really part of your change set so it shouldn't block the merge. I just want to make sure this format isn't due to some codespace rule or some formatting recommendations that I am unaware of.

maxinelasp (approved):

pyproject.toml
greglucas: This is one of my biggest issues with the python packaging trying to move over to this new format now. Each tool/backend will define various subsections and so it is confusing to change back and forth for requirements. This is currently correct for the setuptools backend, but looks like you say it won't work with Poetry?

https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

Everyone wants to define their own standard 😂

maxinelasp: I thought this was a poetry pyproject.toml, I didn't realize it was setuptools! Poetry does use a slightly different format, but it won't be hard to transition over once. It's sort of an ["in addition" thing](https://bbc.github.io/cloudfit-public-docs/packaging/this_way_up.html#pyprojecttoml), rather than replacing it. The [project] section can remain the same (I think?), it looks like poetry operates on top of pyproject.toml, not replacing it. 

I hate python dependency management 🙃

maxinelasp: I would like to move to poetry because it's flexible and powerful enough to completely replace pip, venv, conda, all those managers, while generally [being better than them](https://blogs.sap.com/2022/05/08/why-you-should-use-poetry-instead-of-pip-or-conda-for-python-projects/), particularly at dependency management. 

Dependency management in pip is awful, and in poetry it's still awful but at least you can fix the problems for everyone else? It's not a fix-all but it at least collapses all the dependency management, venv, installation, etc into one thing.

.devcontainer/devcontainer.json
bryan-harter: I wasn't sure if the code was available inside the docker container while it was building.  Some brief googling said to just run the pip install commands after the devcontainer was built.

bryan-harter:
Maybe I can merge this for now, under the assumption that Maxine will re-write our dependency management shortly? The only reason I'm in a rush to get it out is so that the codespaces works with some of the other changes I'm merging shortly.

maxinelasp:
> Maybe I can merge this for now, under the assumption that Maxine will re-write our dependency management shortly? The only reason I'm in a rush to get it out is so that the codespaces works with some of the other changes I'm merging shortly.

Works for me, I have the docker container running locally mostly and will work on dependency management inside of there.

bryan-harter:
Merged into the development branch

PR #29, opened 2023-02-16
Adding a way to create unique stack names each deploy

bryan-harter:
This pull request adds the ability to add "context" to the cdk synth and cdk deploy commands so that stack that's built appends the "SDS_ID" to the S3 buckets and lambda functions created

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # SDS-in-a-box
 
-Hello!  This a project that attempts to capture the minimum requirements for the core of a Science Data System.  
+This project that attempts to capture the minimum requirements for the core of a Science Data System.  
 
 Our goal with the project is that users will only need to modify the file config.json to define the data products stored on the SDC, and the rest should be mission agnostic.  
 
@@ -21,7 +21,7 @@ The development environment uses a GitHub codespace, to ensure that we're all us
 
 Everyone gets 50 free hours per month of github Codespace time.  Alternatively, your organization can pay for it to run longer than this.  
 
-Top start a new development environment, click the button for "Code" in the upper right corner of the repository, and click "Codespaces".  
+To start a new development environment, click the button for "Code" in the upper right corner of the repository, and click "Codespaces".  
 
 
 ### AWS Setup
@@ -43,7 +43,7 @@ cdk bootstrap
 
 Inside the app.py file, there are two important configuration items which you can alter:
 
-1) SDS_ID - This is just a string of 8 random letters that are appendend to each resources.  Alternatively, you can change this to something more meaningful (i.e. "bryan-testing").  Just be aware that this ID needs to be completely unique in each account.  
+1) SDS_ID - This is just a string of 8 random letters that are appended to each resource.  Alternatively, you can change this to something more meaningful (i.e. "bryan-testing").  Just be aware that this ID needs to be completely unique in each account.  
 
 To deploy the SDS, first you'll need to snyth the CDK code with the command:
 
--- a/README.md
+++ b/README.md
@@ -1,22 +1,76 @@
+# SDS-in-a-box
 
-First you need to install nodejs (make sure its newer than version 14!)
+Hello!  This a project that attempts to capture the minimum requirements for the core of a Science Data System.  
 
-https://nodejs.org/en/download/
+Our goal with the project is that users will only need to modify the file config.json to define the data products stored on the SDC, and the rest should be mission agnostic.  
 
-Then you'll need to install the cdk:
+## Architecture
 
-npm install -g aws-cdk
+The code in this repository takes the form of an AWS CDK project. It provides the architecture for:
 
-pip install boto3
-pip install aws-cdk-lib
+1. An HTTPS API to upload files to an S3 bucket (*in development*)
+2. An S3 bucket to contain uploaded files
+3. An HTTPS API to query and download files from the S3 bucket (*in development*)
+4. A lambda function that inserts file metadata into an openseach instance
+5. A Cognito User Pool that keeps track of who can access the restricted APIs.  
 
-The first thing you'll need to do is configure your aws environemnt with 
 
+## Development
+
+The development environment uses a GitHub codespace, to ensure that we're all using the proper libraries as we develop and deploy.  
+
+Everyone gets 50 free hours per month of github Codespace time.  Alternatively, your organization can pay for it to run longer than this.  
+
+Top start a new development environment, click the button for "Code" in the upper right corner of the repository, and click "Codespaces".  
+
+
+### AWS Setup
+The first thing you'll need to do is configure your aws environemnt with:
+
+```
 aws configure
+```
 
-Then you'll need to bootstrap your cdk deployment with 
+Enter in your AWS Access Key ID and AWS Secret Access Key, which can be obtained by setting up a user account in the AWS console. For region, set it to the AWS region you'd like to set up your SDS.  For IMAP, we're using "us-west-2"
 
+**NOTE**-- If this is a brand new AWS account, then you'll need to bootstrap your account to allow CDK deployment with the command: 
+
+```
 cdk bootstrap
+```
+
+### Deploy
+
+Inside the app.py file, there are two important configuration items which you can alter:
+
+1) SDS_ID - This is just a string of 8 random letters that are appendend to each resources.  Alternatively, you can change this to something more meaningful (i.e. "bryan-testing").  Just be aware that this ID needs to be completely unique in each account.  
+
+To deploy the SDS, first you'll need to snyth the CDK code with the command:
+
+```
+cdk synth --context SDSID={insert a unique ID here}
+```
+
+and then you can deploy the architecture with the following command:
+
+```
+cdk deploy --context SDSID={insert a unique ID here}
+```
+for example:
+
+```
+cdk synth --context SDSID=harter-testing
+cdk deploy --context SDSID=harter-testing
+```
+
+After about 20 minutes or so, you should have a brand new SDS set up in AWS.  
 
 
+### Virtual Desktop for Development
+Codespaces actually comes with a fully functional virtual desktop.  To open, click on the "ports" tab and then "open in new browser".  The default password is "vscode".  
 
+
+### Testing the APIs
+Inside of the "scripts" folder is a python script you can use to call the APIs.  It is completely independent of the rest of the project, so you should be able to pull this single file out and run it anywhere.  It only depends on basic python libraries.  
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,22 +1,76 @@
+# SDS-in-a-box
 
-First you need to install nodejs (make sure its newer than version 14!)
+Hello!  This a project that attempts to capture the minimum requirements for the core of a Science Data System.
```
greglucas, line 3:
```suggestion
This a project that attempts to capture the minimum requirements for the core of a Science Data System.  
```

README.md
```diff
@@ -1,22 +1,76 @@
...
 
-The first thing you'll need to do is configure your aws environemnt with 
 
+## Development
+
+The development environment uses a GitHub codespace, to ensure that we're all using the proper libraries as we develop and deploy.  
+
+Everyone gets 50 free hours per month of github Codespace time.  Alternatively, your organization can pay for it to run longer than this.  
+
+Top start a new development environment, click the button for "Code" in the upper right corner of the repository, and click "Codespaces".
```
greglucas, line 24:
```suggestion
To start a new development environment, click the button for "Code" in the upper right corner of the repository, and click "Codespaces".  
```

README.md
```diff
@@ -1,22 +1,76 @@
...
 
+```
 cdk bootstrap
+```
+
+### Deploy
+
+Inside the app.py file, there are two important configuration items which you can alter:
+
+1) SDS_ID - This is just a string of 8 random letters that are appendend to each resources.  Alternatively, you can change this to something more meaningful (i.e. "bryan-testing").  Just be aware that this ID needs to be completely unique in each account.
```
greglucas, line 46:
```suggestion
1) SDS_ID - This is just a string of 8 random letters that are appended to each resource.  Alternatively, you can change this to something more meaningful (i.e. "bryan-testing").  Just be aware that this ID needs to be completely unique in each region.  
```

README.md
```diff
@@ -1,22 +1,76 @@
...
+### Deploy
+
+Inside the app.py file, there are two important configuration items which you can alter:
+
+1) SDS_ID - This is just a string of 8 random letters that are appendend to each resources.  Alternatively, you can change this to something more meaningful (i.e. "bryan-testing").  Just be aware that this ID needs to be completely unique in each account.  
+
+To deploy the SDS, first you'll need to snyth the CDK code with the command:
+
+```
+cdk synth --context SDSID={insert a unique ID here}
```
greglucas, line 51: Is there a way to get the SDSID to persist between commands without having to put the context in each time? I can set `AWS_PROFILE=xyz`, so I'm curious if there is something analogous here we can use? (This looks good as-is, I'm just curious if I can speed it up myself locally)

app.py
```diff
@@ -1,17 +1,23 @@
...
 import os
 import boto3 
+import string
+import random
 import aws_cdk as cdk
 import aws_cdk.assertions as assertions
 from sds_in_a_box.sds_in_a_box_stack import SdsInABoxStack
 
-
+### Implement the code found in "sds_in_a_box_stack.py"
```
greglucas, line 10:
```suggestion
```
I'm not sure what the comment means. Maybe as a module-level docstring instead?

app.py
```diff
@@ -1,17 +1,23 @@
...
 app = cdk.App()
-SdsInABoxStack(app, "SdsInABoxStack")
 
-app.synth()
+SDS_ID = app.node.try_get_context("SDSID")
+
+if SDS_ID is None:
+    raise Exception("ERROR: Need to specify an ID to name the stack (ex - production, testing, etc)")
```
greglucas, line 16: If it is None do you want to just assign it the random ID and log that out for the user, or warn and say you're using a random ID?

```suggestion
if SDS_ID is None:
    raise ValueError("the context SDS_ID needs to specify an ID to name the stack (ex - production, testing, etc)")
```

app.py
```diff
@@ -1,17 +1,23 @@
...
+    # A random unqiue ID for this particular instance of the SDS
+    SDS_ID = "".join( [random.choice(string.ascii_lowercase) for i in range(8)] )
+    
 
-#app = cdk.App()
-#stack = SdsInABoxStack(app, "sds-in-a-box")
-#template = assertions.Template.from_stack(stack)
-#print(template)
\ No newline at end of file
+SdsInABoxStack(app, "SdsInABoxStack"+SDS_ID)
```
greglucas, line 22:
```suggestion
SdsInABoxStack(app, f"SdsInABoxStack-{SDS_ID}")
```

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -20,10 +20,14 @@ class SdsInABoxStack(Stack):
 
     def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
         super().__init__(scope, construct_id, **kwargs)
+        
+ ########### INIT
+        # Determines the initial configuration
+        SDS_ID = self.node.try_get_context("SDSID")
```
greglucas, line 26:
```suggestion
```
We've already parsed it in "app.py", so should we pass it into this function rather than trying to get it again?

Also, if it is None I don't think this is working because there is no fallback like in the original.

tests/unit/test_sds_in_a_box_stack.py
```diff
@@ -4,13 +4,13 @@
...
 from sds_in_a_box.sds_in_a_box_stack import SdsInABoxStack
 
 # example tests. To run these tests, uncomment this file along with the example
 # resource in sds_in_a_box/sds_in_a_box_stack.py
 def test_sqs_queue_created():
-    app = core.App()
+    app = core.App(context={"SDS_ID": "unit-testing"})
+    app_props = core.AppProps(context={"SDS_ID": "unit-testing"})
     stack = SdsInABoxStack(app, "sds-in-a-box")
     template = assertions.Template.from_stack(stack)
```
greglucas, line 14: I'm not following this test. Are there assertions in here somewhere? It looks like it is just creating the stack and template, but not guaranteeing anything is created. We also don't have sqs_queue yet?

tests/unit/test_sds_in_a_box_stack.py
```diff
@@ -4,13 +4,12 @@
...
-
 from sds_in_a_box.sds_in_a_box_stack import SdsInABoxStack
 
 # example tests. To run these tests, uncomment this file along with the example
 # resource in sds_in_a_box/sds_in_a_box_stack.py
 def test_sqs_queue_created():
-    app = core.App()
+    app = core.App(context={"SDSID": "unit-testing"})
     stack = SdsInABoxStack(app, "sds-in-a-box")
     template = assertions.Template.from_stack(stack)
```
greglucas, line 14: I'm not following this test. Are there assertions in here? It seems like it is just creating the stack and template, but not checking against what is in the stack or template...

Name of the test should be updated as well.

greglucas (commented):
Nice! This is nice to use the context for updating the stack names.

README.md
```diff
@@ -1,22 +1,76 @@
...
+```
+
+and then you can deploy the architecture with the following command:
```
maxinelasp, line 54:
```suggestion
This will generate the changes you are attempting to make, so you can review them if needed.  Then, you can deploy the architecture with the following command:
```

Just wanted to add a note about the difference between `synth` and `deploy`

maxinelasp (commented):
Also curious about sds_in_a_box_stack.py getting SDS_ID from the context rather than being passed in. Otherwise looks good to me!

app.py
tech3371, line 22: Can we add a space or `-` between `SdsInABoxStack` and `SDS_ID`? When I was using random string, it becomes hard to read what is SDS_ID.

app.py
bryan-harter: I'm not sure either tbh

app.py
bryan-harter: I would probably just err on the side of forcing a user to specify the context, but I'm fine either way

sds_in_a_box/sds_in_a_box_stack.py
bryan-harter: I initially had it passed in as variables to the function, but I think there was a reason that I stopped passing it in and just looking at the context again.  I'd need to play around with it again to determine why that way.

tests/unit/test_sds_in_a_box_stack.py
bryan-harter: I actually didn't add the unit test, it comes by default when you set up any new CDK project (including the function name "test_sqs_queue_created") and I just never removed it from the repo.  

This unit test started running after github actions was set up to run the unit tests, so I was just making a small fix here to stop github from complaining about the pull request failing some checks.    

I can either fix the unit tests here, or we could make another ticket for it

greglucas: Sure, we can make another ticket, I was just confused here! We have a lot of test work to do, so I think we can break things down into more specific actions then.

sds_in_a_box/sds_in_a_box_stack.py
greglucas: :+1: The reason to pass it in would be that currently if I set my ID to "random", then that will put the random string here instead of the randomly generated string from the app.

app.py
greglucas: Sounds reasonable to me to force a user to provide one rather than default.

README.md
bryan-harter: I just looked it up, apparently we can't! CDK deploy re-synthesizes each time before each deploy anyway, so there is no saved state between a synth and a deploy.  

I thought cdk synth was necessary in order to deploy, but it turns out it's primarily for finding errors and responding to warnings before you run a deploy.

bryan-harter: See here: https://docs.aws.amazon.com/cdk/v2/guide/hello_world.html#hello_world_tutorial_deploy

It is optional (though good practice) to synthesize before deploying. The AWS CDK synthesizes your stack before each deployment.

greglucas: I was actually wondering about whether I could globally set these values, and it looks like you can...
https://docs.aws.amazon.com/cdk/v2/guide/context.html

They are taken from many different places, so I'd have to update one of them.
```
Context values can be provided to your AWS CDK app in six different ways:

Automatically from the current AWS account.

Through the --context option to the cdk command. (These values are always strings.)

In the project's cdk.context.json file.

In the context key of the project's cdk.json file.

In the context key of your ~/.cdk.json file.

In your AWS CDK app using the construct.node.setContext() method.

```

bryan-harter:
Alrighty I think I responded to everyone's requests in this commit, and just pushed out a new version.  I did just pass in the context to the function rather than look up the context again inside the function, it appeared to synth and deploy just fine so 🤷

maxinelasp (approved):

tech3371 (approved):

PR #30, opened 2023-02-17
Primarily adding more fine-grained permissions to the lambda

bryan-harter:
But also made some small improvements to the lambdas

### DIFF
--- a/sds_in_a_box/sds_in_a_box_stack.py
+++ b/sds_in_a_box/sds_in_a_box_stack.py
@@ -177,7 +177,6 @@ class SdsInABoxStack(Stack):
         # download query API lambda
         download_query_api = lambda_alpha_.PythonFunction(self,
             id="DownloadQueryAPILambda",
-            function_name=f'download-query-api-{SDS_ID}',
             entry=os.path.join(os.path.dirname(os.path.realpath(__file__)), "SDSCode/"),
             index='download_query_api.py',
             handler="lambda_handler",
--- a/sds_in_a_box/sds_in_a_box_stack.py
+++ b/sds_in_a_box/sds_in_a_box_stack.py
@@ -83,6 +83,40 @@ class SdsInABoxStack(Stack):
             actions=["es:*"],
             resources=[domain.domain_arn + "/*"]
             ))
+            
+########### IAM POLICIES
+        opensearch_all_http_permissions = iam.PolicyStatement(
+                                               effect=iam.Effect.ALLOW,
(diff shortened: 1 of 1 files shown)


sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -84,56 +84,85 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
                                             "OS_ADMIN_USERNAME": "master-user", 
                                             "OS_ADMIN_PASSWORD_LOCATION": os_secret.secret_value.unsafe_unwrap(),
-                                            "OS_DOMAIN": domain.domain_endpoint,
+                                            "OS_DOMAIN": sds_metadata_domain.domain_endpoint,
                                             "OS_PORT": "443",
                                             "OS_INDEX": "metadata"
                                             }
                                           )
+        query_api_lambda.add_to_role_policy(
+            iam.PolicyStatement(
```
maxinelasp, line 150: Are all these lambdas expected to have some of the same permissions? It might be worth putting these `iam.PolicyStatement` values into one variable, so you only need to update in one place.

maxinelasp (approved):

sds_in_a_box/sds_in_a_box_stack.py
bryan-harter: Haha yeah that's true, they should all be pretty much the same so it makes sense to put them into variables.  The only potential benefit I could see to do it this way is if each lambda would have unique policies, which they might depending on how deep we want to go with least-privilege restrictions.    

For instance, we could find out the exact actions the query API needs to perform on opensearch from this list: https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonopensearchservice.html, and rather than give it permissions to do all those actions to our opensearch cluster ("es:*"), we determine only the specific actions we need to grant, such as "es:DescribeElasticsearchDomain", "es:ESHttpDelete", "es:ESHttpHead", etc

maxinelasp: Definitely, if the lambdas are expected to have different permissions, we should plan to have them only have what's needed. Then, my comment doesn't really apply, because it's mostly placeholders anyway.

greglucas (approved):

tech3371 (approved):
looks good to me

bryan-harter:
I actually just added a section in the stack specifically for IAM policies and narrowed down their use

bryan-harter:
Merged and closed

PR #31, opened 2023-02-17
Small bug fix that allows the indexer to operate when the file has a path name

bryan-harter:
In other words, the indexer wouldn't work if there was a directory structure that the file was put in, it only works in the base path of the S3 bucket


maxinelasp (approved):

tech3371 (approved):

greglucas (approved):
As a general comment, I think we should try and move towards `pathlib` so the pathhandling gets simplified a bit, but this looks good for now.

bryan-harter:
Merging into the development branch and closing this one

PR #32, opened 2023-02-17
Adding some outputs to the cloudformation stack

bryan-harter:
This is where important outputs will go (such at the URLs of the lambda function APIs)

### DIFF
--- a/sds_in_a_box/sds_in_a_box_stack.py
+++ b/sds_in_a_box/sds_in_a_box_stack.py
@@ -155,12 +155,16 @@ class SdsInABoxStack(Stack):
             timeout=cdk.Duration.seconds(60)
         )
 
-        lambda_.FunctionUrl(self,
-            id="DownloadQueryAPI",
-            function=download_query_api,
-            auth_type=lambda_.FunctionUrlAuthType.NONE,
-
-            cors=lambda_.FunctionUrlCorsOptions(
-                                allowed_origins=["*"],
-                                allowed_methods=[lambda_.HttpMethod.GET])
+        download_api_url = lambda_.FunctionUrl(self,
+                                               id="DownloadQueryAPI",
+                                               function=download_query_api,
+                                               auth_type=lambda_.FunctionUrlAuthType.NONE,
+                                               cors=lambda_.FunctionUrlCorsOptions(
+                                                                   allowed_origins=["*"],
+                                                                   allowed_methods=[lambda_.HttpMethod.GET])
         )
+########### OUTPUTS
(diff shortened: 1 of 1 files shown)


sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -155,12 +155,16 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+                                               function=download_query_api,
+                                               auth_type=lambda_.FunctionUrlAuthType.NONE,
+                                               cors=lambda_.FunctionUrlCorsOptions(
+                                                                   allowed_origins=["*"],
+                                                                   allowed_methods=[lambda_.HttpMethod.GET])
         )
+########### OUTPUTS
+        # This is a list of the major outputs of the stack
+        #cdk.CfnOutput(self, "UPLOAD_API_URL", value=upload_api_url.url)
+        cdk.CfnOutput(self, "QUERY_API_URL", value=lambda_query_api_function_url.url)
```
maxinelasp, line 169: What are these outputs used for? From googling, it looks like they have a few different uses.

bryan-harter: They don't do anything directly within the stack, but after the stack is finished it prints out the values of the generated URLs.  These values are saved in AWS, so you can easily reference the values again from elsewhere.  For example, other Cloudformation stacks can reference these outputs.  So you can link cloudformation stacks in a chain, feeding the output of one stack into the input of another.  

Really the only thing I'm using it for now is so I can copy+paste those URLs into my test script without inspecting the AWS console to see what URLs AWS came up with this time.  But we might have more use for these outputs in the future.

tech3371: I didn't think of doing that. Really makes it easy to look up these values like you said.

tech3371 (approved):

sds_in_a_box/sds_in_a_box_stack.py
maxinelasp: Alright, great! So right now just for testing, but maybe in the future, to be used in other ways. Thanks for explaining!

maxinelasp (approved):

greglucas (approved):
Seems fine to output these for convenience now. In the near-term I think we should look at getting Route53 and domains set up so we can reference static URLs that dynamically point to our deployed resources.

`bryands-sds-id.imap-sdc.com/download_query_api` etc...

bryan-harter:
Alrighty, merging and closing this one

PR #33, opened 2023-02-17
Adding the upload API to the stack and adding the Upload API code

bryan-harter:
This adds an HTTPS api that people can use to upload files to the SDS

### DIFF
--- a/sds_in_a_box/SDSCode/requirements.txt
+++ b/sds_in_a_box/SDSCode/requirements.txt
@@ -1,1 +1,3 @@
+requests
+python-jose
 opensearch-py
--- a/sds_in_a_box/SDSCode/upload_api.py
+++ b/sds_in_a_box/SDSCode/upload_api.py
@@ -1,0 +1,117 @@
+import json
+import os
+import requests
+import time
+import boto3
+from jose import jwk, jwt
+from jose.utils import base64url_decode
+import logging
+
+logger = logging.getLogger()
+logging.basicConfig()
+logger.setLevel(logging.INFO)
+
+def _load_allowed_filenames():
+    '''
+    This function loads the config.json file, and stores the contents as a python dictionary 
+    
+    :return: dictionary object of file types and their attributes. 
+    '''
+
+    current_dir = os.path.dirname(__file__)
+    config_file = os.path.join(current_dir, "config.json")
+    
+    with open(config_file) as f:
+        data = json.load(f)
+    return data
+
+def _check_for_matching_filetype(pattern, filename):
+    '''
+    This function takes in a pattern from config.json and compares it to the desired file name
+    
+    :param pattern: A file naming pattern from the config.json
+    :param filename: Required.  String name of the desired file name.  
+    
+    :return: None if there is no match, or the file_dictionary if there is.  
+    '''
+    split_filename = filename.replace("_", ".").split(".")
+
+    if len(split_filename) != len(pattern):
+        return None
+    
+    i = 0
+    file_dictionary = {}
+    for field in pattern:
+        if pattern[field] == '*':
+            file_dictionary[field] = split_filename[i]
+        elif pattern[field] == split_filename[i]:
+            file_dictionary[field] = split_filename[i]
+        else:
+            return None
+        i += 1
+    
+    return file_dictionary
+
+def _generate_signed_upload_url(filename, tags={}):
+    '''
+    Based on a given filename, this function will open up a presigned url into the correct location on the SDS storage bucket
+    
+    :param filename: Required.  A string representing the name of the object to upload.  
+    :param tags: Optional.  A dictionary object of key:value pairs that will be stored in the S3 object metadata.  
+    
+    :return: None if the filename does not match mission naming conventions.  Otherwise, a URL string.  
+     '''
+    filetypes = _load_allowed_filenames()
+    for filetype in filetypes:
+        path_to_upload_file = filetype['path']
+        metadata = _check_for_matching_filetype(filetype['pattern'], filename)
+        if metadata is not None:
+            break
+        
+    if metadata is None:
+        logger.info(f"Found no matching file types to index this file against.")
+        return None
+    
+    bucket_name = os.environ["S3_BUCKET"]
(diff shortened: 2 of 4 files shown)


sds_in_a_box/SDSCode/requirements.txt
```diff
@@ -1 +1,3 @@
+requests
```
tech3371, line 1: What is this requirements.txt used for?

sds_in_a_box/SDSCode/upload_api.py
```diff
@@ -0,0 +1,117 @@
+import json
+import os
+import requests
+import time
+import boto3
+from jose import jwk, jwt
+from jose.utils import base64url_decode
```
tech3371, line 7: should we remove these? They are not used

sds_in_a_box/SDSCode/upload_api.py
```diff
@@ -0,0 +1,117 @@
...
+import logging
+
+logger = logging.getLogger()
+logging.basicConfig()
+logger.setLevel(logging.INFO)
+
+def _load_allowed_filenames():
```
tech3371, line 14: Should we create a utils folder to store these commonly used function? And explain or give example of what is/should be in `config.json`

sds_in_a_box/SDSCode/upload_api.py
bryan-harter: Ah yeah sorry, these imports are actually a part of the cognito pull request (which I'm saving for last, after all these things are properly reviewed).  These will be added in again for that pull request

sds_in_a_box/SDSCode/upload_api.py
bryan-harter: Yeah I was thinking someone might say we should start up a utilities module to the code, I was just being lazy for this commit and copy+pasted some things from the other lambda functions lol

sds_in_a_box/SDSCode/requirements.txt
bryan-harter: When you use a that new CDK function to build a python lambda function, it automatically installs the libraries listed in the requirements.txt

bryan-harter: This one here: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda_python_alpha/README.html

sds_in_a_box/SDSCode/upload_api.py
```diff
@@ -0,0 +1,117 @@
...
+    url = _generate_signed_upload_url(filename, tags=event['queryStringParameters'])
+    
+    if url is None:
+        return {
+            'statusCode': 400,
+            'body': json.dumps("A pre-signed URL could not be generated. Please ensure that the file name matches mission file naming conventions.")
+        }
+    
+        
+    return {
```
sdhoyt, line 114: what is supposed to happen with the returned url? Is that supposed lead to a page where you can upload files?

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -115,6 +115,33 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
...
+        # Adding a lambda for uploading files to the SDS
+        upload_api_lambda = lambda_alpha_.PythonFunction(self,
+                                      id="APILambda",
+                                      entry=os.path.join(os.path.dirname(os.path.realpath(__file__)), "SDSCode/"),
+                                      index="upload_api.py",
+                                      handler="lambda_handler",
+                                      runtime=lambda_.Runtime.PYTHON_3_9,
+                                      timeout=cdk.Duration.minutes(15),
+                                      memory_size=1000,
+                                      environment={"OS_ADMIN_USERNAME": "master-user",
```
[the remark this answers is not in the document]
bryan-harter: Yeah you're right they aren't currently needed, I'll remove

sds_in_a_box/SDSCode/upload_api.py
bryan-harter: Pretty much, though its not a drag and drop page or anything.  It's a spot where a user can do a http PUT request to upload files.  

For example:
 `with open("file.txt", 'rb') as object_file:
        object_text = object_file.read()
    response = requests.put(upload_url, data=object_text)`

sdhoyt: thanks! I think you made this change in the original larger PR, but the config.json needs to have a "/" added to the end of the path (or add it when you combine the path and filename. When I upload a file it's currently adding l0 as a prefix to the filename instead of it being the directory that the file is contained in.

bryan-harter: Ah yes now that you mention it, I did make changes to the config.json!  I'll be sure to put that in this commit.

sds_in_a_box/sds_in_a_box_stack.py
```diff
@@ -115,6 +115,31 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
                                                       )
                                         )
         indexer_lambda.apply_removal_policy(cdk.RemovalPolicy.DESTROY)
+        
+        # Adding a lambda for uploading files to the SDS
+        upload_api_lambda = lambda_alpha_.PythonFunction(self,
+                                      id="APILambda",
```
tech3371, line 121: Can we name it to `UploadAPILambda`? I was trying to test upload lambda by cdk deploy and looking at the resources created by it. It took me a while to find upload api lambda function.

tech3371 (commented):
besides my minor comments. it looks good to me

sds_in_a_box/sds_in_a_box_stack.py
bryan-harter: Whoops, yeah good catch

sds_in_a_box/SDSCode/upload_api.py
bryan-harter: Sure if we can do that for this ticket that would be great, I'd prefer not to make major changes on this ticket just because I've got like 7 branches of the code at the moment lol

tech3371 (approved):

sdhoyt (approved):

PR #19, opened 2023-02-03
MNT: Style updates

greglucas:
This adds pre-commit hooks for forcing a common style throughout the codebase.

I personally prefer this when working on larger projects to keep consistency, but definitely open to changing or removing some defaults/configurations if people aren't fans of this.

PR #36, opened 2023-02-24
MNT: Remove unnecessary requirements files and update pyproject

greglucas:
This is separate from the poetry work, and mostly just the first commit of #19. It also updates the Actions file to test on pushes to the development branch now as well.


greglucas:
I tried adding an empty pre-commit-file and that did not work. I think we do need that file in the base branch possibly.

Issue #37, opened 2023-02-24
Move project to poetry


maxinelasp:
We should move the project away from pip and setup tools to poetry in order to standardize dependency management and use a more powerful dependency management tool to fix some of the pip issues we saw in the SIT-1 test. 

Poetry also creates a project-specific virtual environment and manages tests and packaging. It allows for optional dependencies. We can export the poetry project to a requirements.txt if required for lambda.

greglucas:
Closed by #38

PR #38, opened 2023-02-27
Adding poetry to the project

maxinelasp:
#37 

This PR is to change the project to poetry. Quick explanation about the features used here:

- You can see that the format is pretty similar to the existing pyproject.toml. Adding dependencies can be done in the actual file, but I found it easiest to use [`poetry add`](https://python-poetry.org/docs/cli/#add). This will automatically pick a good version (the carat indicates that it will do minor version updates, but not major ones) or, you can specify version requirements. You can also specify the group (more on that later)
- The `poetry.lock` file will ensure that everyone is using the same versions. When you run `poetry install`, it will install the versions from the lock file. If you add a dependency, or upgrade, or want to change the version numbers, you can run `poetry lock` to regenerate that file, and commit a new version to the repo. We want to make sure that file is up to date - but if you run `poetry install` and the lock file is mismatched from `pyproject.toml`, it will remind you to update it. 
- this poetry project has a few groups. There's the default one `tool.poetry`, the testing one `tool.poetry.group.tests`, and the cdk one, `tool.poetry.group.cdk`. If you just run `poetry install`, the default and the test dependencies are installed, but not the cdk dependencies. If you want to exclude the test dependencies, you can use the `--without tests` flag. If you want to include the CDK dependencies, you can use the `--with cdk` flag. This way, we can have different groups of dependencies that are installed in different scenarios. 
- Poetry includes a virtual environment, which means you don't need to manually manage that separately. After running `poetry install`, you have to activate that environment using `poetry shell`. If you would prefer to use your own venv, through conda, poetry will use that one instead.[ More info here. ](https://python-poetry.org/docs/managing-environments/)

I will leave a handful of other notes calling out specific changes in the file itself, but that's the overview of poetry.

Since this does conflict with Greg's PR #19, I plan on waiting for that to merge first and resolving the conflicts at that point before merging this PR.


pyproject.toml
```diff
@@ -25,32 +26,32 @@ classifiers = [
...
 
-[project.optional-dependencies]
-dev = [
-    "pytest==6.2.5",
-    "moto"
-]
+[tool.poetry.dependencies]
+python = ">=3.9,<4"
+# Required for aws-cdk.aws-lambda-python-alpha
+aws-cdk-lib = ">=2.66.0,<3.0.0"
```
maxinelasp, line 33: This is the only change to the dependencies - when running, I got an error message about the version of aws-cdk-lib, so I bumped it from `2.48.0` to `2.66.0`. If this causes problems, I can downgrade again, but `aws-cdk.aws-lambda-python-alpha` requires at least this version number.

pyproject.toml
```diff
@@ -25,32 +26,32 @@ classifiers = [
...
+#aws-cdk-lib = "2.48.0"
+boto3 = "^1.26.78"
```
maxinelasp, line 35: Some of these libraries didn't have versions in the requirements.txt file - this is what poetry defaults to, but if the team would prefer to have these always be the most recent version, I can change it.  This just prevents major version bumps without human intervention.

maxinelasp:
I removed the requirements.txt files. If we absolutely need those (for lambdas or something I don't know about) then poetry can generate them. They shouldn't be edited by hand though, all the dependency managmenet should go through poetry.

pyproject.toml
```diff
@@ -25,32 +26,33 @@ classifiers = [
...
+"aws-cdk.aws-lambda-python-alpha" = "^2.66.0a0"
+constructs = ">=10.0.0,<11.0.0"
+black = "^23.1.0"
+
+[tool.poetry.group.tests.dependencies]
+# Installed by default, can be installed without these dependencies using `poetry install --without tests`
+pytest = "==6.2.5"
+moto = "^4.1.3"
+hypothesis = "^6.68.2"
```
maxinelasp, line 44: I didn't see hypothesis in any requirements.txt files - but I needed to install if for the tests to run. I put it under the tests dependency, but this was the only dependency I didn't pull from the previous pyproject.toml or requirements.txt files.

pyproject.toml
```diff
@@ -1,13 +1,14 @@
...
+build-backend = "poetry.core.masonry.api"
 
-[project]
-name = "SDS-in-a-box"
+[tool.poetry]
+name = "sds-data-manager"
 version = "0.1.0"
-description = "IMAP Science Operations Center SDS-in-a-box"
+description = "IMAP Science Operations Center AWS data manager"
+authors = ["Greg Lucas <greg.lucas@lasp.colorado.edu>, Bryan Harter <bryan.harter@lasp.colorado.edu>, Sean Hoyt <sean.hoyt@lasp.colorado.edu>, Tenzin Choedon <tenzin.choedon@lasp.colorado.edu>"]
```
greglucas, line 9:
```suggestion
authors = ["IMAP SDC Developers" <imap.sdc@lists.lasp.colorado.edu>]
```
Or add your name too :)

pyproject.toml
greglucas: :+1: Good catch!

pyproject.toml
greglucas: I don't think we actually use boto3 in the package?

pyproject.toml
```diff
@@ -25,32 +26,33 @@ classifiers = [
...
-]
+[tool.poetry.dependencies]
+python = ">=3.9,<4"
+# Required for aws-cdk.aws-lambda-python-alpha
+aws-cdk-lib = ">=2.66.0,<3.0.0"
+#aws-cdk-lib = "2.48.0"
+boto3 = "^1.26.78"
+"aws-cdk.aws-lambda-python-alpha" = "^2.66.0a0"
+constructs = ">=10.0.0,<11.0.0"
+black = "^23.1.0"
```
greglucas, line 38: I'd add this and ruff to a `dev` or `all` dependency section, which can also depend on the tests dependency below then.

pyproject.toml
greglucas: Hm... I don't think this is needed. This should be brought in via other dependencies since we don't have it anywhere in ours from what I can see either.

pyproject.toml
```diff
@@ -25,32 +26,33 @@ classifiers = [
...
+
+[tool.poetry.group.tests.dependencies]
+# Installed by default, can be installed without these dependencies using `poetry install --without tests`
+pytest = "==6.2.5"
+moto = "^4.1.3"
+hypothesis = "^6.68.2"
+
+[tool.poetry.group.cdk]
```
greglucas, line 46:
```suggestion
[tool.poetry.group.lambda_code]
```
I think this is meant to be if we want to install the dependencies for the lambda code as well? I was a bit confused by the name, and I'm not sure mine is a great suggestion either, so feel free to bikeshed that. But cdk made me think this was installing the aws cdk or something similar to that.

pyproject.toml
```diff
@@ -25,32 +26,33 @@ classifiers = [
...
+# Not installed by default, can be installed using `poetry install --with cdk`
+optional = true
+
+[tool.poetry.group.cdk.dependencies]
```
greglucas, line 50: If you can export this to a `poetry.lock` file in the lambda area that would be great! It should be placed in `sds_data_manager/lambda_code/poetry.lock` where there is a current requirements.txt file I think.

https://docs.aws.amazon.com/cdk/api/v2/docs/aws-lambda-python-alpha-readme.html#packaging

pyproject.toml
```diff
@@ -25,32 +26,33 @@ classifiers = [
...
+opensearch-py = "^2.1.1"
+requests = "^2.28.2"
+python-jose = "^3.3.0"
 
 [project.urls]
 homepage = "https://github.com/IMAP-Science-Operations-Center"
-repository = "https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box"
+repository = "https://github.com/IMAP-Science-Operations-Center/sds-data-manager"
 
-[tool.pytest.ini_options]
```
greglucas, line 49: We should keep all of these other tool sections as they are unrelated to the poetry work. We also need the marker and tests path to be discovered.

greglucas (commented):
Looks good! I tried `pip install .` inside the directory and that failed due to a missing module name, but that is because we are in the name transition still I think. When I changed it to the previous `sds-in-a-box` as the name, the `pip install` worked. So, I believe this will support both poetry and pip, but use the nicer poetry lock file like you mention when packaging/installing!

.github/workflows/test.yml
```diff
@@ -18,15 +18,14 @@ jobs:
...
+      - uses: Gr1N/setup-poetry@v8
 
-      - name: Install dependencies
+      - name: Install dependencies and app
         run: |
-          python -m pip install .[dev]
-      - name: Install the app
-        run: |
-          python -m pip install --no-deps -e .
+          poetry install --with cdk
```
greglucas, line 25: when installing with poetry, do you need to`poetry install .[tests]` or anything similar to signify the optional dependencies?

pyproject.toml
maxinelasp: Yeah, I didn't love the name either. Maybe `lambda_dev`?

pyproject.toml
maxinelasp: Ok, will investigate and add that! Good idea!

pyproject.toml
maxinelasp: Ok, good catch. I tried to keep the sections but must have deleted them at some point while formatting everything.

.github/workflows/test.yml
maxinelasp: So, the tests dependency is automatically installed with `poetry install` (this can be changed if you don't like that default.) The cdk section is not automatically installed, which is why I needed to include it here with `--with`. 

The test dependencies can be excluded with `poetry install --without tests`.

maxinelasp:
> Looks good! I tried `pip install .` inside the directory and that failed due to a missing module name, but that is because we are in the name transition still I think. When I changed it to the previous `sds-in-a-box` as the name, the `pip install` worked. So, I believe this will support both poetry and pip, but use the nicer poetry lock file like you mention when packaging/installing!

Hopefully this will be fixed once I merge your changes for real... 

I don't know if pip install uses the poetry.lock file... as far as I can tell, it just installs from pyproject.toml. Which is still fine, since there are some pretty specific versions set, but something to keep in mind.

pyproject.toml
maxinelasp: It shows up in the lambda code, and the tests. So, I could move it out of the main dependencies and into the lambda and test specific dependencies, if that's preferable.

greglucas: Ahh, I did not realize that! Yeah, I would put it into the test dependencies section then.

I like keeping things organized where the imports are actually needed/used if possible. The "base package" should really be only what is required to `poetry install --without-tests && cdk deploy` to get the full stack.

Then I think we have:
* lambda_dev: code for the lambdas, which actually builds and installs itself separate, so we don't need to specify those dependencies in our main app
* tests: everything to run `pytest`
* dev / all: everything to be able to develop locally (everything but lambda_dev I think?)

.github/workflows/test.yml
greglucas: Nice, thanks for clarifying, I didn't know that!

pyproject.toml
maxinelasp: I don't see it anywhere in the poetry dependencies tree, if I don't install it manually. When checking the other development branch and using the pip install method, I also don't see it in the pip dependencies.

pyproject.toml
maxinelasp: Ok, that sounds reasonable to me! I'll double check the dependencies to see what is actually needed for lambdas vs the cdk deploy, it will be good to have those separate.

pyproject.toml
maxinelasp: Forgot myself 🙃 I was going off the contributor list to the repo!

pyproject.toml
maxinelasp: Looks like poetry will only export to one type of `poetry.lock` file. Then, the lambda constructor will convert that to a `requirements.txt`. So, our options are either to copy the poetry.lock file into the lambda subfolder, or export the poetry project to a requirements.txt file to replace the one that's currently in there. Either way, I think that should be set up as a precommit action ([see here)](https://python-poetry.org/docs/cli/#export) and all dependencies should still be managed through this poetry file.

pyproject.toml
greglucas: and it is needed for you to locally run the tests locally? Can you get a traceback of which library the import is coming from when it fails?

pyproject.toml
greglucas: The thing we don't want is all the CDK dependencies going into the lambda_dev requirements (would include a bunch of AWS package bloat), we only want the lambda runtime requirements over there. So this might be a bit of work to figure out how to thread through only that section's dependencies and not the "project"'s dependencies.

maxinelasp: Yeah, in that case, we'll want to generate a requirements.txt file (which can select different groups to include or exclude) since if AWS generates it we won't have control over what gets installed.

pyproject.toml
maxinelasp: Yeah, I'll check it out

pyproject.toml
greglucas: > since if AWS generates it we won't have control over what gets installed.

Right now, it is just hand-made, so we have to manually keep track of the dependencies. I suppose you could also handmake a poetry.lock file?

maxinelasp: It won't be hand made. You can generate a requirements.txt file from the poetry dependencies (using the command `poetry export --without tests --with lambda_dev`) which will only contain the groups that you want included. 

On the other hand, the poetry.lock file includes everything, without any way to change it, but you control what's installed using the `poetry install` command.

 Since AWS will basically take the poetry.lock file and run `poetry export` without specifying groups, it will install everything out of poetry.lock. So instead, we should set up a pre-commit hook which will generate a requirements.txt file using only the specific groups we want to include. 

That generated file should NOT be changed by hand, all the dependencies should still be managed through pyproject.toml.

pyproject.toml
maxinelasp: You know what, my tests don't fail locally anymore. I removed the dependency. I must have been missing something in my first version

pyproject.toml
maxinelasp: Alright @greglucas, I added the precommit hook and the requirements.txt. Can you take a look? Thanks!

.pre-commit-config.yaml
```diff
@@ -11,3 +11,11 @@ repos:
     hooks:
     - id: ruff
       args: [--fix]
+  - repo: https://github.com/python-poetry/poetry
+    rev: '1.4.0'  # add version here
+    hooks:
+      - id: poetry-check
+      - id: poetry-lock
```
maxinelasp, line 18: The github check timed out on the poetry lock step, but it ran fine locally... is there any way to skip hooks on pull requests, but run them locally?

greglucas:
```
ci:
    skip: [poetry-lock, poetry-export]
```
Something like that, see the skip section here: https://pre-commit.ci/#configuration

pyproject.toml
```diff
@@ -25,26 +26,36 @@ classifiers = [
...
+
+[tool.poetry.group.lambda-dev]
+# Not installed by default, can be installed using `poetry install --with cdk`
+optional = true
+
+[tool.poetry.group.lambda-dev.dependencies]
+opensearch-py = "^2.1.1"
+requests = "^2.28.2"
+python-jose = "^3.3.0"
+boto3 = "^1.26.78"
```
greglucas, line 58:
```suggestion
```

* `python-jose`: We aren't using it yet from what I can tell, so we should add that back in again later.
* `boto3`: It does come pre-installed on Lambda's runtimes so do we want to install it here again? I can see either way on this one, so I don't have a strong preference.

pyproject.toml
```diff
@@ -25,26 +26,36 @@ classifiers = [
...
+ruff = "^0.0.253"
+
+[tool.poetry.group.tests.dependencies]
+# Installed by default, can be installed without these dependencies using `poetry install --without tests`
+pytest = "==6.2.5"
+moto = "^4.1.3"
+boto3 = "^1.26.78"
+
+[tool.poetry.group.lambda-dev]
+# Not installed by default, can be installed using `poetry install --with cdk`
```
greglucas, line 51:
```suggestion
# Not installed by default, can be installed using `poetry install --with lambda-dev`
```

.pre-commit-config.yaml
maxinelasp: Ah grand, thanks!

greglucas (approved):
This looks great to me and works well locally! I tested with my standard workflow of `pip` and `pytest`, so it is robust :)

I really like the way you auto-created the requirements file, that is a very nice touch! When testing it out, I did run into a minor annoyance that I'm not sure if you know how to get around, but I had to do it in 3 steps.
1. Modify pyproject.toml and add my new dependency in. Then run git commit
2. git commit fails the pre-commit hook, so then I add the newly generated poetry.lock file and git commit again.
3. git commit fails the pre-commit hook again, so I then add the newly generated sds_data_manager/lambda_code/requirements.txt file and commit again.

Thoughts on somehow combining that all into one failure/action?

maxinelasp:
> Thoughts on somehow combining that all into one failure/action?

Yeah, I ran into this to. For the record, if you add the dependency using `poetry add --group <groupname>` it updates the lock file right there ;)

I don't know how to fix this but I will do some googling and see if there's an easy fix.

maxinelasp:
Yeah, thinking about it a little more, there's no way for the precommit check to know if the requirements.txt needs to be updated, until after the poetry lock is updated. So, I don't think there's any way to fix this, unfortunately.

bryan-harter (approved):
Looks good!

bryan-harter:
I just opened up a codespaces, deployed everything, and tested everything out.  It all worked as I expected, with the exception that I needed to run the `source $(poetry env info --path)/bin/activate` command again when I first started Codespaces, even though it's in the postCreateCommand. I don't mind doing that each time though.

PR #40, opened 2023-03-01
Adding Cognito to the services that can be spun up in the CDK stack

bryan-harter:


bryan-harter:
pre-commit.ci autofix

bryan-harter:
pre-commit.ci autofix

bryan-harter:
pre-commit.ci autofix

app.py
```diff
@@ -1,14 +1,57 @@
...
 from sds_data_manager.sds_data_manager_stack import SdsDataManagerStack
 
 app = cdk.App()
 
+# Grab context from cdk synth and cdk deploy commands
 SDS_ID = app.node.try_get_context("SDSID")
+initial_user = app.node.try_get_context("initial_user")
+userpool_name = app.node.try_get_context("userpool_name")
+app_client_name = app.node.try_get_context("app_client_name")
+cognito_only = app.node.try_get_context("cognito_only")
```
greglucas, line 54: This seems like it is adding a lot of complication and confusion for deployment. We'll now have to remember all of these options if we want to deploy something.

sds_data_manager/lambda_code/SDSCode/cognito_signup_message.py
```diff
@@ -0,0 +1,80 @@
...
+                "Your Username is {username} and your password is {####}."
+                + "Continue signing up at: "
+                + os.environ["COGNITO_DOMAIN"]
+                + "/login?client_id="
+                + command_line_client
+                + "&redirect_uri=https://example.com&response_type=code"
+            )
+            event["response"]["emailMessage"] = message
+
+            print(event)
```
greglucas, line 79:
```suggestion
```

sds_data_manager/lambda_code/SDSCode/cognito_signup_message.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+    command_line_client = clients["UserPoolClients"][0]["ClientId"]
+
+    if "triggerSource" in event:
+        if event["triggerSource"] == "CustomMessage_AdminCreateUser":
+            # Modify the email a new user receives to contain a
+            # link to the cognito domain.
+            subject = f"Science Data System {os.environ['SDS_ID']} sign up"
+            event["response"]["emailSubject"] = subject
+            event["response"]["smsMessage"] = "We don't use this"
```
greglucas, line 68: Can you disable it then?

sds_data_manager/lambda_code/SDSCode/cognito_signup_message.py
```diff
@@ -0,0 +1,80 @@
...
+            message = (
+                "Your Username is {username} and your password is {####}."
+                + "Continue signing up at: "
+                + os.environ["COGNITO_DOMAIN"]
+                + "/login?client_id="
+                + command_line_client
+                + "&redirect_uri=https://example.com&response_type=code"
```
greglucas, line 75: Why are we going to `example.com`?

sds_data_manager/lambda_code/SDSCode/cognito_signup_message.py
```diff
@@ -0,0 +1,80 @@
+"""
```
greglucas, line 1: After reading through this, I'm wondering why we need a custom message at all? I think it is generally for formatting, but I don't see much formatting below, so should we just use the default and live with that? I see adding formatting as a follow-up ticket.

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -12,13 +12,10 @@
 def _load_allowed_filenames():
     """
     Load the config.json file as a python dictionary.
-
```
greglucas, line 15: There should be a blank line here before you get to the "return" section I think, same for all of these below. They shouldn't be squashed together.

sds_data_manager/sds_cognito_stack.py
```diff
@@ -0,0 +1,131 @@
...
+        self,
+        scope: Construct,
+        construct_id: str,
+        sds_id: str,
+        initial_email: str = "",
+        **kwargs,
+    ) -> None:
+        super().__init__(scope, construct_id, **kwargs)
+
+        ########### INIT
```
greglucas, line 25:
```suggestion
```

sds_data_manager/sds_cognito_stack.py
```diff
@@ -0,0 +1,131 @@
...
+        initial_user_context = initial_email
+
+        ########### IAM POLICIES
```
greglucas, line 28:
```suggestion
```
Your variable names are good already, I don't think we need comments reiterating the variable names.

sds_data_manager/sds_cognito_stack.py
```diff
@@ -0,0 +1,131 @@
...
+        cognito_admin_policy = iam.PolicyStatement(
+            effect=iam.Effect.ALLOW,
+            actions=["cognito-idp:*"],
+            resources=["*"],
+        )
+
+        ########### COGNITO
```
greglucas, line 35:
```suggestion
```

sds_data_manager/sds_cognito_stack.py
```diff
@@ -0,0 +1,131 @@
...
+        # Add a client sign in for the userpool
+        command_line_client = cognito.UserPoolClient(
+            user_pool=userpool,
+            scope=self,
+            id="sds-command-line",
+            user_pool_client_name=f"sdscommandline-{sds_id}",
+            id_token_validity=cdk.Duration.minutes(60),
+            access_token_validity=cdk.Duration.minutes(60),
+            refresh_token_validity=cdk.Duration.minutes(60),
+            auth_flows=cognito.AuthFlow(
```
greglucas, line 59: I'm not following what all of this does, but anything with admin always screams out to me. Should we only be using the srp workflow?

https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-authentication-flow.html#amazon-cognito-user-pools-admin-authentication-flow

sds_data_manager/sds_cognito_stack.py
```diff
@@ -0,0 +1,131 @@
...
+                admin_user_password=True, user_password=True, user_srp=True, custom=True
+            ),
+            prevent_user_existence_errors=True,
+        )
+
+        # Add aunique domain name where users can sign up / reset passwords
```
greglucas, line 65:
```suggestion
        # Add a unique domain name where users can sign up / reset passwords
```

sds_data_manager/sds_cognito_stack.py
```diff
@@ -0,0 +1,131 @@
...
+        # Users will be able to reset their passwords at:
+        # https://sds-login-{sds_id}.auth.us-west-2.amazoncognito.com/login?client_id={}&redirect_uri=https://example.com&response_type=code
```
greglucas, line 67:
```suggestion
        # https://sds-login-{sds_id}.auth.us-west-2.amazoncognito.com/login?client_id={}&redirect_uri=https://example.com&response_type=code
```
Should we get a unique domain name to use here? Each user/stack could get deployed to subdomains to keep them separate:
`stack1.imap-sdc.com`

sds_data_manager/sds_cognito_stack.py
```diff
@@ -0,0 +1,131 @@
...
+                force_alias_creation=False,
+                user_attributes=[
+                    cognito.CfnUserPoolUser.AttributeTypeProperty(
+                        name="email", value=initial_user_context
+                    )
+                ],
+                username=initial_user_context,
+            )
+
+        ########### LAMBDA
```
greglucas, line 90:
```suggestion
```

sds_data_manager/sds_data_manager_stack.py
```diff
@@ -15,7 +15,13 @@
 
 class SdsDataManagerStack(Stack):
     def __init__(
-        self, scope: Construct, construct_id: str, sds_id: str, **kwargs
+        self,
+        scope: Construct,
+        construct_id: str,
+        sds_id: str,
+        userpool_id: str = "",
+        app_client_id: str = "",
```
greglucas:
```suggestion
        userpool_id: str
        app_client_id: str,
```
Can the ID be an empty string? Perhaps push the validation up a higher level and make this an argument without a default.

tests/unit/test_sds_cognito_validity.py
```diff
@@ -0,0 +1,16 @@
...
+from sds_data_manager.sds_cognito_stack import SdsCognitoStack
+
+
+# This test just ensures the stack is able to be created
+# Does not currently check the products that were created
+def test_sds_cognito_validity():
+    app = core.App(context={"SDSID": "unit-testing"})
+    sds_id = "".join([random.choice(string.ascii_lowercase) for i in range(8)])
+    stack = SdsCognitoStack(app, f"sds-cognito-{sds_id}", sds_id)
+    assertions.Template.from_stack(stack)
```
greglucas, line 16: Can you add a test to actually assert what you want to verify for the Cognito service? I tried to put an example in #42 for how to do that.

Going forward we want to add robust tests for every infrastructure piece we are adding.

greglucas (commented):
I'm still not convinced we actually want to handle Cognito like this, I think we should be doing some architecture diagrams with this to demonstrate what we want before jumping into code. It seems very complicated for what should be a pretty stationary authentication feature we're adding. i.e. I view this as being basically standalone and deployed once/twice, perhaps a separate App entirely.

If others like this approach that is fine though and I'm happy to be let this go in :)

app.py
bryan-harter: If we do put the cognito in a separate stack, how would we tell the data-management stack what the userpool/app_clients are? Would they be hard-coded?

sds_data_manager/lambda_code/SDSCode/cognito_signup_message.py
bryan-harter: sms is not enabled by default

sds_data_manager/lambda_code/SDSCode/cognito_signup_message.py
bryan-harter: We don't currently have anywhere else to redirect to, and unfortunately the sign-up URL requires a redirect_uri

sds_data_manager/lambda_code/SDSCode/cognito_signup_message.py
bryan-harter: The custom message contains the sign-up link, but the default message doesn't

sds_data_manager/lambda_code/SDSCode/upload_api.py
bryan-harter: Yeah sorry I need to re-format this section, I'm not sure why it got all weird

sds_data_manager/sds_cognito_stack.py
bryan-harter: I would say yes definitely, but Route53 stuff is probably another ticket

bryan-harter:
pre-commit.ci autofix

sds_data_manager/sds_cognito_stack.py
```diff
@@ -0,0 +1,131 @@
...
+        signup_lambda.apply_removal_policy(cdk.RemovalPolicy.DESTROY)
+        # Adding Cognito Permissions
+        signup_lambda.add_to_role_policy(cognito_admin_policy)
+        userpool.add_trigger(cognito.UserPoolOperation.CUSTOM_MESSAGE, signup_lambda)
+        # Add attributes to the class for other stacks to reference
+        self.userpool_id = userpool.user_pool_id
+        self.app_client_id = command_line_client.user_pool_client_id
+
+        ########### OUTPUTS
+        cdk.CfnOutput(self, "COGNITO_USERPOOL_ID", value=self.userpool_id)
```
laspsandoval, line 122: I've been told before to try to not use CloudFormation directory since it is better to use higher level constructs that provide more abstraction, better semantics, and sensible defaults. https://bobbyhadz.com/blog/cfn-resources-aws-cdk. Just wondering what your thoughts are on using another method (if one is possible)?

bryan-harter: As in, we shouldn't use the functions with the "Cfn" prefix on the CDK? I generally agree, though as far as I know there is no other way to output variables from a stack

laspsandoval: Yea. I did a brief Google search and you are right. I think there isn't another way.

bryan-harter:
Its been way too long since this review, I'm gonna close it and redo it to fit with the current version of the code base

PR #41, opened 2023-03-01
Cognito Token Verification 

bryan-harter:
This is the pull request to get the APIs to verify provided cognito tokens


bryan-harter:
pre-commit.ci autofix

bryan-harter:
pre-commit.ci autofix

bryan-harter:
pre-commit.ci autofix

greglucas (changes requested):
Before going in-depth on the review, I'd like to discuss the architecture. I don't think this is what we want. I'd much prefer an APIGateway and letting that handle the auth. This seems much more complicated than it needs to be and prone to complications.

Our current API responses should be extremely fast because we are returning s3 pre-signed URLs. So the slow part is hitting those pre-signed urls after the response and downloading that data, not the initial responses.

bryan-harter:
Fair enough, I generally prefer adding the complications into the python code rather than complicating the architecture.  But we can make the switch to API Gateway, I'll make a new ticket and close this one.

PR #44, opened 2023-03-03
MNT: Ignore deprecation warnings

greglucas:
Ignore deprecation warnings from `importlib` and `jsii` that we can't control. I put in a PR to `jsii` to fix it upstream https://github.com/aws/jsii/pull/3986, but until that is fixed, just ignore them for now.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,9 +31,8 @@ classifiers = [
 python = ">=3.9,<4"
 
 [tool.poetry.group.cdk-install.dependencies]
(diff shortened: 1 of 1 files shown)


laspsandoval (approved):

greglucas:
The upstream PR has already been merged, so that was much faster than I expected. I don't know when they'll push a new release, but we could pin to that as well if it is soon.

tech3371 (approved):

PR #50, opened 2023-04-03
1 create a spot to store configjson in a centralized location on aws

### LINKED ISSUE #1: Create a spot to store config.json in a centralized location on AWS
I was thinking a tiny dynamoDB table that all services can read, but if we already have an opensearch cluster, maybe theres a spot on opensearch where we can put these config files?

sdhoyt:
# Change Summary  

## Overview
- Uploading config.json file to S3 bucket on stack creation
- Retrieving config.json from S3 bucket in indexer lambda 

## Updated Files
- sds_data_manager_stack.py
   - Added S3 bucket for config.json
   - added method to upload new config folder containing the config.json to the bucket on stack creation
   - added S3 read policy to indexer lambda
- indexer.py
   - Updated load_allowed_filenames method to pull config.json from S3 bucket
- upload_api.py
   - Updated load_allowed_filenames method to pull config.json from S3 bucket


tests/infrastructure/test_data_manager_stack.py
```diff
@@ -8,7 +8,7 @@ def test_sds_data_manager_stack(app, sds_id):
     stack = SdsDataManagerStack(app, stack_name, sds_id)
     template = Template.from_stack(stack)
 
-    template.resource_count_is("AWS::S3::Bucket", 1)
+    template.resource_count_is("AWS::S3::Bucket", 2)
```
greglucas, line 11: Can you add another `has_resource_properties()` test below confirming the bucket name of your new bucket?

sds_data_manager/sds_data_manager_stack.py
```diff
@@ -31,6 +32,32 @@ def __init__(
...
+
+        s3_deploy.BucketDeployment(
+            self,
+            "DeployConfig",
+            sources=[
+                s3_deploy.Source.asset(
+                    os.path.join(
+                        os.path.dirname(os.path.realpath(__file__)),
+                        "config",
+                    )
```
greglucas: We should use pathlib to make our lives easier if possible: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module
`pyupgrade` may even handle this for you if you want to try adding that to the project too: https://github.com/asottile/pyupgrade

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,16 +21,12 @@
...
-    # For example, a dynamodb table or a section in opensearch
-    current_dir = os.path.dirname(__file__)
-    config_file = os.path.join(current_dir, "config.json")
-
-    with open(config_file) as f:
-        data = json.load(f)
-    return data
+    # get the config file from the S3 bucket
+    config_object = s3.get_object(
+        Bucket=f'sds-config-{os.environ["SDSID"]}', Key="config.json"
```
greglucas: What about passing the entire bucket name as the environment variable?
`os.environ["SDS_CONFIG_BUCKET"]` or something similar, so you don't need to keep `sds-config-` hard-coded in both places (in this lambda and below in the CDK stack).

greglucas (approved):
Looks good, just a few minor suggestions.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,16 +21,12 @@
...
-    # For example, a dynamodb table or a section in opensearch
-    current_dir = os.path.dirname(__file__)
-    config_file = os.path.join(current_dir, "config.json")
-
-    with open(config_file) as f:
-        data = json.load(f)
-    return data
+    # get the config file from the S3 bucket
+    config_object = s3.get_object(
+        Bucket=os.environ['S3_CONFIG_BUCKET_NAME'], Key="config.json"
```
tech3371: Are we storing multiple `config.json` files for different missions? or do we keep one config file and have information of different missions in there?

If first case, then we may want to change `Key="config.json"` to use variable otherwise, ignore my comment

tech3371 (commented):
looks good to me

tests/infrastructure/test_data_manager_stack.py
sdhoyt: Added another has_resource_properties test for the new bucket

sds_data_manager/sds_data_manager_stack.py
sdhoyt: Updated to use pathlib. I didn't want to try adding the pyupgrade now since the pre-commit is still failing

sds_data_manager/lambda_code/SDSCode/indexer.py
sdhoyt: Yeah, that would have been smart to do. Updated to pass in the bucket name as an environment variable rather than the sdsid

sds_data_manager/lambda_code/SDSCode/indexer.py
sdhoyt: My understanding is all file naming patterns needed in the deployment would be in the config.json. I don't think there would be a reason for someone to want to have multiple, but I might be wrong.

Do @greglucas or @bryan-harter have any thoughts on this?

greglucas: I have no major preference. I think it is fine to keep it like this for now and then if we need to allow more later we should be able to update it here too I think?

sds_data_manager/sds_data_manager_stack.py
greglucas: Sounds good, I agree with holding off for now. But, you should try and get pre-commit passing here still. It looks like running black again will possibly fix it looking at the pre-commit error log.

You should be able to install pre-commit locally so it will fix those automatically for you as well if you want.

sdhoyt: Ah ok, I didn't realize that was working now. I'll fix those issues

greglucas: You can try adding a message `pre-commit.ci autofix` as a comment on this PR for it to do the work for you too if it is an easy fix.

pyproject.toml
```diff
@@ -29,6 +29,7 @@ classifiers = [
 
 [tool.poetry.dependencies]
 python = ">=3.9,<4"
+pathlib = "^1.0.1"
```
greglucas:
```suggestion
```
pathlib is builtin, so we don't want an external dependency.

sdhoyt: removed path lib from dependencies

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,16 +20,12 @@
...
-    # can be changed on the fly.
-    # For example, a dynamodb table or a section in opensearch
-    current_dir = os.path.dirname(__file__)
-    config_file = os.path.join(current_dir, "config.json")
-
-    with open(config_file) as f:
-        data = json.load(f)
-    return data
+    # get the config file from the S3 bucket
+    config_object = s3.get_object(
```
maxinelasp, line 24: Will this fail if the configuration isn't in S3? Is that ever expected to happen, or does the step below which deploys config.json always run first?

sds_data_manager/sds_data_manager_stack.py
```diff
@@ -31,6 +32,29 @@ def __init__(
...
+            auto_delete_objects=True,
+            block_public_access=s3.BlockPublicAccess.BLOCK_ALL,
+        )
+
+        s3_deploy.BucketDeployment(
+            self,
+            "DeployConfig",
+            sources=[
+                s3_deploy.Source.asset(
+                    str(pathlib.Path(__file__).parent.joinpath("config").resolve())
```
maxinelasp, line 67: Does this upload everything in the `config` folder, or only things named `config.json`? A comment about that might be good here, I see there were some comments about similar things, so being super clear would help future devs.

maxinelasp (approved):

sds_data_manager/lambda_code/SDSCode/indexer.py
sdhoyt: Yeah, this will fail if the file isn't there, but the file gets pulled into the S3 bucket during the stack creation. Although I'm now realizing that since the S3 BucketDeployment method that uploads the file only takes directories as inputs, it wouldn't error if the config.json isn't in the config folder. Would it be considered bad practice to have a check for whether config.json exists and some error handling in the stack init?

sds_data_manager/sds_data_manager_stack.py
sdhoyt: It uploads any file in `config`. I will add a comment to make that clear.

sds_data_manager/lambda_code/SDSCode/indexer.py
maxinelasp: You could check for an empty bucket in the stack check... if the file is required/expected, I think having it fail on deploy or just after deploy is the right way to go. So adding some handling seems reasonable to me, although I would keep it lightweight.

sdhoyt: It seems like the stack runs through all the code before actually creating any of the resources, so I couldn't check if the bucket is empty in the stack. I'm now checking if that config folder contains a config.json file and raising an error if it doesn't.

tech3371 (approved):

laspsandoval (approved):
Looks good to me.

PR #52, opened 2023-04-17
Improve stack creation testing

sdhoyt:
- Added 49 unit tests for sds_data_manager stack. (50/50 tests currently passing)
- set app and sds_id pytest fixtures in conftest to module scope so the template fixture in the sds_data_manager stack test is only run once for the entire module.
- added domain_name input and appending sds_id for opensearch instance.


tests/infrastructure/test_data_manager_stack.py
```diff
@@ -1,23 +1,29 @@
-from aws_cdk.assertions import Template
+import pytest
+from aws_cdk.assertions import Match, Template
 
 from sds_data_manager.sds_data_manager_stack import SdsDataManagerStack
 
 
-def test_sds_data_manager_stack(app, sds_id):
+@pytest.fixture(scope="module")
+def template(app, sds_id):
```
greglucas, line 8: This is a really nice way to do this!

tests/infrastructure/test_data_manager_stack.py
```diff
@@ -1,23 +1,29 @@
...
     stack_name = f"stack-{sds_id}"
     stack = SdsDataManagerStack(app, stack_name, sds_id)
     template = Template.from_stack(stack)
+    return template
 
+
+def test_s3_data_bucket_resource_count(template):
     template.resource_count_is("AWS::S3::Bucket", 1)
-    # Delete and update are outside of the Properties section
```
greglucas, line 12: I thought we had two s3 buckets, but I guess you haven't merged your other PR yet. So you'll likely need to update this again once that other PR gets merged. We should try and get this in sooner rather than later so you aren't having to do all of the rebasing/updating.

tests/infrastructure/test_data_manager_stack.py
```diff
@@ -26,3 +32,824 @@ def test_sds_data_manager_stack(app, sds_id):
...
+    template.has_resource(
+        "AWS::SecretsManager::Secret",
+        {
+            "DeletionPolicy": "Delete",
+            "UpdateReplacePolicy": "Delete",
+        },
+    )
+
+
+def test_opensearch_domain_resource_count(template):
```
greglucas, line 250: I guess this is a personal preference for how much to put into each individual test. I probably would have went for `def test_opensearch_domain(template, sds_id)` and combined the count and properties into a single test... This is fine though and should give more informative failure messages.

tests/infrastructure/test_data_manager_stack.py
```diff
@@ -26,3 +32,824 @@ def test_sds_data_manager_stack(app, sds_id):
...
+        },
+    )
+    template.has_resource_properties("AWS::Logs::LogGroup", {"RetentionInDays": 30})
+
+
+def test_iam_roles_resource_count(template):
+    template.resource_count_is("AWS::IAM::Role", 7)
+
+
+def test_s3_auto_delete_iam_role_resource_properties(template):
```
greglucas: The next 7 tests are all asserting the same thing I think. "Does the entire template have a resource with this text". I'm not sure if you can do a loop over all 7 roles to get the nested resource policy? Or is there actually only one resource policy that is the shared by all 7 users, in which case we could delete the following 6 tests.

tests/infrastructure/test_data_manager_stack.py
```diff
@@ -26,3 +32,824 @@ def test_sds_data_manager_stack(app, sds_id):
...
+def test_aws_lambda_function_resource_properties(template):
+    template.has_resource_properties(
+        "AWS::Lambda::Function",
+        props={
+            "Handler": "index.handler",
+            "Runtime": "nodejs14.x",
+            "Timeout": 120,
+            "Role": {
+                "Fn::GetAtt": [
+                    "AWS679f53fac002430cb0da5b7982bd2287ServiceRoleC1EA0FF2",
```
greglucas: Do we want to assert this specific hash role, or just leave it as a `Match.Any()` / remove the role check altogether?

tests/infrastructure/test_data_manager_stack.py
```diff
@@ -26,3 +32,824 @@ def test_sds_data_manager_stack(app, sds_id):
...
+    template.has_resource_properties(
+        "AWS::Lambda::Function",
+        props={
+            "Handler": "__entrypoint__.handler",
+            "Runtime": "nodejs14.x",
+            "Timeout": 900,
+            "MemorySize": 128,
+            "Role": {
+                "Fn::GetAtt": [
+                    "CustomS3AutoDeleteObjectsCustomResourceProviderRole3B1BD092",
```
greglucas: Same here, do we want this specific role checked or is it an auto-generated hash we aren't too concerned with?

greglucas (commented):
Wow, this is fantastic Sean! Really nice job adding all of this. I know tests can be tedious, so thank you for taking the time to invest in this upfront here!

I just have some pretty minor comments inline to consider, but nothing blocking at all.

greglucas:
Also, if you're feeling really ambitious we could add an explicit coverage report. See here for an example:
https://github.com/SWxTREC/pymsis/commit/efba2abf2a0498cb5b259f6b6df541566c0534d7

https://app.codecov.io/gh/SWxTREC/pymsis

tests/infrastructure/test_data_manager_stack.py
sdhoyt: This is one thing I wasn't sure about. In the stack's json, these showed as 7 separate resources, but they're all exactly the same. I can condense this down to one function and a loop if you think that would make more sense.

tests/infrastructure/test_data_manager_stack.py
sdhoyt: Sounds good. I think it's good practice to have one test per function and to have very descriptive names so it's clear what the issue is if one fails.

tests/infrastructure/test_data_manager_stack.py
sdhoyt: Agreed. I only added a couple tests for the s3 config PR so this shouldn't be a big deal to resolve merge conflicts

tests/infrastructure/test_data_manager_stack.py
sdhoyt: Thanks!

tests/infrastructure/test_data_manager_stack.py
sdhoyt: I'll remove the Role check

tests/infrastructure/test_data_manager_stack.py
sdhoyt: Maybe I'll change this to Match.string_like_regexp(CustomS3AutoDeleteObjectsCustomResourceProviderRole*)? It might be good to check that there's a role that's being created for the S3 deletion lambda? But that role is auto-generated.

tests/infrastructure/test_data_manager_stack.py
greglucas: Interesting that they are seven separate resources... I'm not sure how we could capture that in these tests, so I'm not sure there is anything else we can do here unless you can track something down for seeing if these are somehow nested or dependent on something else that you could check. But, basically I am still reading this as "in the entire template, does the string 'hello' exist", and even if it is repeated 7 times 'hello, hello, hello' that will be true every time and not check that each new 'hello' is unique.

I'm indifferent on how to handle this then. I can see reasons to do it both ways. Leaving the 7 tests in so that if something gets updated in the future we can easily change/adapt it, but at the same time, it is kind of odd to duplicate the same "check" 7 times which will give us 7 failures if one thing changes.

tests/infrastructure/test_data_manager_stack.py
greglucas: :+1: I like that idea. Maybe the same above `Match.string_like_regexp("*ServiceRole*")` instead of `.Any()` too?

maxinelasp (approved):

tests/infrastructure/test_data_manager_stack.py
sdhoyt: That's a good idea. I'll use string_like_regexp() for the other one

tests/infrastructure/test_data_manager_stack.py
sdhoyt: My thought was that even though repeating is useless when running the test, it at least makes it explicit to people using the file that there are these seven tests and the test name indicates what each of the seven roles are named even though the tests can't be mapped to a specific one since all the key, values are the same. I was hoping there was a way to test the role name against an expected string, but I haven't found a way to do that. I'll leave them as separate tests for now, but I don't have a strong opinion on this and am kind of torn myself, so I can change this if others feel differently.

I did just add properties to the lambda functions that say which role is attached to each, so at least there is a mapping of Role name and Lambda function

tests/infrastructure/test_data_manager_stack.py
```diff
@@ -26,3 +32,850 @@ def test_sds_data_manager_stack(app, sds_id):
...
+        "Custom::OpenSearchAccessPolicy",
+        {
+            "ServiceToken": {"Fn::GetAtt": [Match.string_like_regexp("AWS*"), "Arn"]},
+            "Create": {
+                "Fn::Join": [
+                    "",
+                    [
+                        '{"action":"updateDomainConfig","service":"OpenSearch","parameters":{"DomainName":"',
+                        {"Ref": Match.string_like_regexp("SDSMetadataDomain*")},
+                        '","AccessPolicies":"{\\"Statement\\":[{\\"Action\\":\\"es:*\\",\\"Effect\\":\\"Allow\\",\\"Principal\\":{\\"AWS\\":\\"*\\"},\\"Resource\\":\\"',
```
greglucas, line 295: I just saw this serialized JSON method, would that help for things like this with all the backslashes? https://docs.aws.amazon.com/cdk/api/v1/python/aws_cdk.assertions/README.html#serialized-json

tests/infrastructure/test_data_manager_stack.py
greglucas: I would if we could do something like `def test_make_sure_7_roles_created()` and use the `findXxx()` (with the properties) method and assert against the size of the returned set.

> Beyond assertions, the module provides APIs to retrieve matching resources. The findResources() API is complementary to the hasResource() API, except, instead of asserting its presence, it returns the set of matching resources.

greglucas (approved):
Looks like codecov is moderately happy. We should look at improving that in a follow-up PR. I wouldn't look at adding more content to this one because it is already quite large :)

tests/infrastructure/test_data_manager_stack.py
sdhoyt: Sounds good! I'll give that a shot

tests/infrastructure/test_data_manager_stack.py
sdhoyt: So far I haven't been able to use serialized_json() to get this to pass. The way this is broken up into serialized and deserialized json makes it so some serialized parts aren't valid json on their own, so I tried parsing this all into a dictionary and wrapping that with the serialized_json(), so I'm thinking I'm either missing something when I covert this all to a dictionary or because of the weird way this is broken up in the stack's json, the serialized_json() isn't going to work. I was trying to find why this is formatted like this (since other policy properties aren't) and haven't been able to find anything.

greglucas: OK, I wouldn't spend too much time on this. I was more curious if there was anything that stuck out to you as a way to make it look a little nicer. It is just a test and we can reformat it later if someone comes along with a better understanding of the differences.

PR #58, opened 2023-05-08
Update issue templates to include spike

maxinelasp:


laspsandoval (approved):
Nice. I've never used this before. Looks like it will be very helpful.

.github/ISSUE_TEMPLATE/spike-template.md
```diff
@@ -0,0 +1,58 @@
+---
+name: Spike Template
+about: Template for issues focused on research (spikes)
+title: 'SPIKE - '
+labels: Spike
+assignees: ''
+
+---
+
+## Description
```
greglucas, line 10:
```suggestion
## Description

```
Extra linebreak after the headers.
https://www.markdownguide.org/basic-syntax/#heading-best-practices

.github/ISSUE_TEMPLATE/spike-template.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Nice to have or Goal Requirements
+- Nice to have 1
+- Nice to have 2
+
+## Additional notes
+Any additional notes on the issue, if you have them. 
+
+## Related tickets
+- #123
```
greglucas: Do you know if this will always automatically reference issue "123" once we get that many? I'm not sure the best way to go about not linking something but still giving an example of what you want, maybe just removing the "#"?
```suggestion
- 123 
```

greglucas (approved):
Nice, looks great!

One thing I didn't think about is that this is just a template for the initial "ISSUE" creation and not the findings. So, one question that comes to mind is how we want to handle the findings portion. Do you want to edit the main body of the issue template with new information, or do we want to just follow the template in the issue body, but put the findings as comments in the thread?

.github/ISSUE_TEMPLATE/spike-template.md
maxinelasp: Yeah, good point. I'll change it.

maxinelasp:
> So, one question that comes to mind is how we want to handle the findings portion. Do you want to edit the main body of the issue template with new information, or do we want to just follow the template in the issue body, but put the findings as comments in the thread?

I was imagining that the template for the reply is always part of the issue, and then once the person is ready to reply, they can copy the template out of the issue and into a comment, and update that.

PR #69, opened 2023-06-02
24 mock opensearch for unit testing

### LINKED ISSUE #24: Mock opensearch for unit testing
From @greglucas [comment](https://github.com/IMAP-Science-Operations-Center/SDS-in-a-box/pull/21#discussion_r1102033196) in PR #21 

Look into mocking opensearch for unit tests

sdhoyt:
# Change Summary

## Overview
- Added OpenSearch mocking for opensearch_utils.client.py unit tests.
- Updated OpenSearch client initialization in client.py to work better with mocking library.
- Skipped one test in test_client due to needed changes to openmock
- Updated a function in query.py to output a dictionary instead of a json string.

## New Dependencies
- [openmock](https://github.com/matthewdeanmartin/openmock)

## New Files
N/A

## Updated Files
- client.py
   - Updated the opensearchpy's OpenSearch class initialization to default to None for authorization which is needed for mocking.
   - Change to only import the library and use the full opensearchpy.OpenSearch call which is needed for the mock patching.
- query.py 
   - Updated query_dsl method to output a dictionary instead of a json string 
   - Updated __repr__ to output the json string
- queries.py
   - updated a log to account for query_dsl() now outputting a dictionary instead of a string (query_dsl() output used in the opensearchpy search function which works with both dictionaries and json strings, but I think it's better to have this function return the dictionary.)
- test_client.py
   - added opensearch mocking decorator
   - skipping test_send_document_delete which is failing because the exists function mock in openmock isn't behaving as expected. I have a fix for this and will open another pull request for openmock.
- test_query.py
   - updated to account for query_dsl() now outputting a dictionary instead of a string.

## Testing
- Apart from the notes above for test_client.py, there are minor inconsistencies between opensearchpy returns and openmock return in OpenSearch fields that we don't use. I manually added/deleted these fields to get the tests working since they aren't used and shouldn't affect what we're trying to test. I added comments in the test explaining it when it was needed.
- A solution for mocking opensearch is still needed for the lambdas(queries.py and indexer.py). The opensearch client is created in the lambdas and the state of that mocked object can't be accessed in the test. I'm not sure if maybe we should add a return for that object when testing or if there's a better solution.


sds_data_manager/lambda_code/SDSCode/queries.py
```diff
@@ -32,7 +32,7 @@ def lambda_handler(event, context):
     # create the opensearch query from the API parameters
     query = Query(event["queryStringParameters"])
     client = _create_open_search_client()
-    logger.info("Query: " + query.query_dsl())
+    logger.info("Query: " + str(query.query_dsl()))
```
greglucas, line 35:
```suggestion
    logger.info("Query: %s", str(query.query_dsl()))
```
Rather than formatting our own statements, it might be beneficial to let the printf format it for us, especially if we start logging longer statements.

(Just noticed this here, so it can be a follow-up PR to address these)

tests/unit/opensearch_utils/test_client.py
```diff
@@ -14,47 +11,19 @@
 from sds_data_manager.lambda_code.SDSCode.opensearch_utils.query import Query
 
 
-@pytest.mark.network()
+# @pytest.mark.network()
```
greglucas:
```suggestion
```
Just remove it if we aren't using it

tests/unit/opensearch_utils/test_client.py
```diff
@@ -315,159 +322,168 @@ def test_scroll_search(self):
...
+        # openmock gives a different _score value than opensearch. This field is not
+        # being used and does not affect the test, so it is being removed from the
+        # output. The _version field is output by openmock, but not by the real
+        # opensearch client, so that is being removed from the output as well and does
+        # not affect the test
+        for i in range(0, len(search_out)):
+            search_out[i].pop("_score")
+            search_out[i].pop("_version")
+        for i in range(0, len(search_true)):
+            search_true[i].pop("_score")
```
greglucas:
```suggestion
        for search in search_out:
            search.pop("_score")
            search.pop("_version")
        for search in search_true:
            search.pop("_score")
```

greglucas (approved):
Nice! Mocking is so much nicer :) Just a few minor suggestions on the code inline.

One thought that I had on the test you weren't able to update is if you can mokeypatch the mock library and add the method you need at runtime? Not ideal, but if it is only that one piece missing, then we can add it for now and if your patch gets picked up then remove it.

maxinelasp (approved):
Looks great! I agree with Greg's suggestions, but don't have anything else to add.

sdhoyt:
> One thought that I had on the test you weren't able to update is if you can mokeypatch the mock library and add the method you need at runtime? Not ideal, but if it is only that one piece missing, then we can add it for now and if your patch gets picked up then remove it.

I just had a really quick turnaround with my new updates being merged into openmock, so no need to skip this test anymore. Thanks for the suggestion though!

sds_data_manager/lambda_code/SDSCode/queries.py
sdhoyt: Agreed. I'll create a new issue for this

tests/unit/opensearch_utils/test_client.py
sdhoyt: Ah thanks, I missed that. Removed

tests/unit/opensearch_utils/test_client.py
sdhoyt: Updated to the suggested change

tests/unit/opensearch_utils/test_client.py
```diff
@@ -122,6 +89,13 @@ def test_send_document_create(self):
         self.client.send_document(document)
         document_out = self.client.get_document(document)
 
+        # openmock does not output the _primary_term and _seq_no. These are two
+        # opensearch fields that we're not using. Adding these fields to the output to
+        #  highlight that these do exist in a real opensearch document, but are missing
+        # from openmock. These missing fields do not impact the test
```
tech3371, line 95: Thanks for comments like this and PR overview. Really helpful to understand what's being changed.

tests/unit/opensearch_utils/test_client.py
```diff
@@ -143,6 +117,9 @@ def test_send_document_delete(self):
         exists_true = False
 
         exists_confirm = self.client.document_exists(document)
+        print(self.client.get_document(document))
```
tech3371: Did you mean to keep these print?

tests/unit/opensearch_utils/test_client.py
```diff
@@ -293,11 +293,18 @@ def test_search(self):
...
-        # need to give opensearch a second to receive the payload before searching
-        time.sleep(1)
+
         ## Act ##
         search_out = self.client.search(query, self.index)
 
+        # openmock gives a different _score value. This field is not being used and does
+        # not affect the test, so it is being removed from the output. The _version
+        # field is output by openmock, but not by the real opensearch client, so that
+        # is being removed from the output as well and does not affect the test
```
tech3371, line 301: good to know!

tech3371 (commented):
It looks good to me besides one question regarding the print statement.

tests/unit/opensearch_utils/test_client.py
sdhoyt: Thanks, good catch! I didn't mean to keep those. Removed

tech3371 (approved):
Looks good!

Issue #43, opened 2023-03-03
Move to API Gateway 


bryan-harter:
Spin up API Gateway architecture in the CDK rather than Lambda URLs

laspsandoval:
An API Gateway is beneficial because spinning up an API Gateway architecture in the CDK rather than using Lambda URLs directly gives you better control over your API's routing, security, and other features:

1. Efficient routing and versioning: With API Gateway, you can route different URL paths or request types to different Lambda functions. You can also manage different versions of your API more effectively.
2. Throttling and scaling: API Gateway provides built-in throttling to protect your back-end services. It also helps manage the scaling of your services.
3. Security: API Gateway provides several features that allow you to create more secure APIs, such as AWS IAM roles and policies, Amazon Cognito user pools, AWS WAF (a web application firewall), and resource policies.
4. Monitoring and troubleshooting: API Gateway integrates with Amazon CloudWatch and AWS X-Ray allowing you to monitor API usage and troubleshoot potential issues.

laspsandoval:
To transfer the domain:
https://repost.aws/knowledge-center/account-transfer-route-53#:~:text=You%20can%20transfer%20a%20domain,domain%20by%20contacting%20AWS%20Support.&text=Create%20a%20new%20hosted%20zone%20in%20an%20AWS%20account%20that%20you%20own.

laspsandoval:
Here are the steps to request a new SSL certificate in your new account:

Navigate to AWS Certificate Manager in your AWS Console.
Click on "Request a certificate".
Select "Request a public certificate" and click "Request a certificate".
On the next screen, in the "Add domain names" section, add your domain name. If you plan to have subdomains and you want the certificate to cover them, you can use a wildcard by entering "*.imap-mission.com". If you don't plan to have subdomains, just enter "imap-mission.com".
Click "Next".
On the next page, select "DNS validation".
Click "Review", review your settings, and then click "Confirm and request".
On the validation page, you'll see a CNAME record that you need to add to your Route 53 DNS records for imap-mission.com. This proves to ACM that you have control over the domain.
After you add the CNAME record to Route 53, the Certificate Manager will automatically validate it and issue the SSL certificate, which you can use in your API Gateway setup.

Once the certificate is issued, you'll be able to select it in API Gateway custom domain setup in the same account.

Remember, SSL certificates are region-specific, so make sure to request the certificate in the same region where you're setting up your API Gateway.

[...]

PR #79, opened 2023-06-29
Move to API Gateway #43

laspsandoval:
# Change Summary

## Overview
- Modularized code by creating OpenSearch stack
- Created API Gateway (using api_gateway_stack.py) that uses subdomains (for example https://query.imap-mission.com)
- Created stack that obtains hosted_zone (as seen in Route 53) and certificate (Certificate Manager) (using domain_stack.py)
- Made minor changes to lambda to get them functional with the new setup
- Created stackbuilder.py which organizes our stacks
- Created template (**I will since this via Slack since I can't attach it here**) for users so we don't accidentally deploy to the wrong AWS account
- Updated docstrings and unit tests (**IMPORTANT: I will come back to test_data_manager_stack.py after we create a strategy for test development**)

## New Dependencies
I created a domain imap-mission.com in the dev account and then transferred it using the instructions posted here to my individual account:
https://repost.aws/knowledge-center/account-transfer-route-53#:~:text=You%20can%20transfer%20a%20domain,domain%20by%20contacting%20AWS%20Support.&text=Create%20a%20new%20hosted%20zone%20in%20an%20AWS%20account%20that%20you%20own

**IMPORTANT**: you must create a SSL certificate for the domain.
**IMPORTANT**: the domain servers of the registered domain must match that of the hosted zone.

## New Files
- Created API Gateway (using api_gateway_stack.py) that uses subdomains (for example https://query.imap-mission.com)
- Created stack that obtains hosted_zone (as seen in Route 53) and certificate (Certificate Manager) (using domain_stack.py)
- Created stackbuilder.py which organizes our stacks

## Deleted Files
- Just moved sds-data-manager.py for organization (which made it appear that it was deleted and re-created)

## Updated Files
-Minor updates to lambdas. Changed Secrets Manager so it doesn't use the "unsafe" was of achieving password anymore.

## Testing
**IMPORTANT: I will come back to test_data_manager_stack.py after we create a strategy for test development**
- Updated all relevant unit tests so they would pass.

## Important Instructions
You must test this in dev account since that is where the registered domain currently lives, otherwise you must set use_custom_domain to False in build_sdc.  If you are not deploying this to the dev AWS account then you can still test using the url in (for example APIs > My Query Service > Stages > dev > query > GET)

stacks = build_sdc(app,
                   env=env,
                   sds_id=params["sds_id"],
                   use_custom_domain=False)

- Add this to your Invoke URL: /query?instrument=mag&start_date=20230101
- Note: if you are not deploying this to the dev AWS account then you can still test using 

**To make a download in Postman**
- https://download.imap-mission.com/download?s3_uri=s3://sds-data-dev/imap_l0_sci_glows_20230602_v01.pkts
- And then click on the url

**To upload using Postman**
- first POST (attach file)
https://upload.imap-mission.com/upload?filename=imap_l0_sci_glows_20230602_v01.pkts
- then PUT using returned url (do not attach file)

**Instructions for deployment**
- Put the file template in the same directory as app.py and follow the instructions


laspsandoval:
#79

laspsandoval:
#43

cdk.context.json
```diff
@@ -0,0 +1,6 @@
+{
+  "hosted-zone:account=301233867300:domainName=imap-mission.com:region=us-west-2": {
```
tech3371: one quick comment that i wanted to make. Is that an aws account number? you may not want it there.

laspsandoval: Added this file to .gitignore and removed it from the repository with the command git rm --cached cdk.context.json

sds_data_manager/lambda_code/SDSCode/download_query_api.py
```diff
@@ -33,19 +41,31 @@ def http_response(header_type="text/html", status_code=200, body="Success"):
...
+    Returns
+    -------
+    dict
+        The response from the function which could either be a pre-signed S3 URL in case of successful
+        operation or an error message with corresponding status code in case of failure.
     """
-
     logger.info(event)
 
+    print(event)
```
sdhoyt, line 64: Were these print statements left on purpose?

sds_data_manager/lambda_code/SDSCode/queries.py
```diff
@@ -1,23 +1,51 @@
...
+
+    Note
+    ----
+    This function is currently using hard-coded parameters for the AWS Secrets Manager session.
+    """
     logger.info("OS DOMAIN: " + os.environ["OS_DOMAIN"])
     hosts = [{"host": os.environ["OS_DOMAIN"], "port": int(os.environ["OS_PORT"])}]
-    auth = (os.environ["OS_ADMIN_USERNAME"], os.environ["OS_ADMIN_PASSWORD_LOCATION"])
+    # TODO: remove hard-coded parameters
+    session = boto3.session.Session()
```
sdhoyt, line 42: Thanks for fixing this! For some reason I couldn't get this to work here with the opensearchpy API which is why I did that terrible unwrap to the lambda environment variable.

sds_data_manager/lambda_code/SDSCode/queries.py
```diff
@@ -28,6 +56,19 @@ def _create_open_search_client():
...
+
+    Parameters
+    ----------
+    event : dict
+        The JSON formatted document with the data required for the lambda function to process
+    context : LambdaContext
+        This object provides methods and properties that provide information about the invocation, function,
+        and runtime environment.
+    """
+    print(event)
```
sdhoyt: Did you mean to leave these print statements in?

sds_data_manager/lambda_code/SDSCode/queries.py
```diff
@@ -36,4 +77,14 @@ def lambda_handler(event, context):
     # search the opensearch instance
     search_result = client.search(query, Index(os.environ["OS_INDEX"]))
     logger.info("Query Search Results: " + json.dumps(search_result))
-    return search_result
+
+    # Format the response
+    response = {
```
sdhoyt, line 85: Not saying this should happen in this PR, but would it make sense to create a Response class with different formatted responses that we could instantiate with body messages, so we can reuse these instead creating these dictionaries in the lambdas?

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -102,6 +102,8 @@ def lambda_handler(event, context):
 
     :return: A pre-signed url where users can upload a data file to the SDS.
     """
+    print(event)
```
sdhoyt: just checking again if these print statements were left on purpose.

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -0,0 +1,215 @@
...
+            function_name=f"file-indexer-{sds_id}",
+            entry=str(pathlib.Path(__file__).parent.joinpath("..", "lambda_code").resolve()),
+            index="SDSCode/indexer.py",
+            handler="lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_9,
+            timeout=cdk.Duration.minutes(15),
+            memory_size=1000,
+            environment={
+                "OS_ADMIN_USERNAME": "master-user",
+                "OS_ADMIN_PASSWORD_LOCATION": opensearch.os_secret.secret_value.unsafe_unwrap(),
```
sdhoyt: OS_ADMIN_PASSWORD_LOCATION can be deleted since you're using the secrets manager client in the lambda now

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -0,0 +1,215 @@
...
+            function_name=f"query-api-handler-{sds_id}",
+            entry=str(pathlib.Path(__file__).parent.joinpath("..", "lambda_code").resolve()),
+            index="SDSCode/queries.py",
+            handler="lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_9,
+            timeout=cdk.Duration.minutes(1),
+            memory_size=1000,
+            environment={
+                "OS_ADMIN_USERNAME": "master-user",
+                "OS_ADMIN_PASSWORD_LOCATION": opensearch.os_secret.secret_value.unsafe_unwrap(),
```
sdhoyt: Same comment as above. OS_ADMIN_PASSWORD_LOCATION is no longer needed

cdk.json
```diff
@@ -15,6 +15,16 @@
     ]
   },
   "context": {
```
sdhoyt, line 17: It seems like this changes how we deploy? (I'm getting an error about an invalid bucket name when I do cdk synth: `RuntimeError: Invalid S3 bucket name (value: cdk-hnb659fds-assets-<account_id>-us-west-2`). If so, the readme instructions should be updated.

laspsandoval: Yes. There is a template that I will message you that you put in the same directory as app.py. I will update the README.md.

sds_data_manager/lambda_code/SDSCode/queries.py
laspsandoval: Yes. The reason I left them in is that the only place they are printing is to CloudWatch logs, which is helpful, especially if we decide to use Step Functions.

cdk.json
```diff
@@ -15,6 +15,16 @@
...
+    "deploy-where": "dev",
+    "dev": {
+      "sds_id": "dev",
```
sdhoyt: Is there a way to use a different sds_id besides updating this file? For example, we used to set the SDSID to something unique as part of the deploy command like SDSID=hoyt-testing so our resources wouldn't conflict.

cdk.context.json
greglucas: This unfortunately doesn't get it all. You will need to squash the commits down into one to get rid of the old commits that have this number. Right now I can go back and see the old commit where this number is still present, so it was added then removed, but the _history_ still remains.

app.py
```diff
@@ -1,23 +1,34 @@
...
-    )
-elif SDS_ID == "random":
-    # A random unique ID for this particular instance of the SDS
-    SDS_ID = "".join([random.choice(string.ascii_lowercase) for i in range(8)])
+account = params["account"]
 
+env = Environment(account=account, region=sdc_region)
+print(f"Deploying to account {account} in region {sdc_region}.")
+
+# Deploy Libera SDC resources. This is the default with no CLI context variables set.
```
greglucas:
```suggestion
# Deploy IMAP SDC resources. This is the default with no CLI context variables set.
```

Also, we are using `SDS` in most places instead of `SDC` do we want to change all of the new sdc functions over to sds functions?

cdk.json
greglucas: I think you could change this to what you need in your file so it is distinct from others. Although if you wanted to deploy the stacks twice then you'd need to have two versions of this file I think...

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -1,17 +1,19 @@
+# Standard
```
greglucas, line 1: Do we need these comments in all the files? I think it is pretty standard to have this separation, so I'm not sure we need the comments in every file we are importing from.

sds_data_manager/lambda_code/SDSCode/queries.py
greglucas: Can you increase the log level to INFO instead and use the logger? It would be good to use that for outputting to cloudwatch instead of prints.

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -0,0 +1,89 @@
...
+            Hosted zone used for DNS routing.
+        certificate : acm.ICertificate
+            Used for validating the secure connections to API Gateway.
+        use_custom_domain : bool
+            Determines if a custom domain should be used.
+        """
+        super().__init__(scope, construct_id, env=env, **kwargs)
+
+        # Define subdomains
+        subdomains = lambda_functions.keys()
```
greglucas: I don't think we will want each lambda function to have its own subdomain, but rather have each lambda_function be a route on a single APIGateway?

i.e. rather than many separate APIGateways:
```
lambda1.dev.imap.com
lambda2.dev.imap.com
```

We could use one APIGateway with different routes
```
api.dev.imap.com/lambda1
api.dev.imap.com/lambda2
```

I think this mostly just means moving this loop over subdomains after your ARecord creation below and making the routes within the loop, but reusing the same API you created.

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -0,0 +1,89 @@
...
+            if use_custom_domain:
+                # Define a custom domain
+                custom_domain = apigw.DomainName(self,
+                                                 f'{subdomain}-DomainName-{sds_id}',
+                                                 domain_name=f'{subdomain}.imap-mission.com',
+                                                 certificate=certificate,
+                                                 endpoint_type=apigw.EndpointType.REGIONAL
+                                                 )
+
+                # Route domain to api gateway
```
greglucas: Is this required? We don't use it on SWxTREC, so I'm not sure exactly what it is doing here. I couldn't quite grasp the AWS documentation either, so I'm mostly just curious what is going on.

sds_data_manager/stacks/domain_stack.py
```diff
@@ -0,0 +1,53 @@
...
+                                          iam.ManagedPolicy.from_aws_managed_policy_name("AmazonRoute53FullAccess")
+                                      ])
+
+        if use_custom_domain:
+            self.hosted_zone = route53.HostedZone.from_lookup(self,
+                                                              f'HostedZone-{sds_id}',
+                                                              domain_name='imap-mission.com')
+
+            self.certificate = acm.Certificate(self, f'Certificate-{sds_id}',
+                                               domain_name='*.imap-mission.com',
```
greglucas, line 51:
```suggestion
                                               domain_name='*.{sds_id}.imap-mission.com',
```
Do we want to be able to have a `dev.imap-mission.com` and `prod.imap-mission.com` as a separate entities?

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -0,0 +1,215 @@
+# Standard
+import pathlib
+# Installed
```
greglucas, line 4:
```suggestion

```
I think there should be a blank line between each of the import sections.

tests/unit/test_download_query_api.py
```diff
@@ -30,7 +31,7 @@ def setUp(self):
         )
         # upload a file
         self.s3_filepath = "test-data/science_block_20221116_163611Z_idle.bin"
-        self.local_filepath = f"tests/unit/{self.s3_filepath}"
+        self.local_filepath = f"{self.s3_filepath}"
```
greglucas: Was this wrong before? Otherwise it seems like we shouldn't even need the variable `local_filepath` at all and this could be sent to the upload_file as `self.s3_filepath, self.bucket_name, self.s3_filepath`

tests/unit/test_queries.py
```diff
@@ -18,22 +20,28 @@
 class TestQueries(unittest.TestCase):
     def setUp(self):
         # Opensearch client Params
-        os.environ[
-            "OS_DOMAIN"
-        ] = "search-sds-metadata-uum2vnbdbqbnh7qnbde6t74xim.us-west-2.es.amazonaws.com"
+        # TODO: there has to be a better way
```
greglucas: Nice, I agree with this comment :)

With your new domain work here I would think this could also get routed through a base domain name for simpler access?

tests/unit/test_queries.py
```diff
@@ -65,31 +74,36 @@ def setUp(self):
...
         """tests that the queries lambda correctly returns the search results"""
-        ## Arrange ##
+        # Arrange
         response_true = [
             {
                 "_index": "test_data",
                 "_type": "_doc",
                 "_id": "1",
-                "_score": 0.2876821,
+                "_score": 0.18232156,
```
greglucas: Interesting, these scores keep changing in updated PRs! I wonder if this is to be expected, or if we can reliably count on these at some point.

tests/unit/test_queries.py
```diff
@@ -65,31 +74,36 @@ def setUp(self):
...
 
-        ## Act ##
-        response_out = queries.lambda_handler(event, "")
+        # Act
+        response = queries.lambda_handler(event, "")
+        response_out = json.loads(response['body'])
+
+        for res in response_out:
+            res['_score'] = None
+        for res in response_true:
```
greglucas: You updated this above, but now are overriding it here to be None? So, maybe you can put `None` in the true body above? But, I don't understand why we can't count on a score value in unit tests, so this force-to-None seems a bit suspect to me that we are missing something.

greglucas (commented):
I really like the reorganization here, but we will see what others have to say. I find it easier to follow with the separate stacks.

This was a ton of work, thank you for taking it on!!

A minor nit was all of the noise from the added comments around imports which I don't think are needed, but if others prefer that I'm fine with keeping them too.

sds_data_manager/lambda_code/SDSCode/download_query_api.py
```diff
@@ -1,25 +1,33 @@
+# Standard
 import json
 import logging
 import os
-
+# Installed
```
tech3371, line 6: good notes!

sds_data_manager/lambda_code/SDSCode/download_query_api.py
```diff
@@ -1,25 +1,33 @@
...
+    status_code : int, optional
+        HTTP status code indicating the result of the operation, defaults to 200.
+    body : str, optional
+        The content of the response, defaults to 'Success'.
+
+    Returns
+    -------
+    dict
+        A dictionary containing headers, status code, and body, designed to be returned
+        by a Lambda function as an API response.
```
tech3371, line 32: thank you for updating documentation!

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: I liked it personally.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +22,14 @@
...
 
 def _load_allowed_filenames():
+    """Load the allowed filenames configuration from an S3 bucket.
+
+    Returns
+    -------
+    dict
+        The content of 'config.json', parsed into a Python dictionary.
+    """
+
```
tech3371, line 34: again, thank you for updating these!

sds_data_manager/lambda_code/SDSCode/queries.py
tech3371: Does Laura's work take care of the `TODO` then?

sds_data_manager/lambda_code/SDSCode/queries.py
tech3371: There is a function in `download_query_api.py` called `http_response()`, is that what you are envisioning? If so, yeah, in a future PR, we could refactor that code and reference where appropriate.

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -0,0 +1,215 @@
...
+            entry=str(pathlib.Path(__file__).parent.joinpath("..", "lambda_code").resolve()),
+            index="SDSCode/download_query_api.py",
+            handler="lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_9,
+            timeout=cdk.Duration.seconds(60),
+        )
+        download_query_api.add_to_role_policy(opensearch.opensearch_all_http_permissions)
+        download_query_api.add_to_role_policy(s3_read_policy)
+
+        self.lambda_functions = {
```
tech3371, line 199: What is this used for?

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -0,0 +1,215 @@
...
+            },
+            'download': {
+                'function': download_query_api,
+                'httpMethod': 'GET'
+            },
+            'indexer': {
+                'function': indexer_lambda,
+                'httpMethod': 'POST'
+            }
+            }
```
tech3371, line 216: Can we add this back in from previous implementation?
```
########### OUTPUTS
        # This is a list of the major outputs of the stack
        cdk.CfnOutput(self, "UPLOAD_API_URL", value=upload_api_url.url)
        cdk.CfnOutput(self, "QUERY_API_URL", value=lambda_query_api_function_url.url)
        cdk.CfnOutput(self, "DOWNLOAD_API_URL", value=download_api_url.url)
```
Is that what you were trying to above?

sds_data_manager/utils/stackbuilder.py
```diff
@@ -0,0 +1,44 @@
...
+    ----------
+    scope : App
+    env : Environment
+        Account and region
+    sds_id : str
+        Name suffix for stack
+    use_custom_domain : bool, Optional
+        Build API Gateway using custom domain
+    """
+    open_search = opensearch_stack.OpenSearch(scope, f"OpenSearch-{sds_id}",
```
tech3371, line 30: Should we move open_search, and API gateway in the `sds_data_manager_stack.py` too? Because they are components of SDS data manager.

tests/unit/test_download_query_api.py
tech3371: I think that path is relative to where we run the pytest from. I think when I was developing it, I was running test from Git repo root directory. But if I was running that test from inside `unit/` folder, then we don't need that additional path.

tests/unit/test_indexer.py
```diff
@@ -18,22 +19,25 @@
...
-        ] = "search-sds-metadata-uum2vnbdbqbnh7qnbde6t74xim.us-west-2.es.amazonaws.com"
+        os.environ["OS_DOMAIN"] = \
+            "search-sdsmetadatadomain-dev-i3bnjqingkrphg2crwdcwqabqe.us-west-2.es.amazonaws.com"
         os.environ["OS_PORT"] = "443"
         os.environ["OS_INDEX"] = "test_data"
 
         hosts = [{"host": os.environ["OS_DOMAIN"], "port": os.environ["OS_PORT"]}]
 
-        secret_name = "OpenSearchPassword9643DC3D-uVH94BjrbF9u"
+        secret_name = "sdp-database-creds"
```
tech3371: what is `sdp`?

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -0,0 +1,89 @@
...
+            Used for validating the secure connections to API Gateway.
+        use_custom_domain : bool
+            Determines if a custom domain should be used.
+        """
+        super().__init__(scope, construct_id, env=env, **kwargs)
+
+        # Define subdomains
+        subdomains = lambda_functions.keys()
+
+        for subdomain in subdomains:
```
tech3371: A short description in the code would be helpful. For example, how would API endpoint will look like after these operations. Would it look like (copying @greglucas above comment)

```
lambda1.dev.imap.com
lambda2.dev.imap.com
```
 or
```
api.dev.imap.com/lambda1
api.dev.imap.com/lambda2
```
or
```
imap-mission.com/dev/lambda1
imap-mission.com/dev/lambda2
```

tests/unit/test_queries.py
laspsandoval: _score is not deterministic, meaning it can change based on the state of the Elasticsearch cluster, and therefore it's not something that can be reliably asserted in a unit test.

sds_data_manager/utils/stackbuilder.py
laspsandoval: I think it is better to have as modular of code as possible. So my vote is to leave it as is.

sds_data_manager/stacks/sds_data_manager_stack.py
laspsandoval: We shouldn't need it anymore. I think the only benefit is that it showed the returned urls, but now we are using a purchased domain.

sds_data_manager/stacks/sds_data_manager_stack.py
laspsandoval: This is passed to the api_gateway_stack.

sds_data_manager/stacks/api_gateway_stack.py
laspsandoval: Here we link the custom domain (custom_domain) to the API (api). If we use the default AWS API Gateway URL, we don't need to use it. Does SWXTREC have a custom domain?

tests/unit/test_indexer.py
laspsandoval: That is the name of the database secrets

sds_data_manager/stacks/sds_data_manager_stack.py
tech3371: That is so true!

laspsandoval:
Closing to make into multiple PRs.

PR #101, opened 2023-07-19
Modularize CDK code

laspsandoval:
# Change Summary

## Overview
- Modularized code by creating OpenSearch stack
- Made minor changes to lambda to get them functional with the new setup
- Created stackbuilder.py which organizes our stacks
- Created template for users so we don't accidentally deploy to the wrong AWS account
- Modified README.md
- Updated docstrings

## New Files
- Created stackbuilder.py which organizes our stacks
- Created app_template_dev.py to help us deploy

## Deleted Files
- Just moved sds-data-manager.py for organization (which made it appear that it was deleted and re-created)

## Updated Files
-Minor updates to lambdas. Changed Secrets Manager so it doesn't use the "unsafe" was of achieving password anymore.

## Testing
- Testing will appear in a separate PR.

## Important Instructions

- Add this to your Invoke URL: /query?instrument=mag&start_date=20230101. 

**Instructions for deployment**
- Put the file template in the same directory as app.py and follow the instructions in README.md or app_template_dev.py


app.py
```diff
@@ -1,23 +1,35 @@
...
-import string
+# Installed
+from aws_cdk import App, Environment
+# Local
+from sds_data_manager.utils.stackbuilder import build_sds
 
-import aws_cdk as cdk
+"""
+This app is designed to be the dev and production deployment app. It defaults to a dev deployment 
+via a default `deploy-where` value in cdk.json. To deploy to prod, specify `--context deploy-where=prod`.
```
maxinelasp: I like this! If this isn't a cdk term, might I suggest "env" or "environment" instead of "deploy-where"?

sds_data_manager/utils/stackbuilder.py
```diff
@@ -0,0 +1,35 @@
...
+)
+
+# Local
+from sds_data_manager.stacks import (
+    opensearch_stack,
+    sds_data_manager_stack
+)
+
+
+def build_sds(scope: App, env: Environment,
```
maxinelasp: I really like that there's a separate function for defining the entire stack. I think it reduces confusion a lot!

app_template_dev.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
+"""Template for CDK Application"""
```
maxinelasp, line 2: What is this meant to be used for?

sds_data_manager/lambda_code/SDSCode/download_query_api.py
```diff
@@ -33,19 +43,31 @@ def http_response(header_type="text/html", status_code=200, body="Success"):
...
+    Returns
+    -------
+    dict
+        The response from the function which could either be a pre-signed S3 URL in case of successful
+        operation or an error message with corresponding status code in case of failure.
     """
-
     logger.info(event)
 
+    print(event)
```
maxinelasp: Are these for testing and accidentally left in?

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -0,0 +1,216 @@
+# Standard
```
maxinelasp, line 1: Can I request some docs here to just explain what this does? "data management" is a little vague. Actually in general, a lot of these files could use a 1-2 sentence at the top just explaining what it is and what it's used for.

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -0,0 +1,216 @@
...
+                 env: Environment,
+                 **kwargs) -> None:
+        """SdsDataManagerStack
+
+        Parameters
+        ----------
+        scope : App
+        construct_id : str
+        sds_id: str
+        env : Environment
```
maxinelasp, line 57: You're missing opensearch in the doc strings. In general, it makes sense that this class depends on opensearch, but I'd like to see a note about that here, since I know those dependencies can be tricky. Just something like "this class depends on opensearch, which is built with opensearch_stack.py" would make it clear that there is another class needed before you call this one. I don't expect this stack to get used much outside of stackbuilder.py but similar things have really tripped me up in other projects, lol

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -0,0 +1,216 @@
...
+            versioned=True,
+            removal_policy=RemovalPolicy.DESTROY,
+            auto_delete_objects=True,
+            block_public_access=s3.BlockPublicAccess.BLOCK_ALL,
+        )
+
+        # Upload all files in the config directory to the S3 config bucket.
+        # This directory should contain a config.json file that will
+        # be used for indexing files into the data bucket.
+        s3_deploy.BucketDeployment(
```
maxinelasp, line 99: Would it make sense to make the S3 deployment a separate file as well? Genuinely asking, I'm not sure if there are advantages beyond reusing the stacks.

maxinelasp (commented):
Overall I think this looks great, I added a few comments mostly requesting a bit more documentation!

laspsandoval:
`pre-commit.ci.autofix`

sds_data_manager/stacks/sds_data_manager_stack.py
laspsandoval: @maxinelasp that would make sense. Also how we are doing it here makes sense too. I actually tried that quickly which resulted in a circular dependency (the stacks weren't deploying in the order they were supposed to). And that could have opened a can of worms. Since it didn't really seem to matter either way I left them in.

laspsandoval:
`pre-commit.ci.autofix`

tech3371:
[pre-commit.ci](http://pre-commit.ci/) autofix

sds_data_manager/stacks/sds_data_manager_stack.py
laspsandoval: Good catch.

tech3371:
`pre-commit.ci autofix`

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/download_query_api.py
laspsandoval: Yes. Geez. I thought that I got all of those. Fixed it.

app.py
laspsandoval: That makes sense. Changed it.

app_template_dev.py
laspsandoval: So the idea here is that we don't run the risk of committing account numbers, hard-coded initials, etc. So when doing development work we make a copy of this template, rename it, and deploy from there instead of in app.py.

greglucas:
pre-commit.ci autofix

app_template_dev.py
```diff
@@ -0,0 +1,58 @@
...
+import os
+
+# Installed
+from aws_cdk import App, Environment
+
+# Local
+from sds_data_manager.utils.stackbuilder import build_sds
+
+"""
+    This app is designed for individual developer testing.
```
maxinelasp:
```suggestion
    This app is used for individual developer testing, instead of app.py. When deploying locally, you can update the sds_id to include your initials and distinguish your stack from the main dev stack or other developers.
```

app_template_dev.py
maxinelasp: That makes sense! I added a suggestion to make it more clear in the docs. Might I also suggest that it retrieve the sds_id from cdk.json and have the user change their local copy of that instead of having to remember to update here? Basically just remove the .join and add a note in the docs about changing the deploy name.

sds_data_manager/stacks/sds_data_manager_stack.py
maxinelasp: Sounds good! Thanks for checking that. Those circular dependencies are a real pain!

app.py
```diff
@@ -1,23 +1,32 @@
...
+via a default `env` value in cdk.json. To deploy to prod, specify `--context env=prod`.
+"""
 
-app = cdk.App()
+app = App()
 
-SDS_ID = app.node.try_get_context("SDSID")
+# Grab values from context
+sds_region = app.node.try_get_context("sds-region")
+where_to_deploy = app.node.try_get_context("env")
```
maxinelasp, line 18: Might want to update this variable to match "env"... but also, it's fine if you'd prefer to leave it, I think it still makes sense

sds_data_manager/stacks/sds_data_manager_stack.py
laspsandoval: Done.

maxinelasp (approved):

laspsandoval:
pre-commit.ci autofix

laspsandoval:
pre-commit.ci autofix

tech3371:
I did deploy CDK and tested it. I couldn't test api calls yet. We can test that in the API gateway PR.

I was able to deploy CDK successfully. Indexer lambda works once we removed hard coded region on line 94 in `indexer.py`.

Nice work! @laspsandoval My approval is still valid.

laspsandoval:
pre-commit.ci autofix

PR #102, opened 2023-07-20
API Gateway

laspsandoval:
# Change Summary

## Overview
* Created API Gateway (using api_gateway_stack.py) that uses subdomains
* Created stack that obtains hosted_zone (as seen in Route 53) and certificate (Certificate Manager) (using domain_stack.py)
* Added stacks to stackbuilder.py

## New Dependencies
I created a registered domain (imap-mission.com), hosted zones, and SSL certificate in the dev AWS account. 

## New Files
* api_gateway_stack.py
* domain_stack.py

## Deleted Files
* None

## Updated Files
* stackbuilder.py
* In the last PR I had updated pyproject.toml to include pre-commit and thought it was working. But now poetry is unable to resolve it and I have learned that I can use ruff directly. So I removed it from the pyproject.toml.

## Testing
- Add this to your Invoke URL: /query?instrument=mag&start_date=20230101
- Example: https://api.tlcs-dev.imap-mission.com/query?instrument=mag&start_date=20230101

**To make a download in Postman**
- Example: https://api.tlcs-dev.imap-mission.com/download?s3_uri=s3://sds-data-tlcs-dev/imap_l0_sci_glows_20230602_v02.pkts

**To upload using Postman**
- first POST (attach file)
https://api.tlcs-dev.imap-mission.com/upload?filename=imap_l0_sci_glows_20230602_v01.pkts
- then PUT using returned url (do not attach file)


laspsandoval:
pre-commit.ci autofix

pyproject.toml
```diff
@@ -41,7 +41,6 @@ constructs = ">=10.0.0,<11.0.0"
 [tool.poetry.group.dev.dependencies]
 black = "^23.1.0"
 ruff = "^0.0.253"
-pre-commit = "^3.3.3"
```
greglucas: Can you leave this in there and locally just not install the pre-commit hooks if you don't want them? I think it is good to have the option for someone to use these tools if they want because they are running on CI so then they can hopefully replicate CI as best as possible.

p.s. if you were having issues locally I'm happy to hop on a call and try to pair together to figure out what was going wrong.

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -0,0 +1,84 @@
+from aws_cdk import Environment, Stack
+from aws_cdk import aws_apigateway as apigw
+from aws_cdk import aws_certificatemanager as acm
+from aws_cdk import aws_route53 as route53
+from aws_cdk import aws_route53_targets as targets
+from constructs import Construct
+
+
+class ApiGateway(Stack):
+    """Sets up api gateway, creates subdomains, and creates methods that
```
greglucas:
```suggestion
    """Sets up api gateway, creates routes, and creates methods that
```

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -0,0 +1,84 @@
...
+        hosted_zone: route53.IHostedZone = None,
+        certificate: acm.ICertificate = None,
+        use_custom_domain: bool = False,
+        environment_name: str = "dev",
+        **kwargs,
+    ) -> None:
+        super().__init__(scope, construct_id, env=env, **kwargs)
+
+        # Define subdomains
+        subdomains = lambda_functions.keys()
```
greglucas:
```suggestion
        # Define routes
        routes = lambda_functions.keys()
```

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Create a single API Gateway
+        api = apigw.RestApi(
+            self,
+            f"api-RestApi-{sds_id}",
+            rest_api_name=f"api-RestApi-{sds_id}",
+            description="This service serves as my API Gateway.",
```
greglucas:
```suggestion
            description="API Gateway for lambda function endpoints",
```

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -0,0 +1,84 @@
...
+            deploy_options=apigw.StageOptions(stage_name=f"{sds_id}"),
```
greglucas, line 57: I'm always confused by "Stages" of the APIs, are we actually using stages here or is it just a formality to pass this in? i.e. I think that this allows you to set up multiple stages of only the APIs in the same account/deployment, but we are staging things ourselves via different stack deployments. So my question is whether we need this deployment option or not?

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -0,0 +1,84 @@
...
+            endpoint_types=[apigw.EndpointType.REGIONAL],
+        )
+
+        # Define a custom domain
+        if use_custom_domain:
+            custom_domain = apigw.DomainName(
+                self,
+                f"api-DomainName-{sds_id}",
+                domain_name=f"{environment_name}.imap-mission.com",
```
greglucas:
```suggestion
                domain_name=f"api.{sds_id}.imap-mission.com",
```

What about using the sds_id as our subdomain identifier? "environment" can mean a lot of different things, so it isn't clear what you're asking the user for to me.

Another note is that I think we'll want a few different things here for routing off of this location, so I think we should also probably put `api.*` (or some other qualifier) in the domain route so we know this is where our APIGateway is living. (We will also have the `dev.imap-mission.com` HTML site being served, possibly some `opensearch.dev.imap-mission.com` database routing too.

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -0,0 +1,84 @@
...
+                f"api-AliasRecord-{sds_id}",
+                zone=hosted_zone,
+                record_name=f"{environment_name}.imap-mission.com",
+                target=route53.RecordTarget.from_alias(
+                    targets.ApiGatewayDomain(custom_domain)
+                ),
+            )
+
+        # Loop through the lambda functions to create resources (routes)
+        for subdomain in subdomains:
```
greglucas:
```suggestion
        for route in routes:
```

Nice, I like this update of having these as routes!

sds_data_manager/stacks/domain_stack.py
```diff
@@ -0,0 +1,60 @@
...
+
+        if use_custom_domain:
+            self.hosted_zone = route53.HostedZone.from_lookup(
+                self, f"HostedZone-{sds_id}", domain_name="imap-mission.com"
+            )
+
+            self.certificate = acm.Certificate(
+                self,
+                f"Certificate-{sds_id}",
+                domain_name="*.imap-mission.com",
```
greglucas:
```suggestion
                domain_name=f"*.{sds_id}.imap-mission.com",
```
If you make the other updates, we might also need to change this cert here, I'm not sure if this is too broad.

greglucas (commented):
Nice, I think this looks good and is a nice and easy way to hit the lambda functions! Just a few minor comments inline.

sds_data_manager/stacks/domain_stack.py
laspsandoval: If we do this change we will need to add more SSL certifications, which is ok but might take more time for me to figure out. Let me know if you want to leave it for now or think it needs to be changed for this pr.

greglucas: I don't know enough about this stuff, my concern would be I deploy a stack with `sds-id="greg"` and create a certificate for `*.imap-mission.com`. What then happens if you want to deploy a stack with `sds-id="laura"`? Are we able to create two different certificates for the same exact domain name? Possibly (because the resource ids would be different), but it'd be worth verifying this too. Did you and @tech3371 deploy your stacks at the same time? That may already give us the answer to this question.

pyproject.toml
tech3371: What if we add it to extras like we did on `imap_processing`?

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -0,0 +1,84 @@
...
+from aws_cdk import aws_route53 as route53
+from aws_cdk import aws_route53_targets as targets
+from constructs import Construct
+
+
+class ApiGateway(Stack):
+    """Sets up api gateway, creates subdomains, and creates methods that
+    are linked to the lambda function.
+
+    An example of the format of the url: https://dev.imap-mission.com/query
```
tech3371: Thank you for giving examples like this!

sds_data_manager/stacks/domain_stack.py
tech3371: I haven't deployed and tested for this PR.

sds_data_manager/stacks/api_gateway_stack.py
laspsandoval: We don't really need it, but it is nice.

If we remove it AWS CDK will still create a stage, but it will be named "prod" by default and won't be specific to the sds_id. And if we remove it apparently we lose the ability to define specific configurations (like caching, throttling, logging, etc.) for different environments/stages of your API at the deployment level.

greglucas: Nice, thanks for clarifying and explaining what is going on behind the scenes!

greglucas (approved):

tech3371:
I tested all the APIs and it works! Reallly nice work, Laura!!!

tech3371 (approved):

Issue #96, opened 2023-07-17
Implementation of Track Data Availability


tech3371:
**Summary**
Create schema for pipeline dependencies

**Requirements**
The more specific the better.
- [ ]
- [ ]

**Related tickets*
-

maxinelasp:
Blocked on #95

PR #104, opened 2023-07-21
Implementation of Track Data Availability

tech3371:
# Change Summary
This is initial implementation of issue [Data Tracket Implementation](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/96)

## Overview

We need a database to track processing. In this review, there are two ways to track processing.
1. Use DynamDB to track processing
2. Use existing opensearch database

## New Files
- sds_data_manager/lambda_code/SDSCode/dynamodb_utils/processing_status.py
   - Defines processing status enum
- sds_data_manager/stacks/dynamodb_stack.py
   - Creates DynamoDB resources using this stack
  
## Updated Files
   - Added code in indexer lambda to take injested data information and initialized data-tracker with default processing status.
   - Created new opensearch index to send same default processing status.
- updated file 2
   - descipriton of change 1 in file 2
   - Updated it take DynamoDB stack as input and use it to get and set environment variable of table name for indexer lambda
   - Updated stack builder to create DynamoDB stack and updated input parameter of sds_data_manager_stack.
## Testing
- tests/infrastructure/test_dynamo_db_stack.py
   - This tests if DynamoDB stack creates resources as expected and checks some required settings.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/dynamodb_utils/processing_status.py
+++ b/sds_data_manager/lambda_code/SDSCode/dynamodb_utils/processing_status.py
@@ -1,0 +1,12 @@
+from enum import Enum
+
+
+class ProcessingStatus(Enum):
+    """
+    Enum for the processing status.
+    """
+    PENDING = 0
+    IN_PROGRESS = 1
+    COMPLETED = 2
+    FAILED = 3
+    CANCELLED = 4
--- a/sds_data_manager/lambda_code/SDSCode/dynamodb_utils/processing_status.py
+++ b/sds_data_manager/lambda_code/SDSCode/dynamodb_utils/processing_status.py
@@ -5,6 +5,7 @@ class ProcessingStatus(Enum):
     """
     Enum for the processing status.
     """
+
     PENDING = 0
     IN_PROGRESS = 1
     COMPLETED = 2
--- a/sds_data_manager/lambda_code/SDSCode/indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/indexer.py
@@ -8,13 +8,14 @@ import sys
 import boto3
 from opensearchpy import RequestsHttpConnection
 
+from .dynamodb_utils.processing_status import ProcessingStatus
+
 # Local
 from .opensearch_utils.action import Action
 from .opensearch_utils.client import Client
 from .opensearch_utils.document import Document
 from .opensearch_utils.index import Index
 from .opensearch_utils.payload import Payload
-from .dynamodb_utils.processing_status import ProcessingStatus
 
 # Logger setup
 logger = logging.getLogger()
@@ -122,17 +123,17 @@ def initialize_data_processing_status(filename: str):
     # Get instrument name by parsing the filename.
     # Eg. filename = imap_l0_sci_codice_20230602_v02.pkts
     # Get instrument name, codice, from the filename.
-    instrument_name = filename.split('_')[3]
+    instrument_name = filename.split("_")[3]
 
     return {
-        'instrument': instrument_name,
-        'filename': filename,
-        'processing_status': ProcessingStatus.PENDING.name,
-        'l1a_status': ProcessingStatus.PENDING.name,
-        'l1b_status': ProcessingStatus.PENDING.name,
-        'l1c_status': ProcessingStatus.PENDING.name,
-        'l2_status': ProcessingStatus.PENDING.name,
-        'l3_status': ProcessingStatus.PENDING.name
+        "instrument": instrument_name,
+        "filename": filename,
+        "processing_status": ProcessingStatus.PENDING.name,
+        "l1a_status": ProcessingStatus.PENDING.name,
+        "l1b_status": ProcessingStatus.PENDING.name,
+        "l1c_status": ProcessingStatus.PENDING.name,
+        "l2_status": ProcessingStatus.PENDING.name,
+        "l3_status": ProcessingStatus.PENDING.name,
     }
 
 
@@ -144,8 +145,8 @@ def write_data_to_dynamodb(item: dict):
     item : dict
         data for database
     """
-    dynamodb = boto3.resource('dynamodb')
-    table = dynamodb.Table(os.environ['DYNAMODB_TABLE'])
+    dynamodb = boto3.resource("dynamodb")
+    table = dynamodb.Table(os.environ["DYNAMODB_TABLE"])
     table.put_item(Item=item)
 
 
@@ -218,8 +219,8 @@ def lambda_handler(event, context):
         # TODO: Decide if we want to keep both or keep one after SIT-2
         # Right now, we can write processing status of injested data to both databases.
         # In the future, we can decide which one to write to.
-        # Initialize processing status for injested data to pending. This will be updated
-        # when the data is processed.
+        # Initialize processing status for injested data to pending. This will be
+        # updated when the data is processed.
         item = initialize_data_processing_status(filename)
 
         # Write processing status data to DynamoDB.
--- a/sds_data_manager/lambda_code/SDSCode/indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/indexer.py
@@ -107,11 +107,17 @@ def _create_open_search_client():
     )
 
 
-def initialize_data_processing_status(filename: str):
+def initialize_data_processing_status(metadata: dict, filename):
     """Generate data that will be sent to database.
 
     Parameters
     ----------
+    metadata : dict
+        metadata from filename. metadata comes from this
+        _check_for_matching_filetype function call.
+        Dictionary returned from that function call
+        can be used to get data level or instrument name
+        via metadata['instrument'] and metadata['level'].
     filename : str
         filename of injested data.
 
@@ -120,16 +126,12 @@ def initialize_data_processing_status(filename: str):
     dict
         data for database
     """
-    # Get instrument name by parsing the filename.
-    # Eg. filename = imap_l0_sci_codice_20230602_v02.pkts
-    # Get instrument name, codice, from the filename.
-    instrument_name = filename.split("_")[3]
 
     return {
-        "instrument": instrument_name,
+        "instrument": metadata["instrument"],
         "filename": filename,
-        "data_level": "l0",
-        "version": "1.0.0",
+        "data_level": metadata["level"],
+        "version": metadata["version"],
         "status": ProcessingStatus.PENDING.name,
     }
 
@@ -218,7 +220,7 @@ def lambda_handler(event, context):
         # In the future, we can decide which one to write to.
         # Initialize processing status for injested data to pending. This will be
         # updated when the data is processed.
-        item = initialize_data_processing_status(filename)
+        item = initialize_data_processing_status(metadata=metadata, filename=filename)
 
         # Write processing status data to DynamoDB.
         write_data_to_dynamodb(item)
--- a/sds_data_manager/lambda_code/SDSCode/indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/indexer.py
@@ -14,6 +14,7 @@ from .opensearch_utils.client import Client
 from .opensearch_utils.document import Document
 from .opensearch_utils.index import Index
 from .opensearch_utils.payload import Payload
+from .dynamodb_utils.processing_status import ProcessingStatus
 
 # Logger setup
 logger = logging.getLogger()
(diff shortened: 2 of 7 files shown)


sds_data_manager/stacks/dynamodb_stack.py
```diff
@@ -0,0 +1,139 @@
...
+        write_capacity : int
+            Write capacity for provisioned DynamoDb table.
+
+        Returns
+        -------
+        Construct
+            DynamoDb table construct
+        """
+        if read_capacity == 1 and write_capacity == 1:
+            raise Warning(
```
greglucas: Is there a problem with using defaults, or do we just want to warn? This should raise a `ValueError` or `warnings.warn(message)` rather than raising the warning.

sds_data_manager/stacks/dynamodb_stack.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+        self.sds_id = sds_id
+        self.table_name = table_name
+        self.partition_key = partition_key
+        self.sort_key = sort_key
+        self.on_demand = on_demand
+        if self.on_demand:
+            self._create_on_demand_dynamodb()
+        else:
+            self._create_provisioned_dynamodb(read_capacity, write_capacity)
```
greglucas: The extra functions seem like they use the same `Table()` constructor, just with a few different arguments. Can you just modify the arguments in the if-block here and then right below call one `Table(...)` constructor? (might need to set `read_capacity=None` for the default into this rather than 1, but I'm not sure)

I think this would help in the case of us wanting to update one of the other arguments like `removal_policy` or something like that to only need to update it in one place rather than both the provisioned and ondemand blocks.

If you do take this suggestion, keep the comments from the functions and put them inline, they are really nice!

tests/infrastructure/test_dynamo_db_stack.py
```diff
@@ -0,0 +1,47 @@
+import pytest
+from aws_cdk.assertions import Template
+
+from sds_data_manager.stacks.dynamodb_stack import DynamoDB
+
+
+@pytest.fixture(scope="module")
+def template(app, sds_id):
```
greglucas: We might want two templates, one also to test the provisioned version?

sds_data_manager/stacks/dynamodb_stack.py
```diff
@@ -0,0 +1,139 @@
...
+            ),
+            billing_mode=dynamodb.BillingMode.PROVISIONED,
+            write_capacity=write_capacity,
+            read_capacity=read_capacity,
+            removal_policy=RemovalPolicy.DESTROY,
+            point_in_time_recovery=True,
+        )
+
+    def get_table_arn(self, table_name):
+        return dynamodb.Table.from_table_name(self, table_name, table_name)
```
greglucas: Is this used for anything?

sds_data_manager/stacks/dynamodb_stack.py
```diff
@@ -0,0 +1,139 @@
...
+        scope: Construct,
+        construct_id: str,
+        sds_id: str,
+        table_name: str,
+        partition_key: str,
+        sort_key: str,
+        env,
+        on_demand=True,
+        read_capacity=1,
+        write_capacity=1,
```
greglucas:
```suggestion
        env: Environment,
        on_demand: bool = True,
        read_capacity: int = 1,
        write_capacity:int = 1,
```
I'd suggest to keep the typing going on all of the arguments.

sds_data_manager/lambda_code/SDSCode/dynamodb_utils/processing_status.py
```diff
@@ -0,0 +1,12 @@
+from enum import Enum
+
+
+class ProcessingStatus(Enum):
```
greglucas, line 4:
```suggestion
class ProcessingStatus(IntEnum):
```
Do we want to use IntEnum? Probably doesn't make a huge difference here.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -105,6 +107,49 @@ def _create_open_search_client():
...
+    return {
+        "instrument": instrument_name,
+        "filename": filename,
+        "processing_status": ProcessingStatus.PENDING.name,
+        "l1a_status": ProcessingStatus.PENDING.name,
+        "l1b_status": ProcessingStatus.PENDING.name,
+        "l1c_status": ProcessingStatus.PENDING.name,
+        "l2_status": ProcessingStatus.PENDING.name,
+        "l3_status": ProcessingStatus.PENDING.name,
+    }
```
greglucas: Is this the schema of the table?

Looking at this, I wonder if we want to make the level be a field instead of each being called out separately? i.e. what happens when some instrument wants to add a value to L3a, L3b, ... If we make that a separate field I think we could just extend that enumeration more easily than continually adding new fields.

(instrument, data_level, version, filename, status)

tech3371:
Note about this PR:
In this PR, I have implemented two ways to track data availability.
1. One through DynamoDB. We write ingested data information to DynamoDB and initialize processing status as PENDING.
2. Second through using existing OpenSearch database. I created new index in OpenSearch and write same processing status data of ingested data to that new index.

I wrote two options for now as place holder for data tracker feature. But post SIT-2, it will go through formal SPIKE review and updated as needed post SPIKE discussion.

In SIT-2, processing trigger will be using DynamoDB to check for data availability.

sds_data_manager/stacks/dynamodb_stack.py
tech3371: Ah ok. I didn't read carefully before. I just saw this `Can only be provided if billingMode is Provisioned.` and thought I need to set those two to some value when provisioned. But it looks like CDK defaults to 5 which is good but not idea. I feel like it might be give a warning of some sort if someone wants to create provisioned version, with an assumption that user may be needing higher read and write capacity than 5 if they chose to provisioned resources.

sds_data_manager/lambda_code/SDSCode/dynamodb_utils/processing_status.py
tech3371: not particularly

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: This is rough draft/view of what data may go in the database. 

I do see that scenario. Want to make sure if I am capture what are thinking. Will data look something like this?
```
 return {
        "instrument": instrument_name,
        "filename": filename,
        "data_level": {
            'l1a': ProcessingStatus.PENDING.name,
            'l1b': ProcessingStatus.PENDING.name,
            'l1c': ProcessingStatus.PENDING.name,
            ....
            'l3': ProcessingStatus.PENDING.name,
        },
        "version": ProcessingStatus.PENDING.name,
        "status": ProcessingStatus.PENDING.name
    }
```

```
 return {
        "instrument": instrument_name,
        "filename": filename,
        "data_level": 'l1a' or 'l1b' or 'l1c'....'l3', (indicating which level it's at)
        },
        "version": ProcessingStatus.PENDING.name,
        "status": ProcessingStatus.PENDING.name (status of current data_level)
    }
```

sds_data_manager/stacks/dynamodb_stack.py
tech3371: That makes sense. Will change that.

sds_data_manager/stacks/dynamodb_stack.py
tech3371: Not yet. Now that I am thinking, I am may be over doing it. It's one line of code and we can call this line directly if we need to look up table arn by table name.

tests/infrastructure/test_dynamo_db_stack.py
tech3371: sure thing.

sds_data_manager/lambda_code/SDSCode/indexer.py
greglucas: I think I was interpreting this table as being "keep track of file status". So, I was just thinking that each file was only associated with one level. Meaning we could take out all of the other level fields and just store whichever single one we are on with this individual file. So, has this file not been created, currently being created, processing failed, etc... But all for only the status of this single file.

But, what I think you're describing is that each file may want to keep track of what the processing status is of the l1a/l1b before it, and the l3 after it that are all dependencies or dependent on this single file?

It might be good to lay out the schema here or link to the schema somewhere externally, just so we are explicitly noting what should go into this table and how a processing job may update it etc... I think I'm just missing the overall workflow. Happy to hop on a call and discuss if that is easier too rather than async via the review.

tech3371: I think this is a discussion we need to have as a team. I can create a subtask issue to discuss this.

For the time being, I think what you are thinking is correct. I think second return statement suggestion goes well with what you are thinking. what do you think?

tech3371: Created new issue for that. https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/105

sds_data_manager/stacks/dynamodb_stack.py
tech3371: It accepts None. I think None makes sense. I am adding check and raise ValueError if user wants to create provisioned DB but didn't have read and write capacity value.

sds_data_manager/stacks/dynamodb_stack.py
tech3371: removed it in upcoming commit.

sds_data_manager/stacks/dynamodb_stack.py
```diff
@@ -0,0 +1,100 @@
...
+        table_name: str,
+        partition_key: str,
+        sort_key: str,
+        env: Environment,
+        on_demand: bool = True,
+        read_capacity: int = None,
+        write_capacity: int = None,
+        **kwargs,
+    ):
+        super().__init__(scope, construct_id, **kwargs)
```
laspsandoval, line 22: It should be:
super().__init__(scope, construct_id, env=env, **kwargs)

tests/infrastructure/test_dynamo_db_stack.py
```diff
@@ -0,0 +1,96 @@
...
+        env=None,
+        on_demand=False,
+        write_capacity=100,
+        read_capacity=100,
+    )
+    template = Template.from_stack(stack)
+    return template
+
+
+def test_table_count(on_demand_dynamodb, provisioned_dynamodb):
```
greglucas, line 47:
```suggestion
@pytest.mark.parametrize("template", [on_demand_dynamodb, provisioned_dynamodb])
def test_table_count(template):
    template.resource_count_is("AWS::DynamoDB::Table", 1)
```

I think this would work with parametrization on most of these tests (the test_billing_mode would be trickier though, so probably not worth doing).

laspsandoval:
Looks good! I didn't deploy. I'll probably play around with it later unless you need me to deploy. Just let me know. Nice work!

sds_data_manager/stacks/dynamodb_stack.py
tech3371: I don't use env. should I still set env for future use?

tests/infrastructure/test_dynamo_db_stack.py
tech3371: Yeah, billing mode test was little tricky. But it's good to know we can do that.

greglucas (approved):

greglucas:
Needs a rebase, but the content looks good to me!

tech3371:
pre-commit.ci autofix

tech3371:
> Needs a rebase, but the content looks good to me!

@greglucas I tried to rebase. It's first time I am doing rebase. I hope it did it right.

tech3371:
> Looks good! I didn't deploy. I'll probably play around with it later unless you need me to deploy. Just let me know. Nice work!

@laspsandoval I tried to test it again after rebasing but if you can double check, I appreciate it.

tech3371:
pre-commit.ci autofix

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -105,6 +107,46 @@ def _create_open_search_client():
...
+    """
+    # Get instrument name by parsing the filename.
+    # Eg. filename = imap_l0_sci_codice_20230602_v02.pkts
+    # Get instrument name, codice, from the filename.
+    instrument_name = filename.split("_")[3]
+
+    return {
+        "instrument": instrument_name,
+        "filename": filename,
+        "data_level": "l0",
```
bryan-harter: Should we gather the data level from the file name as well?

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -105,6 +107,46 @@ def _create_open_search_client():
...
+
+    Returns
+    -------
+    dict
+        data for database
+    """
+    # Get instrument name by parsing the filename.
+    # Eg. filename = imap_l0_sci_codice_20230602_v02.pkts
+    # Get instrument name, codice, from the filename.
+    instrument_name = filename.split("_")[3]
```
bryan-harter: Maybe pass in the `metadata` variable gathered from `_check_for_matching_filetype` function.  Then you can get the level/instrument via `metadata['instrument']` and `metadata['level']`.  That would be useful if the file format ever changed (or for future, non-IMAP missions)

tech3371: That's a good idea! It looks like we get metadata in above lines. I will use that `metadata` and pass that as input to this function.

tech3371:
Can I merge this? I will need to bring one minor thing from this for next processing pipeline task.

sds_data_manager/stacks/dynamodb_stack.py
greglucas: I'd actually just remove it from your listed keyword arguments above, and then `**kwargs` will forward it automatically if you pass in `env=env` during creation, but won't require it here because you don't actually use it below.

tech3371: That's what I figured. I didn't add it yet for this review. I will do same(leave out env if not used) for my upcoming PR too.

bryan-harter:
Looks good to me!

bryan-harter (approved):

tech3371:
Opening new PR from `dev` branch

PR #110, opened 2023-07-27
Changing the Upload API to use a GET rather than a POST

bryan-harter:
# Change Summary
Updating the API Gateway functions 

## Overview
Removed the "indexer" API, and changed the "upload" API to GET

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
- api_gateway_stack.py
   - Removed the "indexer" API.  The indexer lambda function triggers from new files arriving in the bucket, so it only gets invoked internally and doesn't need an API
   - Changed the upload from a "POST" to a "GET".  The reason why it is a "GET" is because the upload API actually returns a signed url to the user to upload a data file.

### DIFF
--- a/sds_data_manager/stacks/sds_data_manager_stack.py
+++ b/sds_data_manager/stacks/sds_data_manager_stack.py
@@ -233,8 +233,7 @@ class SdsDataManager(Stack):
         download_query_api.add_to_role_policy(s3_read_policy)
 
         self.lambda_functions = {
-            "upload": {"function": upload_api_lambda, "httpMethod": "POST"},
+            "upload": {"function": upload_api_lambda, "httpMethod": "GET"},
             "query": {"function": query_api_lambda, "httpMethod": "GET"},
             "download": {"function": download_query_api, "httpMethod": "GET"},
-            "indexer": {"function": indexer_lambda, "httpMethod": "POST"},
         }


greglucas (approved):
I agree this seems right. I assume that the Javascript SDK (i.e. eventual frontend website) uses a GET as well when interacting with this?

It looks like all methods are GET, do we anticipate anything else, otherwise we could get rid of the httpMethod here altogether and just make it a collection of tuples `(endpoint_name, function)`

laspsandoval:
looks good

laspsandoval (approved):

bryan-harter:
> I agree this seems right. I assume that the Javascript SDK (i.e. eventual frontend website) uses a GET as well when interacting with this?
> 
> It looks like all methods are GET, do we anticipate anything else, otherwise we could get rid of the httpMethod here altogether and just make it a collection of tuples `(endpoint_name, function)`

It's possible there could be post commands in the future so we might want to keep it.  For example, if we do include a science impacting event database (where scientists can post solar events to a DB that shows up on the website, similar to the MAVEN SDC), that would be a POST.  We might also make a POST that updates rows in opensearch, for example we could have an API that we use to handle public release process (setting status=public or public=True or something to the file in the opensearch instance)

PR #112, opened 2023-07-28
MNT: Add more information in the app template file

greglucas:
This moves the places you need to modify code up to the top for easy reference. It outputs the profile you're trying to use and the expected profile for information when you're using the wrong profile.

## Updated Files

`app_template_dev.py`: Some refactoring and details above.


laspsandoval (approved):
Nice updates.

bryan-harter (approved):
Looks good!

PR #113, opened 2023-07-28
BLD: Adding extras for supporting pip installs

greglucas:
# Change Summary

## Overview

Currently `pip install .` does not install all of the required packages. Move the runtime required dependencies up into the main section. Then `pip install .[dev]` does not pick up the group dev, so we need to add an extras section for dev. That requires optional dependencies to be added to the `main` dependencies section, and to keep them in groups that means we need to duplicate those, once with optional and once in the group.

## New Dependencies

None, just rearranged the dependency section and added an extras.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -61,7 +61,7 @@ tests-no-zope = ["cloudpickle", "hypothesis", "mypy (>=1.1.1)", "pympler", "pyte
 name = "aws-cdk-asset-awscli-v1"
 version = "2.2.200"
 description = "A library that contains the AWS CLI for use in Lambda Layers"
-category = "dev"
+category = "main"
 optional = false
 python-versions = "~=3.7"
 files = [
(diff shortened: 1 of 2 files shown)


bryan-harter (approved):
Looks good

tech3371 (approved):

greglucas:
I'm actually going to close this. @maxinelasp and I had a discussion a few weeks back about this and this was just an easy way to demo what the issue was. I think that we do need to rework the dependencies to make sure `pip install .` works as expected, which means we need to move a few dependencies out of the dev block of poetry. But, I also think @maxinelasp's suggestion about moving towards a `pip install -r requirements-dev.txt` or something like that for our pip friends would be acceptable since there is a lot of duplicated "version" stuff in the different sections with this approach.

@maxinelasp do you agree with the above, and do you want to do the rework of the dependencies according to that? Or do you want me to tackle that approach? To sum it up:
* Make sure our actual minimal functioning dependencies are included in the main dependency section.
* Export the dev and optional groups into a new `requirements-dev.txt` (or some other name) following the same pattern we already do for the lambda dependencies.

maxinelasp:
@greglucas  Yes, that sounds good. I can rework the dependencies and see if I can find a solution we can both be happy with, lol. Is there an issue tracking this yet?

greglucas:
> I can rework the dependencies and see if I can find a solution we can both be happy with, lol

:) I actually don't think it will be too hard. It is just the unfortunate Python packaging ecosystem we are fighting.

> Is there an issue tracking this yet?

I don't believe so. I will create one.

PR #118, opened 2023-08-01
Updated infrastructure tests

laspsandoval:
# Change Summary

## Overview
Modified test_data_manager_stack.py and moved some tests to test_open_search_stack.py. This was necessary given the separation of stacks in a previous PR.

## New Dependencies
The test_data_manager_stack.py tests sds_data_manager_stack.py. The test_open_search_stack.py tests opensearch_stack.py.

## New Files
- test_open_search_stack.py
   - Created as a result to making the creation of opensearch in its own stack.

## Deleted Files
None

## Updated Files
- test_data_manager_stack.py
   - We are no longer using lambda urls so those tests were removed.
   - s3 bucket names were changed
   - Testing of opensearch was moved to its own test (along with resources)

## Testing
Everything in the PR is a test.

## Note
I left some comments to explain which tests were removed and a description of why they were removed for the purpose of the review. I will delete those comments before merging.

### DIFF
--- a/tests/infrastructure/test_data_manager_stack.py
+++ b/tests/infrastructure/test_data_manager_stack.py
@@ -4,8 +4,6 @@ from aws_cdk.assertions import Match, Template
 from sds_data_manager.stacks.opensearch_stack import OpenSearch
 from sds_data_manager.stacks.sds_data_manager_stack import SdsDataManager
 
-pytest.skip("Skipping tests temporarily", allow_module_level=True)
-
 
 @pytest.fixture(scope="module")
 def opensearch_stack(app, sds_id, env):
@@ -37,7 +35,7 @@ def test_s3_data_bucket_resource_properties(template, sds_id):
     template.has_resource_properties(
         "AWS::S3::Bucket",
         props={
-            "BucketName": f"sds-data-{sds_id}",
+            "BucketName": f"sds-config-bucket-{sds_id}",
             "VersioningConfiguration": {"Status": "Enabled"},
             "PublicAccessBlockConfiguration": {"RestrictPublicBuckets": True},
         },
@@ -56,7 +54,7 @@ def test_s3_config_bucket_resource_properties(template, sds_id):
     template.has_resource_properties(
         "AWS::S3::Bucket",
         {
-            "BucketName": f"sds-config-{sds_id}",
+            "BucketName": f"sds-config-bucket-{sds_id}",
             "VersioningConfiguration": {"Status": "Enabled"},
             "PublicAccessBlockConfiguration": {
                 "BlockPublicAcls": True,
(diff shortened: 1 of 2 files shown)


laspsandoval:
pre-commit.ci autofix

tests/infrastructure/test_data_manager_stack.py
```diff
@@ -1022,57 +805,60 @@ def test_indexer_lambda_permission_resource_properties(template):
                 "Fn::GetAtt": [Match.string_like_regexp("IndexerLambda*"), "Arn"]
             },
             "Principal": "s3.amazonaws.com",
-            "SourceAccount": {"Ref": "AWS::AccountId"},
+            "SourceAccount": Match.string_like_regexp("\\d{10}"),
```
greglucas: If you pass in `test_func(template, account)` you should be able to get the testing account ID to use for the matches as well if that is more straightforward for the comparison.
```suggestion
            "SourceAccount": account,
```

tests/infrastructure/test_open_search_stack.py
```diff
@@ -0,0 +1,282 @@
...
+        },
+    )
+
+
+def test_iam_roles_resource_count(template):
+    template.resource_count_is("AWS::IAM::Role", 1)
+
+
+def test_expected_properties_for_iam_roles(template):
+    found_resources = template.find_resources(
```
greglucas:
```suggestion
    found_resources = template.has_resource_properties(
```
since there is only one I think we can just use the assertion on resource properties?

tests/infrastructure/test_open_search_stack.py
```diff
@@ -0,0 +1,282 @@
...
+        "AWS::Logs::LogGroup",
+        {
+            "DeletionPolicy": "Retain",
+            "UpdateReplacePolicy": "Retain",
+        },
+    )
+    template.has_resource_properties("AWS::Logs::LogGroup", {"RetentionInDays": 30})
+
+
+def test_sdsmetadatadomain_app_logs_resource_properties(template):
```
greglucas: This test is identical to the above. I think this is where you could use the `find_resources()` and count if you wanted to say there should be exactly 3 resources found with these same exact properties. Otherwise, this is not really testing what we think it is here because if one of them changes, this will still pass because the other one will be found.

tests/infrastructure/test_open_search_stack.py
```diff
@@ -0,0 +1,282 @@
...
+                        'AccessPolicy"}}',
+                    ],
+                ]
+            },
+            "InstallLatestAwsSdk": True,
+        },
+    )
+
+
+def test_sdsmetadatadomain_slow_index_logs_resource_properties(template):
```
greglucas: Another duplicate test that can be removed.

tests/infrastructure/test_open_search_stack.py
```diff
@@ -0,0 +1,282 @@
...
+        "AWS::Logs::LogGroup",
+        {
+            "DeletionPolicy": "Retain",
+            "UpdateReplacePolicy": "Retain",
+        },
+    )
+    template.has_resource_properties("AWS::Logs::LogGroup", {"RetentionInDays": 30})
+
+
+def test_iam_roles_resource_count(template):
```
greglucas: These IAM roles and properties below this are also already tested earlier in this file I believe and can be removed here.

greglucas (approved):
Very nice, @laspsandoval! I've made a few minor comments inline, and I suggest removing a few duplicated tests that may have been a duplicate copy/paste over to the new file (which I think will make pre-commit happy then as well).

laspsandoval:
pre-commit.ci autofix

PR #120, opened 2023-08-01
Implementation of Track Data Availability

tech3371:
# Change Summary
This is initial implementation of issue [Data Tracket Implementation](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/96)

## Overview

We need a database to track processing. In this review, there are two ways to track processing.
1. Use DynamDB to track processing
2. Use existing opensearch database

## New Files
- sds_data_manager/lambda_code/SDSCode/dynamodb_utils/processing_status.py
   - Defines processing status enum
- sds_data_manager/stacks/dynamodb_stack.py
   - Creates DynamoDB resources using this stack
  
## Updated Files
   - Added code in indexer lambda to take injested data information and initialized data-tracker with default processing status.
   - Created new opensearch index to send same default processing status.
- updated file 2
   - descipriton of change 1 in file 2
   - Updated it take DynamoDB stack as input and use it to get and set environment variable of table name for indexer lambda
   - Updated stack builder to create DynamoDB stack and updated input parameter of sds_data_manager_stack.
## Testing
- tests/infrastructure/test_dynamo_db_stack.py
   - This tests if DynamoDB stack creates resources as expected and checks some required settings.


sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -105,6 +107,48 @@ def _create_open_search_client():
...
+    return {
+        "instrument": metadata["instrument"],
+        "filename": filename,
+        "data_level": metadata["level"],
+        "version": metadata["version"],
+        "status": ProcessingStatus.PENDING.name,
+    }
+
+
+def write_data_to_dynamodb(item: dict):
```
laspsandoval, line 139: The code creates a new DynamoDB resource inside the write_data_to_dynamodb function every time it's called. If this function is called frequently, it could be more efficient to create the resource once and reuse it.

laspsandoval:
Everything deploys and works. I had one additional comment that might be implemented in another PR. I also have a general question which is : Do you know why we chose to use OpenSearch for the APIs and then a separate DynamoDB for tracking processing? Somehow I missed that information along the way….

Besides that it all looks good. Nice work!

laspsandoval (approved):

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: Yeah, for sure! like a global variable or create resource in the handler function and pass as argument?

tech3371:
pre-commit.ci autofix

tech3371:
I don't know why `test_indexer_lambda_iam_policy_resource_properties` test is failing. I tried fixing it but can't seem to fix it. It looks correct me but can't seems to pass it. Any suggestions?

tests/infrastructure/test_data_manager_stack.py
```diff
@@ -411,8 +424,16 @@ def test_indexer_lambda_iam_policy_resource_properties(template):
                             },
                         ],
                     },
+                    {"Action": "dynamodb:PutItem", "Effect": "Allow", "Resource": "*"},
                     {
```
sdhoyt, line 428:
```suggestion
                    {
                        "Action": [
                            "secretsmanager:GetSecretValue",
                            "secretsmanager:DescribeSecret"
                        ],
                        "Effect": "Allow",
                        "Resource": {"Fn::Join": ["", ["arn:", {"Ref": "AWS::Partition"}, Match.string_like_regexp(":secretsmanager:.*:secret:.*")]]},
                    }
```

sdhoyt: I was able to get the test passing by replacing the secrets manager part with this. I got that format from writing the template json to a file and looking at it. I tried Matching the whole Resource entry, but that didn't work for some reason. I think it also didn't like that the GetSecretValue and DescribeSecret were out of order in that list. I'm not sure if I don't know what I'm doing or the cdk is just really annoying to test. But the code above should get the test passing and it seems to be what you were expecting. Let me know if you have any issues

sdhoyt (changes requested):

sdhoyt (approved):

tests/infrastructure/test_data_manager_stack.py
tech3371: Thank you for inspecting that. I wouldn't have done `Fn::Join` because I thought it's not needed since it was not getting arn or anything. 

Test did pass!!!

tech3371:
pre-commit.ci autofix

tech3371:
Thank you @laspsandoval and @sdhoyt for testing it! I am going to merge this PR now.

PR #121, opened 2023-08-02
Opensearch snapshot

### LINKED ISSUE #98: Setup Manual OpenSearch Snapshots
**Summary**
Manual OpenSearch snapshots should be taken when the indexer lambda is finished running

**Requirements**
- Take manual opensearch snapshots during metadata indexing
- store snapshots in S3 bucket
**Related tickets*
-

sdhoyt:
# Change Summary

## Overview
This update adds functionality to backup the OpenSearch database each time a set of files is indexed. The OpenSearch version was also updated to version 2.7 and the CDK version was updated to 2.88. CDK version 2.88 was required to deploy OS 2.7 through the CDK.

## New Dependencies
- requests-aws4auth (used to send authentication through requests)
- requests-mock (used to mock requests in unit tests)
- freezegun (used to keep file timestamp consistent in unit tests)

## New Files
- snapshot.py (mostly copied from Libera. Big thanks to Brian M.)
   - This file includes the code for registering a backup repository (S3 bucket) and taking the OpenSearch snapshot.
   - I think it would be good to update this file in the future to better fit in with the existing opensearch_utils classes, but this should work for now.

## Deleted Files
None

## Updated Files
- sds_data_manager_stack.py
   - Updated OpenSearch to 2.7
   - Added Snapshot Role and policies for Indexer lambda to allow it to take the snapshots
   - Added a new S3 bucket to store snapshots
- indexer.py
   - Added function to take snapshot at end of lambda

## Testing
A new unit test file `test_snapshot.py` was created, `sds_data_manager_stack.py` and `test_open_search_stack.py` were updated. All files are passing all tests.

NOTE: When deploying and testing the snapshots, there are a few extra steps to give proper permissions to OS to take the snapshots. 
1. Log in to your OpenSearch instance dashboard
2. Back in AWS, go to IAM -> Roles and copy your IndexerServiceLamba role's ARN
3. In the OpenSearch dashboard menu, go to Security -> Roles and click on the manage_snapshots role.
4. Go to the Mapped Users tab and click on the Map Users
5. Paste the ARN that you copied into the backed role text bock and click Map

Now when you upload files to the data bucket, a snapshot should be taken. You can confirm those files were created in the snapshot bucket.


pyproject.toml
```diff
@@ -31,15 +31,18 @@ classifiers = [
...
 pytest-cov = "^4.0.0"
 openmock = "^2.3.0"
+requests-mock = "^1.11.0"
+freezegun = "^1.2.2"
 
 # Optional dependencies - install with `poetry install -E docs`
 sphinx = {version="^7.1.0", optional=true}
 myst-parser = {version="^2.0.0", optional=true}
 pydata-sphinx-theme = {version="^0.13.3", optional=true}
+requests-aws4auth = "^1.2.3"
```
maxinelasp: This shouldn't be necessary since `requests-aws4auth` is in the lamda-dev dependency group

sds_data_manager/lambda_code/SDSCode/opensearch_utils/snapshot.py
```diff
@@ -0,0 +1,101 @@
...
+    # Register the snapshot, this can be run every time, if the
+    # repo is registered will return 200
+    try:
+        path = f"_snapshot/{snapshot_repo_name}"  # the OpenSearch API endpoint
+        url = "https://" + host + "/" + path
+
+        payload = {
+            "type": "s3",
+            "settings": {
+                "bucket": f"{snapshot_s3_bucket}",
```
maxinelasp, line 105: Is the eventual plan for this bucket to get backed up to the Backup account?

sds_data_manager/lambda_code/SDSCode/opensearch_utils/snapshot.py
```diff
@@ -0,0 +1,101 @@
...
+    ----------
+    url : str
+        OpenSearch domain URL endpoint including https:// and trailing /.
+    """
+
+    r = requests.put(url, auth=awsauth)
+    return r
+
+
+def run_backup(host, region, snapshot_repo_name, snapshot_s3_bucket, snapshot_role_arn):
```
maxinelasp, line 72: Needs a docstring

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -16,6 +16,7 @@
 from .opensearch_utils.document import Document
```
maxinelasp, line 16: Where does this indexer run?

maxinelasp (commented):
Overall looks good to me. Can you add some documentation somewhere (maybe in indexer.py?) about the manual steps to add the role to opensearch? Is there any way to do that in CDK? I found I could assume roles in CDK, but that might be different than what you need to do here. 

If we're going to have manual steps, I'd really prefer at least having them written down somewhere! You could also write a rst doc covering all of that info.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: It runs on a lambda. This lambda get's kickoff after we drop a file on data bucket.

pyproject.toml
tech3371: Are you saying we don't need this because lambda is running inside AWS environment?

sds_data_manager/lambda_code/SDSCode/opensearch_utils/snapshot.py
```diff
@@ -0,0 +1,101 @@
...
+        region,
+        service,
+        session_token=credentials.token,
+    )
+
+    return awsauth
+
+
+def register_repo(payload: dict, url: string, awsauth):
+    """Register the snapshot repo
```
tech3371: Can you add some comments about repo? Repo is used for so many things that it's confusing sometimes. Is it ECR repo?

sds_data_manager/lambda_code/SDSCode/opensearch_utils/snapshot.py
```diff
@@ -0,0 +1,101 @@
...
+    payload : dict
+             S3 bucket and AWS region to store the manual snapshots
+             The role ARN that has S3 permissions to store the new snapshot
+    url : str
+        OpenSearch domain URL endpoint including https:// and trailing /.
+    """
+
+    headers = {"Content-Type": "application/json"}
+
+    r = requests.put(url, auth=awsauth, json=payload, headers=headers)
```
tech3371, line 53: Ah, ok. We need authentication for the http requests. That makes sense.

sds_data_manager/lambda_code/SDSCode/opensearch_utils/snapshot.py
tech3371: Yeah, looks like other functions needs update to their docstring.

sds_data_manager/lambda_code/SDSCode/opensearch_utils/snapshot.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+
+    r = requests.put(url, auth=awsauth)
+    return r
+
+
+def run_backup(host, region, snapshot_repo_name, snapshot_s3_bucket, snapshot_role_arn):
+    awsauth = get_auth(region)
+    snapshot_start_time: datetime = datetime.utcnow().strftime("%Y-%m-%d-%H:%M:%S")
+    snapshot_name = f"os_snapshot_{snapshot_start_time}"
```
tech3371: can we spell out `os` to `opensearch`?

sds_data_manager/lambda_code/SDSCode/indexer.py
maxinelasp: Got it, thanks! That was my guess, just making sure.

sds_data_manager/lambda_code/requirements.txt
```diff
@@ -95,3 +95,5 @@ six==1.16.0 ; python_version >= "3.9" and python_version < "4" \
 urllib3==1.26.16 ; python_version >= "3.9" and python_version < "4" \
     --hash=sha256:8d36afa7616d8ab714608411b4a3b13e58f463aee519024578e062e141dce20f \
     --hash=sha256:8f135f6502756bde6b2a9b28989df5fbe87c9970cecaa69041edcce7f0589b14
+requests-aws4auth==1.2.3 ; python_version >= "3.9" and python_version < "4" \
```
tech3371: I meant to comment on past PR but didn't get to it. What is this requirements.txt used for?

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -114,15 +114,33 @@ def __init__(
             destination_bucket=config_bucket,
         )
 
+        ########### OpenSearch Snapshot Storage
+        snapshot_bucket = s3.Bucket(
+            self,
+            f"SnapshotBucket-{sds_id}",
+            bucket_name=f"sds-os-snapshot-{sds_id}",
```
tech3371: same here. can we spell out `os`?

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -136,6 +154,26 @@ def __init__(
             resources=["*"],
         )
 
+        snapshot_role_policy = iam.PolicyStatement(
+            effect=iam.Effect.ALLOW,
+            actions=[
+                "s3:ListBucket",
+                "s3:GetObject",
+                "s3:PutObject",
+                "s3:DeleteObject",
```
tech3371, line 166: should we have ability to delete on this bucket? Is it safe?

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -176,6 +217,27 @@ def __init__(
...
+        # Add permissions for Lambda to access OpenSearch
+        indexer_lambda.add_to_role_policy(
+            iam.PolicyStatement(
+                effect=iam.Effect.ALLOW,
+                actions=["es:*"],
+                resources=[f"{opensearch.sds_metadata_domain.domain_arn}/*"],
+            )
+        )
+
+        # PassRole allows services to assign AWS roles to resources and services
```
tech3371, line 239: Thank you for adding clarification!

pyproject.toml
maxinelasp: The dependency is created again a few lines down, so it's a duplicate. One or the other should probably be removed.

sdhoyt: I do use this library in the unit test. If I keep it in pyproject.toml and remove it from requirements.txt, it will still be available in the lambdas?

sds_data_manager/lambda_code/SDSCode/opensearch_utils/snapshot.py
sdhoyt: Correct, the plan is to backup the snapshot s3 bucket to the backup account

sds_data_manager/lambda_code/SDSCode/opensearch_utils/snapshot.py
sdhoyt: Agreed. I'll add the docstring and update the others

sds_data_manager/lambda_code/SDSCode/opensearch_utils/snapshot.py
sdhoyt: Yeah, I'll add comments about repo. OpenSearch uses the term "repository" for where the snapshots are stored

sds_data_manager/lambda_code/SDSCode/opensearch_utils/snapshot.py
sdhoyt: Agreed. I'll spell out those variables

sds_data_manager/lambda_code/requirements.txt
sdhoyt: The requirements.txt is provides those libraries for the lambdas to access. Although it sounds like I might be able to just put that in the pyproject.toml

sds_data_manager/stacks/sds_data_manager_stack.py
sdhoyt: Yes, I'll spell that out

sds_data_manager/stacks/sds_data_manager_stack.py
sdhoyt: From what I read, it seems like this is needed for snapshots to work. I can try removing it and see what happens, but in the documentation it says to setup the policy with delete permissions: https://docs.aws.amazon.com/opensearch-service/latest/developerguide/managedomains-snapshots.html

sdhoyt:
> Overall looks good to me. Can you add some documentation somewhere (maybe in indexer.py?) about the manual steps to add the role to opensearch? Is there any way to do that in CDK? I found I could assume roles in CDK, but that might be different than what you need to do here.
> 
> If we're going to have manual steps, I'd really prefer at least having them written down somewhere! You could also write a rst doc covering all of that info.

I agree. I'll document this somewhere. I'm hoping this is just temporary for this SIT and we can find a way to do this in the CDK in the future. The documentation only had the manual steps (this is only needed because we are using fine-grain access controls for OpenSearch). I talked to Brian M about this and Libera hasn't tightened permissions for their OpenSearch instance yet, so they haven't had to deal with this. But I'm hoping these manual steps are temporary, but I haven't found a way to do this in the CDK yet.

pyproject.toml
maxinelasp: Yes, the requirements.txt export statement in `.pre-commit-config.yaml` does export the `lambda-dev` dependency group. Since the dependency is in that group, it is still available for install. I'm not sure what groups the unit tests use, but it should be easy to install the lambda-dev group for the tests as well. 

Or, if you just want to make this a greater project dependency, which would also be fine, it should be moved up out of the optional dependencies section, and the duplicate should be removed from `lambda-dev`. 

It's ok to have duplicate dependencies in different dependency groups. But, if it's a full-project dependency (under `[tool.poetry.dependencies]` then it is unnecessarily duplicated by putting it in a group. The full project dependencies are installed every time, while the groups can be included or excluded depending on the use case.

sds_data_manager/lambda_code/requirements.txt
maxinelasp: This requirements.txt file is used for pip installs. It is automatically generated in the pre-commit steps, and should not be updated by hand. It is created from the dependencies specified in pyproject.toml. Let me know if you have other questions about it!

tech3371: cool. Yeah, I noticed those were updated automatically but didn't know their uses. Thank you!

sds_data_manager/stacks/sds_data_manager_stack.py
tech3371: interesting. Thank you for clarifying

docs/source/cdk/opensearch-setup
```diff
@@ -0,0 +1,20 @@
...
+OpenSearch has the ability to take `Snapshots <https://opensearch.org/docs/2.7/tuning-your-cluster/availability-and-recovery/snapshots/snapshot-restore/>`_ of itself which can be used to recover from a failure or to migrate from one cluster to another. These Snapshots are stored in a "repository" which is just a storage location. In the case of SDS Data Manager the storage location is an AWS S3 Bucket. 
+
+Required Manual Setup Steps
+---------------------------
+Because the OpenSearch instance used in the SDS Data Manager is setup with fine-grain access controls, a few manual steps need to be taken to allow the Indexer lambda to intitate the Snapshot in Opensearch. If these steps are not taken, OpenSearch will not be able to create Snapshots. The steps below assume that all SDS Data Manager stacks have been deployed.
+
+The steps for setting up Snapshot permissions are as follows:
+1. Log in to your OpenSearch instance dashboard
+2. Back in AWS, go to IAM -> Roles and copy your IndexerServiceLamba role's ARN
+3. In the OpenSearch dashboard menu, go to Security -> Roles and click on the manage_snapshots role.
```
tech3371, line 18: I didn't realized I need to click on `Roles` to see `managed_snapshots`. Besides that, everything was very clear. I was able to follow these steps!

tech3371 (approved):

pyproject.toml
sdhoyt: I think I'm doing something wrong here/not understanding. I have the requests-as4auth under the lambda section in the pyproject.toml:
```
[tool.poetry.group.lambda-dev.dependencies]
opensearch-py = "^2.1.1"
requests = "^2.28.2"
requests-aws4auth = "^1.2.3"
```
but when I deploy and run the lambda, it's saying that requests-aws4auth can't be found when it tries to import. It looks like the requests and opensearch-py are in the requirements.txt file. Is requests-aws4auth supposed to be automatically added to that file at some point?

maxinelasp: The whole lambda-dev dependency group is automatically added to the requirements.txt file on pre-commit, which you can see [here](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/dev/.pre-commit-config.yaml#L32C26-L32C26). It could be that you need to manually run that export before you deploy - which you could do with the command `poetry export -f requirements.txt -o sds_data_manager/lambda_code/requirements.txt --only lambda-dev`. Or, by committing your change prior to deploying and triggering the pre-commit checks.

sdhoyt: Ah, fixed it! Thank you!

template.json
```diff
@@ -0,0 +1 @@
...4", "Roles": [{"Ref": "DownloadQueryAPILambdaServiceRole082F4736"}]}}, "DownloadQueryAPILambdaE3D6D46B": {"Type": "AWS::Lambda::Function", "Properties": {"Code": {"S3Bucket": "cdk-hnb659fds-assets-1234567890-us-east-1", "S3Key": "0da65ff8b611cbd242828b50b32351a366221c349a77039c012a482f11c4eaaf.zip"}, "FunctionName": "download-query-api-sdsid-test", "Handler": "SDSCode.download_query_api.lambda_handler", "Role": {"Fn::GetAtt": ["DownloadQueryAPILambdaServiceRole082F4736", "Arn"]}, "Runtime": "python3.9", "Timeout": 60}, "DependsOn": ["DownloadQueryAPILambdaServiceRoleDefaultPolicy3490E4E4", "DownloadQueryAPILambdaServiceRole082F4736"]}}, "Parameters": {"BootstrapVersion": {"Type": "AWS::SSM::Parameter::Value<String>", "Default": "/cdk-bootstrap/hnb659fds/version", "Description": "Version of the CDK Bootstrap resources in this environment, automatically retrieved from SSM Parameter Store. [cdk:skip]"}}, "Rules": {"CheckBootstrapVersion": {"Assertions": [{"Assert": {"Fn::Not": [{"Fn::Contains": [["1", "2", "3", "4", "5"], {"Ref": "BootstrapVersion"}]}]}, "AssertDescription": "CDK bootstrap stack version 6 required. Please run 'cdk bootstrap' with a recent version of the CDK CLI."}]}}}
```
bryan-harter, line 1: Remove this file?

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -233,6 +241,10 @@ def lambda_handler(event, context):
 
     # send the paylaod to the opensearch instance
     client.send_payload(document_payload)
+
+    # take OpenSearch Snapshot
+    run_backup(host, region, snapshot_repo_name, snapshot_s3_bucket, snapshot_role_arn)
```
bryan-harter, line 246: So does this run a backup every single time a new file is added to the SDC?

bryan-harter (approved):

sds_data_manager/lambda_code/SDSCode/indexer.py
sdhoyt: Yeah, each time the indexer is called it creates a new snapshot. Snapshots only capture the difference between what exists at the present and the previous snapshot, so more frequent snapshots will take less time to run and won't take up much more space as infrequent snapshots.

bryan-harter: Ah that's interesting! So when you restore, it automatically figures out how to aggregate the million snapshots? 

Do you know what would happen if we received like 100 files at once and opensearch was told to backup 100 times in rapid succession? Does it just listen to one at a time and queue up the others? I'm just curious if opensearch would get stressed out from all the requests or if it does some magic on the backend.

Issue #107, opened 2023-07-26
Implement S3 Backups


maxinelasp:
### Summary

Per the spike #53, we need a backup system to back up the data S3 bucket to the backup account. To do this, we will need to create a backup bucket in the different account, and set up both the main bucket and the backup bucket to backup data. 

### Requirements

The more specific the better.
  - [x] Create a CDK deployment scheme for the backup account which creates and sets up the S3 bucket
  - [x] Modify the main CDK code to send back ups to the other account
  - [x] Update tests (if possible? Might only be possible in integration testing)
  - [x] Confirm that any files in S3 are copied over

### Related tickets

#53 
#67 
#93

PR #125, opened 2023-08-04
S3 backups

maxinelasp:
# Change Summary
Per ticket #107, create the CDK system for automatically backing up all files in the S3 data bucket. 

## Overview

This system is designed to be run in two accounts, although you can run both steps in the same account. You change the `CONTEXT` variable in the personal app template to change the context to deploy to - the backup account is set to "backup". If you are deploying to the backup account, you then have to change your profile to the backup account credentials.

Also required is setting the "CDK_S3_BACKUPS_SOURCE_ACCOUNT" environment variable to specify what account your source bucket is set up in. 

There are two manual steps that I haven't yet figured out. 
1. When you deploy the SdsDataManager stack in the dev or source account, you need to manually input the role name that is created into backup_bucket_stack.py. This name is generated with different characters each time and I haven't figured out what the template is yet to replicate it, or how to pass the ARN across different deploy steps. 
2. You need to manually create the replication rule in the source bucket. This takes about 2 minutes, and I think there is some way to do this in CDK, but I don't want to delay getting this PR out ahead of SIT-2. 

If you'd like to test cross-account replication, you can use my data bucket in the backup account (`sds-data-mh-backup`). Or, you can deploy both pieces to the same account and they should work. 

There aren't any dependencies across the new backup bucket stack and the sds-data-manager stack so they can be deployed separately. However, this does mean I make some manual assumptions. I would be open to passing some information back and forth to reduce those dependencies if people have any ideas

## New Dependencies
None

## New Files

- backup_bucket_stack.py
   - Creates the backup bucket and sets up permissions using the new role in sds_data_manager.py
   
## Updated Files

- sds_data_manager.py
   - Created a new role for replication permissions
- stackbuilder.py
   - added new util function to create backup bucket stacks
- app_template_dev.py
   - created new flow for accessing backup bucket stack  

## Testing

Skipped testing for now to get this up to review. I would expect this to just check the permissions on each role. I think checking the actual backups themselves are for integration testing - is that necessary at this point? What do people think?


maxinelasp:
Tests failed because it checks the number of roles and I didn't update it. I will fix now.

maxinelasp:
Here is what I see as potential solutions to not knowing the Role name: 
1. Just figuring out a way to get the Role arn - this is obviously best, but may not be possible. I'll continue looking into this.
2. Having the backup stack depend on the sds-data-manager stack - this isn't ideal because they otherwise are not really connected. This is beneficial because the backup stack doesn't need to be deployed more than once, and is in an entirely different account, so I don't want to add more dependencies than needed. I also am not sure if it work to deploy to two different accounts and have them connected - I think that's generally something CDK discourages. Also, since the stacks kind of depend on each other (sds-data-manager needs the backup bucket name), it seems like it would cause problems. 
3. Passing via "[Outputs](https://bobbyhadz.com/blog/aws-cdk-outputs)" - CDK has a method for outputting information which is then accessible to other stacks. These outputs can be written to a local file and then accessed from there. This would require that the sds-data-stack was run before backup-buckets stack on the same machine, which is kind of annoying, but also fine. This is probably my preferred method, but I could see it introducing issues later with automatic deployments (but the backup account probably won't be deployed as much and maybe we can manually move that output file around.)
4. writing to the context - I could write the role arn to cdk.json, but that would probably cause problems with git deployments if it ever gets committed by accident or something. 
5. Guessing at the name of the role - this is how I get the bucket name, but that's easier to guess. There is some template out there somewhere in cloudformation which determines this name, but I'm not sure if I could consistently get the correct name. Also, guessing at the name kind of sucks. 

Thoughts?

sds_data_manager/stacks/backup_bucket_stack.py
```diff
@@ -0,0 +1,96 @@
...
+class BackupBucket(Stack):
+    """
+    This stack creates the destination bucket for data backups. It can be run in
+    the same account as SdsDataManager, or in a separate account.
+    The source_account is a required parameter. This source account should be the
+    AWS account for the source bucket.
+
+    For replication to work, you also need to deploy SdsDataManager and create
+    the source bucket and replication role. Then, you need to manually update
+    the role_arn variable with the replication role created.
```
tech3371, line 24: I saw a question on your PR comment about how to add dependency. In CDK, there is a way to create a resource only after its dependent resources are created. For example, something like 
```
backup_bucket.node.add_dependency(sds_data_manager)
backup_bucket.node.add_dependency(source_bucket)
backup_bucket.node.add_depencency(replication_role)
```
What this does is, it makes sure all the source for the `backup_bucket` is created before it creates backup_bucket. I hope this helps.

sds_data_manager/stacks/backup_bucket_stack.py
```diff
@@ -0,0 +1,96 @@
...
+        sds_id: str
+        env : Environment
+            Account and region
+        source_account : str
+            Account number for the source S3 bucket
+        """
+        super().__init__(scope, construct_id, env=env, **kwargs)
+
+        # FOR NOW: Deploy other stack, update this name with the created role.
+        role_arn = (
```
tech3371, line 59: To get role_arn, I think you can use this `from_role_name()` from this link: https://docs.aws.amazon.com/cdk/api/v1/python/aws_cdk.aws_iam/Role.html#aws_cdk.aws_iam.Role.from_role_name.

Above returns an IRole. Then get `role_arn` from it?

Eg. 
```
iam_role_by_look_up = iam.Role.from_role_name(self, 'roleLookup', role_name=role_name)
role_arn = iam_role_by_look_up.role_arn
```
I think that's the syntax. I haven't tested it.

sds_data_manager/stacks/backup_bucket_stack.py
```diff
@@ -0,0 +1,96 @@
...
+            "arn:aws:iam::449431850278:"
+            "role/SdsDataManager-mh-dev-BackupRoleF43CFD90-11WK8QFUD7HTG"
+        )
+
+        # This is the S3 bucket used by upload_api_lambda
+        backup_bucket = s3.Bucket(
```
tech3371, line 65: Is this creating bucket again or is it creating same bucket in the backup account or region? If second case, should we add something to its name that it's a backup bucket?

sds_data_manager/stacks/backup_bucket_stack.py
maxinelasp: I can look into this, but I don't think it will work, because the sds_data_manager is created in a different account. When  deploying the backup bucket, it won't be able to see the sds_data_manager in the dev account.

sds_data_manager/stacks/backup_bucket_stack.py
maxinelasp: Unfortunately, that still requires the role name, which is the piece of information I don't know.

sds_data_manager/stacks/backup_bucket_stack.py
maxinelasp: It is creating a new bucket in the backup account. This does result in the name being `sds-data-[initial]-backup`, due to the way the sds-id is created. Is that sufficient, or would you like me to change the name of the bucket to something else? Maybe `sds-backup-[initial]-backup`?

tech3371: I see. We could add comment saying that this part is creating backup bucket in a different account to minimize confusion because they look exactly same when we look at it first. :)

or add suffix `backup`. You know how sds_id is set to `dev` or `prod`. Will `sds_id` be set to `backup` for backup account? If that's the case, then we are good.

A question. If we keep same exact bucket name convention for both source account and backup account, will it not give "bucket name is not unique" error?

sds_data_manager/stacks/backup_bucket_stack.py
tech3371: Ah, I see. Because it lives in different account?

sds_data_manager/stacks/backup_bucket_stack.py
tech3371: I think there is way to add this dependency at very high level like in `app.py` but we can look at it post SIT-2.

sds_data_manager/stacks/backup_bucket_stack.py
tech3371: ignore my comments. You already do this!

app_template_dev.py
```diff
@@ -52,20 +55,39 @@
...
-    app,
-    env=env,
-    sds_id=sds_id,
-    use_custom_domain=True,
-)
+if params["sds_id"] == "backup":
+    if not s3_source_account:
+        raise KeyError(
+            "No source account is set for the backup deploy."
+            "Please define the CDK_SOURCE_ACCOUNT environment variable."
```
tech3371: should this be `CDK_S3_BACKUPS_SOURCE_ACCOUNT` instead?

app.py
```diff
@@ -9,6 +9,9 @@
 This app is designed to be the dev and production deployment app.
 It defaults to a dev deployment via a default `env` value in cdk.json.
 To deploy to prod, specify `--context env=prod`.
+To deploy to the backup account (only deploys required backup stacks),
+specify `--context env=backup`.
+
```
tech3371, line 14: It's not related to your PR but wanted to mention that at one point, we need to update this `app.py` to match `app_template_dev.py`. Also, I don't know when we use `app.py` vs `app_template_dev.py`. I will create ticket for this task.

tech3371 (approved):
It looks good to me! Let me know if need anything else.

sds_data_manager/stacks/backup_bucket_stack.py
maxinelasp: Yes, the role is in a different account (dev) while this IAM role is in the backup account. 

That method actually also requires the role to be in the same account, so even if I did know the name it wouldn't work.

sds_data_manager/stacks/backup_bucket_stack.py
maxinelasp: Yes, the sds_id is set to `backup` for the backup account, so all the deployed resources have "backup" in the name! Does that all make sense?

app_template_dev.py
maxinelasp: Yes! Thanks for the catch, I changed the name

app.py
maxinelasp: I wasn't clear what parts of app.py needed to be updated to match with app_template_dev.py so I left it as is. I'm happy to copy over app_template_dev.py and remove all the initial specific stuff if that's preferable!

sds_data_manager/stacks/backup_bucket_stack.py
tech3371: yes. It does now!

app.py
tech3371: me too. I created a ticket for this https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/127

sdhoyt (approved):

bryan-harter:
I tried it out and it worked great! I uploaded three files, and saw them replicate into another account.  I delete the files on the primary account, and saw they still stayed put in the backup account.  

My only comment is that there's a lot of manual steps.  It's probably unavoidable, but it would be nice if they were somehow integrated into the code.

PR #151, opened 2023-08-28
SIT-2 Release [dev]

bourque:
This PR marks the release of version `2.0.0`, which corresponds to the completion of SIT-2. This PR is being merged back into `dev` per the nominal [git & GitHub workflow](https://imap-processing.readthedocs.io/en/latest/development/style-guide/git-and-github-workflow.html#git-and-github-workflow) and [release workflow](https://imap-processing.readthedocs.io/en/latest/development/release-workflow.html).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry]
 name = "sds-data-manager"
-version = "0.1.0"
+version = "2.0.0"
 description = "IMAP Science Operations Center AWS data manager"
 authors = ["IMAP SDS Developers <imap.sdc@lists.lasp.colorado.edu>"]
 readme = "README.md"
--- a/sds_in_a_box/sds_in_a_box_stack.py
+++ b/sds_in_a_box/sds_in_a_box_stack.py
@@ -26,7 +26,8 @@ class SdsInABoxStack(Stack):
                                 #bucket_name="DataBucket",
                                 versioned=True,
                                 removal_policy=RemovalPolicy.DESTROY,
-                                auto_delete_objects=True
+                                auto_delete_objects=True,
+                                block_public_access=s3.BlockPublicAccess.BLOCK_ALL
                                 )
 
         # Need to make a secret username/password for OpenSearch


greglucas:
Seeing how we missed commits from the previous time of doing this, how would you feel about instead of going from `v2.0.0 -> dev` (currently done here), we go from `main -> dev` **after** `v2.0.0 -> main`? This would make the release a two-step dependent PR process instead of in parallel. Thoughts?

bourque:
@greglucas I can't say I have a strong opinion here, and I think your proposed method would work just fine, though I think the workflow we are using here is sound. I think the main reason we missed commits this time around was because there were direct commits to `main`, which is something we should be really trying to avoid.

FWIW, I did a quick search of ['GitHub release workflow'](https://www.google.com/search?sca_esv=560734445&q=github+release+workflow&tbm=isch&source=lnms&sa=X&ved=2ahUKEwjw6LGWnYCBAxWvm2oFHUrkC8MQ0pQJegQIDBAB&biw=1920&bih=967&dpr=1) and the vast majority of the diagrams have the release branch being merged into `dev` and `main`, which makes me think that this is an industry standard (I think people refer to this as "git Flow"). Not saying this is a good reason to stick with 'the norm' necessarily, but just wanted to point it out.

greglucas:
Sounds good to me! Lets stick with what you've got here then.

I think part of my confusion here may be this piece of text from the Atlassian doc "Once the release is ready to ship, it will get merged it into main and develop, then the release branch will be deleted." I'm generally used to keeping multiple release branches around, so we'd have branches: `v1.x`, `v2.x`, ... with our releases that we could backport things to if need be. The `main` branch here is really saying we are only going to have one supported release at all times I think? So in my mind I can switch "main" with "v2.x" and those would be analagous in our current situation.

Regardless, I say we're good to go here!

greglucas (approved):

bourque:
@greglucas I agree, I also like to keep release branches around for those reasons you mention. I'm not sure why Atlassian likes to delete theirs 😂

laspsandoval (approved):

PR #179, opened 2023-10-12
S3 Bucket Triggers

laspsandoval:
# Change Summary

## Overview
Simple way to kick-off the Step Functions based on objects created in s3 buckets. 

## New Files
- dependencies.json
   - Each instrument team will populate this document based on their dependencies from other instruments, their own instrument, or spice kernels. 
- get_dependency.py
   - This just provides a list of the dependencies provided in dependencies.json

## Deleted Files
None

## Updated Files
- stackbuilder.py : made a string into a list so that each instrument could have multiple dependencies
- processing_stack.py : added more rules to EventBridge
- l1b_codice.py and l1c_codice.py: can handle multiple buckets now
- instrument_lambdas.py: can handle multiple buckets now

## Testing
1. Deploy the  L1bCodiceProcessing-<sds_id>-dev and L1cCodiceProcessing-<sds_id>-dev stacks.
2. Create s3 buckets based on dependencies listed in dependencies.json (see my bucket for example: [s3://sds-data-ls9-dev])
3. Make certain step functions begin when objects are dropped in appropriate directories.

Note: The step function will fail because there is not Docker image in the ECR repos (unless you put one there). That doesn't really matter. This PR just shows that we have a way to trigger the Step Function. Also, this is not the last of the trigger code. The manual trigger will be in a near-future PR.

### DIFF
--- a/sds_data_manager/constructs/instrument_lambdas.py
+++ b/sds_data_manager/constructs/instrument_lambdas.py
@@ -1,5 +1,5 @@
 """Module containing constructs for instrumenting Lambda functions."""
-
+import json
 from pathlib import Path
 
 from aws_cdk import Duration
@@ -20,7 +20,7 @@ class InstrumentLambda(Construct):
         data_bucket: s3.Bucket,
         code_path: str or Path,
         instrument_target: str,
-        instrument_sources: str,
+        instrument_sources: list,
     ):
         """
         InstrumentLambda Constructor.
@@ -37,7 +37,7 @@ class InstrumentLambda(Construct):
             Path to the Lambda code directory
         instrument_target : str
             Target data product (i.e. expected product)
-        instrument_sources : str
+        instrument_sources : list
             Data product sources (i.e. dependencies)
         """
 
@@ -47,7 +47,7 @@ class InstrumentLambda(Construct):
         # TODO: if we need more variables change so we can pass as input
         lambda_environment = {
             "S3_BUCKET": f"{data_bucket.bucket_name}",
-            "S3_KEY_PATH": instrument_sources,
+            "S3_KEY_PATH": json.dumps(instrument_sources),
             "INSTRUMENT_TARGET": instrument_target,
             "PROCESSING_NAME": processing_step_name,
             "OUTPUT_PATH": f"s3://{data_bucket.bucket_name}/{instrument_target}",
--- a/sds_data_manager/lambda_code/SDSCode/instruments/l1b_codice.py
+++ b/sds_data_manager/lambda_code/SDSCode/instruments/l1b_codice.py
@@ -2,6 +2,7 @@
 """
 import logging
 import os
+import json
 from datetime import datetime
 
 import boto3
@@ -15,7 +16,6 @@ logger = logging.getLogger(__name__)
 
 def lambda_handler(event: dict, context):
     """Handler function"""
-    """Handler function"""
     logger.info(f"Event: {event}")
     logger.info(f"Context: {context}")
 
@@ -24,15 +24,16 @@ def lambda_handler(event: dict, context):
 
     # Get the environment variables
     bucket = os.environ["S3_BUCKET"]
-    prefix = os.environ["S3_KEY_PATH"]
+    prefixes = json.loads(os.environ["S3_KEY_PATH"])
 
     # Retrieves objects in the S3 bucket under the given prefix
-    try:
-        s3 = boto3.client("s3")
-        object_list = s3.list_objects_v2(Bucket=bucket, Prefix=prefix)["Contents"]
-        logger.info(f"Object list: {object_list}")
-    except KeyError:
-        logger.warning("No files present.")
+    for prefix in prefixes:
+        try:
+            s3 = boto3.client("s3")
+            object_list = s3.list_objects_v2(Bucket=bucket, Prefix=prefix)["Contents"]
+            logger.info(f"Object list: {object_list}")
+        except KeyError:
+            logger.warning("No files present.")
 
     # TODO: this state will change based on availability of data
     # TODO: we need to think about what needs to be passed into the container
--- a/sds_data_manager/lambda_code/SDSCode/instruments/l1b_codice.py
+++ b/sds_data_manager/lambda_code/SDSCode/instruments/l1b_codice.py
@@ -1,8 +1,8 @@
 """Lambda runtime code that triggers off of arrival of data into S3 bucket.
 """
+import json
 import logging
 import os
-import json
 from datetime import datetime
 
 import boto3
--- a/sds_data_manager/lambda_code/SDSCode/instruments/l1c_codice.py
+++ b/sds_data_manager/lambda_code/SDSCode/instruments/l1c_codice.py
@@ -1,8 +1,8 @@
 """Lambda runtime code that triggers off of arrival of data into S3 bucket.
 """
+import json
 import logging
 import os
-import json
 from datetime import datetime
 
 import boto3
--- a/sds_data_manager/lambda_code/SDSCode/instruments/l1c_codice.py
+++ b/sds_data_manager/lambda_code/SDSCode/instruments/l1c_codice.py
@@ -2,6 +2,7 @@
 """
 import logging
 import os
+import json
 from datetime import datetime
 
 import boto3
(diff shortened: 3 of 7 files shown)


sds_data_manager/utils/dependencies.json
```diff
@@ -0,0 +1,5 @@
+{
+    "l1a_Codice": ["l0_Codice"],
+    "l1b_Codice": ["l1a_Codice", "l1a_Mag"],
+    "l1c_Codice": ["l1b_Codice", "l2_Ultra", "spice"]
```
maxinelasp, line 4: Is this intended for all dependencies eventually? I'm a little worried this file will get massively out of hand. Also, as an imap specific piece of information, it might not belong in this repository. I'm thinking more and more this should eventually become a database. As for right now for the initial pass, this is totally cool, but I could see it getting out of hand...

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -37,7 +37,7 @@ def __init__(
             Path to the Lambda code directory
         instrument_target : str
             Target data product (i.e. expected product)
-        instrument_sources : str
+        instrument_sources : list
```
maxinelasp, line 40: should this type be `list[str]`? also what does this look like, is it a S3 path? Can you add that to the docstring here?

sds_data_manager/lambda_code/SDSCode/instruments/l1b_codice.py
```diff
@@ -24,15 +24,16 @@ def lambda_handler(event: dict, context):
...
-        object_list = s3.list_objects_v2(Bucket=bucket, Prefix=prefix)["Contents"]
-        logger.info(f"Object list: {object_list}")
-    except KeyError:
-        logger.warning("No files present.")
+    for prefix in prefixes:
+        try:
+            s3 = boto3.client("s3")
+            object_list = s3.list_objects_v2(Bucket=bucket, Prefix=prefix)["Contents"]
+            logger.info(f"Object list: {object_list}")
+        except KeyError:
```
maxinelasp, line 35: Can we make the logger description more descriptive? (i.e. "missing the following files: ...") If this is temporary and planned to be removed, then it's fine as-is

sds_data_manager/lambda_code/SDSCode/instruments/l1c_codice.py
```diff
@@ -1,5 +1,6 @@
 """Lambda runtime code that triggers off of arrival of data into S3 bucket.
 """
+import json
 import logging
 import os
```
maxinelasp, line 5: What is the difference between these lambdas? Can they just be the same?

sds_data_manager/stacks/processing_stack.py
```diff
@@ -94,18 +94,19 @@ def __init__(
...
-                detail_type=["Object Created"],
-                detail={
-                    "bucket": {"name": [data_bucket.bucket_name]},
-                    "object": {"key": [{"prefix": f"{instrument_sources}"}]},
-                },
-            ),
-        )
+        # Kicks off Step Function as a result of object ingested into directories in
+        # s3 bucket (instrument_sources).
+        for source in instrument_sources:
```
maxinelasp, line 99: Nice, this looks good to me!

maxinelasp (commented):
Looks like a great first pass, I'm excited to see it! There's definitely a few outstanding questions but those can be answered in the future.

sds_data_manager/lambda_code/SDSCode/instruments/l1b_codice.py
```diff
@@ -24,15 +24,16 @@ def lambda_handler(event: dict, context):
...
     # Retrieves objects in the S3 bucket under the given prefix
-    try:
-        s3 = boto3.client("s3")
-        object_list = s3.list_objects_v2(Bucket=bucket, Prefix=prefix)["Contents"]
-        logger.info(f"Object list: {object_list}")
-    except KeyError:
-        logger.warning("No files present.")
+    for prefix in prefixes:
+        try:
+            s3 = boto3.client("s3")
```
greglucas, line 32:
```suggestion
    s3 = boto3.client("s3")
    for prefix in prefixes:
        try:
```
Move the client up so we only get one client outside the loop and not request a new one each time. I'm not sure if it would also make sense to get the bucket right away outside the loop and then filter on prefixes after that?

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -47,7 +47,7 @@ def __init__(
         # TODO: if we need more variables change so we can pass as input
         lambda_environment = {
             "S3_BUCKET": f"{data_bucket.bucket_name}",
-            "S3_KEY_PATH": instrument_sources,
+            "S3_KEY_PATH": json.dumps(instrument_sources),
```
greglucas, line 50: Change the name to `S3_INSTRUMENT_SOURCE_PREFIXES` or something similar since now there could be more than one path. Is it actually prefixes, or do we need explicit individual keys for each source? If it is individual keys you could avoid the list objects and go to something like Object.load(): https://stackoverflow.com/questions/33842944/check-if-a-key-exists-in-a-bucket-in-s3-using-boto3

sds_data_manager/utils/get_dependency.py
```diff
@@ -0,0 +1,25 @@
...
+    ----------
+    key : str
+        Target name
+
+    Returns
+    -------
+    packets : list
+        List of all the dependencies
+    """
+    dependency_path=Path(sys.modules[__name__.split(
```
greglucas, line 20: Avoid sys.modules if possible. `Path(__file__)` should work I think?

sds_data_manager/stacks/processing_stack.py
greglucas: Is this kicking off a new step function each time one of the dependencies gets a new file? Or is it kicking off an event and a single step function would wait for all of the dependencies to trigger their events?

I'm wondering if we'll have a lot of step functions kicked off and most of them just fail until one of them meets the conditions, or is it more like one step function is kicked off the first time and then has some Waits associated with it that check for new data?

laspsandoval: Yes. It is a new step function each time. We had a "WAIT" in the original step function design that I can easily add back in. I agree. I think it is the right thing to do.

sds_data_manager/utils/dependencies.json
laspsandoval: This is for the s3 bucket data triggers only. Calibration checks will happen in the lambas. I see it as being static and not complicated. But if it does get that way then we can look at the information going into a database. Or it could go into the database regardless of how complicated it is if it helps to keep us organized. Good for now at least.

sds_data_manager/lambda_code/SDSCode/instruments/l1b_codice.py
laspsandoval: The lambas may be changed entirely. I'll put a todo in case we want to eventually keep this to have a more descriptive logger.

sds_data_manager/lambda_code/SDSCode/instruments/l1c_codice.py
laspsandoval: They are the same. But eventually each instrument will modify their own lambda. Right now they exist only for the purposes of demonstration.

sds_data_manager/stacks/processing_stack.py
laspsandoval: Ok, thought about this more. Right now we are using event-based triggering for the step functions. The dependencies will be listed in dependencies.json so:

"l1b_Codice": ["l1a_Codice", "l1a_Mag"]

l1b codice processing will kick off anytime new objects are dropped in the l1a_Codice or l1a_Mag buckets. The lambda within the step function will check to see if requirements for processing are met and end (not fail) the step function if not. If we add a "Wait" in the step function there will still be step functions that are triggered since we are using event-based processing.

sds_data_manager/lambda_code/SDSCode/instruments/l1c_codice.py
maxinelasp: What kind of differences will exist in instrument-specific lambdas? Could we have one base lambda that everyone uses and maybe change that for specific instrument changes? I'm not sure what differences there will be across instruments, but I would prefer to try very hard to keep the instrument processing steps the exact same for all instruments.

sds_data_manager/utils/dependencies.json
maxinelasp: Ah yes, I understand. Ok, that is probably alright for now, although I agree with you that we can keep an eye on it and decide if it's overly complicated. It shouldn't affect this PR either way, though.

maxinelasp (approved):
Looks good as a first pass, although there are a few outstanding questions this PR brought up for discussion

laspsandoval:
We have changes that we are working on for Trigger code. Closing this PR and will open another.

PR #180, opened 2023-10-12
Starting point for RDS

laspsandoval:
# Change Summary

## Overview
This is a start of the RDS. It creates the RDS but also has permissions added so that the lambda and Batch Job can access it. It also allow a user to access it if that user has a LASP IP address (see attachment). 

## New Files
- database_stack.py : database stack

## Updated Files
- stackbuilder.py: passes in parameters for RDS; where RDS stack is built
- processing_stack: passes in parameters for RDS
- networking_stack.py: add a security group for RDS, add ingress role for RDS
- instrument_lambdas.py: add environment variable, pass in parameters, grants lambdas access to RDS
- batch_compute_resources: add RDS access to Batch Job
- sdc_step_function.py: add environment variable to Batch Job
- l1b_codice: provides info so that lambda will eventually be able to access RDS
- l1c_codice: provides info so that lambda will eventually be able to access RDS

## Testing
I confirmed that this could deploy. More work is needed: create schemas, tables, make certain lambda can access database, etc.


laspsandoval:
Screen Shot 2023-10-12 at 4 26 05 PM

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -39,6 +45,14 @@ def __init__(
             Target data product (i.e. expected product)
         instrument_sources : str
             Data product sources (i.e. dependencies)
+        rds_security_group : ec2.SecurityGroup
+            RDS security group
+        subnets : ec2.SubnetSelection
+            RDS subnet selection.
+        db_secret_name : str
+            RDS secret name for secret manager access
```
maxinelasp, line 53: Is this variable just going to be passed around like this? Would it make sense to combine this into some kind of "config.py" file or something that can hold all of these specific names? Then we could pass around that file or an instance of the class or something and have all the info in there.

sds_data_manager/lambda_code/SDSCode/instruments/l1c_codice.py
```diff
@@ -24,6 +25,23 @@ def lambda_handler(event: dict, context):
     # Get the environment variables
     bucket = os.environ["S3_BUCKET"]
     prefix = os.environ["S3_KEY_PATH"]
+    secret_name = os.environ['SECRET_NAME']
```
maxinelasp, line 28: Rather than putting this secret info in the lambda handler, we should try and connect it directly to the lambda while we're building it. It looks like AWS provides a lambda option with RDS, so that might work. But we shouldn't put this info into the lambda itself if we can avoid it, I think. https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/rds-lambda-tutorial.html

sds_data_manager/stacks/database_stack.py
```diff
@@ -0,0 +1,97 @@
...
+        secret_name : str,
+            Database secret_name for Secrets Manager
+        database_name : str,
+            Database name
+        """
+        super().__init__(scope, construct_id, env=env, **kwargs)
+
+        self.secret_name = "sdp-database-creds"
+        database_name = "imap"
+        username = "imap_user"
```
maxinelasp, line 63: Would it make sense (long term, not in this PR) to have instrument-specific users?

sds_data_manager/stacks/processing_stack.py
```diff
@@ -33,6 +33,10 @@ def __init__(
         instrument_target: str,
         instrument_sources: str,
         repo: ecr.Repository,
+        batch_security_group: ec2.SecurityGroup,
```
maxinelasp, line 36: this constructor has a lot of inputs, does it make sense to combine any of them into a `NamedTuple` or something?

maxinelasp (commented):
Overall this looks great, I'm concerned about the secret handling inside the lambda handler function

sds_data_manager/utils/stackbuilder.py
```diff
@@ -19,7 +21,13 @@
 
 
 def build_sds(
-    scope: App, env: Environment, sds_id: str, use_custom_domain: bool = False
+        scope: App,
+        env: Environment,
+        sds_id: str,
+        rds_size: str = 'SMALL',
```
maxinelasp, line 27: I'm concerned with how many arguments this function has. Can you think of a good way to reduce the number of arguments in here? Maybe by putting all the database settings into a config class?

sds_data_manager/lambda_code/SDSCode/instruments/l1b_codice.py
```diff
@@ -25,6 +25,23 @@ def lambda_handler(event: dict, context):
...
+        region_name="us-west-2")
+    secret_value_response = client.get_secret_value(
+        SecretId=secret_name
+    )
+
+    secret_object = json.loads(secret_value_response['SecretString'])
+    secret_object["host"]
+    secret_object["username"]
+    secret_object["dbname"]
+    secret_object["password"]
```
tech3371, line 44: I was wondering how we construct db uri to connect and query. Looks like this secret_object contains all information we need to construct db uri. Thanks for pulling these information out.

sds_data_manager/lambda_code/SDSCode/instruments/l1c_codice.py
tech3371: I think we are creating secret at CDK build time(I am unsure of where it happens). It gets passed as env variable to the lambda at CDK build time. Then we are referencing env variable here to get below information to connect to RDS. I think you both are thinking similar thing if I understood correctly.

This is from link Maxine shared:
```
# rds settings
user_name = os.environ['USER_NAME']
password = os.environ['PASSWORD']
rds_proxy_host = os.environ['RDS_PROXY_HOST']
db_name = os.environ['DB_NAME']
```
This is from Laura's code:
```
    secret_object = json.loads(secret_value_response['SecretString'])
    secret_object["host"]
    secret_object["username"]
    secret_object["dbname"]
    secret_object["password"]
```

sds_data_manager/stacks/database_stack.py
```diff
@@ -0,0 +1,97 @@
...
+from aws_cdk import (
+    aws_ec2 as ec2,
+)
+from aws_cdk import (
+    aws_rds as rds,
+)
+from constructs import Construct
+
+
+class SdpDatabase(Stack):
```
tech3371, line 16: What is `Sdp`?

sds_data_manager/stacks/database_stack.py
```diff
@@ -0,0 +1,97 @@
...
+            connection=ec2.Port.tcp(5432),
+            description="Ingress RDS",
+        )
+
+        # Lambda was put into the same security group as the RDS, but we still need this
+        rds_security_group.connections.allow_internally(ec2.Port.all_traffic(),
+                                                        description="Lambda ingress")
+
+        # Secrets manager credentials
+        rds_creds = rds.DatabaseSecret(self, "RdsCredentials",
```
tech3371, line 77: Is this where it's creating RDS password using secrets manager?

sds_data_manager/utils/stackbuilder.py
tech3371: I do agree here. Those new argument are RDS specific and it make sense to refactor it out somewhere else.

tech3371 (commented):
It looks good as first pass. I just had minor comment. Can you create issue ticket for

- remove LASP ip address access
- create instrument specific user in RDS (?)

sds_data_manager/constructs/batch_compute_resources.py
```diff
@@ -48,6 +51,10 @@ def __init__(
             S3 bucket.
         repo : ecr.Repository
             Container repo
+        batch_security_group: ec2.SecurityGroup
+            Batch security group
+        db_secret_name : str
```
greglucas, line 56:
```suggestion
        rds_secret_name : str
```
Since you're calling it rds_secret below it'd be good to be consistent in the naming everywhere.

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -21,6 +23,10 @@ def __init__(
         code_path: str or Path,
         instrument_target: str,
         instrument_sources: str,
+        rds_security_group: ec2.SecurityGroup,
+        subnets: ec2.SubnetSelection,
+        db_secret_name: str,
+        vpc: ec2.Vpc,
```
greglucas, line 29: Can you pass in the networking stack completely and then access what you need after that?

sds_data_manager/utils/stackbuilder.py
greglucas: :+1: to the above. I guess the question is "how many of these knobs do we actually want to mess with?" versus just hard-coding them and putting in comments about why you chose something. If we do want the knobs, I think we could put them in the `cdk.json` configuration area possibly so we could choose different settings for each account. But, I'm not sure how many of these we really need versus just selecting a value and going with it.

greglucas (commented):
Needs to have black run on it to satisfy pre-commit hooks. But, it looks good to me. I am OK with opening up LASP IPs if that makes development easier. I was just assuming there was a way to assume an IAM role and have my IAM role access a database with the IAM credentials rather than username/password credentials. something like this to run my script with my access credentials locally `AWS_PROFILE=imap run_sql_query.py` is what I was hoping for.

laspsandoval:
Per Sean's request I am merging in.

PR #181, opened 2023-10-16
Share resources in a common environment/account and auto deploy resources

### LINKED ISSUE #127: Update `app.py` to match `app_template_dev.py`
### Topic:
Update `app.py`
### Description:
`app.py` is behind on changes made recently. Most recent changes are reflected in `app_template_dev.py`. We need to update `app.py` to match changes in `app_template_dev.py`. 

### Follow-up comments:
When do we use `app.py`?

greglucas:
# Change Summary

## Overview

This removes the `sds_id` parameter passed into all the resources and moves to a "shared" style of stacks, where we can all point to resources already deployed a little easier. Additionally, it adds an automatic deployment mechanism upon merge to the `dev` branch so our resources stay up-to-date.

closes #127

## New Dependencies
N/A

## New Files
N/A

## Deleted Files
`app_template_dev.py`: We aren't using the app_template files anymore, we are moving over to putting the configuration information into the cdk.json as extra parameters.

## Updated Files
Lots of updates... Mostly removing `sds_id`, and requiring an explicit `env` to be passed, just let `**kwargs` pick those up and pass them to the super() stack.

## Testing
Updated the unit tests to match the new resource ids.


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,48 @@
...
+        run: |
+          # Activate the poetry virtualenv where we installed our package
+          source $(poetry env info --path)/bin/activate
+          cdk synth --context account_name=dev
+
+      - name: Deploy
+        run: |
+          # Activate the poetry virtualenv where we installed our package
+          source $(poetry env info --path)/bin/activate
+          cdk deploy --all --context account_name=dev --require-approval never
```
tech3371, line 48: Does this argument `--require-approval never` also allows destroy without needing approval? For example, in the near future, if we remove opensearch stack from the code, would this be able to destroy the opensearch stack? just asking out of curiosity.

app.py
```diff
@@ -3,34 +3,41 @@
...
-
-account = params["account"]
-
-env = Environment(account=account, region=sds_region)
-print(f"Deploying to account {account} in region {sds_region}.")
-
-# Deploy SDS resources. This is the default with no CLI context variables set.
-stacks = build_sds(app, env=env, sds_id=params["sds_id"], use_custom_domain=True)
+account_name = app.node.get_context("account_name")
+account_config = app.node.get_context(account_name)
```
tech3371: Why are we assigning `app.node.get_context("account_name")` to those two variable?

app.py
```diff
@@ -3,34 +3,41 @@
...
+# Add the account_name to the account_config for later lookup
+account_config["account_name"] = account_name
+account = account_config["account"]
+region = account_config["region"]
```
tech3371: I think I follow it now but it was little confuse at first glance. 

This first `app.node.get_context("account_name")` is getting command line input. Then second `app.node.get_context("account_name")` is getting context from `cdk.json`. Then we add the command line input  `account_name`'s name to account_config. Then grab account number and region from the config.

docs/source/cdk/cdk-deployment.rst
```diff
@@ -1,32 +1,70 @@
...
+To deploy the application, you must have local credentials to be able to deploy to
+the specified account and region. This can be done through environment variables
+and AWS access key credentials::
 
         export AWS_PROFILE=<profile>
 
 You'll then need to synthesize the CDK code with the command::
 
-        cdk synth --app "python app_<initials>_dev.py"
+        cdk synth --context account_name="dev"
```
tech3371, line 29: If we want to skip command line input and go straight to cdk.json, will it app.py fail or give warning?

tech3371 (dismissed):
I just had some questions but otherwise it looks good to me!

cdk.json
```diff
@@ -15,18 +15,22 @@
     ]
   },
   "context": {
-    "env": "backup",
+    "account_name": "dev",
     "usernames": ["sandoval"],
+    "backup": {
```
laspsandoval, line 20: Does the account number for the backup need to be changed to dev?

.github/workflows/deploy.yml
greglucas: Yes, but I was thinking we would manually run the destroys for now. I have also had issues with still needing to be notified if trying to delete s3 buckets or something like that I think too. So there are still _some_ guardrails I believe.

app.py
greglucas: I'll add an extra comment about this to try and clarify, but there is also an `account_name` in the cdk.json already. So specifying it on the command line overwrites that value.

docs/source/cdk/cdk-deployment.rst
greglucas: `cdk synth` will use the defaults, which is currently "dev". But, it is better to be explicit here in case someone else changes something on you. So if you update `cdk.json` with the `"account_name": "tenzin"` then `cdk synth` will use that value.

app.py
greglucas: Just added additional comments, let me know if those still don't make sense.

tech3371: That is more clear. Thank you! The code still looks good to me.

sds_data_manager/utils/stackbuilder.py
```diff
@@ -51,65 +44,60 @@ def build_sds(
     # other step function processing steps
     processing_step_function = step_function_stack.ProcessingStepFunctionStack(
         scope,
-        f"ProcessingStepFunctionStack-{sds_id}",
-        sds_id,
+        "ProcessingStepFunctionStack",
         dynamodb_table_name=dynamodb.table_name,
         env=env,
```
laspsandoval, line 52: I think that you can now remove env=env.

app.py
```diff
@@ -3,34 +3,48 @@
...
+# once we have the account_name, get that section out of cdk.json
+account_config = app.node.get_context(account_name)
+# Add the account_name to the account_config for later lookup
+# and using it within the stacks without needing to pass an extra variable
+account_config["account_name"] = account_name
+account = account_config["account"]
+region = account_config["region"]
+
+env = Environment(account=account, region=region)
+print(f"Using the {account_name} account in the {region} region.")
```
laspsandoval: I find it helpful to print the account number here too.

laspsandoval (dismissed):
I had a few comments, but it looks good to me.

tech3371 (dismissed):
forgot to click approve

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::449431850278:role/GitHubDeploy
+          aws-region: us-west-2
+
+      - name: Synth
+        run: |
+          # Activate the poetry virtualenv where we installed our package
+          source $(poetry env info --path)/bin/activate
```
maxinelasp: suggestion: Change this line to `poetry shell`

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,48 @@
...
+          cdk synth --context account_name=dev
```
maxinelasp, line 42: How do you imagine eventually handling the account name in these deploys? Not super relevant for this PR, just curious if you have a vision

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,48 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - dev
+
+jobs:
```
maxinelasp, line 8: Did you go through and double check that no sensitive information is printed or logged as part of the deploy? It's something we're going to need to be vigilant about with these github deploys.

cdk.json
maxinelasp: We have a backup account which was used for testing, but for dev I think it should be dev.

cdk.json
```diff
@@ -15,18 +15,22 @@
     ]
   },
   "context": {
-    "env": "backup",
+    "account_name": "dev",
     "usernames": ["sandoval"],
```
maxinelasp, line 19: Should Laura's username be here still? What is this used for?

docs/source/cdk/cdk-deployment.rst
```diff
@@ -1,32 +1,70 @@
...
+Automatic Deployments
+~~~~~~~~~~~~~~~~~~~~~
+The CDK code is automatically deployed to the dev account when a pull request is
+merged into the dev branch. This is done through a GitHub Action that runs
+the ``cdk deploy`` command. The GitHub Action is defined in the
+``.github/workflows/deploy.yml`` file. It uses short lived credentials to deploy
+the code to the dev account. The credentials are obtained through an OIDC
+authentication process and an aws provided credentials action step.
+
+With a new account, there are several steps that are required to set this up
```
maxinelasp, line 55: Does this only need to be done once?

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -18,70 +19,58 @@ def __init__(
         self,
         scope: Construct,
         construct_id: str,
-        sds_id: str,
```
maxinelasp, line 21: :) I love seeing this method signature get smaller

sds_data_manager/stacks/backup_bucket_stack.py
```diff
@@ -64,8 +58,8 @@ def __init__(
         # This is the S3 bucket used by upload_api_lambda
         backup_bucket = s3.Bucket(
             self,
-            f"BackupDataBucket-{sds_id}",
-            bucket_name=f"sds-data-{sds_id}",
+            "BackupDataBucket",
+            bucket_name=f"sds-data-{source_account}-backup",
```
maxinelasp, line 64: This will result in a bucket name with a long string of numbers, which isn't really ideal... I like that it indicates the source bucket, but I don't think this is going to be useful.

sds_data_manager/stacks/lambda_stack.py
```diff
@@ -34,8 +33,6 @@ def __init__(
         ----------
         scope : Construct
             The scope of the construct.
-        sdd_id : str
```
maxinelasp, line 37: Needs a docstring for construct_id

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -54,25 +52,24 @@ def __init__(
...
             This class depends on dynamodb_stack, which is built with
             opensearch_stack.py
         processing_step_function_arn:
             This has step function arn
-        env : Environment
-            Account and region
         """
         super().__init__(scope, construct_id, env=env, **kwargs)
+        # Get the current account number so we can use it in the bucket names
+        account = env.account
```
maxinelasp, line 67: Is there some way to put an account alias in this env variable that gets passed around all over the place?

sds_data_manager/stacks/step_function_stack.py
```diff
@@ -25,13 +24,10 @@ class ProcessingStepFunctionStack(Stack):
     def __init__(
         self,
         scope: Construct,
-        id: str,
```
maxinelasp, line 28: These removed parameters need a removed docstring too, and construct_id needs a new one

sds_data_manager/utils/stackbuilder.py
```diff
@@ -51,65 +44,60 @@ def build_sds(
...
     instrument_list = ["Codice"]  # etc
 
-    lambda_code_directory = Path(__file__).parent / ".." / "lambda_code" / "SDSCode"
+    lambda_code_directory = Path(__file__).parent.parent / "lambda_code" / "SDSCode"
     lambda_code_directory_str = str(lambda_code_directory.resolve())
 
     for instrument in instrument_list:
         ecr = ecr_stack.EcrStack(
             scope,
-            f"{instrument}Processing-{sds_id}",
```
maxinelasp, line 128: Is this something we want to share? Don't we want to have a distinguishing name here?

maxinelasp (commented):
Few questions and comments, overall this looks like a good design

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -54,25 +52,24 @@ def __init__(
...
+        region = env.region
 
         # This is the S3 bucket used by upload_api_lambda
         self.data_bucket = s3.Bucket(
             self,
-            f"DataBucket-{sds_id}",
-            bucket_name=f"sds-data-{sds_id}",
+            "DataBucket",
+            bucket_name=f"sds-data-{account}",
```
laspsandoval, line 74: Right now this bucket_name is sds-data-account number. Change it to account name instead.

.github/workflows/deploy.yml
greglucas: Done. I thought I read something that didn't work if it was a remote CI environment or non-interactive shell.

.github/workflows/deploy.yml
greglucas: I was thinking we would make different steps for these cases and put an `if: branch_name == "main"` or `if: branch_name == "dev"`. But, I haven't thought real hard about this. I guess we could just put in `account_name=${GITHUB_REF##*/}` to get the git branch name to use. Then we'd need a dev and prod branch on this project for deploys.

.github/workflows/deploy.yml
greglucas: Discussed with you via slack.

cdk.json
greglucas: I didn't want to touch this stuff in this PR, but I agree I don't think this is what we should do. I think we should add tags to users or something like that which can be added to roles instead.

docs/source/cdk/cdk-deployment.rst
greglucas: Yes, correct.

sds_data_manager/stacks/backup_bucket_stack.py
greglucas: I don't think bucket names really matter at all, usually, you just get a random hash, so we could just use the default even and not specify a name. Here I am just saying I think we want a unique name for the bucket and one that we can point back to where it is coming from would possibly be helpful.

sds_data_manager/stacks/lambda_stack.py
greglucas: Thanks! I updated this and also updated all other scope/construct_id docstrings to be exactly the same text now since they were all different and the inconsistency bothered me.

sds_data_manager/stacks/sds_data_manager_stack.py
greglucas: This is an instance of the AWS Environment class, so I'm not sure we can add other names to it. We could just pass in the account_config here as well... @tech3371 said she was going to rearrange all of this, so I'd prefer leaving this alone for now.

sds_data_manager/utils/stackbuilder.py
greglucas: I figure each instrument only wants to create a single ECR stack for now, the images within those registries should be tagged differently though. I think this is something we should discuss about how to prototype as a team though without squashing people's work. We can add a configuration parameter to the `cdk.json` file somewhere with a unique id to pass around... But the problem is if I'm not developing on this specific piece I don't want that unique id to be passed around everywhere, only the specific piece I'm working on. So I figured it was easier to just manually while I develop go in and add my own "sds-id" to the stack I'm working on if I need it.

Other thoughts are welcome on this!

sds_data_manager/stacks/sds_data_manager_stack.py
greglucas: I don't think `sds-data-dev` is unique enough, and may be taken by someone else. I chose the account number because it is sort of a unique-ish value we can count on being available within each account (if you want to deploy this stack to your own "dev" account for instance). We don't care what the bucket is named really as long as we can access it programmatically.

app.py
greglucas: OK, I thought that would just be more confusing because I have no idea what account numbers belong to what, but I added it back in.

cdk.json
greglucas: I'm not following. We want to backup our dev to dev for now? Or do we want to put the backup account number into this configuration now? I figured that could come later as well, or do this step manually to hide that account specifically if we need it more locked down.

sds_data_manager/utils/stackbuilder.py
greglucas: I think we actually want to pass the `env=env` in all of these Stack class instance creations to make sure they are getting the proper environment passed in, it is just that subsequently we handle that via the `**kwargs` without needing to explicitly handle them inside the downstream stacks.

cdk.json
laspsandoval: I can remove it later. I added it for permission to push to the ECR.

cdk.json
laspsandoval: I just didn't understand why under back-up we listed "source_account" as dev and then put in the prod account number.

sds_data_manager/stacks/sds_data_manager_stack.py
laspsandoval: Alright that's fine.

sds_data_manager/utils/stackbuilder.py
maxinelasp: I suppose I can see it both ways. If we're expecting to regularly deploy large parts of the stack with unique names, then it would be really annoying to go through and add an sds-id to everything.

 On the other hand, if you are only deploying one or two things, then it's fine. I'm concerned that excessively sharing resources will cause problems, but at least for our development, I don't think that many people are working on the CDK code at the same time... and hopefully it's decoupled enough that we could share buckets and everything. 

I'm not sure what the best answer  is or what my preference is, so I guess defaulting to removing it and seeing if it causes problems makes sense to make the code cleaner.

tech3371 (approved):
I thought I approved a while ago but looks like I didn't. sorry!

PR #188, opened 2023-10-26
Ancillary data access from EFS

### LINKED ISSUE #120: [L5] Setup Access to Ancillary data
### Summary of the L5 requirement

Create EFS, add files to EFS, and setup batch job and lambda access for SPICE files. 

### Parent requirement

SOC-SDC-L4-29 | SDC Execution of L1 Processing Software | The SDC shall be capable of executing L1 processing software to produce L1 data products within seven (7) days of receipt of all requisite data inputs. | Requirement | The SDC produces L1 data products via Instrument Team supplied algorithms and SDC produced processing code. | SOC-L3-57 | Test | SIT-3: L0 -> L1 processing | SIT-3
-- | -- | -- | -- | -- | -- | -- | -- | --

### Tasks

- [x] Create EFS
- [x] Setup batch job access
- [x] Setup lambda access
- [x] Setup SPICE symlink for latest version

tech3371:
# Change Summary
Created EFS stack and added access to it from Lambda and Batch job. Related to this issue https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/120
 
## Overview
This review:
* create EFS
* Setup batch job access
* Setup lambda access
* Setup SPICE symlink for latest version

## New Files
- sds_data_manager/batch/efs-access-batch/
   - Added Dockerfile and batch job's code to read from EFS.
- sds_data_manager/lambda_code/efs_lambda/lambda_function.py
   - EFS write lambda code
- sds_data_manager/stacks/efs_lambda_stack.py and efs_stack.py
   - Stacks that creates EFS stack and lambda stack that writes data to EFS

## Updated Files
- sds_data_manager/constructs/batch_compute_resources.py
    - Added logic to mount EFS as volume on batch job
- cdk.json
   - Added EFS config variable there
   - Added above stacks and update postgres engine version since 14.2 has been deprecated.


tech3371:
All of changes on this PR is visible on this commit: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/188/commits/b5f21c3af4d1a7484b7a41471e5e1ea23c345a7d

.gitignore
```diff
@@ -9,6 +9,7 @@ __pycache__
 external
 .idea/
 .vscode/
+.DS_Store
```
greglucas: FYI, it might be helpful to set a global gitignore for stuff like this so you don't have to do that on every project, something like this if you have a `.gitignore` file in your home directory.
`git config --global core.excludesFile '~/.gitignore'`

cdk.json
```diff
@@ -31,6 +31,13 @@
       "domain_name": "imap-mission.com",
       "region": "us-west-2"
     },
+    "efs_mount_config": {
+      "volume_name": "EFS",
+      "efs_path": "/data",
+      "mount_path": "/mnt/spice",
+      "efs_access_point_id": "efs-access-pt-id",
+      "efs_fs_id": "efs-filesystem-id"
+    },
```
greglucas: I don't think these need to be config parameters. We should probably just set them inside the code to fixed values because they won't be changing between the different accounts.

sds_data_manager/constructs/batch_compute_resources.py
```diff
@@ -148,14 +154,50 @@ def __init__(
         # processing_step_name tag. I think this will require
         # setting up a lambda. Maybe there's another way?
         self.job_definition_name = f"fargate-batch-job-definition{processing_step_name}"
+
+        # Get EFS mount information
+        efs_mount_config = scope.node.get_context("efs_mount_config")
```
greglucas: I think you can pass in the efs stack and reference the filesystem information from there. I'm not sure if Batch requires the workaround with exporting/importing the function value but you could get the information that way too rather than needing to put it in the configuration. Something like this:

```
        # NOTE: Workaround to overcome EFS circular dependency when mounting a filesystem
        # https://github.com/aws/aws-cdk/issues/18759
        efs_access_point_id = Fn.import_value("efs-access-pt-id")
        lambda_efs_access = aws_lambda.FileSystem(
            arn=f"arn:aws:elasticfilesystem:{self.region}:{self.account}:access-point/{efs_access_point_id}",
            local_mount_path="/mnt/data",
        )
```

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
```diff
@@ -0,0 +1,140 @@
+import os
+import shutil
+
+import boto3
+
+# Define the paths
+mount_path = os.environ.get("EFS_MOUNT_PATH")
```
greglucas: Should we hard-code these across the project and specify something like "The SPICE kernels are always accessed from `/mnt/spice`? I don't see a huge benefit to making this flexible and could just cause other referencing issues if group permissions are set up differently because it changes or something like that.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
```diff
@@ -0,0 +1,140 @@
...
+
+attitude_symlink_path = f"{mount_path}/latest_attitude_kernel.ah.a"
```
greglucas: What date format do we get these from our partners in? If the date is more like this: `attitude_kernel_YYYYMMDD.ah.a` I would think we might want to change this to be `attitude_kernel_latest.ah.a` to put `latest` in the date area.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
```diff
@@ -0,0 +1,140 @@
...
+    source_path : str
+        Source path of the symlink
+    destination_path : str
+        Destination path of the symlink
+    """
+
+    # Check if the symlink exists
+    if os.path.islink(destination_path):
+        # Remove the old symlink
+        os.unlink(destination_path)
```
greglucas: Do you have to remove the old one or can you just overwrite it?

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # Create a new symlink pointing to the new file
+    os.symlink(source_path, destination_path)
```
greglucas: Use pathlib for all the path manipulations: https://docs.python.org/3/library/pathlib.html#pathlib.Path.symlink_to

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
```diff
@@ -0,0 +1,140 @@
...
+    s3_client = boto3.client("s3")
+
+    try:
+        # Download the file from S3
+        s3_client.download_file(s3_bucket, s3_key, download_path)
+        print(f"File downloaded: {download_path}")
+    except Exception as e:
+        print(f"Error downloading file: {e!s}")
+
+    # Move the file to the mount directory. Eg. /mnt/efs
```
greglucas: Can we just write/download directly to the mountpoint? I don't think we should need to do both operations here.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
```diff
@@ -0,0 +1,140 @@
...
+    # kernel has different prefix. Eg.
+    #   imap_recon_yyyymmdd_yyyymmdd_v##.bsp
+    #   imap_burn_yyyymmdd_yyyymmdd_v##.bsp
+    #   imap_pred_yyyymmdd_yyyymmdd_v##.bsp
+    elif filename.startswith("imap_nom") and filename.endswith(".bsp"):
+        create_symlink(filename, ephemeris_symlink_path)
+
+    print("Done with symlink creation")
+
+    print("After : ", os.listdir(mount_path))
```
greglucas: Remove these debugs? Or change to logging if you think we'll need it.

sds_data_manager/stacks/efs_lambda_stack.py
```diff
@@ -0,0 +1,147 @@
...
+            managed_policies=[
+                aws_iam.ManagedPolicy.from_aws_managed_policy_name(
+                    "service-role/AWSLambdaBasicExecutionRole"
+                ),
+                aws_iam.ManagedPolicy.from_aws_managed_policy_name(
+                    "service-role/AWSLambdaVPCAccessExecutionRole"
+                ),
+                aws_iam.ManagedPolicy.from_aws_managed_policy_name(
+                    "AmazonElasticFileSystemClientReadWriteAccess"
+                ),
```
greglucas: Is this already handled by the below `FullAccess` policy?

sds_data_manager/stacks/efs_lambda_stack.py
```diff
@@ -0,0 +1,147 @@
...
+            local_mount_path=efs_mount_config["mount_path"],
+        )
+
+        lambda_code = aws_lambda.Code.from_asset(
+            "sds_data_manager/lambda_code/efs_lambda"
+        )
+
+        boto3_layer_arn = (
+            f"arn:aws:lambda:{self.region}:770693421928:layer:Klayers-p311-boto3:2"
+        )
```
greglucas: Why do we need this layer? I thought boto3 was included with python runtimes already.

sds_data_manager/stacks/efs_lambda_stack.py
```diff
@@ -0,0 +1,147 @@
...
+            self,
+            "EfsWriteLambdaS3EventRule",
+            event_pattern=events.EventPattern(
+                source=["aws.s3"],
+                detail_type=["Object Created"],
+                detail={
+                    "bucket": {"name": [data_bucket.bucket_name]},
+                    "object": {
+                        "key": [
+                            {"prefix": "imap"},
```
greglucas:
```suggestion
                            {"prefix": "imap/spice"},
```
Can we add another prefix to just the spice folder as well?

sds_data_manager/stacks/efs_lambda_stack.py
```diff
@@ -0,0 +1,147 @@
...
+                            {"suffix": "ah.a"},
+                            {"suffix": ".bsp"},
+                        ]
+                    },
+                },
+            ),
+        )
+
+        # Add the Lambda function as the target for the rule
+        event_rule.add_target(targets.LambdaFunction(self.efs_list_runs_lambda))
```
greglucas:
```suggestion
        event_rule.add_target(targets.LambdaFunction(self.efs_spice_ingest_lambda))
```
I think this was a name holdover :)

sds_data_manager/stacks/efs_stack.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+        efs_mount_config = scope.node.get_context("efs_mount_config")
+
+        self.efs = efs.FileSystem(
+            self,
+            "EFS",
+            vpc=vpc,
+            file_system_name=efs_mount_config["volume_name"],
+            # This will automatically downscale infrequently accessed
+            # data to a cheaper storage tier after 7 days of inactivity.
```
greglucas:
```suggestion
            # data to a cheaper storage tier after 14 days of inactivity.
```

sds_data_manager/utils/stackbuilder.py
```diff
@@ -149,6 +163,14 @@ def build_sds(scope: App, env: Environment, account_config: dict):
         )
         # etc
 
+    # create EFS
+    efs_stack.EFSStack(scope, "EFSStack", networking.vpc, env=env)
```
greglucas: Create this earlier and then pass the stack information to later steps instead of needing the configuration parameters to be passed in earlier.

greglucas:
In the commit message could you put something about EFS in the first line so it is obvious at a glance what this commit contains? Right now it shows up as: `In this commit:`, I think that could be moved down a line and the header something like "Adding EFS for SPICE data handling"

sds_data_manager/constructs/batch_compute_resources.py
```diff
@@ -31,6 +32,8 @@ def __init__(
         batch_max_vcpus=10,
         job_vcpus=0.25,
         job_memory=512,
+        efs_mount=False,
```
laspsandoval: These are not used and not described in docstring.

sds_data_manager/stacks/efs_stack.py
```diff
@@ -0,0 +1,105 @@
...
+            "AccessPoint",
+            create_acl=efs.Acl(owner_gid="1000", owner_uid="1000", permissions="750"),
+            path=f"{efs_mount_config['efs_path']}/spice",
+            posix_user=efs.PosixUser(gid="1000", uid="1000"),
+        )
+
+        # Export the access point value to avoid a circular dependency between
+        # the EFS stack and API/Lambda stack
+        # https://github.com/aws/aws-cdk/issues/18759
+        CfnOutput(
```
laspsandoval, line 113: Interesting. I didn't know you could do that to avoid circular dependencies. I wonder if that would work for s3 buckets too. Thanks for the idea!

sds_data_manager/stacks/efs_lambda_stack.py
laspsandoval: We kind of brainstormed on this. Rather than setting up two EventBridges we talked about combining them. Maybe put as a TODO here for another PR. Anyways, attaching the picture I took for reference. 

IMG-5080

sds_data_manager/batch/efs-access-batch/efs_check.py
```diff
@@ -0,0 +1,32 @@
+import os
```
laspsandoval: I think you will need to explain to me the purpose of this code here. 

I'm also wondering if once the dust settles with Opensearch being replaced we might want to optimize our Dockerfiles so that we only have one for all Dockerized lambdas. But that would be for another PR.

laspsandoval (dismissed):
Minor comments. I would like to understand the purpose of efs_check.py a little bit better. I'll chat with you today. But overall I think everything looks really good.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
greglucas: There is a pathlib version as well with a `missing_ok` keyword argument so you can ignore the missing error: https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink

sds_data_manager/constructs/batch_compute_resources.py
tech3371: good catch. This was remains from previous change.

.gitignore
tech3371: removed that and added to global ignore file now.

cdk.json
tech3371: I wasn't sure if I should do that because carrying that mount path consistent across different resources is important. If someone make changes to the EFS mount path but forgot to change it the in resources it was being used by, then that would cause issues. But I do see your point that these things won't change as much between different accounts. I will change those to fixed value for now and then we can revisit if that becomes an issue.

sds_data_manager/batch/efs-access-batch/efs_check.py
tech3371: I will add some comment in the function about its purpose.

sds_data_manager/constructs/batch_compute_resources.py
tech3371: Yes. That gives us the access point id. I do that in line 184 and 192. Mine was just reading the config that contains those `import_value`'s name like 'efs-access-pt-id' and other informations as well.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: I could do that.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: filenaming convention looks like this for attitude kernel
```
Attitude History Report:

File Name: imap_yyyy_doy_yyyy_doy_##.ah.bc and
    imap_yyyy_doy_yyyy_doy_##.ah.a
    (where yyyy_doy defines the time span of
    the data in the file)

Server Location: imap/data_products/attitude_history/yyyy/

Attitude Predict Report
File Name: imap_pred_yyyy_doy_yyyy_doy_##.ah.bc or
    imap_pred_yyy_doy_yyyy_doy_##.ah.a
    (where yyyy_doy defines the time span
    of the data in the file)

Server Location: imap/data_products/attitude_history/yyyy/

Attitude Metakernel
File Name: imap_yyyy_doy_a##.spice.mk or
    imap_yyyy_doy_a##.stk_a.mk (where yyyy_doy
    is the day of creation)

```

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: ok. will add that.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: how does this know the source file that we want to link to the target symlink? Also curious of preference of Path over os. Can we use os and path in combination? Eg. create link using Path but read in its real path using os or do you need to do everything by one way?

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: I feel like there was some reason but I don't know for sure. I can check if we can do that.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: I will remove them. It was helpful for development.

sds_data_manager/stacks/efs_lambda_stack.py
tech3371: It may be covered by below. I will test it out.

sds_data_manager/stacks/efs_lambda_stack.py
tech3371: That comes now with Python but was doing that to show how we can add layers to lambda. I feel like there is enough resources out there if we need layer later. I will remove that.

sds_data_manager/stacks/efs_lambda_stack.py
tech3371: btw, this prefix is related to filename. all kernel file's name starts with `imap`. I know we use `imap` as root folder of our data bucket. I can see the confusion here.

sds_data_manager/stacks/efs_lambda_stack.py
tech3371: That's better name

cdk.json
greglucas: What I more-so meant is that these should just be living within the code itself. So inside your stack you could define these: `efs_stack.mount_path`, and re-use them elsewhere. Then if the EFS gets updated or things get changed that stack should carry that info along rather than needing to update any configuration parameters.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
greglucas: I think it would be this: `destination_path.symlink_to(source_path)`

> Can we use os and path in combination? Eg. create link using Path but read in its real path using os or do you need to do everything by one way?

Yes, you can use os and pathlib together, but it is nice to be consistent and pathlib is the more modern replacement. In particular, pathlib handles `/` vs `\` on *nix vs windows nicely so you don't have to think about the path separator. Everything you formerly did with os.path should be doable with pathlib I think... So I try to only use pathlib when I'm writing new code.

I am often looking at this reference for the corresponding uses: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

sds_data_manager/stacks/efs_lambda_stack.py
greglucas: Ahh, so I think you'd need to put `prefix: imap/spice/imap` in that case with the double imap for the nested filename?

cdk.json
tech3371: ok. I see. Yeah, I can totally do that.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: I am trying those changes. Yeah, I saw that comparison link and that was helpful to look up how it would map between os and Path.

sds_data_manager/stacks/efs_lambda_stack.py
tech3371: true. since our folder structure will looks like that. will do that.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: did that in upcoming changes

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: Let me know if you still want date in the symlink file. I personally like to avoid date in the symlink name. When we get real file that symlink is pointing, it should have date in it and that file should be what's written to the CDF metadata.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: I was able download directly to the mountpoint.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: removed and changed some to logging debug level

sds_data_manager/stacks/efs_lambda_stack.py
tech3371: we can remove it.

sds_data_manager/constructs/batch_compute_resources.py
```diff
@@ -148,14 +156,45 @@ def __init__(
         # processing_step_name tag. I think this will require
         # setting up a lambda. Maybe there's another way?
         self.job_definition_name = f"fargate-batch-job-definition{processing_step_name}"
+
+        # Get account_name to use as ECR's tag.
+        # account_name would be 'dev' or 'prod' or user specified.
+        # This value is defined by command line input and can
+        # be accessed from the cdk.json file.
+        account_name = scope.node.get_context("account_name")
```
greglucas: Access in the `account_config[account_name]` that can be passed in through the arguments.
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/039137b88083b6e7a4a5e0e2099be1a16551af3c/app.py#L28-L30

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
greglucas: Sorry for the confusion, I agree with you here that we definitely don't want a date in the symlink name. I was trying to say where the `latest` would go in the name, first like you have it or just replaying the YYYY date format.
`imap_yyyy_doy_a##.stk_a.mk` -> `imap_latest.stk_a.mk`

I think what you have is good, and it is a descriptive name :+1:

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
```diff
@@ -0,0 +1,133 @@
...
+import boto3
+
+# Define the paths
+mount_path = os.environ.get("EFS_MOUNT_PATH")
+
+attitude_symlink_path = f"{mount_path}/latest_attitude_kernel.ah.a"
+ephemeris_symlink_path = f"{mount_path}/latest_ephemeris_kernel.bsp"
+
+
+def create_symlink(source_path: str, destination_path: str) -> None:
```
greglucas:
```suggestion
def create_symlink(source_path: Path, destination_path: Path) -> None:
```
If you change over to `Path` objects early (even in the string paths globally above this) you could use these directly down below without having to create the new Path objects.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
```diff
@@ -0,0 +1,133 @@
...
+    source_path : str
+        Source path of the symlink
+    destination_path : str
+        Destination path of the symlink
+    """
+
+    # Check if the symlink exists
+    if Path(destination_path).is_symlink():
+        # Remove the old symlink
+        Path(destination_path).unlink(missing_ok=True)
```
greglucas: You already know it exists with the if check. So, just go directly to removing the path, or take out the missing_ok I think.

```suggestion
    # Remove the old symlink
    Path(destination_path).unlink(missing_ok=True)
```

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
```diff
@@ -0,0 +1,133 @@
...
+        S3 object key
+    s3_bucket : str
+    """
+    filename = os.path.basename(s3_key)
+
+    # Create an S3 client
+    s3_client = boto3.client("s3")
+
+    # Download the file to the mount directory. Eg. /mnt/efs
+    destination_path = f"{mount_path}/" + os.path.basename(s3_key)
```
greglucas: you can re-use the filename from above.
```suggestion
    destination_path = f"{mount_path}/" + filename
```
This could also be `mount_path / filename` if you make it a Path object above.

greglucas (commented):
Looks good to me, nice work! I'm a bit unsure why the tests are failing for you. Did pulling the s3 bucket into the event-bridge rules cause issues for the s3 bucket resource tests?

I put some more comments in about pathlib stuff, but all of them are just nits and I'm happy to leave that work for a follow-up PR.

The one other architecture question I have is whether you want to put the EFSWriteLambda stack code into the EFSStack itself? i.e. is the Lambda that writes to the EFS kind of directly related to the EFS and therefore doesn't need to be a separate piece.

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
tech3371: I see. ok. I will leave like it is for now.

sds_data_manager/constructs/batch_compute_resources.py
tech3371: will do that

greglucas (approved):

PR #191, opened 2023-10-31
Processing Trigger Mechanisms

laspsandoval:
# Change Summary

## Overview
The most important part of this PR is what happens in batch_starter.py. It does the following:
1. Get details of the object dropped in the s3 bucket (level, version, process_dates). 
2. Query dependents of the instrument (who had an object dropped in its bucket) to see if they have been ingested
3. Decide if we have sufficient dependencies for each dependent to process. 
4. If we have dependencies to process, then send relevant information to the Step Function. 

This is for normal processing. I have also thought about how we would reprocess:
1. Manually submit relevant information to Step Function.
2. If we only want to process a single Step Function we would manually turn off the Rules for instrument dependencies for EventBridge.

## New Files
**instruments/batch_starter.py** : lambda that queries the database to decide whether or not a step function should kick off and what information to send it.
**instruments/dependents.py** : My best shot at creating a json of dependents based on diagram Tenzin created.
**instruments/Dockerfile** : Creates environment for Dockerized lambda.

**utils/dependencies.json** : My best shot at creating a json of dependencies based on diagram Tenzin created.
**utils/get_dependency.py** : Gets instrument dependents based on dependencies.json.

## Deleted Files
**lambda_code/SDSCode/instruments/l1b_codice.py** : we only need a single lambda for checking the database.
**lambda_code/SDSCode/instruments/l1c_codice.py**: we only need a single lambda for checking the database.

## Updated Files
**batch_compute_resources** : remove unnecessary batch_security_group
**instrument_lambdas**: dockerize lambda; improve lambda environment
**sdc_step_function**: remove lambda, change what is passed to batch job
**database_stack**: minor changes to allow rds_creds to be passed to other stacks. This allowed batch_starter lambda to access the database.
**networking_stack**: added endpoint so that lambda could access secrets manager
**processing_stack**: updated the eventbridge rules
**stackbuilder**: reformatting

## Testing
**lambda_endpoints/test_batch_starter.py** : Creates mock database and mock event that can be used for testing functions in batch_starter.py
**test-data/codicehi_event.json** : Contains example of json created by an event
**process_kickoff_test_data.csv** : Contains example of how an instrument table in the database may be setup

To test after deployment:
1. Create directory in s3 bucket: sds-data-<acct #>/CodiceHi/l1a/ and place an object in that bucket
2. Navigate to CloudWatch > log groups > InstrumentLambdas and click on logstream. This is the error you should see. That shows that the lambda connected to the database which is not yet populated with schema (sdc) and table (codicehi).

[ERROR] UndefinedTable: relation "sdc.codicehi" does not exist

## Notes
1. I decided to use a single lambda for all instruments since they would all have common tools. The lambda does not include any lookup for calibration tables, but I thought that might be performed in the batch job instead.
2. Likely needs many more tests once we have RDS setup and code for Docker image for Batch Job complete.

Dependency diagram used:

Common Processing steps (Draft copies)


maxinelasp:
As a note for the reprocessing, one idea that I discussed with Greg and Tenzin is having a column or two in the database which indicate whether the file should trigger downstream processing or not. This would allow you to manually edit the database to prevent later steps reprocessing on a case by case basis.

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -18,14 +19,16 @@ def __init__(
...
         construct_id: str,
-        processing_step_name: str,
         data_bucket: s3.Bucket,
         code_path: str or Path,
-        instrument_target: str,
-        instrument_sources: str,
+        instrument: str,
+        instrument_dependents: dict,
+        step_function_arn: str,
+        step_function_policy: iam.PolicyStatement,
```
maxinelasp: This is a lot of inputs, could we combine some of them? Like maybe instrument and instrument_dependents? some of the policies?

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -37,22 +40,26 @@ def __init__(
...
         code_path : str or Path
             Path to the Lambda code directory
-        instrument_target : str
-            Target data product (i.e. expected product)
-        instrument_sources : str
-            Data product sources (i.e. dependencies)
+        instrument : str
+            Instrument
+        instrument_dependents : dict
+            Instrument dependents
```
maxinelasp: What is the structure of this dictionary? Is it only the dependencies for the one instrument, or all instrument dependencies? Can you make the documentation more clear?

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -62,33 +69,34 @@ def __init__(
...
             self,
-            id=f"InstrumentLambda-{processing_step_name}",
-            function_name=f"{processing_step_name}",
-            entry=str(code_path),
-            index=f"instruments/{instrument_target.lower()}.py",
-            handler="lambda_handler",
-            runtime=lambda_.Runtime.PYTHON_3_11,
-            timeout=Duration.seconds(10),
-            memory_size=512,
+            "InstrumentLambdas",
```
maxinelasp: Is this one lambda for all instruments and processing levels or does each instrument get the same lambda? If there are different ones, maybe having a different name will be useful. The plural here is also confusing to me - it's just one lambda, why is it called "InstrumentLambdas"?

sds_data_manager/constructs/sdc_step_function.py
```diff
@@ -45,41 +39,28 @@ def __init__(
...
         batch_resources: FargateBatchResources
             Fargate compute environment
-        instrument_target : str
-            Target data product
-        data_bucket : str
+        data_bucket : s3.Bucket
             S3 bucket
         db_secret_name : str
+            Db secret name
+        dependents : str
```
maxinelasp: Hear me out... what if we make dependents a class? I have a feeling we will want to pass them around a lot and it would be nice to have it always be a consistent dataclass. Then we could also swap out the JSON file for a different style of storing dependencies if we decide we need something more complicated.

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+        If the necessary fields are not found in the event.
+    """
+    try:
+        full_path = event["detail"]["object"]["key"]
+        return full_path.split("/")[-1]
+    except KeyError as err:
+        raise KeyError("Invalid event format: Unable to extract filename") from err
+
+
+def db_connect(db_secret_arn):
```
maxinelasp: Maybe we should make a database accesser library? I assume other pieces of the infrastructure will want to access it. Not necessary for this code review, just something to consider.

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+    process_dates : list
+        Dates to process
+    """
+
+    query = f"""SELECT * FROM sdc.{instrument.lower()}
+                WHERE filename = %s
+                LIMIT 1;"""
+    params = (filename,)
+
+    cur.execute(query, params)
```
maxinelasp: Can we have database calls and data formatting in different methods? For example, instead of passing around the cursor object, we can have this method call `query_filenames` or whatever, which would then return `record_dict` from line 121. Then the rest of the data rearranging can happen separately from the database call. 

This goes back to my note from earlier about making the database methods a separate file, as well.

pyproject.toml
```diff
@@ -33,6 +33,7 @@ pytest-cov = "^4.0.0"
 openmock = "^2.3.0"
 requests-mock = "^1.11.0"
 freezegun = "^1.2.2"
+psycopg2 = "^2.9.3"
```
greglucas: Is the database needed in the cdk package itself, or is it needed in the `lambda-dev` section (only the lambda dependencies need it) and therefore should be optional here?

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -18,14 +19,16 @@ def __init__(
...
-        instrument_target: str,
-        instrument_sources: str,
+        instrument: str,
+        instrument_dependents: dict,
+        step_function_arn: str,
+        step_function_policy: iam.PolicyStatement,
         rds_security_group: ec2.SecurityGroup,
         subnets: ec2.SubnetSelection,
         db_secret_name: str,
+        db_secret_arn: str,
```
greglucas: I feel like we are passing in a lot of individual things in these stacks, where it'd be perhaps a bit cleaner to pass in the containing stack above this.

```python
step_function_stack: StepFunctionStack,
```
Then reference `step_function_stack.policy`, and `step_function_stack.security_group` below. Same with passing the `db_stack` in and then looking up the values like secret name and arn. I think that will make it a bit more clear that the instrument lambda are dependent on step functions, dbs, ... without needing quite so many individual arguments.

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -62,33 +69,34 @@ def __init__(
...
-            runtime=lambda_.Runtime.PYTHON_3_11,
-            timeout=Duration.seconds(10),
-            memory_size=512,
+            "InstrumentLambdas",
+            function_name="InstrumentLambdas",
+            code=docker_image_code,
             environment=lambda_environment,
+            retry_attempts=0,
+            memory_size=512,
+            timeout=Duration.minutes(10),
```
greglucas: 10-30 seconds instead? This seems like a long timeout for a dependency lambda, or is there more going on in the lambda.

sds_data_manager/lambda_images/instruments/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+
+# Install development tools and required libraries
+RUN pip install psycopg2-binary
+
+# Install pip and install imap-processing module
+RUN python -m pip install --upgrade pip
+RUN pip install git+https://github.com/IMAP-Science-Operations-Center/imap_processing.git@dev
+
+WORKDIR ${LAMBDA_TASK_ROOT}
+COPY batch_starter.py .
```
greglucas:
```suggestion

# Install development tools and required libraries
RUN pip install psycopg2-binary imap_processing
WORKDIR ${LAMBDA_TASK_ROOT}
COPY batch_starter.py .
```

We should use the released versions of `imap_processing` if we can and reduce layers by putting all RUN commands on a single line.

This brings up the question of what happens if we update the package. Do we also need to redeploy the CDK to get that new version in here? Looking at the `batch_starter.py` file though I don't think you import imap_processing there so I actually think it can be removed altogether...

sds_data_manager/lambda_images/instruments/dependents.json
```diff
@@ -0,0 +1,196 @@
+{
+    "CodiceHi": {
```
greglucas: I think all instrument names should be lowercased to align with our s3 bucket naming. `codicehi` or `codice-hi` @tech3371 or @bourque I think one of you had a document identifying what we were going to do for consistent naming?

sds_data_manager/lambda_images/instruments/dependents.json
greglucas, line 1: Is this a file that should live with the CDK, or should it be in an s3 bucket that we can change without having to redeploy infrastructure? I think we did something like that with the old opensearch work where there was a configuration.json that was deployed to a bucket instead of stored within the CDK...

sds_data_manager/stacks/networking_stack.py
```diff
@@ -51,6 +51,13 @@ def __init__(
             ],
         )
 
+        self.vpc.add_interface_endpoint(
```
greglucas: Can you add a comment in the code about what this does and why it is needed?

sds_data_manager/stacks/processing_stack.py
```diff
@@ -77,47 +80,53 @@ def __init__(
...
+            db_secret_name=db_secret_name,
+            rds_security_group=rds_security_group,
+            subnets=subnets,
+            db_secret_arn=db_secret_arn,
+            vpc=vpc,
         )
 
-        rule.add_target(event_targets.SfnStateMachine(self.step_function.state_machine))
+        # Kicks off Step Function as a result of object ingested into directories in
+        # s3 bucket (instrument_sources).
```
greglucas, line 114: I'm not sure I am following this part. I thought we were going to be watching the data bucket instrument prefixes themselves for events. So the rules should be set up to listen on `"prefix": f"{instrument}/{level}"`? Is this adding another directory structure?

sds_data_manager/stacks/processing_stack.py
```diff
@@ -77,47 +80,53 @@ def __init__(
...
+        for source in instrument_dependents:
+            rule = events.Rule(
+                self,
+                f"Rule-{processing_step_name}-{source}",
+                event_pattern=events.EventPattern(
+                    source=["aws.s3"],
+                    detail_type=["Object Created"],
+                    detail={
+                        "bucket": {"name": [data_bucket.bucket_name]},
+                        "object": {"key": [{"prefix": f"{instrument}_{source}"}]},
```
greglucas:
```suggestion
                        "object": {"key": [{"prefix": f"{instrument}_{source}"} for source in instrument_dependents]},
```
I think you could squash all of the rules down into a single one with multiple prefixes like this? Putting the for-loop into the list of prefixes instead of outside.

sds_data_manager/utils/dependencies.json
greglucas, line 1: What is the difference between dependents and dependencies? Do we need to track both of them, or can we invert one to get the other? I fear these have the possibility to get out of sync otherwise.

sds_data_manager/utils/get_dependency.py
greglucas, line 1: Is this needed as a separate functionor could we just open the file once and save it as a module-level dictionary somewhere?

Do we need to return an empty dict, or can we fail with `KeyError` if we index into somewhere that doesn't exist so we get an immediate failure?

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -0,0 +1,258 @@
...
+    get_process_details,
+    prepare_data,
+    query_dependencies,
+    query_dependents,
+    remove_ingested,
+)
+
+
+@pytest.fixture(scope="session")
+def mock_event():
```
greglucas, line 20: Should we be using moto here to mock the s3 event instead?
https://docs.getmoto.org/en/latest/docs/services/events.html

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -0,0 +1,258 @@
...
+    """Example of the type of event that will be passed to
+    the instrument lambda (in our case batch_starter.py).
+    """
+    with open("../test-data/codicehi_event.json") as file:
+        event = json.load(file)
+    return event
+
+
+@pytest.fixture()
+def setup_database(postgresql):
```
```suggestion
def database(postgresql):
```
Just call it `database`? You are setting up here, but the fixture is what you are using elsewhere.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -0,0 +1,258 @@
...
+    """Populate test database."""
+
+    cursor = postgresql.cursor()
+    cursor.execute("CREATE SCHEMA IF NOT EXISTS sdc;")
+
+    # Drop the table if it exists, to start with a fresh table
+    cursor.execute("DROP TABLE IF EXISTS sdc.codicehi;")
+
+    sql_command = """
```
greglucas, line 40: Is this going to be used in the RDS code as well? It seems like it should be abstracted out into the code somehow so that you are initializing your test DB table with the same schema as the real one and that they don't get out of sync here and your tests are still passing even though you updated your code schema.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -0,0 +1,258 @@
...
+        spice_id INTEGER,
+        parent_codicehi_id INTEGER,
+        status TEXT
+    );
+    """
+
+    cursor.execute(sql_command)
+
+    # Insert mock data:
+    with open("../test-data/process_kickoff_test_data.csv") as f:
```
greglucas: This seems like a lot of formatting work below. Is there a way you can just hardcode the rows inside here to make it more obvious what is going on, or store the stripped values directly in the csv.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # Cleanup: close the connection after tests
+    postgresql.close()
+
+
+def test_get_filename_from_event(mock_event):
+    # Use mock event from the fixture
+    filename = get_filename_from_event(mock_event)
+
+    assert filename == "imap_l3a_sci_codicehi_20230602_v01.pkts"
```
greglucas: the instrument name should be first.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # Use the cursor to execute SQL and fetch results
+    cursor.execute("SELECT COUNT(*) FROM sdc.codicehi")
+    count = cursor.fetchone()[0]
+    cursor.close()
+
+    with open("../test-data/process_kickoff_test_data.csv") as f:
+        reader = csv.reader(f)
+        next(reader, None)
+        row_count = sum(1 for _ in reader)
```
```suggestion
    with open("../test-data/process_kickoff_test_data.csv") as f:
        row_count = len(f.readlines())
```

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -0,0 +1,258 @@
...
+            "parent_codicehi_id": 3,
+            "status": "INCOMPLETE",
+        },
+    ]
+
+    res_true = all_dependency_present(result, dependencies_true)
+    res_false = all_dependency_present(result, dependencies_false)
+
+    assert res_true is True
+    assert res_false is False
```
```suggestion
    assert all_dependency_present(result, dependencies_true)
    assert not all_dependency_present(result, dependencies_false)
```

greglucas (commented):
I am not a DB expert so it'd be good for someone else to take a close look at that portion.

I'm not sure I followed the rule to batch_starter integration. It looked like it was requiring a new directory structure, but I might not have been following that correctly.
Do we need a full docker lambda integration on that, or can it be a simpler lambda function that queries the DB like we were doing on opensearch with a separate `requirements.txt` file if necessary?

sds_data_manager/utils/dependencies.json
sdhoyt: If we could get that to be read in as a matrix, I think taking the transpose would get us the other one.


Rows show the dependents and Columns show the dependencies:
  | Codice Hi L0 | Codice Hi l1a | Codice Hi l1b | Codice Hi l2 | Codice Hi L3a
-- | -- | -- | -- | -- | --
Codice Hi L0 |   |   |   |   |  
Codice Hi l1a | 1 |   |   |   |  
Codice Hi l1b |   | 1 |   |   |  
Codice Hi l2 |   |   | 1 |   |  
Codice Hi L3a |   |   |   | 1 |  
Codice Hi L3b |   |   |   |   | 1
Codice Hi L3c |   |   |   |   | 1

sds_data_manager/lambda_images/instruments/batch_starter.py
sdhoyt: I'll make the change on my branch since I'll probably need to pull this PR in, but I'm using the same function in my updates, so we should pull this function out into a database utility file

I'm mostly commenting so I can link this comment to an issue so I don't forget that this is a spot I will need to update.

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+    -------
+    level : str
+        Instrument level
+    version : int
+        Version
+    process_dates : list
+        Dates to process
+    """
+
+    query = f"""SELECT * FROM sdc.{instrument.lower()}
```
sdhoyt: Does the LIMIT 1 mean there might be multiple files with the same name? Should there be some kind of ordering here if that's the case?

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+        grouped_data[instrument][level].append(date)
+
+    grouped_dict = {
+        instrument: dict(levels) for instrument, levels in grouped_data.items()
+    }
+
+    return grouped_dict
+
+
+def lambda_handler(event: dict, context):
```
sdhoyt: I think I'm confusing myself with the dependents vs dependencies. 

This is how I'm interpreting this function:
- An event is received because a new file was added to the S3
- That event contains the filename which is used to query for information about the file: level, version, etc.
- Another query is made to get the file's dependents
   - _This is the first place I'm getting confused. It seems like the dependents are all of the things that were needed to create this file, but it seems like it should be getting the dependencies which are all the files that this files is used to help create?_
- It checks to see if those dependents have been ingested and retains the ones that have not
- Another query is made to all of the uningested files' dependencies
   - _Same confusion as above, I thought dependencies were all the files that this file was used to help make, but it seems like it should be querying for all the files that are needed to create the uningested file which would be the dependents?_
 - Those files that come back from the previous query are input into the step function to be created
    - _What happens if somewhere down the dependency line it's missing something that it needs? Like the step function is trying to process an L1c file, but an L1a that it needs doesn't exist? It kind of seems like the dependency checks for uningested files should be recursive?_

I'm probably reading the json wrong and not totally understanding this function, but wanted to confirm.

sdhoyt (commented):
This looks great and was surprisingly easy to follow considering how complicated all this processing dependency stuff is. Great job! I just left a few comments.

sds_data_manager/lambda_images/instruments/batch_starter.py
maxinelasp: This is a slightly outdated comment now. Per our discussion in the meeting, this is great to leave for now and when we have a more definite database design we can revisit.

sds_data_manager/lambda_images/instruments/dependents.json
```diff
@@ -0,0 +1,196 @@
+{
```
maxinelasp: Maybe rename this file to "upstream_dependencies.json"?

maxinelasp: Also, a comment on the structure of this file is probably a good thing to add, because there was a fair amount of confusion about it

sds_data_manager/stacks/processing_stack.py
```diff
@@ -28,13 +28,14 @@ def __init__(
         processing_step_name: str,
         lambda_code_directory: str or Path,
         data_bucket: s3.Bucket,
-        instrument_target: str,
-        instrument_sources: str,
+        instrument: str,
+        instrument_dependents: dict,
+        dependents: str,
```
maxinelasp: `dependents` is a confusing name for this IMO. maybe `dependency_filepath` is more clear?

sds_data_manager/stacks/processing_stack.py
```diff
@@ -28,13 +28,14 @@ def __init__(
         processing_step_name: str,
         lambda_code_directory: str or Path,
         data_bucket: s3.Bucket,
-        instrument_target: str,
-        instrument_sources: str,
+        instrument: str,
+        instrument_dependents: dict,
```
maxinelasp: Downstream dependents? could be `instrument_downstream` or something. Also comment should include the expected format of this dict (instrument name -> list of downstream dependents?)

sds_data_manager/stacks/processing_stack.py
```diff
@@ -77,47 +80,53 @@ def __init__(
...
+            vpc=vpc,
         )
 
-        rule.add_target(event_targets.SfnStateMachine(self.step_function.state_machine))
+        # Kicks off Step Function as a result of object ingested into directories in
+        # s3 bucket (instrument_sources).
+        # TODO: Right now these directories are created manually.
+        #  Add code so that they are not.
+        for source in instrument_dependents:
+            rule = events.Rule(
```
maxinelasp: Would it make sense to move this into a different file for triggering-specific info?  Seems like this might get quite complicated with all the lambdas, reprocessing, etc. Not needed for this review, just a general question

sds_data_manager/utils/dependencies.json
maxinelasp: Definitely I agree that there should be ONE source of truth for this. It will get very confusing otherwise. I vote for the other file, deleting this one. Maybe we will need to rework the format like Sean suggested

sds_data_manager/utils/get_dependency.py
maxinelasp: plus one on saving it as a module wide data structure, although I'm sure you'll be unsurprised to hear me say we should use a dataclass 😆 For the future, perhaps we can have an issue to rework the dependency management in this code, including consolidating to one file?

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -0,0 +1,258 @@
...
+
+    assert records[0]["filename"] == "imap_l3b_sci_codicehi_20230531_v01.cdf"
+
+    # Since there are 2 instruments x 3 dates and one record to remove = 5
+    output = remove_ingested(records, instrument_dependents, process_dates)
+
+    assert len(output) == 5
+
+
+def test_query_dependencies(setup_database):
```
maxinelasp: Easy testing is an argument _against_ consolidating all the database calls. Something to consider.

pyproject.toml
maxinelasp: Agreed with Greg, if this is only used in the lambda, please move it to `lambda-dev`. If you're installing with requirements.txt then you shouldn't need to do anything else, except run pre-commit locally to update your local version.

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -18,14 +19,16 @@ def __init__(
         self,
         scope: Construct,
         construct_id: str,
-        processing_step_name: str,
         data_bucket: s3.Bucket,
         code_path: str or Path,
-        instrument_target: str,
-        instrument_sources: str,
+        instrument: str,
+        instrument_dependents: dict,
```
maxinelasp: rename to `instrument_downstream` (or `instrument_upstream` if it's some kind of mapping of instruments to upstream dependencies?)

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -62,33 +69,34 @@ def __init__(
...
         # TODO: if we need more variables change so we can pass as input
         lambda_environment = {
-            "S3_BUCKET": f"{data_bucket.bucket_name}",
-            "S3_KEY_PATH": instrument_sources,
-            "INSTRUMENT_TARGET": instrument_target,
-            "PROCESSING_NAME": processing_step_name,
-            "OUTPUT_PATH": f"s3://{data_bucket.bucket_name}/{instrument_target}",
-            "SECRET_NAME": db_secret_name,
+            "INSTRUMENT": f"{instrument}",
+            "INSTRUMENT_DEPENDENTS": f"{instrument_dependents}",
```
maxinelasp: Ditto for name change

sds_data_manager/constructs/sdc_step_function.py
```diff
@@ -29,11 +24,10 @@ def __init__(
         scope: Construct,
         construct_id: str,
         processing_step_name: str,
-        processing_system: InstrumentLambda,
         batch_resources: FargateBatchResources,
-        instrument_target: str,
         data_bucket: s3.Bucket,
         db_secret_name: str,
+        dependents: str,
```
maxinelasp: Rename to `dependent_config_path` or something

sds_data_manager/lambda_images/instruments/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# TODO I think we should have a single Dockerfile and pyproject.toml
+# for all dockerized lambdas
+FROM public.ecr.aws/lambda/python:3.11
+
+# Install git
+RUN yum install -y git
+
+# Install development tools and required libraries
+RUN pip install psycopg2-binary
```
maxinelasp: This should be in the poetry project, so you either pip install the existing requirements.txt file for `lambda-dev` or we make a new dependency group and a new requirements.txt file. Let me know if you have other questions on how to do that.

maxinelasp (dismissed):
Overall looks great! I'd like to see some of the name changes I suggested and maybe a fix for the dependencies, just because those rarely get fixed after PR ends. Other than that, my comments can be taken as suggestions for future improvement.

sds_data_manager/utils/stackbuilder.py
```diff
@@ -92,62 +93,49 @@ def build_sds(scope: App, env: Environment, account_config: dict):
         env=env,
         vpc=networking.vpc,
         rds_security_group=networking.rds_security_group,
-        engine_version=rds.PostgresEngineVersion.VER_14_2,
+        engine_version=rds.PostgresEngineVersion.VER_15,
```
tech3371: If you rebase with latest change, it will look like this. The latest version CDK supports is version 15.3. Sean and we both tried and that seems to the latest at the moment.
```suggestion
        engine_version=rds.PostgresEngineVersion.VER_15_3,
```

sds_data_manager/utils/stackbuilder.py
```diff
@@ -92,62 +93,49 @@ def build_sds(scope: App, env: Environment, account_config: dict):
...
         instance_size=ec2.InstanceSize[rds_size],
         instance_class=ec2.InstanceClass[rds_class],
         max_allocated_storage=rds_storage,
-        username="postgres",
-        secret_name="sdp-database-creds",
-        database_name="imapdb",
+        username="imap_user",
+        secret_name="sdp-database-cred",
```
tech3371, line 103: We don't use these variable inside the SdpDatabase() stack. Can you remove these hard code value with these passed in values?
```
        self.secret_name = "sdp-database-creds"
        database_name = "imap"
        username = "imap_user"
```

sds_data_manager/utils/stackbuilder.py
```diff
@@ -92,62 +93,49 @@ def build_sds(scope: App, env: Environment, account_config: dict):
...
+    )
     lambda_code_directory_str = str(lambda_code_directory.resolve())
+    dependents_file_path = lambda_code_directory / "dependents.json"
+    dependents_file_path_str = str(dependents_file_path.resolve())
 
     for instrument in instrument_list:
         ecr = ecr_stack.EcrStack(
             scope,
-            f"{instrument}Processing",
+            f"{instrument}Ecr",
```
tech3371, line 152: Thank you for changing that. I thought `Ecr` made more sense.

sds_data_manager/utils/stackbuilder.py
```diff
@@ -92,62 +93,49 @@ def build_sds(scope: App, env: Environment, account_config: dict):
...
-        )
-
-        processing_stack.ProcessingStep(
-            scope,
-            f"L1c{instrument}Processing",
+            f"{instrument}Processing",
             env=env,
             vpc=networking.vpc,
-            processing_step_name=f"l1c-{instrument}",
+            processing_step_name=f"{instrument}",
```
tech3371:
```suggestion
            processing_step_name=instrument,
```

sds_data_manager/utils/stackbuilder.py
```diff
@@ -92,62 +93,49 @@ def build_sds(scope: App, env: Environment, account_config: dict):
...
             lambda_code_directory=lambda_code_directory_str,
             data_bucket=data_manager.data_bucket,
-            instrument_target=f"l1c_{instrument}",
-            instrument_sources=f"l1b_{instrument}",
+            instrument=instrument,
+            instrument_dependents=get_dependency(instrument),
+            dependents=dependents_file_path_str,
```
tech3371: Repeating Maxine's comment of use of upstream or downstream dependency name.

tests/lambda_endpoints/test_batch_starter.py
tech3371: This is the file format: 
`imap_<instrument>_<datalevel>_<mode>_<descriptor>_<startdate>_<version>.cdf`

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+    cur : psycopg2.extensions.cursor
+        A psycopg2 database cursor object to execute
+        database operations.
+    instrument : str
+        The name of the instrument for which details
+        are to be retrieved.
+    filename : str
+        The filename associated with the instrument.
+    process_range : int
+        Numbers of days backwards to process
```
tech3371: Is this where we specify how many days ago we go to get certain data? Like 2 means get all data from 2 days ago? If so, can you add that clarification to the doc string of this parameter?

sds_data_manager/lambda_images/instruments/batch_starter.py
tech3371: same thinking here. or since all files should be unique name, may be we should remove `limit 1` and make sure that it only returns one value when you call `fetchall()`

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+    if not records:
+        raise ValueError(f"No records found for filename: {filename}")
+
+    record_dict = dict(zip(column_names, records[0]))
+
+    level = record_dict["level"]
+    version = record_dict["version"]
+
+    dt = record_dict["date"]
+    dt_start = dt - timedelta(days=process_range)
```
tech3371: can we spell out `dt` to `date`?

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+    ----------
+    cur : psycopg2.extensions.cursor
+        A psycopg2 database cursor object to execute database operations.
+    version : int
+        Version of the instrument to be queried.
+    process_dates : list
+        A list containing start and end date to filter records on their ingestion date.
+    instrument_dependents : list of dict
+        A list containing dictionaries of dependent instruments and their levels.
+        Each dictionary should have keys 'instrument' and 'level'.
```
tech3371: It's not clear here if it's upstream dependent or downstream dependent. I will avoid dependent comments from now on since all those may get taken care of once you make changes from Maxine's comment.

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+
+        # Map the column names to the records
+        records_dicts = [dict(zip(column_names, record)) for record in records]
+
+        all_records.extend(records_dicts)
+
+    return all_records
+
+
+def remove_ingested(records, dependents_level, process_dates):
```
tech3371: I see now what you are doing in `remove_ingested()`. It's looking for any missing dependent data. Is that right?

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+
+    """
+
+    output = []
+
+    records_set = {
+        (rec["date"].date(), rec["instrument"], rec["level"]) for rec in records
+    }
+
+    for dependent in dependents_level:
```
tech3371: can you add comment of why there could be multiple data levels in dependents_level?

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    json_path = os.path.join(dir_path, "dependents.json")
+
+    with open(json_path) as f:
+        data = json.load(f)
+
+    instruments_to_process = []
+
+    for record in uningested:
```
tech3371: uningested part is little confusing but I think it will go away with upcoming changes.

sds_data_manager/lambda_images/instruments/batch_starter.py
```diff
@@ -0,0 +1,425 @@
...
+                return
+
+            # decide if we have sufficient dependencies
+            # for each dependent to process
+            instruments_to_process = query_dependencies(cur, uningested, version)
+
+            # No instruments to process
+            if not instruments_to_process:
+                logger.info("No instruments_to_process. Skipping further processing.")
+                return
```
tech3371: If you can add some example in the comments of what it will look like at each of these operation, it will help so much to follow and understand. I know it's a complicated logic and it may help us even when we come back to look at this logic again after some time.

sds_data_manager/constructs/instrument_lambdas.py
laspsandoval: Just a single lambda. I'll change the name.

sds_data_manager/lambda_images/instruments/dependents.json
laspsandoval: Changed so that directories in bucket are lowercase. Will revisit once we have completely finalized naming. I might just have each instrument lead look through this. I'll put that as a todo and note it is really only a first pass attempt.

laspsandoval:
> I am not a DB expert so it'd be good for someone else to take a close look at that portion.
> 
> I'm not sure I followed the rule to batch_starter integration. It looked like it was requiring a new directory structure, but I might not have been following that correctly. Do we need a full docker lambda integration on that, or can it be a simpler lambda function that queries the DB like we were doing on opensearch with a separate `requirements.txt` file if necessary?

The rule for batch_starter_integration assumed that within the s3 bucket there was a directory called codicehi_l1a. I have changed it to now be codicehi/l1a. I think we should weigh pros and cons of Dockerizing the lambdas versus using the alpha lambda and be consistent with whatever we do. We have 1/2 the project doing it one way and the other half the other way.

sds_data_manager/lambda_images/instruments/batch_starter.py
laspsandoval: It's looking to see if the data product that we are planning on processing has already been ingested. This is just to prevent duplicates, but I am thinking that we might just remove this part of the code. We can get together are discuss next week.

tests/lambda_endpoints/test_batch_starter.py
laspsandoval: I couldn't figure out a way to generate an event payload using moto. I basically just wanted to test if we could get the filename from a event from an object dropped in the s3 bucket.

pyproject.toml
laspsandoval: Correct me if I am wrong, but the Dockerized lambdas require a Dockerfile to exist within their same directory. So we need to revisit whether or not we want to use Dockerized lambdas or the regular lambda_alpha_.PythonFunction.

Here is the ChatGPT Conclusion:

Choose AWS Lambda's Built-in Runtimes if you want simplicity, ease of use, and are working within the constraints of the supported languages and package size limits.

Opt for Dockerized Lambdas if you need greater control over the runtime environment, have large dependencies, or use languages/runtimes not natively supported by AWS Lambda.

greglucas: Either way you'll need this in the requirements file that is used to build the lambdas because either the Dockerfile will need to install the dependencies or the PythonLambdaAlpha will need to install them.

PythonLambdaAlpha is just a nice wrapper on the ContainerLambda for you so all you need is a `requirements.txt` file and it will create the Dockerfile automagically for you under the hood and build the container image and push it up to an ECR for you.

sds_data_manager/lambda_images/instruments/Dockerfile
laspsandoval: Wait, for Dockerized lambdas we actually require a Dockerfile. Or am I missing something?

sds_data_manager/lambda_images/instruments/dependents.json
laspsandoval: I think that no matter where it lives the infrastructure will need to be redeployed if it changes (though it should not change). The reason is that it is used to created rules for EventBridge.

laspsandoval:
**Processing Trigger Mechanisms ticket summary:**
1. I addressed PR comments
2. Rebased with current dev branch. When I deploy I have errors related to efs: "**No export named efs-filesystem-id found**", "**No export named spice-access-point-id found**". This error occurs on line 179 in batch_compute_resources.py. I checked out the upstream/dev branch and tried to deploy it, but had the same error. @tech3371 do you have ideas about why this might be? Did you deploy the stacks in a certain order that could have prevented this error?
3. I will definitely have another round of updates for this code that include: update batch_starter.py to query pointing number for certain instruments, change entrypoint for batch job, modify code to use ORM.
4. codecov requires more tests for batch_starter.py. I think that I would rather get a more final version of that code finished before perfecting the tests. 

I can put more work into this PR, but it might be better to merged first and then improve, but I will leave that vote up to the team. Please let me know how to proceed.

tech3371:
> **Processing Trigger Mechanisms ticket summary:**
> 
> 1. I addressed PR comments
> 2. Rebased with current dev branch. When I deploy I have errors related to efs: "**No export named efs-filesystem-id found**", "**No export named spice-access-point-id found**". This error occurs on line 179 in batch_compute_resources.py. I checked out the upstream/dev branch and tried to deploy it, but had the same error. @tech3371 do you have ideas about why this might be? Did you deploy the stacks in a certain order that could have prevented this error?
> 3. I will definitely have another round of updates for this code that include: update batch_starter.py to query pointing number for certain instruments, change entrypoint for batch job, modify code to use ORM.
> 4. codecov requires more tests for batch_starter.py. I think that I would rather get a more final version of that code finished before perfecting the tests.
> 
> I can put more work into this PR, but it might be better to merged first and then improve, but I will leave that vote up to the team. Please let me know how to proceed.

I deployed couple times to reproduce the error and I haven't come across this error `"No export named efs-filesystem-id found", "No export named spice-access-point-id found".`. I wonder if it's a one time thing. I remember we both were creating resources at the same time in dev account and I wonder if it caused some issue. But if you come across this issue again, I am happy to look into it.

sds_data_manager/constructs/batch_compute_resources.py
```diff
@@ -63,7 +61,7 @@ def __init__(
             Dependent on Docker image contents.
         job_memory : int: Optional
             Memory required per Batch job in MB. Dependent on Docker image contents.
-        efs: efs.Filesystem
+        efs_instance: efs.Filesystem
```
tech3371, line 64: Thank you for catching this.
```suggestion
        efs_instance: EFSStack
```

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -62,35 +65,34 @@ def __init__(
         # Define Lambda Environment Variables
         # TODO: if we need more variables change so we can pass as input
         lambda_environment = {
-            "S3_BUCKET": f"{data_bucket.bucket_name}",
-            "S3_KEY_PATH": instrument_sources,
-            "INSTRUMENT_TARGET": instrument_target,
-            "PROCESSING_NAME": processing_step_name,
-            "OUTPUT_PATH": f"s3://{data_bucket.bucket_name}/{instrument_target}",
-            "SECRET_NAME": db_secret_name,
+            "INSTRUMENT": f"{instrument}",
```
tech3371:
```suggestion
            "INSTRUMENT": instrument,
```

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -62,35 +65,34 @@ def __init__(
...
+            "InstrumentLambda",
+            function_name="InstrumentLambda",
             entry=str(code_path),
-            index=f"instruments/{instrument_target.lower()}.py",
+            index="batch_starter.py",
             handler="lambda_handler",
-            runtime=lambda_.Runtime.PYTHON_3_11,
-            timeout=Duration.seconds(10),
-            memory_size=512,
+            runtime=lambda_.Runtime.PYTHON_3_9,
```
tech3371: Was there reason for reverting Python version?

tech3371 (approved):
I only had minor comments. We can address in future PR. A white board session would be helpful.

PR #196, opened 2023-11-16
RDS with Schema Creation Custom Resource

sdhoyt:
# Change Summary

## Overview
This PR adds a custom resource lambda that will create the rds database schema on app deployment.

## New Dependencies
None

## New Files
- create_schema_stack.py
   - stack for creating the custom resource lambda that creates a table on deployment.

## Deleted Files
None

## Updated Files
- database_stack.py
   - created new instance variables for use outside the stack
- networking_stack.py
   - added endpoint interface to allow lambda to access RDS
 - sds_data_manager.py
    - added vpc/subnets to indexer lambda
    - granted lambda permission to get secrets from secret manager
 - stack_builder.py
    - added new inputs to sds-data-manager stack
    - added custom resource stack

## Testing
- Deployed and confirmed custom resource lambda ran using CloudWatch
- updated test_data_manager_stack
- added test_create_schema_stack
- added test_database_stack


pyproject.toml
```diff
@@ -95,5 +97,5 @@ ignore = ["D203", "D212", "PLR0913", "PLR2004"]
 
 [tool.ruff.per-file-ignores]
 # TODO: Too many statements, this could be refactored to separate
-#       the single stack out into a few smaller pieces
+#       the single stack out into a few  smaller pieces
```
greglucas: revert the extra space here

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -0,0 +1,95 @@
...
+# TODO: The code in this lambda will be updated in the next PR
+# to reflect the schema that will be needed by the SDC.
+# Its current state is the bare minimum to get a table
+# created to allow files to be indexed.
+
+# Add code to get Secret instead
+s3 = boto3.client("s3")
+
+
+def send_response(event, context, response_data, response_status):
```
greglucas, line 15: Add a docstring indicating what this is doing. I'm not following at first glance what is going on.

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    json_response_body = json.dumps(response_body)
+
+    headers = {"content-type": "", "content-length": str(len(json_response_body))}
+
+    try:
+        response = requests.put(response_url, data=json_response_body, headers=headers)
+        response.raise_for_status()
+    except Exception as e:
```
greglucas: We shouldn't have a bare except here. Do we want to just let this error out instead of printing?

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -0,0 +1,95 @@
...
+def lambda_handler(event, context):
+    secret_name = os.environ["SECRET_NAME"]
+    session = boto3.session.Session()
+    client = session.client(service_name="secretsmanager")
+    secret_string = client.get_secret_value(SecretId=secret_name)["SecretString"]
+    secret = json.loads(secret_string)
+
+    try:
+        # Establish a connection to the PostgreSQL database
+        connection = psycopg2.connect(
```
greglucas: I think this try-except block could actually be made much simpler with the new psycopg3 package because they have context managers to handle the connections and clean-up for you.
https://www.psycopg.org/psycopg3/docs/basic/usage.html

Can do this as a follow-up PR as well though.

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -0,0 +1,95 @@
...
+        )
+
+        # Create a cursor object to interact with the database
+        cursor = connection.cursor()
+
+        # SQL query to create a table
+        create_table_query = """
+            CREATE TABLE IF NOT EXISTS metadata (
+                id VARCHAR(100) PRIMARY KEY,
+                mission VARCHAR(100),
```
greglucas: Do we want to restrict some of these fields to the lengths we know they will actually be?
```suggestion
                mission VARCHAR(4),
```
Also, is the mission going to change, or is this field kind of extraneous? (Maybe it would change for I&T / commissioning?)

sds_data_manager/lambda_code/SDSCode/rds_utils/db_ingest_query.py
```diff
@@ -0,0 +1,89 @@
...
+import sys
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
+
+
+class DbIngestQuery:
+    """
+    Class to represent the connection to the ingest query.
```
greglucas: If I'm following this right, this class actually constructs the query, and then allows itself to be filled, but doesn't do any DB connections, that is handled in your other `DbConnection` class right? If that is the case, I wonder if this would make more sense as a dataclass or similar. So this class would create the query information to be submitted to psycopg (which you already have here), but then it would be more of a function call to `records = db_connection.send_query(DbIngestQuery)` that would return the cursor records instead of filling a tuple you give it?

sds_data_manager/stacks/networking_stack.py
```diff
@@ -61,6 +61,13 @@ def __init__(
             self, "FargateInstanceSecurityGroup", vpc=self.vpc
         )
 
+        self.vpc.add_interface_endpoint(
```
greglucas, line 64: Could you add a comment about why this is added. Is it that our public lambdas need to access our secrets manager when getting the rds secret to query the db?

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -249,6 +256,10 @@ def __init__(
             runtime=lambda_.Runtime.PYTHON_3_9,
             timeout=cdk.Duration.minutes(15),
             memory_size=1000,
+            allow_public_subnet=True,
```
greglucas, line 262: Why would the indexer lambda need to run on a public subnet? Is this for the DB connection/secrets manager?

In the future I think we should look at tightening some of these up to within the private VPC routes if we can. Not necessary right now though.

sds_data_manager/utils/stackbuilder.py
```diff
@@ -21,7 +22,15 @@
...
 
-def build_sds(scope: App, env: Environment, account_config: dict):
+def build_sds(
+    scope: App,
+    env: Environment,
+    account_config: dict,
+    rds_size: str = "SMALL",
+    rds_class: str = "BURSTABLE3",
+    rds_storage: int = 200,
+    use_custom_domain: bool = False,
```
greglucas: Could these be in the `account_config` dict? Maybe `rds_size = account_config.get("rds_size", "SMALL")` to get a default value inside here if you want.

tests/rds_utils/test_db_ingest_query.py
```diff
@@ -0,0 +1,69 @@
...
+        "level": "l1a",
+        "date": "20231030",
+    }
+    return DbIngestQuery(file_path, metadata)
+
+
+def test_db_ingest_init(create_db_query):
+    """Test that DBIngestQuery object correctly initializes"""
+    ## Arrange
+    query_true = """
```
greglucas: Minor nit, but I find the `true` to be confusing in this context, I see `expected` most of the time.

greglucas (commented):
Looks great! Thank you for breaking this up into pieces :)

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -0,0 +1,95 @@
...
+# Add code to get Secret instead
+s3 = boto3.client("s3")
+
+
+def send_response(event, context, response_data, response_status):
+    response_url = event["ResponseURL"]
+    response_body = {
+        "Status": response_status,
+        "Reason": "See the details in CloudWatch Log Stream: "
+        + context.log_stream_name,
```
tech3371, line 35: cool. I didn't know you could get those values from `context`. That input has been a mystery for me and never saw it used.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -180,11 +180,8 @@ def lambda_handler(event, context):
...
-    host = os.environ["OS_DOMAIN"]
-    snapshot_repo_name = os.environ["SNAPSHOT_REPO_NAME"]
-    snapshot_s3_bucket = os.environ["S3_SNAPSHOT_BUCKET_NAME"]
-    snapshot_role_arn = os.environ["SNAPSHOT_ROLE_ARN"]
-    region = os.environ["REGION"]
+    os.environ["OS_DOMAIN"]
+    os.environ["REGION"]
 
     # create opensearch client
     client = _create_open_search_client()
```
tech3371: Are those going to get removed in a separate PR? just curious. And separate PR might be easier and it was lot easier to remove :)

sds_data_manager/lambda_code/SDSCode/rds_utils/db_ingest_query.py
```diff
@@ -0,0 +1,89 @@
...
+        self.metadata["year"] = self.metadata["date"][:4]
+        self.metadata["month"] = self.metadata["date"][4:6]
+        self.metadata["day"] = self.metadata["date"][6:8]
+        # remove date from metadata
+        self.metadata.pop("date")
+        logger.info(f"metadata: {self.metadata}")
+        # create a string for the fields in query
+        fields = ", ".join(self.metadata.keys())
+        # create placeholders for the query data to put into query string
+        placeholders = ", ".join(["%s"] * len(self.metadata))
```
tech3371: I think I understand the others but not sure of this `placeholders`. Can you give an example of what it's doing with `len(self.metadata)` and what the input to query would look like?

sds_data_manager/stacks/create_schema_stack.py
```diff
@@ -0,0 +1,66 @@
...
+        vpc_subnets,
+        rds_security_group,
+        **kwargs,
+    ) -> None:
+        super().__init__(scope, construct_id, env=env, **kwargs)
+
+        schema_create_lambda = lambda_alpha_.PythonFunction(
+            self,
+            id="CreateMetadataSchema",
+            function_name="create-schema-hoyt-test-2",
```
tech3371: should we rename this now?

sds_data_manager/stacks/create_schema_stack.py
```diff
@@ -0,0 +1,66 @@
...
+            entry=str(
+                pathlib.Path(__file__).parent.joinpath("..", "lambda_code").resolve()
+            ),
+            index="SDSCode/create_schema.py",
+            handler="lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_9,
+            timeout=aws_cdk.Duration.minutes(15),
```
tech3371: I think Greg's been commenting about reducing this timeout duration and it's a good idea. Can you change it to somewhat close to how long it takes to complete task in this lambda?

sds_data_manager/stacks/sds_data_manager_stack.py
```diff
@@ -44,6 +47,10 @@ def __init__(
         dynamodb_stack: DynamoDB,
         processing_step_function_arn: str,
         env: cdk.Environment,
+        db_secret_name: str,
+        vpc: ec2.Vpc,
+        vpc_subnets,
```
tech3371, line 54: I was wondering why we pass in both `vpc` and `vpc_subnets` but later saw that it was RDS's subnet selection. may be, we could rename this to something along there to indicate what these subnets are from and what it's used for? does RDS uses private subnet? I wonder if we can get private subnet from vpc and use that. I have not tried so it's ok. That's something we can look at later time.

tests/rds_utils/test_db_ingest_query.py
```diff
@@ -0,0 +1,69 @@
...
+    ## Assert
+    assert create_db_query.query == query_true
+    assert create_db_query.data == query_data_true
+
+
+def test_ingest_query(create_db_query):
+    """Test that the ingest_query() function returns the correct query"""
+
+    ## Arrange
+    query_true = """
```
tech3371: I see. This is what the input to query looks like. A placeholder to the inputs. Now what you wrote above makes sense. Could you put this example in the function as comment?

tech3371 (dismissed):
minor comments. But looks good to me! Nice work!

sds_data_manager/lambda_code/SDSCode/rds_utils/db_ingest_query.py
```diff
@@ -0,0 +1,89 @@
...
+                    INSERT INTO metadata (
+                        {fields})
+                    VALUES ({placeholders})
+                """
+        logger.info(f"query: {self.query}")
+        # create tuple to hold query data
+        self.data = tuple(self.metadata.values())
+        logger.info(f"query values: {self.data}")
+
+    def get_query(self):
```
maxinelasp: These getters are more for the Java style of classes, in Python, you would expect to access these attributes through `db_query.query` and `db_query.data`

sds_data_manager/lambda_code/SDSCode/rds_utils/db_ingest_query.py
```diff
@@ -0,0 +1,89 @@
...
+        gets the ingest query string.
+    get_data():
+        gets the ingest data tuple.
+    """
+
+    def __init__(self, s3_path: str, metadata: dict):
+        self.query: str
+        self.data: tuple
+        self.s3_path = s3_path
+        self.metadata = metadata
```
maxinelasp: This database is a dictionary sourced entirely from a different piece of code, which if I was modifying, I would not know it would also have the knock-on affect of changing the database creation query. I would instead suggest adding a new dataclass for the metadata with very explicit fields for creating this table (or, just use a file like we discussed.)

This way, it is immediately clear what is required by that metadata attribute when you're creating this class.

```
@dataclass
class FileMetadata:
    id: str 
    year: str 
    day: str
    month: str
    # whatever else is in that original metadata database
```

Then, when you're initializing the class in indexer.py:

```
 db_ingest_query = DbIngestQuery(s3_path, FileMetadata(None, metadata["date"][:4], metadata["date"][4:6], metadata["date"][6:8]))
```

sds_data_manager/lambda_code/SDSCode/rds_utils/db_ingest_query.py
```diff
@@ -0,0 +1,89 @@
...
+
+        self._construct_ingestion_query()
+
+    def _construct_ingestion_query(self):
+        """Constructs the ingestion query string and corresponding data structure."""
+        logger.info("Constructing metadata ingestion query")
+        # add s3 path to the id field
+        self.metadata["id"] = self.s3_path
+        # break up date into year, month, day
+        self.metadata["year"] = self.metadata["date"][:4]
```
maxinelasp: Instead of modifying this dictionary directly, this should be a new dictionary just for this method which uses values from self.metadata. This works with or without the FIleMetadata class I suggested above. 

It's confusing to change the fields in a dictionary. If I passed in a dictionary like: 
`{filename: "foobar",
date: "20231119"}`

I would expect that `db_ingest_query.date` would return something. 

Instead, this code can be vastly simplified by leaving the self.metadata modification out of it: 

```
columns = {
            "id": self.s3_path, 
            "year": self.metadata["date"][:4],
            "month": self.metadata["date"][4:6],
            "day": self.metadata["date"][6:8]
        }
 ```
 
 Alternatively, if you go with the FileMetadata class, you can just [convert it to a dictionary](https://docs.python.org/3/library/dataclasses.html?highlight=field#dataclasses.asdict)

`columns = self.metadata.asdict()`

File: sds_data_manager/lambda_code/SDSCode/rds_utils/db_ingest_query.py [hunk 21]
```diff
@@ -0,0 +1,89 @@
...
+        placeholders = ", ".join(["%s"] * len(self.metadata))
+        # format query
+        self.query = f"""
+                    INSERT INTO metadata (
+                        {fields})
+                    VALUES ({placeholders})
+                """
+        logger.info(f"query: {self.query}")
+        # create tuple to hold query data
+        self.data = tuple(self.metadata.values())
```
maxinelasp: It doesn't make a lot of sense to me to have self.query and self.data here. It just requires that later, when you're calling `cursor.execute`, you are accessing two attributes from the same dataclass. I would remove `self.data` and just insert the values directly: 

```
 values = ", ".join(self.metadata.values())

self.query = f"""
                    INSERT INTO metadata (
                        {fields})
                    VALUES ({values})
                """
```

If you want to have the flexibility to add different placeholder values, it would make more sense to me to have a `generate_empty_insert_query()` method or something which uses the metadata table (another reason not to modify it!) and this placeholder style, with the expectation that the end user will provide their own values to insert. But I don't think that's necessary right now.

sds_data_manager/lambda_code/SDSCode/create_schema.py
sdhoyt: The lengths and value types were something I'm planning to get right in the next PR when I focus on the schema. 

Yeah, I guess mission isn't really necessary. I was thinking it might be nice to have it in case we need to reference it somewhere for the files to avoid having to hardcode "imap" since sds data manager is supposed to be mission agnostic. But I guess that issue might only come up in the processing repo where it wouldn't matter.

sds_data_manager/lambda_code/SDSCode/create_schema.py
sdhoyt: Yeah, I had no idea what that was before this either. You can get a few things from that object, like the name of the function, the resource arn that invoked the lambda, memory, etc. : https://docs.aws.amazon.com/lambda/latest/dg/python-context.html

sds_data_manager/lambda_code/SDSCode/indexer.py
sdhoyt: Yeah, I'm going to remove all the OpenSearch code in a separate PR once RDS is developed enough to fully replace it.

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -0,0 +1,95 @@
...
+                type VARCHAR(100),
+                instrument VARCHAR(100),
+                level VARCHAR(10),
+                year INT,
+                month INT,
+                day INT,
+                version VARCHAR(100),
+                extension VARCHAR(10)
+            )
+        """
```
maxinelasp: Not necessary for this PR, I just want to document that we discussed in our team meeting having this query in a file.

sds_data_manager/lambda_code/SDSCode/rds_utils/db_ingest_query.py
sdhoyt: Yeah, I will add a comment explaining that. For psycopg you can pass a query string and query data to its execute method to send the query. The query string looks something like this:
```python
"""
INSERT INTO metadata (
    mission, instrument, level, id, year, month, day)
VALUES (%s, %s, %s, %s, %s, %s, %s)
"""
```

So you put the field names in the query string, but leave placeholders (the `%s`s) for the actual data. Then you pass the query string along with a tuple containing the data for each field where each value in the tuple will map to one of the placeholders (`%s`). 

I'm realizing that you might not need to pass in a separate data tuple and could add those values to the string before calling execute, so the placeholders might not be needed. But that's what they're doing here.

sds_data_manager/stacks/create_schema_stack.py
sdhoyt: Oops, yeah forgot to change that.

maxinelasp (commented):
Added some notes on the database stuff, if you're planning on changing that section in a future PR then I'm ok with you merging it. LMK if you have questions about any of my comments, they were kind of convoluted

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -0,0 +1,95 @@
...
+import psycopg2
+import requests
+from psycopg2 import Error
+
+# TODO: The code in this lambda will be updated in the next PR
+# to reflect the schema that will be needed by the SDC.
+# Its current state is the bare minimum to get a table
+# created to allow files to be indexed.
+
+# Add code to get Secret instead
```
bourque: Could you expand on this a bit? Why is S3 in particular needed for this? And what is the alternative method we are avoiding by doing it this way?

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -0,0 +1,95 @@
+import json
```
bourque, line 3: It would be good to include a module docstring here that describes its purpose

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -0,0 +1,95 @@
...
+    headers = {"content-type": "", "content-length": str(len(json_response_body))}
+
+    try:
+        response = requests.put(response_url, data=json_response_body, headers=headers)
+        response.raise_for_status()
+    except Exception as e:
+        print(f"Error sending response: {e}")
+
+
+def lambda_handler(event, context):
```
bourque, line 51: Missing docstring here

sds_data_manager/lambda_code/SDSCode/rds_utils/db_connection.py
```diff
@@ -0,0 +1,115 @@
+import json
```
bourque: It would be nice to add a module docstring here (and for the other modules that are missing one)

sds_data_manager/lambda_code/SDSCode/rds_utils/db_connection.py
```diff
@@ -0,0 +1,115 @@
...
+    cursor: psycopg2.cursor
+        psycopg2 cursor obejct that allows code to execute PostgreSQL
+        commands in the database session.
+
+    Methods
+    -------
+    send_query(query):
+        sends query to the database.
+    close():
+        closes the database connection.
```
bourque: :100: I'm a big fan of including the ``Attributes`` and ``Methods`` in the docstring, nice work!

sds_data_manager/lambda_code/SDSCode/rds_utils/db_connection.py
```diff
@@ -0,0 +1,115 @@
...
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
+
+
+class DbConnection:
+    """
+    Class to represent the connection to the RDS database.
+
+    ...
```
bourque: I see that this is in the `numpydoc` convention, but I don't understand its purpose 🤔

sds_data_manager/stacks/create_schema_stack.py
```diff
@@ -0,0 +1,66 @@
+"""SdpDatabase Stack"""
```
bourque, line 1: What is Sdp?

bourque (commented):
Great work here! I just had a few nit-picky suggestions, mostly because I get real uppity about docstrings 😂

sds_data_manager/lambda_code/SDSCode/create_schema.py
sdhoyt: That comment and line of code are confusing because they're completely unrelated and neither are needed lol Good catch, I'll delete these two lines:

```python
# Add code to get Secret instead
s3 = boto3.client("s3")
```

tech3371 (dismissed):
It still looks good to me

maxinelasp (approved):
LGTM, I like the new tests

PR #197, opened 2023-11-18
ENH: Add cloudwatch monitoring

### LINKED ISSUE #160: Setup Cloudwatch notification to send email alert if APIs have "slower" performance
The current suggestion is to send a notification if Upload/Download API performance drops below 70% of average rate for an hour

### LINKED ISSUE #173: Setup cloudwatch for processing monitoring
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

greglucas:
# Change Summary

## Overview
This adds a slack integration to send cloudwatch alarms to via an SNS Topic. The slack integration required a manual setup due to authorization permissions, but it wasn't too onerous and the steps are documented in the docstring.

One question I have is whether we want to do it this way and pass in things we want monitored (i.e. we are passing in APIGateway). Or do we want to go the other way and pass our SNS Topic and set up the cloudwatch alarms/metrics everywhere we want to integrate them so they are closer to what we are monitoring?

Is Slack integration OK for now, or do we also want to add e-mail notification? That is easy enough to add by just subscribing an e-mail to the topic as well.

closes #160
closes #173

Example of it deployed:
image
And what it looks like in Slack:
image

## New Files

- monitoring_stack.py
   - Adds an SNS Topic for Slack integration and a cloudwatch alarm based on the APIGateway.

## Updated Files

- api_gateway
   - Needed to output the api name (I couldn't figure out how to get this as a wildcard unfortunately)


greglucas:
I reworked this to put monitoring at the top now and pass the sns topic to a method called on the APIGateway stack now.

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -1,5 +1,7 @@
-from aws_cdk import Stack
```
bourque, line 1: I notice that a fair number of the modules in this repo don't have module docstrings. I try to make it a habit to include these, but perhaps I am in the minority. Do we think it is important to include module docstrings?

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -86,7 +88,40 @@ def __init__(
...
+        metric = cw.Metric(
+            namespace="AWS/ApiGateway",
+            metric_name="Latency",
+            dimensions_map={"ApiName": self.api.rest_api_name},
+            period=Duration.minutes(1),
+            statistic="Maximum",
+        )
+
+        # Define the alarm
+        cw_alarm = cw.Alarm(
```
bourque:
```suggestion
        cloudwatch_alarm = cw.Alarm(
```
Unnecessarily abbreviated variable names is a pet peeve of mine, but maybe this one gets a pass if `cw` is just the standard way to name things like this?

sds_data_manager/stacks/api_gateway_stack.py
```diff
@@ -86,7 +88,40 @@ def __init__(
             http_method = lambda_info["httpMethod"]
 
             # Define the API Gateway Resources
-            resource = api.root.add_resource(route)
+            resource = self.api.root.add_resource(route)
 
             # Create a new method that is linked to the Lambda function
             resource.add_method(http_method, apigw.LambdaIntegration(lambda_fn))
+
+    def subscribe_to_sns(self, sns_topic: aws_sns.Topic):
```
bourque: Missing docstring here

sds_data_manager/stacks/monitoring_stack.py
```diff
@@ -0,0 +1,26 @@
...
+class MonitoringStack(Stack):
+    """Monitoring stack
+
+    Set up an integration with Slack to send notifications to a channel.
+
+    This had to be done manually in the AWS console to authorize the
+    AWS account to use the Slack app.
+    Just follow the steps here to create a new Slack authorization.
+    https://us-east-2.console.aws.amazon.com/chatbot/home
+    Choosing the channel and permissions as you go through the steps.
```
bourque: Thanks for adding these instructions here!

bourque (commented):
This looks good to me, nice work! I just had some docstring-related comments.

I don't think I know enough about what is going on / I don't have strong enough opinions to comment on your first question, but in regards to email notifications -- My personal preference would be to just stick with Slack alerts. I tend to instantly ignore automated emails 😅

sds_data_manager/stacks/api_gateway_stack.py
greglucas: I agree, this repo is in a pretty inconsistent state right now :-/ I am +0.5 on module docstrings. So I added them in. I think in most of these cases it actually made sense to move the class docstrings up to the module level, so that is what I did.

sds_data_manager/stacks/api_gateway_stack.py
greglucas: Agreed, thanks for bringing this up. I updated this and the imports. I think we could actually remove a bunch of the `from aws_cdk import aws_xxx as xxx` and just change them to `from aws_cdk import aws_xxx`, the `aws_xxx` is actually somewhat helpful for disambiguating resources we create vs import from cdk. But, I'll leave that for a follow-up cleanup PR.

sds_data_manager/stacks/api_gateway_stack.py
greglucas: Thanks! Fixed.

sds_data_manager/stacks/monitoring_stack.py
greglucas: Moved this info up to the module level now.

sds_data_manager/stacks/api_gateway_stack.py
tech3371: I have notice that and have been wondering if we should start doing that. I like both ways. Not particular about it.

sds_data_manager/stacks/monitoring_stack.py
```diff
@@ -0,0 +1,43 @@
...
+        Attributes
+        ----------
+        sns_topic_notifications : aws_sns.Topic
+            SNS Topic for sending notifications to that external
+            resources can subscribe to for alerts.
+        """
+        super().__init__(scope, construct_id, **kwargs)
+        self.sns_topic_notifications = aws_sns.Topic(
+            self, "sns-notifications", display_name="sns-notifications"
+        )
```
tech3371, line 43: I don't any subscription to the `SNS`. Is it because we have to integrate to Slack manually?

tech3371 (approved):
Looks good to me. It was easy to follow.

sds_data_manager/stacks/monitoring_stack.py
greglucas: Good catch! I actually hadn't deployed this new version when you commented because I was waiting until people voted on which method they'd prefer. I just deployed it though. But you are correct, you have to manually go in and subscribe the chatbot to this Topic. Do you want to do that now to see what it looks like?

tech3371: I see. We can add more subscription in future PR too

greglucas: Exactly.

PR #198, opened 2023-12-01
fix to upload folder path

### LINKED ISSUE #194: Update upload lambda to use filename to create upload path
Link to code that needs update: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/1f6109cd03eb35d773e0b6336627cff0785efe79/sds_data_manager/lambda_code/SDSCode/upload_api.py#L56

tech3371:
# Change Summary

## Overview
Fix upload folder path

## Updated Files
- sds_data_manager/config/config.json
   - update and added new product
   - Added fixes for upload folder path

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,3 +100,4 @@ ignore = ["D203", "D212", "PLR0913", "PLR2004"]
 # TODO: Too many statements, this could be refactored to separate
 #       the single stack out into a few smaller pieces
 "sds_data_manager/stacks/sds_data_manager_stack.py" = ["PLR0915"]
+"sds_data_manager/lambda_code/SDSCode/path_helper.py" = ["B008"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -100,4 +100,3 @@ ignore = ["D203", "D212", "PLR0913", "PLR2004"]
 # TODO: Too many statements, this could be refactored to separate
 #       the single stack out into a few smaller pieces
 "sds_data_manager/stacks/sds_data_manager_stack.py" = ["PLR0915"]
-"sds_data_manager/lambda_code/SDSCode/path_helper.py" = ["B008"]
--- a/sds_data_manager/config/config.json
+++ b/sds_data_manager/config/config.json
@@ -1,8 +1,67 @@
 [
-{"product": "IMAP-L0-File",
-"pattern": {"mission":"imap", "level":"l0", "type":"sci", "instrument":"*", "date":"*", "version":"*", "extension":"pkts"},
-"path": "imap/l0/"},
-{"product": "IMAP-L1-File",
-"pattern": {"mission":"imap", "level":"l1", "instrument":"*", "date":"*", "version":"*", "extension":"fits"},
-"path": "imap/l1/"}
+    {
+        "product": "IMAP-L0-File",
+        "pattern": {
+            "mission": "imap",
+            "instrument": "*",
+            "level": "l0",
+            "descriptor": "*",
+            "startdate": "*",
+            "enddate": "*",
+            "version": "*",
+            "extension": "pkts"
+        }
+    },
+    {
+        "product": "IMAP-L1-File",
+        "pattern": {
+            "mission": "imap",
+            "instrument": "*",
+            "level": "l1",
+            "descriptor": "*",
+            "startdate": "*",
+            "enddate": "*",
+            "version": "*",
+            "extension": "cdf"
+        }
+    },
+    {
+        "product": "IMAP-L1A-File",
+        "pattern": {
+            "mission": "imap",
+            "instrument": "*",
+            "level": "l1a",
+            "descriptor": "*",
+            "startdate": "*",
+            "enddate": "*",
+            "version": "*",
+            "extension": "cdf"
+        }
+    },
+    {
+        "product": "IMAP-L1B-File",
+        "pattern": {
+            "mission": "imap",
+            "instrument": "*",
+            "level": "l1b",
+            "descriptor": "*",
+            "startdate": "*",
+            "enddate": "*",
+            "version": "*",
+            "extension": "cdf"
+        }
+    },
+    {
+        "product": "IMAP-L1C-File",
+        "pattern": {
+            "mission": "imap",
+            "instrument": "*",
+            "level": "l1c",
+            "descriptor": "*",
+            "startdate": "*",
+            "enddate": "*",
+            "version": "*",
+            "extension": "cdf"
+        }
+    }
 ]
--- a/sds_data_manager/lambda_code/SDSCode/path_helper.py
+++ b/sds_data_manager/lambda_code/SDSCode/path_helper.py
@@ -96,6 +96,10 @@ class FilenameParser:
 
         # Validate if it's a real date
         try:
+            # This checks if date is in YYYYMMDD format.
+            # Sometimes, date is correct but not in the format we want
+            if not re.match(r"^\d{8}$", input_date):
+                raise ValueError("Invalid date format.")
             datetime.strptime(input_date, "%Y%m%d")
             return True
         except ValueError:
@@ -117,6 +121,21 @@ class FilenameParser:
         bool
             Whether filename format is valid or not
         """
+        # First check if any of parameter is missing
+        if any(
+            attr is None or attr == ""
+            for attr in [
+                self.mission,
+                self.instrument,
+                self.data_level,
+                self.descriptor,
+                self.startdate,
+                self.enddate,
+                self.version,
+                self.extension,
+            ]
+        ):
+            return False
 
         # Dictionary to map fields to their valid values and error messages
         validation_checks = {
@@ -138,7 +157,6 @@ class FilenameParser:
                     f"{file_pattern_config.data_level}"
                 ),
             ),
-            "descriptor": (self.descriptor is not None, "Descriptor is required."),
             "startdate": (
                 self.check_date_input(self.startdate),
                 "Invalid start date format. Please use YYYYMMDD format.",
@@ -169,7 +187,7 @@ class FilenameParser:
             if not is_valid:
                 self.message = error_message
                 return False
-
+        print("done")
         return True
 
     def create_path_to_upload(self) -> str:
--- a/sds_data_manager/lambda_code/SDSCode/path_helper.py
+++ b/sds_data_manager/lambda_code/SDSCode/path_helper.py
@@ -98,7 +98,7 @@ class FilenameParser:
         try:
             # This checks if date is in YYYYMMDD format.
             # Sometimes, date is correct but not in the format we want
-            if not re.match(r"^\d{8}$", input_date):
+            if len(input_date) != 8:
                 raise ValueError("Invalid date format.")
             datetime.strptime(input_date, "%Y%m%d")
             return True
--- a/sds_data_manager/lambda_code/SDSCode/path_helper.py
+++ b/sds_data_manager/lambda_code/SDSCode/path_helper.py
@@ -105,7 +105,7 @@ class FilenameParser:
         except ValueError:
             return False
 
-    def validate_filename(self, file_pattern_config=FilenamePatternConfig()) -> bool:
+    def validate_filename(self, file_pattern_config=None) -> bool:
         """Validate filename patterns.
 
         Parameters
@@ -121,6 +121,9 @@ class FilenameParser:
         bool
             Whether filename format is valid or not
         """
+        if file_pattern_config is None:
+            file_pattern_config = FilenamePatternConfig()
+
         # First check if any of parameter is missing
         if any(
             attr is None or attr == ""
@@ -187,7 +190,6 @@ class FilenameParser:
             if not is_valid:
                 self.message = error_message
                 return False
-        print("done")
         return True
 
     def create_path_to_upload(self) -> str:
--- a/sds_data_manager/lambda_code/SDSCode/path_helper.py
+++ b/sds_data_manager/lambda_code/SDSCode/path_helper.py
@@ -1,3 +1,5 @@
+"""Functions to store file pattern, validate and construct upload path
+"""
 import re
 from dataclasses import dataclass, field
 from datetime import datetime
@@ -78,6 +80,18 @@ class FilenameParser:
         self.message = None
 
     def check_date_input(self, input_date: str) -> bool:
+        """Check input date string is in valid format and is correct date
+
+        Parameters
+        ----------
+        input_date : str
+            Date in YYYYMMDD format.
+
+        Returns
+        -------
+        bool
+            Whether date input is valid or not
+        """
         # Check if the pattern matches 8 digits (YYYYMMDD)
         if not re.match(r"^\d{8}$", input_date):
             return False
@@ -90,6 +104,13 @@ class FilenameParser:
             return False
 
     def validate_filename(self) -> bool:
+        """Validate filename patterns.
+
+        Returns
+        -------
+        bool
+            Whether filename format is valid or not
+        """
         pattern = FilenamePatternConfig()
 
         # Dictionary to map fields to their valid values and error messages
@@ -150,6 +171,13 @@ class FilenameParser:
         return path_to_upload_file
 
     def upload_filepath(self):
+        """Return upload path or error message.
+
+        Returns
+        -------
+        dict
+            Upload path or error message
+        """
         if not self.validate_filename():
             default_message = (
                 f"Invalid filename. Filename convention is {self.filename_convention}"
--- a/sds_data_manager/lambda_code/SDSCode/path_helper.py
+++ b/sds_data_manager/lambda_code/SDSCode/path_helper.py
@@ -60,11 +60,12 @@ class FilenameParser:
         filename : str
             Filename
         """
+        self.filename = filename
         self.filename_convention = (
             "<mission>_<instrument>_<datalevel>_<descriptor>_"
             "<startdate>_<enddate>_<version>.<extension>"
         )
-        self.split_filename = filename.replace("_", ".").split(".")
+        self.split_filename = filename.split("_")
         (
             self.mission,
             self.instrument,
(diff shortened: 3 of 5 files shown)


sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -53,6 +56,35 @@ def _check_for_matching_filetype(pattern, filename):
     return file_dictionary
 
 
+def create_path_to_upload(metadata: dict) -> str:
```
greglucas: Do we need this new function or could you use this to replace the `check_for_matching_filetype` function above this? Since we have such a rigid structure now, I'm not sure we even need the config dictionary file.

```python
mission, instrument, data_level, descriptor, startdate, enddate = filename.split("_")
# create your path from these, maybe raise if something doesn't match our requirements
```

tech3371: `check_for_matching_filetype` checks that the filename follows one of the pattern in the `config.json`. Once it finds that format, then it stops there and continue. If we remove `check_for_matching_filetype`, then having `config.json` is not useful.

greglucas: Yeah, so up to you whether you think we'll need that check/config, but now that we are more rigid in the filename requirements I'm not so sure we will was my suggestion to merge them together. Adding onto my above suggestion we could add checks from the config file directly here as well:
```python
mission, instrument, data_level, descriptor, startdate, enddate = filename.split("_")
# create your path from these, maybe raise if something doesn't match our requirements
if not enddate.endswith((".pkts", ".cdf")):
    raise ValueError("Bad filetype, the SDC requires '.pkts' or '.cdf' filetypes to be uploaded")
if instrument not in (...):
    raise ValueError("Acceptable instruments are: ...")
# Even more validation on things if you want as well.
file_path_to_upload = same_as_you_have_below
```

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -53,6 +56,35 @@ def _check_for_matching_filetype(pattern, filename):
...
+    folder path in S3 bucket.
+
+    Parameters
+    ----------
+    metadata : dict
+        metadata derived from filename.
+
+    Returns
+    -------
+    str
```
bourque:
```suggestion
    path_to_upload_file : str
```

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -53,6 +56,35 @@ def _check_for_matching_filetype(pattern, filename):
...
+        Folder path
+    """
+    # path to upload file follows this format:
+    # mission/instrument/data_level/descriptor/year/month/filename
+    # NOTE: year and month is from startdate and startdate format is YYYYMMDD.
```
bourque: I suggest to add this info to the function docstring rather than an inline comment here. I think then it could be more easily referenced within the sphinx docs.

bourque (dismissed):
Looks good! I just had a few non-blocking suggestions.

tech3371:
@greglucas and @maxinelasp Can you review the `path_helper.py` to see if I captured your suggestions? After your approval, I will work towards adding tests and incorporate into `indexer.py` too.

sds_data_manager/config/config.json
greglucas, line 1: I think we were going to remove this file. Another benefit I thought of is that we won't need to download this onto the Lambda every single time if we put the configuration directly into the code.

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -0,0 +1,186 @@
+"""Functions to store file pattern, validate and construct upload path
+"""
+import re
+from dataclasses import dataclass, field
+from datetime import datetime
+
+
+@dataclass
+class FilenamePatternConfig:
```
greglucas, line 9: I sort of worry we are going too class-heavy on IMAP... Is there a reason we need this stored/instantiated as an object versus using chains of functions? I think you could chain a bunch of functions together and it would be easier to re-use for other purposes too then.

My personal preference is to use global constants for your valid options and then check against those:
```python
VALID_INSTRUMENTS = {"codice", "glows", ...}
# Could make this more robust later and restrict it to specific data levels for each instrument
VALID_DATALEVELS = {"l0", ...}

def is_filename_valid(filename):
    instrument, datalevel, ... = filename.split("_")
    if instrument not in VALID_INSTRUMENTS:
        raise ValueError(f"{instrument} is not a valid instrument on IMAP, must be one of {VALID_INSTRUMENTS})
```

You could also break those checks out into individual functions if you wanted too.

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -0,0 +1,186 @@
...
+        Parameters
+        ----------
+        filename : str
+            Filename
+        """
+        self.filename_convention = (
+            "<mission>_<instrument>_<datalevel>_<descriptor>_"
+            "<startdate>_<enddate>_<version>.<extension>"
+        )
+        self.split_filename = filename.replace("_", ".").split(".")
```
greglucas: I don't trust that people won't put `.` in the middle of their data product names... I guess we could enforce that, but I think we could handle the underscores alone, and then strip off the extension later.
```suggestion
        self.split_filename = filename.split("_")
```

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -0,0 +1,186 @@
...
+            Date in YYYYMMDD format.
+
+        Returns
+        -------
+        bool
+            Whether date input is valid or not
+        """
+        # Check if the pattern matches 8 digits (YYYYMMDD)
+        if not re.match(r"^\d{8}$", input_date):
+            return False
```
greglucas: Do you need this, or can you just let your try/except below catch this case too?

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -0,0 +1,186 @@
...
+
+        # Iterate through each validation check
+        for _field, (is_valid, error_message) in validation_checks.items():
+            if not is_valid:
+                self.message = error_message
+                return False
+
+        return True
+
+    def create_path_to_upload(self) -> str:
```
greglucas, line 195: Rather than just creating the path to the upload, should we just create the entire s3 uri here? We have the filename too, so I think that would save us the headache later of having to append the filename to this path.

sds_data_manager/lambda_code/SDSCode/path_helper.py
maxinelasp: Agree with Greg, below will catch only 8 digit times

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -124,3 +78,9 @@ def lambda_handler(event, context):
         }
 
     return {"statusCode": 200, "body": json.dumps(url)}
+
+
+if __name__ == "__main__":
```
maxinelasp: Is this intentionally included? Or just for testing?

sds_data_manager/lambda_code/SDSCode/path_helper.py
maxinelasp: I like the class, personally. That being said, I would probably prefer to pass it in to the functions that use it as a parameter so other users could override it easily. (eg in line 114, maybe you could have `file_pattern_config` as an optional keyword and default to this FilePatternConfig class.)

maxinelasp (dismissed):
I like your changes, they look good to me. If config.json isn't needed anymore, then it should be removed before merging.

sds_data_manager/config/config.json
tech3371: Yes, True.
Yeah, we will need to remove this file. I didn't do that in this PR yet, because removing this is going to effect other CDK stack and I didn't want to bring in more changes in this PR. I plan to remove this and stacks that are associated with this in upcoming PR:
* S3 bucket stack
* resources that uploads this file
* and may be others as well that I don't about.

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: I do like Class for this case because it groups all associated parameters together and easy to access.

So @maxinelasp Are you suggesting something like this?
```
    def validate_filename(self, file_pattern_config=FilenamePatternConfig()) -> bool:

```

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: ok. will do. Changing to this:
```
        self.split_filename = filename.split("_")
        (
            self.mission,
            self.instrument,
            self.data_level,
            self.descriptor,
            self.startdate,
            self.enddate,
            self.last_value
        ) = self.split_filename
        (self.version, self.extension) = self.last_value.split(".")
```

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: cool. I will remove this.

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: Will do that.

sds_data_manager/lambda_code/SDSCode/upload_api.py
tech3371: yes. just for testing. good catch!

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: I push new code with these changes. I hope I address your comments.

maxinelasp: Yes, that's what I had in mind!

tech3371:
@bourque @maxinelasp @greglucas Can you review it again? No rush. I have to ping because this repo undoes your approval.

pyproject.toml
```diff
@@ -98,3 +98,4 @@ ignore = ["D203", "D212", "PLR0913", "PLR2004"]
 # TODO: Too many statements, this could be refactored to separate
 #       the single stack out into a few smaller pieces
 "sds_data_manager/stacks/sds_data_manager_stack.py" = ["PLR0915"]
+"sds_data_manager/lambda_code/SDSCode/path_helper.py" = ["B008"]
```
greglucas: I don't think we should avoid this error.

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -0,0 +1,225 @@
...
+        bool
+            Whether date input is valid or not
+        """
+
+        # Validate if it's a real date
+        try:
+            # This checks if date is in YYYYMMDD format.
+            # Sometimes, date is correct but not in the format we want
+            if not re.match(r"^\d{8}$", input_date):
+                raise ValueError("Invalid date format.")
```
greglucas: The below strptime should take care of all the regex work for you.
```suggestion
```

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -0,0 +1,225 @@
...
+            datetime.strptime(input_date, "%Y%m%d")
+            return True
+        except ValueError:
+            return False
+
+    def validate_filename(self, file_pattern_config=FilenamePatternConfig()) -> bool:
```
greglucas:
```suggestion
    def validate_filename(self, file_pattern_config=None) -> bool:
        ...
        if file_pattern_config is None:
            file_pattern_config = FilenamePatternConfig()
```

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -0,0 +1,225 @@
...
+                ),
+            ),
+        }
+
+        # Iterate through each validation check
+        for _field, (is_valid, error_message) in validation_checks.items():
+            if not is_valid:
+                self.message = error_message
+                return False
+        print("done")
```
greglucas:
```suggestion
```

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -92,39 +93,70 @@ def check_date_input(self, input_date: str) -> bool:
...
         """
-        # Check if the pattern matches 8 digits (YYYYMMDD)
-        if not re.match(r"^\d{8}$", input_date):
-            return False
 
         # Validate if it's a real date
         try:
+            # This checks if date is in YYYYMMDD format.
+            # Sometimes, date is correct but not in the format we want
+            if not re.match(r"^\d{8}$", input_date):
```
maxinelasp: This can just be removed, as the `strptime` will also fail if the input date isn't 8 digits exactly.

maxinelasp (approved):

tests/lambda_endpoints/test_path_helper.py
```diff
@@ -0,0 +1,80 @@
...
+
+from sds_data_manager.lambda_code.SDSCode.path_helper import FilenameParser
+
+
+def test_date_checker():
+    """Tests date inputs"""
+    filename = "imap_glows_l0_raw_20231010_20231011_v01-01.pkts"
+    file_parser = FilenameParser(filename)
+    assert file_parser.check_date_input("20200101")
+    assert file_parser.check_date_input("2020-01-01") is False
```
greglucas:
```suggestion
    assert not file_parser.check_date_input("2020-01-01")
```

tests/lambda_endpoints/test_path_helper.py
```diff
@@ -0,0 +1,80 @@
...
+    assert file_parser.check_date_input("202301") is False
+    # NOTE: the date is correct but not in the format YYYYMMDD
+    assert file_parser.check_date_input("2023105") is False
+
+
+def test_filename_validator():
+    """Validate filenames"""
+    filename = "imap_glows_l0_raw_20231010_20231011_v01-01.pkts"
+    assert FilenameParser(filename).validate_filename() is True
```
greglucas: comparisons should be `assert something_is_true` or `assert not something_is_false` and not comparing against the specific object.
```suggestion
    assert FilenameParser(filename).validate_filename()
```

tests/lambda_endpoints/test_path_helper.py
```diff
@@ -0,0 +1,80 @@
+import pytest
+
+from sds_data_manager.lambda_code.SDSCode.path_helper import FilenameParser
+
+
+def test_date_checker():
+    """Tests date inputs"""
+    filename = "imap_glows_l0_raw_20231010_20231011_v01-01.pkts"
+    file_parser = FilenameParser(filename)
+    assert file_parser.check_date_input("20200101")
```
greglucas, line 10: Why do you need to input a date to the fileparser? Shouldn't this be acting on `self.date`?

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: I did remove it originally based on feedback from you and Greg. But when I input correct date such as `2023105`, `strptime` accepted because it took that as valid date even though its document says otherwise.
https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior

tests/lambda_endpoints/test_path_helper.py
```diff
@@ -0,0 +1,80 @@
...
+    assert file_parser.check_date_input("2020-01-01") is False
+    assert file_parser.check_date_input("202301") is False
+    # NOTE: the date is correct but not in the format YYYYMMDD
+    assert file_parser.check_date_input("2023105") is False
+
+
+def test_filename_validator():
```
greglucas: These tests might be easier to parameterize over your assertions.
https://docs.pytest.org/en/7.1.x/example/parametrize.html

```suggestion
@pytest.mark.parametrize("filename,expected", [("imap_glows_l0_raw_20231010_20231011_v01-01.pkts", True), ...]
def test_filename_validator(filename, expected):
    assert FilenameParser(filename).validate_filename() == expected
```

greglucas (commented):
I'm fine with this and won't block, but I do find this harder to follow by creating a class with a bunch of methods than individual functions would be. Since others like it this way that is fine though.

```python
def create_upload_key(fname, date_validator=is_valid_date):
    # Validate the pieces of our filename
    mission, instrument, level = fname.split()
    # call out to other validation functions
    if not date_validator(start_date):
        return False
    ...
    # assemble parts after validations

def is_valid_date(datestr, date_format="YYYYMMDD"):
    try:
        datetime.strptime(datestr, date_format)
    except ValueError:
        return False
    return True
```

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: I will remove it since you guys are sure of this. As I mentioned above, `strptime` takes in seven digit date as long as it's a valid date such as `2023101`. I feel like I am missing something here.

tests/lambda_endpoints/test_path_helper.py
tech3371: Yes. but I wrote that function to take in an input date since we need to check for both startdate and enddate. And since I wanted to check the date format only, I figure I could call the function directly instead of calling class again and again.

sds_data_manager/lambda_code/SDSCode/path_helper.py
maxinelasp: Odd! Well, good job verifying 😆

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: let me know if you like me to remove it still.

greglucas: Nope, it looks like I was wrong here! Good catch on your part and sorry for both of us mentioning you could remove it 🐑 

Another option to avoid regex would be to just check the length: `if len(input_date) != 8` since the following sequence will check the numbers.

tech3371: true. that's much simpler.

PR #201, opened 2023-12-07
CI: Remove version pin on npm cdk version installation

greglucas:
# Change Summary

This was out of sync with the Python cdk code version and couldn't be used together. Unpin for now to hope that the latest updates keep us in sync. We can pin again later if needed.

https://github.com/IMAP-Science-Operations-Center/sds-data-manager/actions/runs/7121173929/job/19389864426#step:8:795

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Install cdk
         run: |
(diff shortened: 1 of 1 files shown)


tech3371 (approved):
sounds good. Latest release has caused some issue locally and lot of issues was resolved by installing latest aws-cdk using npm. Let's hope it does it here too.

PR #204, opened 2023-12-15
Create spin table

tech3371:
# Change Summary
Create universal spin table using SQLAlchemy ORM

## New Files
- sds_data_manager/lambda_code/SDSCode/database/
   - This folder contains main files that setups database information.
   - It creates and stores information such as engine, schema definitions and db uri

## Updated Files
- pyproject.toml
   - add SQLAlchemy
- sds_data_manager/lambda_code/SDSCode/create_schema.py
   - Added logic to create all tables

## Testing
TODO: add tests


sds_data_manager/lambda_code/SDSCode/database/database.py
```diff
@@ -0,0 +1,5 @@
+"""Create engine that will be used to create schema or make query"""
+from SDSCode.database import config
+from sqlalchemy import create_engine
```
bourque, line 3:
```suggestion
from sqlalchemy import create_engine

from SDSCode.database import config
```

sds_data_manager/lambda_code/SDSCode/database/config.py
```diff
@@ -0,0 +1,21 @@
...
+import json
+import os
+
+import boto3
+
+
+def get_db_uri():
+    """Create DB URI from secret manager.
+    Returns:
+        str: DB URI
```
bourque:
```suggestion

    Returns
    --------
        str : DB URI
```

sds_data_manager/lambda_code/SDSCode/database/config.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    secret_name = os.environ["SECRET_NAME"]
+    session = boto3.session.Session()
+    client = session.client(service_name="secretsmanager")
+    secret_string = client.get_secret_value(SecretId=secret_name)["SecretString"]
+    secret = json.loads(secret_string)
```
bourque:
```suggestion
    db_config = json.loads(secret_string)
```

I think this name is a bit more descriptive on what kind of data it is storing, but that is just personal preference

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -0,0 +1,33 @@
...
+    Column,
+    DateTime,
+    Integer,
+)
+from sqlalchemy.orm import DeclarativeBase
+
+
+class Base(DeclarativeBase):
+    pass
+
```
bourque, line 13:
```suggestion
Base = sa.orm.declarative_base()
```
If you want to do this in one less line!  In my experience, I don't think we should ever need to define things in the `Base` class(?)

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -0,0 +1,33 @@
...
+    __tablename__ = "universal_spin_table"
+    id = Column(Integer, primary_key=True)
+    spin_number = Column(Integer, primary_key=True)
+    spin_start_sc_time = Column(Integer)
+    spin_start_utc_time = Column(DateTime(timezone=True))
+    star_tracker_flag = Column(Boolean)
+    spin_duration = Column(Integer)
+    thruster_firing_event = Column(Boolean)
+    repointing = Column(Boolean)
+    repointing_number = Column(Integer)
```
bourque: Can the value for these ever be `NULL`? For any ones that cannot be `NULL`, you could add the `nullable=False` parameter.

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -5,14 +5,16 @@
 import sys
 
 import requests
+from SDSCode.database.database import engine
+from SDSCode.database.models import Base
```
bourque, line 9:
```suggestion
import requests

from SDSCode.database.database import engine
from SDSCode.database.models import Base
```

bourque (approved):
Nice work! I just have a few non-blocking comments/suggestions.

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: Should this be a foreign key? Like we want a repointing table with information about repoints too? When a repoint started/stopped for example.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def __str__(self):
+        """Returns a string representation of this object."""
+        return "%s %s %d" % (self.directory_path, self.file_name, self.file_size)
+
+    __repr__ = __str__
```
greglucas: What is this used for? I don't see the directory path, file name, or file size anywhere. I figured it was all done touching a server somewhere and not anything with our local directory / files.

sds_data_manager/lambda_code/SDSCode/create_schema.py
tech3371: it seems like pre-commit checks removes that space between `requests` and `engine` lines.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: That's what I saw in others but I found this way to be more consistence with object oriented.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: Yeah, I think all of the field should have a value and shouldn't be null. But we can change that later if things changes. For now, I will set them to not be null.

About foreign key, I have not thought about it. That's a good point though.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: I wonder about that too. I don't know enough.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: @greglucas can I revisit foreign key in upcoming PR because I like to think about that little bit.

sds_data_manager/lambda_code/SDSCode/database/database.py
tech3371: again pre-commit undoes this change.

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: Yes, no problem at all! Maybe add a quick comment/TODO above those lines so we don't completely forget about it.

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: Can you remove it then?

sds_data_manager/lambda_code/SDSCode/database/database.py
bourque: Hmm, that is odd to me. Since one is a third-party import and the other is a local package import.

tech3371: yep

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -0,0 +1,29 @@
...
+    Integer,
+)
+from sqlalchemy.orm import DeclarativeBase
+
+
+class Base(DeclarativeBase):
+    pass
+
+
+class UniversalSpinTable(Base):
```
sdhoyt, line 15: It looks like there's a mocking library for sqlalchemy. Is this something we might want to use for unit testing? https://pypi.org/project/alchemy-mock/

sdhoyt (approved):
Looks great!

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: That or may be this: https://pypi.org/project/pytest-sqlalchemy-mock/

tech3371:
I addressed all comments. I am going to merge this PR.

PR #206, opened 2023-12-18
ENH: Add redirect response from download url

### LINKED ISSUE #144: Add Redirect on download API
### Summary

Rather than returning a link which can then be used to download a file, the download API should redirect automatically to the download link

### Requirements

The more specific the better.
  - [ ] Download API redirects to download link instead of returning the link

### Related tickets

_No response_

greglucas:
# Change Summary

## Overview

Add an HTTP 302 redirect with the location for the s3 presigned url so that the browser/download utility can immediately go to the location and not require an additional request on the user's part.

If you go to a file that exists in s3 through the download api you should automatically start downloading the file. For example, this url now.
https://api.dev.imap-mission.com/download?s3_uri=s3://sds-data-449431850278/CodiceHi/l1a/l2_Codice.h5

closes #144

## New Dependencies
N/A

## New Files
N/A

## Deleted Files
N/A

## Updated Files

- download_query_api.py
   - Changed response to return statusCode 302 and the Location in the header

## Testing
Updated tests to reflect the expected response


bourque (approved):

tech3371 (approved):
looks great! Good to know about 302

PR #208, opened 2023-12-20
I-ALiRT: First pass at architecture for IOIS

### LINKED ISSUE #225: I-ALiRT: Create initial EC2 for POC
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
First pass at creating way to run IOIS. For this version I am only running the container on EC2. Next steps will be mounting the s3 bucket to the container and adding ECS. I expect the setup of things like the database to change so try not to focus too much on that part of the code. ECS will add the capability to not have to manually pull the latest image. We can use what Greg suggested: https://stackoverflow.com/questions/34840137/how-do-i-deploy-updated-docker-images-to-amazon-ecs-tasks

ECS abstracts and reduces much of the management overhead associated with running containers and offers autoscaling, but not load balancing. 
Autoscaling - launching additional EC2 instances as load increases
Load balancing - balances network traffic across instances
* I think we may only need autoscaling, but this should be discussed more to confirm.

## Updated Files
- stack_builder.py
   - Used to deploy dynamodb, ecr, and processing stacks for ialirt.
- compute_resources.py
   - EC2 compute environment.
- processing_stack.py
   - Processing system for ialirt
- ecr_stack.py
   - Created ECR policy
- Renaming:
   - instrument_lambdas.py
   - sdc_step_function.py
   - database_stack.py
   - test_step_function_starter.py

## Testing
   - Verify the successful deployment and operation in an ECR and EC2 setup.
- Dockerfile
   - Used to Dockerize test_app.py
- test_app.py
   - Simple Flask web application designed to be Dockerized
and deployed on an EC2 instance.


laspsandoval:
Not certain if I should add Flask to pyproject.toml if it is only used for testing in the Docker container. What do you think?

sds_data_manager/constructs/compute_resources.py
```diff
@@ -204,3 +204,101 @@ def __init__(
             },
             tags={"Purpose": "Batch Processing"},
         )
+
+
+class IalirtEC2Resources(Construct):
```
greglucas: I'd suggest moving everything into its own file like `ialirt_stack.py` so this can really be separated more easily.

sds_data_manager/constructs/compute_resources.py
```diff
@@ -204,3 +204,101 @@ def __init__(
...
+        # Allow ingress to LASP IP address range and specific port
+        security_group.add_ingress_rule(
+            ec2.Peer.ipv4("128.138.131.0/24"),
+            ec2.Port.tcp(8080),
+            "Allow inbound traffic on TCP port 8080",
+        )
+
+        # Create an IAM role for the EC2 instance
+        # - Read-only access to AWS ECR
+        # - Basic instance management via AWS Systems Manager
```
greglucas: Thanks for these comments! Maybe add a why too, like: "AWS ECR to pull the image onto the instance"

What is the AWS Systems Manger needed for?

sds_data_manager/constructs/compute_resources.py
```diff
@@ -204,3 +204,101 @@ def __init__(
...
+            managed_policies=[
+                iam.ManagedPolicy.from_aws_managed_policy_name(
+                    "AmazonEC2ContainerRegistryReadOnly"
+                ),
+                iam.ManagedPolicy.from_aws_managed_policy_name(
+                    "AmazonSSMManagedInstanceCore"
+                ),
+            ],
+        )
+        ec2_role.add_to_policy(ecr_policy)
```
greglucas: Do you need the `ecr_policy` here too? You're adding the ECR ReadOnly above, so I think this might be expanding the permissions.

sds_data_manager/constructs/compute_resources.py
```diff
@@ -204,3 +204,101 @@ def __init__(
...
+
+        # Create an EC2 instance
+        ec2_instance = ec2.Instance(
+            self,
+            "IalirtEC2Instance",
+            instance_type=ec2.InstanceType(instance_type),
+            machine_image=ec2.MachineImage.latest_amazon_linux2(),
```
greglucas: If you use the ECS-optimized AMI to start from I _think_ you may get Docker installed already and be able to minimize some of the commands needed in the user-data script.

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-optimized_AMI.html

We should probably be using the latest 2023 versions as well so we aren't too far behind later.

sds_data_manager/constructs/compute_resources.py
```diff
@@ -204,3 +204,101 @@ def __init__(
...
+            vpc=vpc,
+            vpc_subnets=ec2.SubnetSelection(subnet_type=ec2.SubnetType.PUBLIC),
+            security_group=security_group,
+            role=ec2_role,
+            user_data=user_data,
+        )
+
+        ec2_instance.add_user_data(user_data.render())
```
greglucas: You have the `user_data=user_data` above, so I think this may not be needed.

sds_data_manager/constructs/instrument_lambdas.py
greglucas, line 1: Avoid updating these names in this PR. Feel free to open up one-line PRs if you want to change names. I'm all for small PRs rather than lumping them into this.

sds_data_manager/stacks/ecr_stack.py
```diff
@@ -52,3 +52,13 @@ def __init__(
             ecr_authenticators.add_user(user)
 
         self.container_repo.apply_removal_policy(RemovalPolicy.DESTROY)
+
+        # Create ECR Policy
```
greglucas: If these other policies are needed can you add them over in the IalirtEC2 stack location since this isn't used directly here?
I'm not sure this is needed though, or maybe you could find a managed policy that enables the PULL for a specific repository easier?

sds_data_manager/utils/stackbuilder.py
```diff
@@ -192,6 +193,40 @@ def build_sds(
         env=env,
     )
 
+    # I-ALiRT IOIS ECR
```
greglucas, line 195: Make a separate `build_ialirt` function so it can be deployed independently?

Although you are using the SDS VPC, so maybe that is not possible...

sds_data_manager/utils/stackbuilder.py
```diff
@@ -192,6 +193,40 @@ def build_sds(
...
+        env=env,
+        instrument_name="IalirtEcr",
+    )
+
+    # I-ALiRT IOIS DynamoDB
+    # ingest-ugps: ingestion ugps - 64 bit
+    # sct-vtcw: spacecraft time ugps - 64 bit
+    # src-seq-ctr: increments with each packet (included in filename?)
+    # ccsds-filename: filename of the packet
+    dynamodb_stack.DynamoDB(
```
greglucas: I think you'll need to rebase because Tenzin got rid of this stack. I'd suggest just directly making your DynamoDB Table in the I-ALiRT stack and not bothering with a separate Stack() construct.

We should put `on_demand=True` for now since it really won't be doing much of anything at first and doesn't need read/write capacity always on.

tests/infrastructure/test_ialirt_processing_stack.py
```diff
@@ -0,0 +1,44 @@
...
+
+    url = f"http://{EC2_IP}:8080/"
+    try:
+        response = requests.get(url, timeout=60)
+        # Check if the response content is "Hola Mundo."
+        assert response.text == "Hola Mundo."
+    except requests.exceptions.ConnectionError:
+        pytest.fail("Failed to connect to the Flask application.")
+    except requests.exceptions.Timeout:
+        pytest.fail("Connection to the Flask application timed out.")
```
greglucas: These errors are pretty self explanatory so probably don't need the except / fail and just let pytest fail for you. (remove the entire try/except block)

tests/infrastructure/test_ialirt_processing_stack.py
```diff
@@ -0,0 +1,44 @@
...
+EC2_IP = os.getenv("EC2_IP_ADDRESS")
+
+
+def test_flask_app_response():
+    """Test the Flask application response."""
+    assert EC2_IP is not None, "EC2 IP Address is not set in environment variables."
+
+    url = f"http://{EC2_IP}:8080/"
+    try:
+        response = requests.get(url, timeout=60)
```
greglucas: Probably can set this to 5-10 seconds so you fail much sooner?

greglucas:
> Not certain if I should add Flask to pyproject.toml if it is only used for testing in the Docker container. What do you think?

Definitely not. You put it in the Dockerfile to be installed that way and that is perfect since it is contained purely to that image and not used anywhere else. So 👍 to how you have it now.

sds_data_manager/constructs/compute_resources.py
laspsandoval: Yes. That is a benefit to the ECS. I'll include the latest version in the next code revision.

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -0,0 +1,63 @@
...
+
+        # Setup the EC2 resources
+        self.ec2_resources = IalirtEC2Resources(
+            self,
+            "IalirtEC2Environment",
+            vpc=vpc,
+            repo=repo,
+        )
+
+        # I-ALiRT IOIS DynamoDB
```
greglucas: I feel like there are a lot of options in these stacks that we probably don't need and can expand later. Our init functions are unnecessarily long if things are really just hard-coded and not switched via options somewhere else.

Right now this is creating stacks of stacks, but would it make sense to just create a single I-ALiRT Stack here with the specific resources made within this class?

```python
class IalirtProcessing(Stack):
    def __init__(...)
        self.add_compute_resources()
        self.add_dynamodb_table()

    def add_compute_resources(self, ...):
        """Add EC2 components"""

    def add_dynamodb_table(self, ...):
        """Add DynamoDB Table here"""
```

tests/infrastructure/test_ialirt_processing_stack.py
greglucas, line 1: This will need to be pytest.skip()'d if it won't run in a CI environment.

tests/test-data/ialirt_ec2/test_app.py
greglucas, line 1: This will need to be avoided in the test collection. Something like this I think.

```bash
[tool.pytest.ini_options]
ignore = ["tests/test-data"]
```

greglucas (commented):
Sorry, I had this review done earlier and must never have hit submit :(

I think it looks good, my broader architecture comment is to move the resources into a single stack and start hard-coding some of the values since they won't really be changed, or if they are we can just change that explicit line rather than through a keyword argument in some other place in the code.

sds_data_manager/stacks/ialirt_processing_stack.py
laspsandoval: I think what I would prefer to do is make the dynamodb resource into a construct since we will have multiple DynamoDB tables. I also think that keeping the compute resource as a stack improves modularization.

greglucas: > I think what I would prefer to do is make the dynamodb resource into a construct

I'm not sure I'm completely following this. That is what AWS gives us already (an L2 construct). My main concern is we don't need to write a wrapper around AWS's wrapper unless it greatly simplifies things because that will make it harder to follow what is set where rather than going to AWS' documentation. So if you need multiple tables, just call the `dynamo_db.Table()` constructor multiple times in the various stacks where you want them.

> I also think that keeping the compute resource as a stack improves modularization.

Maybe listing out what resources you are planning in each of the stacks would help me follow. But, I-ALiRT is only a few resources total, which is why I wasn't sure we would need modularization (is this going to be re-used anywhere else, is it nice to deploy standalone, etc.) I was thinking this would be nice to just `cdk deploy i-alirt` rather than `cdk deploy i-alirt-compute i-alirt-storage i-alirt-networking ...`

laspsandoval: The resources expected is something like this (although I'm sure this list will change some). I would rather start with really modular code and move towards one Stack if we think that makes sense than the opposite.
- ECS w/ autoscaling (maybe some type of load balancing...I'm not sure)
- ECR
- s3 bucket for mounting to container
**- Eventbridge for populating DynamoDB for IOIS binary blob 
- lambda for populating IOIS DynamoDB 
- DynamoDB for IOIS**
- lambda for parsing packet data
- DynamoDB for parsed packet data (table for each instrument?)
- Eventbridge for kicking off next lambda once previous lambda is complete
- lambda: check for complete dataset before applying algorithms; apply algorithms
- DynamoDB for final data product (table for each instrument?)
- s3 bucket for final data product

**bolded = maybe not needed if we are archiving CCSDS packets in s3

If we deploy IalirtProcessing and IalirtProcessing uses other stacks then those stacks will automatically be deployed.

greglucas: > I would rather start with really modular code and move towards one Stack if we think that makes sense than the opposite.

It would be good to get other eyes/thoughts on this, because I would go the other way. Start out with something contained and then expand later. It is much easier to grow scope than to shrink IMO.

The one I'm really thinking about is the DynamoDB table Stack. That just seems like you should hard-code it right into here. Currently, I have to go over to this project's DynamoDB constructor to figure out what everything means, and that just calls the aws_dynamodb table constructor, so there is an indirection and the options aren't used yet. I'd rather go to AWS' documentation and figure out what the defaults are.

I think that entire Stack class could be put inline here in ~15 lines of code.
```python
ialirt_table = dynamodb.Table(
            self,
            "DynamoDB-ialirt",
            table_name="ialirt-packets",
            partition_key=dynamodb.Attribute(
                name="ingest-ugps", type=dynamodb.AttributeType.STRING
            ),
            sort_key=dynamodb.Attribute(
                name="sct-vtcw", type=dynamodb.AttributeType.STRING
            ),
            # on-demand
            billing_mode=dynamodb.BillingMode.PAY_PER_REQUEST,
            point_in_time_recovery=True,
        )
```

Then, if later you want to add in the read/write units thats OK and should be simple to add, but we aren't using those now so it adds confusion IMO.

pyproject.toml
```diff
@@ -87,6 +87,7 @@ filterwarnings = [
     "ignore::DeprecationWarning:importlib*",
     "ignore::DeprecationWarning:jsii*",
 ]
+ignore = ["tests/test-data"]
```
greglucas: It is still flagging the other files. Maybe easiest is to move these to an `examples/ialirt` directory, because it isn't really run in our test suite? That way pytest wouldn't pick them up

You could also add `norecursedirs` or `addopts=--ignore` flags instead.

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -0,0 +1,63 @@
...
+        # ingest-ugps: ingestion ugps - 64 bit
+        # sct-vtcw: spacecraft time ugps - 64 bit
+        # src-seq-ctr: increments with each packet (included in filename?)
+        # ccsds-filename: filename of the packet
+        ialirt_dynamodb_stack.DynamoDB(
+            scope,
+            construct_id="IalirtDynamoDB",
+            table_name="ialirt-iois",
+            partition_key="ingest-ugps",
+            sort_key="sct-vtcw",
```
greglucas: I think we should make these more descriptive names for us. The iois team isn't going to use this table, you'll be the one interacting with it.

```suggestion
            # Are we storing individual packets in this table, or the binary file delivered every minute?
            table_name="ialirt-packets",
            partition_key="ingest-time",
            sort_key="spacecraft-time",
```

bmcclellan-cu: Is the plan to have all queries be ingest-time = X and spacecraft_time between Y and Z? Is there ever a need to query by a range of ingest-time values?

bmcclellan-cu (commented):
I don't have any major concerns with the code. I question the primary key for the DynamoDB table, have we identified how the table will be queried and by who/what? If so, what do those queries look like?

 The use of constructs over stacks seems to be more of a "preference", I can certainly see both points: constructs == re-usable code, stacks == simplified code. 

I can see Laura's point of if a resource is used multiple times, then it could be beneficial to define once. I can also see Greg's point, in that if we are only using those resources a few times with limited input options, then the AWS L2 may serve the purpose.

I can say we are going the construct route on Libera ITDC so the patterns developed of using multiple services to accomplish a task, can be used for other projects. I've never used constructs the way they are defined in this IOIS code, but it may technically be the best "Computer Science"/"Software Engineering" approach. :)

sds_data_manager/constructs/ialirt_compute_resources.py
```diff
@@ -0,0 +1,102 @@
...
+This module provides the IalirtEC2Resources class which sets up I-Alirt
+compute resources utilizing EC2 as the compute environment (for now).
+"""
+from aws_cdk import aws_ec2 as ec2
+from aws_cdk import aws_ecr as ecr
+from aws_cdk import aws_iam as iam
+from constructs import Construct
+
+
+class IalirtEC2Resources(Construct):
```
tech3371: I was in favor of construct before but in this case, I feel like this could be turned into Stack like others since it only creates EC2 stuffs. construct is helpful if things are very custom.

sds_data_manager/stacks/ialirt_processing_stack.py
tech3371: I like this breakdown. It looks cleaner. I wonder if we should do same in `stackbuilder.py` it's getting crowded.
```
class IalirtProcessing(Stack):
    def __init__(...)
        self.add_compute_resources()
        self.add_dynamodb_table()

    def add_compute_resources(self, ...):
        """Add EC2 components"""

    def add_dynamodb_table(self, ...):
        """Add DynamoDB Table here"""
```

tech3371:
It looks great though!

sds_data_manager/stacks/ialirt_processing_stack.py
laspsandoval: That's the plan. Data will be ingested every minute (ingest time) and spacecraft_time will be the actual timestamp of the packets. I do not anticipate a need to query the ingest-time values. They are only to be used for populating the database.

bmcclellan-cu: Ok, just to clarify, will the ingest_time be known and provided in every query?

With Dynamo we cannot query the sort_key (spacecraft_time) without specifying the partition_key (ingest_time).

laspsandoval: Yes. The ingest_time will be known and provided in every query. Thank you for the heads up.

greglucas: > Yes. The ingest_time will be known and provided in every query.

Maybe this is a naming thing, but what is `ingest_time`? I'm not sure I'd want to keep track of that to make a query. My use-case would be the following query: "Give me all packets between 2023-01-01T00:00 and 2023-01-02T00:00". I don't necessarily know when those packets were ingested I don't think.

laspsandoval: I think we might be getting ahead of ourselves with the database conversation. As I mentioned in the Overview of this PR,  I expect the setup of the database to change so try not to focus too much on that part of the code. The reason I said that is that I want to reevaluate whether or not we even want to put metadata from the CCSDS packets into the database if we are already storing them in S3. It might make sense to decom them first and then put the decom'd information in the database instead. So it is still debatable whether or not we even need it for ingest. But if we do use it for ingest then we will need a primary key and sort key of which the combination is unique. Since OIS is delivering the packets on a minute cadence to me it makes sense to have a primary key that is ingest time. But I am not opposed to changing that.

laspsandoval:
pre-commit.ci autofix

laspsandoval:
For the sake of progress I organized the code as suggested by Tenzin and Greg.

greglucas (approved):
I think this is really good Laura!

> For the sake of progress I organized the code as suggested by Tenzin and Greg.

I hope this isn't discouraging/a resignation of your ideas to what we've suggested. If you feel strongly about going in the other organization style that is totally fine since neither of us were blocking because the content you have here is really good! Feel free to ping me if you want to chat about any of the organization and we can do it offline as well if that'd be better.

PR #209, opened 2023-12-20
Metadata Schema

### LINKED ISSUE #185: Create database schema

sdhoyt:
# Change Summary

## Overview
This PR adds a common file metadata schema that is inherited by all the instrument tables. 

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
   - adds common metadata schema for instrument tables to inherit
   - add instrument tables 

## Testing
deployed and confirmed that create_schema lambda successfully creates the tables and columns.


sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -27,3 +22,84 @@ class UniversalSpinTable(Base):
...
+
+
+class MetadataTable:
+    """Common Metadata table"""
+
+    id = Column(Integer, primary_key=True)
+    file_name = Column(String, nullable=False)
+    # TODO: Do we need instrument since these will be split up into their
+    # separate tables?
+    instrument = Column(String, nullable=False)
```
bourque: I could see the benefit of including `instrument` if we ever want to run a query that joins the results of two or more instrument tables.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -27,3 +22,84 @@ class UniversalSpinTable(Base):
     # TODO: create table for repointing and then make
     # a foreign key to universal_spin_table
     repointing_number = Column(Integer, nullable=False)
+
+
+class MetadataTable:
+    """Common Metadata table"""
+
+    id = Column(Integer, primary_key=True)
+    file_name = Column(String, nullable=False)
```
bourque: Do we have any restrictions on how long the filename can be? If so, you could include that number in the definition here (e.g. `String(100)`)

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -27,3 +22,84 @@ class UniversalSpinTable(Base):
...
+
+
+# TODO: Follow-up PR should add in columns for each instrument
+# for instrument dependency IDs
+
+
+class LoMetadataTable(MetadataTable, Base):
+    """IMAP-Lo Metadata Table"""
+
+    __tablename__ = "lo_metadata_table"
```
bourque: Do we expect any other instrument-specific tables other than metadata tables? If not, I wonder if we should just name the tables `lo`, `hi`, `ultra`, `mag`, etc. It would slightly simplify the queries.

bourque (approved):
Looks good to me! I just have a few questions that don't necessarily need to be figured out in this PR.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -27,3 +22,84 @@ class UniversalSpinTable(Base):
...
+    # TODO: Do we need instrument since these will be split up into their
+    # separate tables?
+    instrument = Column(String, nullable=False)
+    data_level = Column(String, nullable=False)
+    descriptor = Column(String, nullable=False)
+    start_date = Column(DateTime(timezone=True), nullable=False)
+    end_date = Column(DateTime(timezone=True), nullable=False)
+    ingestion_date = Column(DateTime(timezone=True), nullable=False)
+    version = Column(Integer, nullable=False)
+    format = Column(String, nullable=False)
```
tech3371: what is format? is it file extension? If so, can we change it to `file_extension` or something like that?

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: We could so similar thing for other string as well. Like cap them at certain character number since we know they won't be greater than some length.

tech3371 (approved):
Looks great! Had couple very minor comments.

sds_data_manager/lambda_code/SDSCode/database/models.py
sdhoyt: I don't think so since we're going to be putting the dependency info in these tables as well, so it would make more sense to not have metadata in the name. I'll make that update.

sds_data_manager/lambda_code/SDSCode/database/models.py
sdhoyt: yeah, I called it format because that's what in the style guide for naming conventions `imap_<instrument>_<datalevel>_<descriptor>_<startdate>_<enddate>_<version>.<format>`. I'll change it to `extension` to be consistent with the FilenameParser class?

bourque: I was torn on using `format` vs `extension` too. I'm happy to change the filename docs if we prefer `extension`

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -27,3 +22,84 @@ class UniversalSpinTable(Base):
...
+
+
+# TODO: Follow-up PR should add in columns for each instrument
+# for instrument dependency IDs
```
laspsandoval: This TODO answers my questions.

Plus add:     spice_id INTEGER,
    parent_<instrument>_id INTEGER,
    pointing_id INTEGER

laspsandoval (approved):
Looks good.

maxinelasp (approved):
Looks like a good start!

maxinelasp:
Only note would be that "metadata" isn't my favorite name for these tables - could we instead use something like "catalog" or "file catalog?"

sdhoyt:
> Only note would be that "metadata" isn't my favorite name for these tables - could we instead use something like "catalog" or "file catalog?"

I agree with this. I think that's a better description of what this is.

PR #210, opened 2023-12-21
Metadata Ingest

### LINKED ISSUE #192: Setup RDS in indexer lambda

sdhoyt:
# Change Summary

## Overview
This PR adds metadata ingest to the indexer lambda. I still need to write the indexer lambda unit tests, but I wanted to put this out to keep things moving.

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
- indexer.py
   - added code to add file metadata to the database for each instrument
- models.py
   - updated column types in the common file catalog table
      - it looks like we're using a string format for version (ie v1-1) or should this be converted to a decimal after it's read in?
      - updated id field to automatically increment when new data is added.
   
## Testing
uploaded files to S3 and confirmed that they were properly indexed into the database.

TODO: Still need to add unit tests for indexer. Currently trying to figure out how to mock sqlalchemy.


sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -47,17 +51,52 @@ def lambda_handler(event, context):
...
-        # TODO: remove this check since upload api validates filename?
-        # Found nothing. This should probably send out an error notification
-        # to the team, because how did it make its way onto the SDS?
-        if metadata is None:
-            logger.info("Found no matching file types to index this file against.")
-            return None
+        # confirm that the file is valid
+        if filepath["statusCode"] == 400:
+            logger.error(filepath["body"])
+            break
```
```suggestion
        if filepath["statusCode"] != 200:
            logger.error(filepath["body"])
            break
```
Do you want to raise here instead of breaking? Or return the bad statusCode and message?

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -47,17 +51,52 @@ def lambda_handler(event, context):
...
+        # setup a dictionary of metadata parameters to unpack in the
+        # instrument table
+        metadata_params = {
+            "file_path": filepath["body"],
+            "instrument": filename_parsed.instrument,
+            "data_level": filename_parsed.data_level,
+            "descriptor": filename_parsed.descriptor,
+            "start_date": filename_parsed.startdate,
+            "end_date": filename_parsed.enddate,
+            "ingestion_date": datetime.datetime.now(),
```
```suggestion
            "ingestion_date": datetime.datetime.now(datetime.datetime.UTC),
```

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -47,17 +51,52 @@ def lambda_handler(event, context):
...
+            "version": filename_parsed.version,
+            "extension": filename_parsed.extension,
+        }
+
+        # Add data to the corresponding instrument database
+        with Session(engine) as session:
+            if filename_parsed.instrument == "lo":
+                data = models.LoTable(**metadata_params)
```
greglucas: I'd write this as a lookup table outside of the session if you're not relying on the session for this portion.

```python
model_lookup = {"lo": models.LoTable, "hi": models.HiTable}
data = model_lookup[filename_parsed.instrument](**metadata_params)
with Session(engine) as session:
    session.add(data)
    session.commit()
```

greglucas (approved):
Looks good to me, just a few minor suggestions you can take or leave.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: nice!

tech3371 (approved):

PR #212, opened 2024-01-03
update to batch env

laspsandoval:
# Change Summary

## Overview
Update to batch environment to make it Docker container mount to EFS.

## New Files
None

## Deleted Files
None

## Updated Files
- batch_compute_resources.py
   - Now grabs latest image
   - Setup an environment variable so that we can use the imap_processing run_processing.py script both locally (for development) and also in the batch job
- lambda_function.py
   - I added a TODO here regarding which spice kernels we should be pulling. 

## Testing
Testing described in PR for imap_processing: https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/302#issue-2064537491

### DIFF
--- a/sds_data_manager/constructs/batch_compute_resources.py
+++ b/sds_data_manager/constructs/batch_compute_resources.py
@@ -170,7 +170,7 @@ class FargateBatchResources(Construct):
             container_properties={
                 "image": f"{repo.repository_uri}:latest",
                 "environment": [
-                    {"name": "SPICE_DIRECTORY", "value": "/mnt/spice"},
+                    {"name": "IMAP_SPICE_DIR", "value": "/mnt/spice"},
                 ],
                 "mountPoints": [
                     {
--- a/sds_data_manager/constructs/batch_compute_resources.py
+++ b/sds_data_manager/constructs/batch_compute_resources.py
@@ -168,7 +168,10 @@ class FargateBatchResources(Construct):
             type="CONTAINER",
             platform_capabilities=["FARGATE"],
             container_properties={
-                "image": f"{repo.repository_uri}:{account_name}",
+                "image": f"{repo.repository_uri}:latest",
+                "environment": [
+                    {"name": "SPICE_DIRECTORY", "value": "/mnt/spice"},
+                ],
                 "mountPoints": [
                     {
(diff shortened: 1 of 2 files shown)


sds_data_manager/constructs/batch_compute_resources.py
```diff
@@ -168,7 +168,10 @@ def __init__(
             type="CONTAINER",
             platform_capabilities=["FARGATE"],
             container_properties={
-                "image": f"{repo.repository_uri}:{account_name}",
+                "image": f"{repo.repository_uri}:latest",
+                "environment": [
+                    {"name": "SPICE_DIRECTORY", "value": "/mnt/spice"},
```
greglucas: Set this with IMAP as a prefix which is sometimes nice to grep through environment variables during processing.
```suggestion
                    {"name": "IMAP_SPICE_DIR", "value": "/mnt/spice"},
```

sds_data_manager/lambda_code/efs_lambda/lambda_function.py
```diff
@@ -55,12 +55,21 @@ def write_data_to_efs(s3_key: str, s3_bucket: str):
 
     logging.debug("After downloading file: %s", os.listdir(mount_path))
 
-    # Attitude naming convention is this:
+    # TODO: we only want historical attitude kernels delivery
```
greglucas, line 58: Do you want to add an `and not filename.contains("pred")` check below?

Or possibly we would want to have 3 separate symlinks for users to choose between rather than having a single symlink to be used? `reconstructed`, `historical`, `predict`.

sds_data_manager/constructs/batch_compute_resources.py
```diff
@@ -168,7 +168,10 @@ def __init__(
             type="CONTAINER",
             platform_capabilities=["FARGATE"],
             container_properties={
-                "image": f"{repo.repository_uri}:{account_name}",
+                "image": f"{repo.repository_uri}:latest",
```
tech3371, line 171: I would be careful of using `latest` but that's a discussion for another time. Versioning tracking is complicated :)

tech3371 (approved):

PR #214, opened 2024-01-10
Remove Step Functions

laspsandoval:
# Change Summary

## Overview
Removes step function and instead submits batch job using batch_starter.py.

## New Files
None

## Deleted Files
sdc_step_function.py

## Updated Files
instrument_lambdas.py - passes environment variables for batch job instead of step function
batch_starter.py - submits batch job instead of step function
processing_stack.py - removal of step function stack


sds_data_manager/stacks/processing_stack.py
```diff
@@ -87,30 +86,21 @@ def __init__(
             account_name=account_name,
```
tech3371, line 86: I have been wondering if we can delete this whole file and start fresh? meaning de-couple how we create resources. We would still same thing as what we do in this file. But we create batch job, lambda and event rule separately. Create resources by logics of what goes together.

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -441,21 +443,17 @@ def lambda_handler(event: dict, context):
...
-                f"Started Step Function for {instrument_name} with response: {response}"
-            )
+            for data_level in grouped_list[instrument_name]:
+                batch_client.submit_job(
+                    jobName=f"imap_{instrument_name}_{data_level}_<descriptor>_<startdate>_<enddate>_<vxx-xx>.cdf",
+                    jobQueue=job_queue,
+                    jobDefinition=job_definition,
+                    containerOverrides={
+                        "command": [
+                            f"imap_{instrument_name}_{data_level}_<descriptor>_<startdate>_<enddate>_<vxx-xx>.cdf"
```
tech3371, line 456: This is not a comment for this PR but food for thought for later. I was envisioning something like this for future command(of course it's subject to change but rough of what it may look like). 
```
           "command": [
                "python",
                "imap_cli --instrument glows --level l1a --filename 's3://sds-data-449431850278/imap/glows/l0/2023/07/imap_glows_l0_1_20230724_20230724_v02-01.pkts' --dependency {.....}"
            ],
```

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -62,12 +62,24 @@ def __init__(
 
         super().__init__(scope, construct_id)
 
+        # Batch Job Inputs
+        stack = Stack.of(self)
+        job_definition_arn = (
+            f"arn:aws:batch:{stack.region}:{stack.account}:job-definition/"
+            f"{batch_resources.job_definition_name}"
+        )
```
greglucas: I think these could be referenced through the objects `batch_resources.job_definition.job_definition_arn` rather than constructing the ARNs yourself.
https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_events_targets/BatchJob.html

sds_data_manager/lambda_code/batch_starter.py
greglucas: Agree with you @tech3371, minus the `python` portion which we won't want if we have the cli installed properly, and that one long string has to be separate entries in the command I think.

For now though, I don't know if this command will work or not since it is just a filename. We might want `python run-my-script.py filename.cdf` so you can update the script without needing to update the CDK command.

greglucas (approved):

sds_data_manager/stacks/processing_stack.py
laspsandoval: Yes. Especially with event rule. Let's create another ticket for organization.

sds_data_manager/lambda_code/batch_starter.py
laspsandoval: Yep. As long as it fits into the batch job command limits. I'll work on batch_starter.py more in the next PR (hence the todo) and figure out what we can actually pass.

tech3371: yes. Thank you!

tech3371 (approved):

PR #215, opened 2024-01-12
Reorganize sds manager stack

tech3371:
# Change Summary

## Overview
moved resources into other files. No changes made to resources logics. Mainly moved resources into logical groups and update their tests

## New Files
- sds_data_manager/stacks/data_bucket_stack.py
- sds_data_manager/stacks/sds_api_manager_stack.py

## Deleted Files

## Updated Files

## Testing
- tests/infrastructure/test_data_bucket.py
- tests/infrastructure/test_indexer_lambda_stack.py
- tests/infrastructure/test_sds_api_manager_stack.py


sds_data_manager/stacks/indexer_lambda_stack.py
```diff
@@ -0,0 +1,78 @@
...
+                "SECRET_NAME": db_secret_name,
+            },
+        )
+
+        indexer_lambda.apply_removal_policy(cdk.RemovalPolicy.DESTROY)
+
+        rds_secret = secrets.Secret.from_secret_name_v2(
+            self, "rds_secret", db_secret_name
+        )
+        rds_secret.grant_read(grantee=indexer_lambda)
```
greglucas, line 78: Not for this PR, but this is a fairly common pattern, I'm curious if we should add this as a helper function somewhere like on the RDS Stack so we only have the secret in one location.

```

class RDSStack:
    def grant_rds_read(self, construct):
        # give construct read access to the rds secret key
        self.rds_secret.grant_read(indexer_lambda)
```

sds_data_manager/stacks/sds_api_manager_stack.py
greglucas, line 1: Thought on renaming to just `api_stack.py`? I personally don't really like the nested `sds_*`, or the `*_manager` names.

greglucas (approved):
Looks good to break this out into smaller sections. Minor comment on naming to consider, but looks good as-is to me.

sds_data_manager/stacks/indexer_lambda_stack.py
tech3371: That would be good thing to investigate for sure. Let's create ticket for that.

sds_data_manager/stacks/sds_api_manager_stack.py
tech3371: I will merge for now and then change that one thing in upcoming PR.

sds_data_manager/stacks/indexer_lambda_stack.py
tech3371: just created ticket for this.

PR #216, opened 2024-01-12
Replace processing step stack with Batch job stack

tech3371:
# Change Summary
Other changes are from this PR: #215 . This PR only has few files changes

## Overview
processing_stack.ProcessingStep was giving this error:
```
        AttributeError: 'CfnJobDefinition' object has no attribute
        'attr_job_definition_arn'
```
Took out batch job stack into its own stack 

## New Files
- sds_data_manager/stacks/batch_job_stack.py

## Deleted Files
- deleted file 1
   - sds_data_manager/stacks/processing_stack.py

## Updated Files


sds_data_manager/utils/stackbuilder.py
```diff
@@ -145,21 +138,22 @@ def build_sds(
...
-        processing_stack.ProcessingStep(
-            scope,
-            f"{instrument}Processing",
-            env=env,
+        # NOTE: processing_stack.ProcessingStep was giving this
+        # error:
+        # AttributeError: 'CfnJobDefinition' object has no attribute
+        # 'attr_job_definition_arn'
+        # Replaced it with batch job stack
+        batch_job_stack.BatchJobStack(
```
greglucas: Can you just call `FargateBatchResources()` directly here instead of creating a new stack? This new stack is just passing things through immediately.

tech3371: I don't know if I can since that is a construct but let me try. That would simplify it.

greglucas: Turn that construct into a Stack then? Not sure it needs to be one versus the other.

sds_data_manager/stacks/batch_compute_resources.py
```diff
@@ -32,9 +33,11 @@ def __init__(
         db_secret_name: str,
         efs_instance: EFSStack,
         account_name: str,
+        env: cdk.Environment,
```
greglucas: Minor nit, but I'd remove this and let the **kwargs handle it.

```suggestion
```

sds_data_manager/stacks/batch_compute_resources.py
```diff
@@ -69,7 +72,7 @@ def __init__(
             This value can be overwritten by command line input and can
             be accessed from the cdk.json file.
         """
-        super().__init__(scope, construct_id)
+        super().__init__(scope, construct_id, env=env, **kwargs)
```
greglucas:
```suggestion
        super().__init__(scope, construct_id, **kwargs)
```

sds_data_manager/utils/stackbuilder.py
```diff
@@ -145,21 +138,23 @@ def build_sds(
         #     ...
         # )
 
-        processing_stack.ProcessingStep(
+        # NOTE: processing_stack.ProcessingStep was giving this
+        # error:
+        # AttributeError: 'CfnJobDefinition' object has no attribute
+        # 'attr_job_definition_arn'
+        # Replaced it with batch job stack
```
greglucas: I'm not sure what this comment means in the context of the code. This seems more like a comment for the PR itself.

```suggestion
```

greglucas (approved):
I like this approach of making that a Stack and using it directly. Seems to simplify the code. Minor nits that you can ignore if you want.

PR #217, opened 2024-01-12
Indexer events

tech3371:
# Change Summary
Other changes are from this PR: #215 . This PR only has few files changes

## Overview
Added events to the indexer lambda

## Updated Files

### DIFF
--- a/sds_data_manager/stacks/indexer_lambda_stack.py
+++ b/sds_data_manager/stacks/indexer_lambda_stack.py
@@ -9,6 +9,10 @@ from aws_cdk import (
 from aws_cdk import (
     aws_ec2 as ec2,
 )
+from aws_cdk import aws_events as events
+from aws_cdk import (
+    aws_events_targets as targets,
+)
 from aws_cdk import (
     aws_lambda as lambda_,
 )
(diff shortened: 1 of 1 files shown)


sds_data_manager/stacks/indexer_lambda_stack.py
```diff
@@ -76,3 +80,92 @@ def __init__(
...
+            "l0DataArrival",
+            rule_name="l0-data-arrival",
+            event_pattern=events.EventPattern(
+                source=["aws.s3"],
+                detail_type=["Object Created"],
+                detail={
+                    "bucket": {"name": [data_bucket.bucket_name]},
+                    "object": {
+                        "key": [
+                            {"prefix": f"imap/{instrument}/l0/"}
```
greglucas, line 116: Not for this PR, but just a note for discussion in our next tag-up.

I was just chatting with @sdhoyt this morning about whether we have `/<mission>/` as an entry in the DB to search by anymore. I know we talked in the past about getting I&T data and running that through our pipeline. I'm wondering if we should add a new qualifier to our bucket to put things in a different base "mission" area, but still run everything the same if we can.

sds_data_manager/stacks/indexer_lambda_stack.py
```diff
@@ -76,3 +80,92 @@ def __init__(
...
+            self,
+            "batchStarterEvent",
+            rule_name="batch-starter-event",
+            event_pattern=events.EventPattern(
+                source=["imap.lambda"],
+                detail_type=["Batch Job Started"],
+                detail={
+                    "filename": [{"exists": True}],
+                    "status": ["InProgress"],
+                    "depedency": [{"exists": True}],
```
greglucas: Put something as a placeholder for now?

Also, `filename` is a bit unclear to me which file we are talking about. I think it is the file you're creating? Could you add some naming to make that clear in the variable we are passing around?

```suggestion
                    "file_to_create": "imap_mag_l1a_...",
                    "status": ["InProgress"],
                    "depedency": [{"mag": "imap_mag_l0_raw_..."}],
```

Now that I'm thinking looking at this more closely, is it actually saying only trigger the event if the detail has the fields "filename", "status", and "dependency" (the `"exists": True`)? If so, maybe we don't need all of these for the event rule, but just need the detail_type for the match. Ignore my suggestion if this is the case.

tech3371: Yes, latter. It only triggers target only when those field exits. We don't do checks there. We will do the checks in indexer lambda or any target we specify. But I do agree with better naming. I get confused too. I like `file_to_create` name. I will change that.

greglucas (approved):

PR #219, opened 2024-01-16
Api dockerfile for testing

laspsandoval:
# Change Summary

## Overview
Command line program that downloads a file from the s3_uri, modifies it, and uploads the modified file to the same s3 uri. 

## New Files
- imap_api.py
   - Command line program that downloads a file from the s3_uri, modifies it, and uploads the modified file to the same s3 uri.
- Dockerfile.testing
   - Creates image that will be used for testing batch_starter.py.

## Testing
Since the script is only used for testing I did not perform tests on imap_api.py apart from checking that a new file with expected naming appeared in the s3 bucket (from the s3_uri). 

Example: 
docker build -f Dockerfile.testing -t test-image . --no-cache
docker run test-image s3://sds-data-449431850278/imap/hit/l0/2024/01/imap_hit_l0_sci_20240101_20240104_v02-01.pkts


tests/docker_images/Dockerfile.testing
```diff
@@ -0,0 +1,23 @@
+# Dockerfile that installs imap_processing and its dependencies
+FROM public.ecr.aws/docker/library/python:3.10-slim
+USER root
+
+# Upgrade pip to the latest version
+RUN pip install --upgrade pip
+
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y git
```
greglucas:
```suggestion
```
I don't think we need to update any of the OS because you aren't using git below.

tests/docker_images/imap_api.py
```diff
@@ -0,0 +1,128 @@
...
+    logging.info(f"Starting download from S3 URI: {s3_uri}")
+
+    url_with_parameters = f"{api_endpoint}/download?{s3_uri}"
+    response = requests.get(url_with_parameters)
+
+    # Check if running in Docker
+    running_in_docker = os.getenv("DOCKER_ENV") is not None
+
+    # Set the base directory
+    base_directory = "/mnt/data" if running_in_docker else "."
```
greglucas: So you don't have to set anything in your container, would it be easier to just see if that mountpoint exists, and if not use the current directory?
```suggestion
    # Set the base directory
    base_directory = Path("/mnt/data")
    if not base_directory.exists():
        base_directory = Path.cwd()
```

I also think your other commands will be easier below if you switch over to pathlib as well.

```python
# Make parent path
file_name_and_path.parent.mkdir(parents=True, exist_ok=True)
```

tests/docker_images/imap_api.py
```diff
@@ -0,0 +1,128 @@
...
+def main():
+    """Main function for the IMAP API utilities."""
+    args = _parse_args()
+
+    if args.api_endpoint is not None:
+        file_name_and_path = download(args.s3_uri, args.api_endpoint)
+        upload(file_name_and_path, args.api_endpoint)
+    else:
+        file_name_and_path = download(args.s3_uri)
+        upload(file_name_and_path)
```
greglucas: Set the default here instead of relying on the method calls so you don't need the if/else block for different argument lengths.
```suggestion
    endpoint = args.api_endpoint if args.api_endpoint is not None else "https://api.dev.imap-mission.com"
    file_name_and_path = download(args.s3_uri, endpoint)
    upload(file_name_and_path, args.endpoint)
```

tests/docker_images/__init__.py
greglucas, line 1: I don't think you need this file since it isn't a part of the package technically and it is just a standalone script.

greglucas (approved):
Looks good to me. Just minor suggestions that don't change any of the content.

tech3371 (approved):
nice work!

PR #220, opened 2024-01-17
InstrumentLambda from Construct to Stack

laspsandoval:
# Change Summary

## Overview
InstrumentLambda was a construct in the former Processing Stack. When the Processing Stack was removed, InstrumentLambda was no longer able to be deployed. This PR changes InstrumentLambda into a Stack that is now called on in stackbuilder.py.

## New Dependencies
None

## Deleted Files
- /constructs
   - There are no more constructs.

## Updated Files
- instrument_lambdas.py
   - Change from Construct to Stack
- stackbuilder.py
   - Added InstrumentLambda Stack

## Testing
- Made certain deployment went as expected.


greglucas (approved):
Is InstrumentLambda the Kickoff / batch_starter lambda or is it different? It'd help to put an extra verb in the name like "InstrumentStarterLambda" or "InstrumentKickoffLambda" to distinguish where this goes in the process.

Could you also change the instrument_list above to all lower case, maybe `swe` for now since I also think codice may be a single instrument and not have hi broken out. I think that will help with our directory data storage during testing.

sds_data_manager/stacks/instrument_lambdas.py
```diff
@@ -60,7 +61,7 @@ def __init__(
             VPC into which to put the resources that require networking.
         """
 
-        super().__init__(scope, construct_id)
+        super().__init__(scope, construct_id, **kwargs)
 
         # Batch Job Inputs
         job_definition_arn = batch_resources.job_definition.attr_job_definition_arn
```
tech3371: Did this line give you error when you deploy? It was giving me error before and I couldn't figure out why it was.

sds_data_manager/stacks/instrument_lambdas.py
```diff
@@ -14,7 +14,7 @@
 from sds_data_manager.stacks.database_stack import SdpDatabase
 
 
-class InstrumentLambda(Construct):
+class InstrumentLambda(Stack):
```
tech3371: May be we should name it `BatchStarterLambda` since we have been creating other lambda on its own. and carry that name through out this lambda. Right now, I think this one lambda will build dependency for all instrument based on current input. If we decide to break down for each instrument, we can then revisit too.

I am happy to jump on call real quick to discuss what I mean.

laspsandoval: Nice catch!

laspsandoval:
I will update batch_starter.py next.

PR #223, opened 2024-01-18
Update File Catalog and Add Status Table

### LINKED ISSUE #222: Update RDS tables to use single file catalog table and add status table
### Summary

Update the RDS tables to reflect the current design: https://lasp.colorado.edu/galaxy/display/IMAP/SDC+Database+Tables

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

sdhoyt:
# Change Summary

## Overview
This PR updates the File Catalog to contain all instruments rather than having separate tables for each instrument and adds a table for the file statuses.

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
- models.py
   - removed instrument specific tables
   - added status table
- indexer.py
   - updated how the data is added to RDS to reflect changes made to the tables (single table vs instrument tables)
   - fixed small bug where file path was being given to file parser instead of filename when using the API for file uploads.
   - Added TODO to remove Records loop from indexer because it looks like events no longer have a record list. Tenzin already has a PR coming up to change that so I left it as is for now

## Testing
Uploaded files with the python API wrapper and confirmed they were properly indexed into the file catalog

closes #222


sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -24,82 +73,31 @@ class UniversalSpinTable(Base):
     repointing_number = Column(Integer, nullable=False)
 
 
-class FileCatalogTable:
-    """Common file catalog table"""
+class StatusTrackingTable(Base):
+    """Status tracking table"""
+
+    __tablename__ = "status_tracking_table"
```
bourque:
```suggestion
class StatusTracking(Base):
    """Status tracking table"""

    __tablename__ = "status_tracking"
```

I'm not sure we need the word "table" in these names. In my mind, it can be assumed we are talking about tables since this is for a database.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -24,82 +73,31 @@ class UniversalSpinTable(Base):
...
+    file_to_create_path = Column(String, nullable=False)
+    status = Column(statuses, nullable=False)
+    job_definition = Column(String, nullable=False)
+    ingestion_date = Column(DateTime(timezone=True), nullable=True)
+
+
+class FileCatalogTable(Base):
+    """File catalog table"""
+
+    __tablename__ = "file_catalog_table"
```
bourque:
```suggestion
class FileCatalog(Base):
    """File catalog table"""

    __tablename__ = "file_catalog"
```

bourque (approved):
Looks good to me! Thanks for updating this. I just have a nit-picky suggestion.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -1,7 +1,56 @@
...
+    "l3",
+    "l3a",
+    "l3b",
+    "l3c",
+    "l3d",
+    name="data_level",
+)
+
+# status enums for the status tracking table
+statuses = Enum("SUCCESS", "FAILURE", name="status")
```
tech3371:
```suggestion
statuses = Enum("INPROGRESS", "SUCCEEDED", "FAILED", name="status")
```

tech3371 (approved):
added comment about status. otherwise, it looks good.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -24,82 +73,31 @@ class UniversalSpinTable(Base):
...
 
     # TODO: determine cap for strings
     id = Column(Integer, Identity(start=1, increment=1), primary_key=True)
     file_path = Column(String, nullable=False)
-    instrument = Column(String(6), nullable=False)
-    data_level = Column(String(3), nullable=False)
+    instrument = Column(instruments, nullable=False)
+    data_level = Column(data_levels, nullable=False)
     descriptor = Column(String, nullable=False)
     start_date = Column(DateTime(timezone=True), nullable=False)
```
maxinelasp: Do we want timezones?

maxinelasp (approved):
Nice, lgtm!

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -1,7 +1,56 @@
...
+)
+
+# data level enums for the file catalog table
+data_levels = Enum(
+    "l0",
+    "l1a",
+    "l1b",
+    "l1c",
+    "l1ca",
+    "l1cb",
```
greglucas, line 39: where do the "a" and "b" qualifiers come from? The "pre" is a bit non-standard, but was sort of agreed upon to be included for MAG's sake. Did we have the same agreement for ca and cb?

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: I'd say we want to avoid timezones as much as possible and only use UTC. Does that still require a timezone=True here to clarify that this is a UTC start_date?

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -24,82 +73,31 @@ class UniversalSpinTable(Base):
...
+
+
+class FileCatalog(Base):
+    """File catalog table"""
+
+    __tablename__ = "file_catalog"
 
     # TODO: determine cap for strings
     id = Column(Integer, Identity(start=1, increment=1), primary_key=True)
     file_path = Column(String, nullable=False)
```
greglucas, line 95: Can be added in a later PR, but did we want to have the "mission" qualifier as well, so we could have `imap-it`, `imap-dev` things like that to distinguish between dev/test/production data?

sds_data_manager/lambda_code/SDSCode/database/models.py
sdhoyt: That's a good question. I think I made `timezone=True` because when I previously added ingestion time to file catalog (which just got moved to file status table) I specified UTC when getting the current time and it sounds like I need to set timezone to True for it to recognize that python date time with timezone input, but I might be reading the documentation wrong. I added timezone for all the dates in case we need to specify UTC for everything. I'm not sure if I'm interpreting how that works correctly though, so I'm not sure if we really need this.

sds_data_manager/lambda_code/SDSCode/database/models.py
sdhoyt: I wasn't there for that part of the discussion, but I got those from here:https://lasp.colorado.edu/galaxy/display/IMAP/SDC+Database+Tables in the DATA_LEVELS list

sds_data_manager/lambda_code/SDSCode/database/models.py
sdhoyt: Will we have that kind of data mixed together? I guess I was assuming we'd have separate accounts / deployments for each

sds_data_manager/lambda_code/SDSCode/database/models.py
maxinelasp: I'm worried timezone=True will cause more problems than always assuming the timezone is UTC and having timezone=False. I think you can specify timezone when converting from this datetime to python datetimes?

sds_data_manager/lambda_code/SDSCode/database/models.py
maxinelasp: I think we should agree to keep them separate, either in different accounts or potentially in different tables. Mixing dev and production data together seems like not a good idea

sds_data_manager/lambda_code/SDSCode/database/models.py
sdhoyt: Yeah, that's true. I'll remove timezone

sds_data_manager/lambda_code/SDSCode/database/models.py
bourque: And I got those from the diagrams here: https://lasp.colorado.edu/galaxy/display/IMAP/Instrument+Science+Data+Processing+Pipelines+Diagrams

greglucas: @tech3371 are these accurate with the a/b or did we remove those?

But, lets not let this slow us down, we can update the fields later if need be.

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: Sounds reasonable, I think the only issue we'd run into is with `ingest_time`, where we'll need to make sure that is handled the same in all of our Lambda's that write to the database with `datetime.now(datetime.UTC)`.

PR #224, opened 2024-01-19
Update to indexer lambda to make pytest work locally

### LINKED ISSUE #213: Mock Postgres database
Mock postgres database to test indexer data ingest

tech3371:
# Change Summary
This PR include changes to make pytest work for indexer lambda.

* Removed some database files to fix import issue and secret environment variable reference issue.
* Updated conftest.py in lambda_endpoints to create database engine and session for local testing.
* Added minimal test to test that it works.


sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -2,18 +2,35 @@
...
 from SDSCode.database.models import Base
+from sqlalchemy import create_engine
 
 # Logger setup
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
 logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
 
 
+def get_db_uri():
```
sdhoyt: What was the error related to config.py and database.py? These are being used in multiple places (soon including queries.py) so it'd be a lot better if we could import these

tech3371: yes. It was causing some issue with import. But I just re-try and have this in `database.py` might work. I will push that change.

pyproject.toml
```diff
@@ -55,6 +55,9 @@ pytest = "==6.2.5"
 moto = "^4.1.3"
 boto3 = "^1.26.78"
 pytest-postgresql = "==5.0.0"
+pytest-mock = "^3.12.0"
+SQLAlchemy = "<=2.0.23"
```
greglucas: Why do we need these pins? Do newer versions of SQLAlchemy break something for us?

sds_data_manager/lambda_code/SDSCode/create_schema.py
tech3371: I pushed that change. Looks like issue I was having at the beginning was resolved by db_session parameter now. So I moved `get_db_uri()` back into `database.py` and import in lambda(s).

sds_data_manager/lambda_code/SDSCode/database/database.py
```diff
@@ -1,5 +1,21 @@
...
+"""Create database URI that will be used to create engine or make query"""
 
-engine = create_engine(config.db_uri, echo=True)
+import json
+import os
+
+import boto3
+
+
+def get_db_uri():
```
greglucas: Could this be `get_engine()` instead?

So your new function would return `create_engine(name_you_current_return, echo=True)`
Then you would mock the `create_engine` function instead of mocking the `get_db_uri` function?

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -38,11 +37,25 @@ def lambda_handler(event, context):
         This object provides methods and properties that provide
         information about the invocation, function,
         and runtime environment.
+    db_session : sqlalchemy.orm.scoping.scoped_session, optional
+        When you use an in-memory SQLite database (sqlite:///:memory:)
+        during test, the database exists only for the lifetime of the
+        connection. Each new call to create_engine("sqlite:///:memory:")
+        creates a brand new, isolated database. This session resolves
+        that issue for testing locally.
```
greglucas: I don't think we want to be passing in `db_session`... That doesn't seem quite right to me.

If you want the engine to be persistent can you mock that return and use it continuously?
`engine = get_engine()` -> Which you've now mocked to return your test_engine. (Instead of mocking get_db_uri)

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -5,9 +5,10 @@
 import sys
 
 import requests
-from SDSCode.database.database import engine
 from SDSCode.database.models import Base
 
+from .database import database as db
```
greglucas: Can you make this consistent with the above import?
```suggestion
from SDSCode.database import database as db
```

greglucas: I also am not a huge fan of `from datetime import datetime` like imports, so in a follow-up PR maybe we could move this to `initialization.py` or something like that? Or even move all the models and initialization code into a single file if it isn't too much code.

sds_data_manager/lambda_code/SDSCode/database/database.py
```diff
@@ -1,5 +1,32 @@
...
+
+import json
+import os
+
+import boto3
 from sqlalchemy import create_engine
 
-engine = create_engine(config.db_uri, echo=True)
+
+def get_db_uri():
```
greglucas: Can this live in `get_engine()` now? Just put a comment saying you first need to get the DB uri and add the content from here down below in that function.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -23,11 +21,12 @@
 
 
 def lambda_handler(event, context):
-    """Handler function for creating metadata, adding it to the payload,
-    and sending it to the opensearch instance.
+    """Handler function for creating metadata, adding it to the
+    database.
```
```suggestion
    """Handler function for creating metadata, adding it to the database.

```
All one line for the header docstring.

greglucas (approved):
This looks good. Some minor cleanup comments, but that can be done later too.

sds_data_manager/lambda_code/SDSCode/create_schema.py
tech3371: sure thing since we are not testing this file yet.

sdhoyt (approved):

PR #228, opened 2024-01-22
Query RDS with API

### LINKED ISSUE #186: Setup RDS database API access

sdhoyt:
# Change Summary

## Overview
This PR adds the ability to use the query API to query RDS

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
- queries.py
   - added ability to send queries to RDS
- sds_api_manager_stack.py
   - gave permission for queries lambda to access RDS

## Testing
- Added a new unit test for queries.py lambda
- Tested using the sds_api.py wrapper


sds_data_manager/lambda_code/SDSCode/queries.py
```diff
@@ -28,14 +33,62 @@ def lambda_handler(event, context):
...
+    ]
+    # go through each query parameter to set up sqlalchemy query conditions
+    for param, value in query_params.items():
+        # confirm that the search parameter is valid
+        if param not in valid_parameters:
+            response = {
+                "statusCode": 400,
+                "body": json.dumps(
+                    f"{param} is not a valid search parameter. "
+                    + f"Valid search parameters are: {valid_parameters}"
```
greglucas: Do we want to say "query" instead of "search" here? I was wondering if we wanted to change the endpoint to `/search` actually :) But, I guess we are "querying" and not "searching"?

sds_data_manager/lambda_code/SDSCode/queries.py
```diff
@@ -28,14 +33,62 @@ def lambda_handler(event, context):
...
+            return response
+        # check if we're search for start_date or end date to
+        # setup the correct "where" time condition
+        if param == "start_date":
+            query = query.where(
+                models.FileCatalog.start_date
+                >= datetime.datetime.strptime(value, "%Y%m%d")
+            )
+        elif param == "end_date":
+            # TODO: Need to discuss as a team how to handle date queries. For now,
```
greglucas, line 72: I like this way personally, but agree we should discuss it.

sds_data_manager/lambda_code/SDSCode/queries.py
```diff
@@ -28,14 +33,62 @@ def lambda_handler(event, context):
...
+            # the date queries will only look at the file start_date.
+            query = query.where(
+                models.FileCatalog.start_date
+                <= datetime.datetime.strptime(value, "%Y%m%d")
+            )
+        # all non-time string matching parameters
+        else:
+            query = query.where(getattr(models.FileCatalog, param) == value)
```
greglucas, line 80: Out of curiosity and knowing nothing about this, is this the standard way of combining the query conditions, or is it faster to do something like combining all the conditions together into a single `where` clause? `query.where(logical_condition1 and logical_condition2 and logical_condition3 ...)`

sds_data_manager/lambda_code/SDSCode/queries.py
```diff
@@ -28,14 +33,62 @@ def lambda_handler(event, context):
...
+
+    engine = db.get_engine()
+    with Session(engine) as session:
+        search_result = session.execute(query).all()
+
+    logger.info("Query Search Results: " + str(search_result))
```
greglucas:
```suggestion
    logger.info("Found [%s] Query Search Results: %s", len(search_result), search_result)
```
Would it be helpful to have the number of results found for our logs too?

sds_data_manager/lambda_code/SDSCode/queries.py
```diff
@@ -28,14 +33,62 @@ def lambda_handler(event, context):
...
 
     # Format the response
     response = {
         "statusCode": 200,
-        "body": json.dumps(search_result),  # Convert JSON data to a string
+        "body": json.dumps(str(search_result)),
```
greglucas: Could you add a quick comment of what this should look like? Is it a list of tuples, list of dictionaries (one dictionary for each item and a mapping of each quantity), something else entirely?

tests/lambda_endpoints/test_queries.py
```diff
@@ -0,0 +1,708 @@
...
+        "version": "v02-01",
+        "extension": "pkts",
+    }
+
+    # Add data to the file catalog
+    with Session(test_engine) as session:
+        session.add(models.FileCatalog(**metadata_params))
+        session.commit()
+
+    return ""
```
greglucas: Is this just to avoid the pre-commit warning? Can you remove the return altogether or return None.

tests/lambda_endpoints/test_queries.py
```diff
@@ -0,0 +1,708 @@
...
+                )
+            ]
+        )
+    )
+    return expected_response
+
+
+def test_start_date_query(setup_test_data, test_engine, expected_response):
+    """Test that start date can be queried"""
+    event = {
```
greglucas: Can you reduce these down to the required values? I think that might just be the query parameters section and maybe a few others. But it would be a bit easier to parse what we are testing specifically.

Another way to combine things would be to put this in a `pytest.parametrize("event", [start_date_event, end_date_event, ...)` so you could eliminate a little boilerplate that way as well and have pytest do the looping for you with different queries.

sds_data_manager/lambda_code/SDSCode/queries.py
sdhoyt: Yeah, I'll change that to query

tests/lambda_endpoints/test_queries.py
sdhoyt: That's weird. I did have  `yield ""` because I thought ruff would have an issue if it was just `yield`, but it was complaining about it being `yield` instead of `return` and it looks like it might have automatically changed it to `return`. It does look like like it still works as only `return`, so I'll remove the empty string.

tests/lambda_endpoints/test_queries.py
sdhoyt: Yeah, I'm not sure why I thought I had to input a whole event for this to work. I'll remove everything except for query parameters.

tests/lambda_endpoints/test_queries.py
sdhoyt: Nevermind, if I just have `return` or `return None` it looks like either black or ruff removes it from the code and then ruff complains about nothing being returned lol

sds_data_manager/lambda_code/SDSCode/queries.py
sdhoyt: I'm not sure what the standard way is, but I saw both in documentation. Appending wheres like this seemed to me like the best way to add all the query parameters while handling the dates.

sdhoyt: I'm not sure what the difference would be as far as speed, but from my understanding, the query variable is converted to sql when it's executed, so I'm not sure if the resulting sql would be different

bourque (approved):
I agree with the comments made by Greg here, and other than those, this looks good to me. And nice test coverage!

sds_data_manager/lambda_code/SDSCode/queries.py
greglucas: :+1: I agree this seems pretty clean and the documentation indicates that the where returns a new statement connected by an AND, so as long as thats what you want this seems reasonable.
https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.Select.where
> Return a new [select()](https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.select) construct with the given expression added to its WHERE clause, joined to the existing clause via AND, if any.

But, if you are curious, it looks like you could compare the two results and see if they differ at all:
https://stackoverflow.com/questions/5631078/sqlalchemy-print-the-actual-query

tests/lambda_endpoints/test_queries.py
greglucas: `yield session` might make sense here? Which would be followed by a `session.close()` below I think.
https://stackoverflow.com/questions/74921132/how-to-populate-db-from-fixture-before-test

tests/lambda_endpoints/test_queries.py
```diff
@@ -0,0 +1,150 @@
...
+        "queryStringParameters": {"start_date": "20251101", "end_date": "20251201"}
+    }
+
+    returned_query = queries.lambda_handler(event=event, context={})
+
+    assert returned_query["statusCode"] == 200
+    assert returned_query["body"] == expected_response
+
+
+def test_empty_start_date_query(setup_test_data, test_engine, expected_response):
```
greglucas:
```suggestion
def test_empty_start_date_query(setup_test_data, test_engine):
```

greglucas (approved):
Some minor style suggestions, but the content all looks good to me! I agree the test cases are really nice here. I personally also prefer parameterizations for the repetitive tests (can give an `id="multiple-queries"` if you want more descriptive names too), but that is very minor and no need to update here.

PR #229, opened 2024-01-23
Indexer events

### LINKED ISSUE #221: Update Indexer Lambda to handle different event sources
### Summary

Update indexer lambda to handle event from three event sources and write data to correct database table.

Three event sources
  - aws.s3
  - aws.batch
  - imap.lambda - custom PutEvent from lambda

### Requirements

_No response_

### Related tickets

_No response_

tech3371:
# Change Summary
Refactored path_helper.py and updated indexer lambda to handle two event sources. Will update batch job event handler in upcoming PR.

## Overview
Main changes are in indexer.py and path_helper.py

## Updated Files
   - updated status enum and changed job definition can be null
   - updated indexer to handle s3 and custom event.
- sds_data_manager/lambda_code/SDSCode/path_helper.py
   - updated path_helper class 

Minor updates to these two files:

## Testing
- tests/lambda_endpoints/test_path_helper.py

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -124,6 +124,7 @@ class FileCatalog(Base):
         Integer, ForeignKey("status_tracking.id"), nullable=False
     )
 
+
 class PreProcessingDependency(Base):
     """Preprocessing dependency table"""
 
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -1,5 +1,7 @@
 """Main file to store schema definition"""
 
+from enum import Enum
+
 from sqlalchemy import (
     Boolean,
     Column,
@@ -10,12 +12,12 @@ from sqlalchemy import (
     String,
 )
 from sqlalchemy import (
-    Enum as SqlEnum,
+    Enum as SQLEnum,
 )
 from sqlalchemy.orm import DeclarativeBase
 
 # Instrument name Enums for the file catalog table
-instruments = SqlEnum(
+instruments = SQLEnum(
     "codice",
     "glows",
     "hi-45",
@@ -32,7 +34,7 @@ instruments = SqlEnum(
 )
 
 # data level enums for the file catalog table
-data_levels = SqlEnum(
+data_levels = SQLEnum(
     "l0",
     "l1",
     "l1a",
@@ -52,13 +54,22 @@ data_levels = SqlEnum(
 )
 
 # status enums for the status tracking table
-statuses = SqlEnum("INPROGRESS", "SUCCEEDED", "FAILED", name="status")
+statuses = SQLEnum("INPROGRESS", "SUCCEEDED", "FAILED", name="status")
 
 # "upstream" dependency means an instrument's processing depends on the existence
 # of another instrument's data
 # "downstream" dependency means that the instrument's data is used in another
 # instrument's processing
-dependency_directions = SqlEnum("UPSTREAM", "DOWNSTREAM", name="dependency_direction")
+dependency_directions = SQLEnum("UPSTREAM", "DOWNSTREAM", name="dependency_direction")
+
+
+class Status(Enum):
+    INPROGRESS = "INPROGRESS"
+    SUCCEEDED = "SUCCEEDED"
+    FAILED = "FAILED"
+
+
+statuses = SQLEnum(Status)
 
 
 class Base(DeclarativeBase):
@@ -90,7 +101,7 @@ class StatusTracking(Base):
     id = Column(Integer, Identity(start=1, increment=1), primary_key=True)
     file_to_create_path = Column(String, nullable=False)
     status = Column(statuses, nullable=False)
-    job_definition = Column(String, nullable=False)
+    job_definition = Column(String, nullable=True)
     ingestion_date = Column(DateTime, nullable=True)
 
 
@@ -109,8 +120,9 @@ class FileCatalog(Base):
     end_date = Column(DateTime, nullable=False)
     version = Column(String, nullable=False)
     extension = Column(String, nullable=False)
-    status_tracking_id = Column(Integer, ForeignKey("status_tracking.id"))
-
+    status_tracking_id = Column(
+        Integer, ForeignKey("status_tracking.id"), nullable=False
+    )
 
 class PreProcessingDependency(Base):
     """Preprocessing dependency table"""
--- a/sds_data_manager/lambda_code/SDSCode/indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/indexer.py
@@ -5,7 +5,6 @@ import os
 import sys
 
 import boto3
-from sqlalchemy import inspect, select
 from sqlalchemy import select
 from sqlalchemy.orm import Session
 
--- a/sds_data_manager/lambda_code/SDSCode/indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/indexer.py
@@ -6,11 +6,13 @@ import sys
 
 import boto3
 from sqlalchemy import inspect, select
+from sqlalchemy import select
 from sqlalchemy.orm import Session
 
 from .database import database as db
 from .database import models
-from .path_helper import FilenameParser
+from .database_handler import update_file_catalog_table, update_status_table
+from .path_helper import InvalidScienceFileError, ScienceFilepathManager
 
 # Logger setup
 logger = logging.getLogger()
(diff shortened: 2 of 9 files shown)


tech3371:
closing temporarily to rework path_helper.py

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -49,7 +54,15 @@
 )
 
 # status enums for the status tracking table
-statuses = Enum("INPROGRESS", "SUCCEEDED", "FAILED", name="status")
+
+
+class Status(Enum):
```
greglucas, line 66: Why is this one different than the others? Could it just be a SQLEnum directly?

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -5,12 +5,12 @@
 import sys
 
 import boto3
-from sqlalchemy import inspect
 from sqlalchemy.orm import Session
 
 from .database import database as db
 from .database import models
-from .path_helper import FilenameParser
+from .database.models import Status
```
greglucas: below when you use Status I would write `models.Status` since you've already imported `models` above, then you don't need to add this extra import.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +20,146 @@
...
+    """
+    # Retrieve the Object name
+    s3_filepath = event["detail"]["object"]["key"]
+
+    filename = os.path.basename(s3_filepath)
+    # TODO: add checks for SPICE or other
+    # data types
+    science_file = ScienceFile(filename)
+
+    if science_file.is_valid is False:
```
```suggestion
    if not science_file.is_valid:
```

Another idea would be to try and create your `ScienceFile` above and raise an exception and catch it here if it can't create it in the first place.

```python
class InvalidScienceFile(Exception):
    """Indicates a bad file type"""
    pass

try:
    science_file = ScienceFile(filename)
except InvalidScienceFile as e:
    return {"statusCode": 400, "body": str(e)}

File: sds_data_manager/lambda_code/SDSCode/indexer.py [hunk 4]
```diff
@@ -20,6 +20,146 @@
...
+        return {"statusCode": 400, "body": science_file.error_message}
+
+    # setup a dictionary of metadata parameters to unpack in the
+    # file catalog table
+    metadata_params = science_file.get_file_metadata_params()
+    metadata_params["file_path"] = s3_filepath
+
+    if metadata_params["data_level"] == "l0":
```
greglucas: invert this logic and emit the bad message right away and dedent the current if-block.
```suggestion
    if metadata_params["data_level"] != "l0":
        return {"statusCode": 400, "body": "Invalid data level"}
```

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +20,146 @@
...
+        Engine object to interact with the database.
+
+    Returns
+    -------
+    dict
+        HTTP response
+    """
+
+    if event["detail-type"] != "Batch Job State Change":
+        return {"statusCode": 400, "body": "Invalid event"}
```
greglucas: I think in one of your other PRs you added a helper function for all these return types. Do you want to re-use that here/add something similar since it looks like you're calling a lot of these?

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +20,146 @@
...
+    engine : sqlalchemy.engine.Engine
+        Engine object to interact with the database.
+
+    Returns
+    -------
+    dict
+        HTTP response
+    """
+    if event["detail-type"] != "Job Started":
+        return {"statusCode": 400, "body": "Invalid event"}
```
```suggestion
        return {"statusCode": 400, "body": "Invalid event, must have detail-type 'Job Started'"}
```

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -39,54 +179,17 @@ def lambda_handler(event, context):
...
-            # from upcoming PR
-            result = session.query(models.FileCatalog).all()
-            for row in result:
-                print(row.instrument)
-                print(row.file_path)
-
-            inspector = inspect(engine)
-            table_names = inspector.get_table_names()
-            print(table_names)
+    # L0 data or other data type such as SPICE data ingestion
```
greglucas, line 304: I like how these are broken out into helper methods and dispatching out to the other handlers to  do the bulk of the work.

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -1,43 +1,29 @@
...
 """
 import re
-from dataclasses import dataclass, field
 from datetime import datetime
 
+VALID_INSTRUMENTS = {
+    "codice",
+    "glows",
+    "hit",
+    "hi-45",
```
greglucas, line 10: `hi-90` and `ultra-90`?

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -1,43 +1,29 @@
...
-    startdate: str = "YYYYMMDD"
-    enddate: str = "YYYYMMDD"
-    version: str = r"^v\d{2}-\d{2}$"
-    extensions: list = field(default_factory=lambda: ["pkts", "cdf"])
-
-
-class FilenameParser:
+VALID_DATALEVELS = {"l0", "l1", "l1a", "l1b", "l1c", "l1d", "l2"}
+
+VALID_FILE_EXTENSION = ["pkts", "cdf"]
```
greglucas: For consistency with the sets above
```suggestion
VALID_FILE_EXTENSION = {"pkts", "cdf"}
```

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -58,14 +44,18 @@ def __init__(self, filename):
         Parameters
         ----------
         filename : str
-            Filename
+            Science data filename
         """
         self.filename = filename
-        self.filename_convention = (
+        self.split_filename = self.filename.split("_")
```
greglucas: Not really critical at all, but do we need this as an attribute on self here or could it just be a local variable that you use below?
```suggestion
        split_filename = self.filename.split("_")
```
The same goes for the `last_value` below and wondering if that could also be a local variable instead.

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -58,14 +44,18 @@ def __init__(self, filename):
...
+        filename_convention = (
             "<mission>_<instrument>_<datalevel>_<descriptor>_"
             "<startdate>_<enddate>_<version>.<extension>"
         )
-        self.split_filename = filename.split("_")
+
+        if len(self.split_filename) != 7:
+            raise ValueError(f"Invalid filename. Expected - {filename_convention}")
```
greglucas: If you were to take my other try/except suggestion (you don't need to, it is only an idea/suggestion), then this would be:
```suggestion
        if len(self.split_filename) != 7:
            raise InvalidScienceFile(f"Invalid filename. Expected - {filename_convention}")
```

tests/lambda_endpoints/test_indexer_lambda.py
```diff
@@ -10,20 +15,130 @@ def test_batch_job_event(test_engine):
...
+        "detail-type": "Batch Job State Change",
+        "source": "aws.batch",
+        "detail": {
+            "jobName": "test-batch-tenzin",
+            "status": "FAILED",
+            "statusReason": "some error message",
+            "container": {
+                "image": (
+                    "123456789012.dkr.ecr.us-west-2.amazonaws.com/"
+                    "codicehi-repo:latest"
```
greglucas:
```suggestion
                    "codice-repo:latest"
```

tests/lambda_endpoints/test_indexer_lambda.py
```diff
@@ -10,20 +15,130 @@ def test_batch_job_event(test_engine):
...
+                ),
+                "command": [
+                    "python",
+                    (
+                        "imap_cli --instrument glows --level l1a "
```
greglucas: codice here too?

tests/lambda_endpoints/test_indexer_lambda.py
```diff
@@ -10,20 +15,130 @@ def test_batch_job_event(test_engine):
...
+    returned_value = indexer.lambda_handler(event=event, context={})
+    assert returned_value["statusCode"] == 200
+
+    event["detail-type"] = "Testing"
+    returned_value = indexer.lambda_handler(event=event, context={})
+    assert returned_value["statusCode"] == 400
+    assert returned_value["body"] == "Invalid event"
+
+
+def test_custom_lambda_event(test_engine):
```
greglucas, line 187: Could you add a quick description of what you are testing/expecting to happen in each of your assert blocks? I think that will make it easier to follow what I should be looking for.

```
# With an input event with two dependencies there should be XYZ in the database table

# Here we are making sure that a SWAPI file is in the table

# Test with a bad job status (must be INPROGRESS)
```

greglucas (commented):
I like this approach. I just took a quick pass at it to give some coding feedback and didn't look at the specific event logic in detail. Hoping @laspsandoval and @maxinelasp will go over that!

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: I thought of that but I didn't know how to import its value from other files, such as indexer code file. The workaround was this. If you know way to do it, I appreciate it.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: I like that idea. Will this mean that in path_helper.py, I will need to raise error instead of return bool, error_message when filename was not valid.

tech3371: I pushed changes for this change.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +19,176 @@
...
+    dict
+        HTTP response
+    """
+    # Retrieve the Object name
+    s3_filepath = event["detail"]["object"]["key"]
+
+    filename = os.path.basename(s3_filepath)
+    # TODO: add checks for SPICE or other
+    # data types
+    science_file = ScienceFile(filename)
```
maxinelasp: Looks like this is a duplicate line (with line 78) so should this one be removed?

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +19,176 @@
...
+        lambda function to process
+    engine : sqlalchemy.engine.Engine
+        Engine object to interact with the database.
+
+    Returns
+    -------
+    dict
+        HTTP response
+    """
+    if event["detail-type"] != "Job Started":
```
maxinelasp: It might be good to add a note at the top about what these events look like - eg what the valid values are and general structure for the batch files and this custom event handler.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +19,176 @@
...
+        msg = "Invalid event, must have detail-type 'Job Started'"
+        return http_response(status_code=400, body=msg)
+
+    if event["detail"]["status"] != "INPROGRESS":
```
maxinelasp: Should these use the ENUM values for status?

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -1,43 +1,37 @@
...
+VALID_FILE_EXTENSION = {"pkts", "cdf"}
+
+
+class InvalidScienceFileError(Exception):
+    """Indicates a bad file type"""
+
+    pass
+
+
+class ScienceFile:
```
maxinelasp: Suggested name change - "ScienceFilepathManager" or something similar. As it is, this looks like the class is opening the file rather than doing things like verifying the name

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -1,43 +1,37 @@
...
     def __init__(self, filename):
```
maxinelasp, line 33: The docs here should include what attributes you are setting in init (eg `self.filename`, `self.instrument`...)

sds_data_manager/lambda_code/SDSCode/path_helper.py
```diff
@@ -1,43 +1,37 @@
...
-        """Stores common methods to check
-        file patterns and build upload path.
+        """Dataclass to store file pattern
```
maxinelasp: This isn't a dataclass (although I do think it should be one!)

sds_data_manager/lambda_code/SDSCode/path_helper.py
maxinelasp: The usage of this class might be a little more clear if you move a lot of `__init__` into some kind of static method like `validate_filename` so in indexer.py you can run `ScienceFile.validate_filename(filename)` and make it really clear what you're doing with the filename. Then, `__init__` would basically call validate_filename first and then set the attribute variables. 

This isn't something I feel strongly about - I think your current design works well. Just something I wanted to mention since the way you use this class in indexer.py doesn't involve going back to the instance after you create it.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +19,176 @@
...
+    logger.info(f"Attempting to insert {filename} into database")
+
+    try:
+        _ = ScienceFile(filename)
+    except InvalidScienceFileError as e:
+        return http_response(status_code=400, body=str(e))
+
+    # Write event information to status tracking table.
+    logger.info(f"Inserting {filename} into database")
+    with Session(engine) as session:
```
maxinelasp: I'd like this call to be pulled out into a helper function in a different database focused file - so you can just call `database_handler.update_status_table(status_params)` instead of opening a database session here. (Gave similar advice to Laura)

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -39,54 +208,17 @@ def lambda_handler(event, context):
...
-            print(table_names)
+    # L0 data or other data type such as SPICE data ingestion
+    if event["source"] == "aws.s3":
+        return s3_event_handler(event, engine)
+    # Data processing status
+    elif event["source"] == "aws.batch":
+        return batch_event_handler(event, engine)
+    # Data processing initiation
+    elif event["source"] == "imap.lambda":
+        return custom_event_handler(event, engine)
```
maxinelasp: instead of a passing around the engine here, I'd like to see an instance of a database accessor class instead. Then we can share tools across all different database accessors, it's less to change if we end up changing the database, and it generally allows more flexibility in case the database causes us problems.

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -60,12 +60,14 @@ def lambda_handler(event, context):
...
     filename = event["queryStringParameters"]["filename"]
-    filename_parsed = FilenameParser(filename)
-    upload_path = filename_parsed.upload_filepath()
-    if upload_path["statusCode"] != 200:
-        return upload_path
 
-    s3_key_path = upload_path["body"]
+    science_file = ScienceFile(filename)
+
+    if science_file.is_valid is False:
```
maxinelasp:
```suggestion
    if not science_file.is_valid:
```

sds_data_manager/stacks/indexer_lambda_stack.py
```diff
@@ -145,7 +145,7 @@ def __init__(
                 detail_type=["Batch Job Started"],
                 detail={
                     "file_to_create": [{"exists": True}],
-                    "status": ["InProgress"],
+                    "status": ["INPROGRESS"],
```
maxinelasp, line 153: Related to another comment - should this use the status ENUM?

maxinelasp (changes requested):
Overall a great first pass - I mainly would want to see a different way of accessing the database.

sds_data_manager/lambda_code/SDSCode/upload_api.py
maxinelasp: Lol, I didn't see that Greg made the same suggestion 😆

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: I am removing this and above lines because I realized that we don't need this check because eventbridge never let it through based on its rule. So it was extra thing I was doing. but yeah, if I had it then it makes sense.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: yeah, I don't know why I passed it. I don't use it. good catch though.

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: I have heard mixed things. Let me think about this.

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: I know. may be we could revisit that after demo.

sds_data_manager/stacks/indexer_lambda_stack.py
tech3371: I am having scope/access issue. Can I revisit this in future PR?

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +20,226 @@
...
+        update_status_table(status_params)
+    except Exception as e:
+        logger.info(str(e))
+        return http_response(status_code=400, body=str(e))
+
+    # Query to get foreign key id for catalog table
+    status_tracking = None
+    with Session(db.get_engine()) as session:
+        # Query to get foreign key id for catalog table
+        status_tracking = (
```
sdhoyt: I'm still learning about sqlalchemy, but from what I saw in the documentation, it looked like the session.query().filter() is an older way of querying from version 1 and in version 2 it's more focused on select().where(), which is what I used in my query PR. If both are still valid, we should probably pick one way to be consistent unless there are differences between the two where one works better in certain situations that I"m not aware of.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +20,226 @@
...
+        return http_response(status_code=400, body=str(e))
+
+    # setup a dictionary of metadata parameters to unpack in the
+    # file catalog table
+    metadata_params = science_file.get_file_metadata_params()
+    metadata_params["file_path"] = s3_filepath
+
+    if metadata_params["data_level"] != "l0":
+        logger.info("Invalid data level. Expected l0")
+        return http_response(status_code=400, body="Invalid data level")
```
sdhoyt, line 130: Where does this response end up? Does this somehow show in the upload API?

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -100,4 +113,6 @@ class FileCatalog(Base):
     end_date = Column(DateTime, nullable=False)
     version = Column(String, nullable=False)
     extension = Column(String, nullable=False)
-    status_tracking_id = Column(Integer, ForeignKey("status_tracking.id"))
+    status_tracking_id = Column(
+        Integer, ForeignKey("status_tracking.id"), nullable=False
```
sdhoyt, line 124: I remember a conversation about wanting null to be valid here because a file wouldn't have a status id at the time it's entered into the catalogue. Did I misunderstand that?

tech3371: I believe so. The way I understood was that by the time we get to writing to catalog table, we would already have record in the status table and therefore, it shouldn't be null. That's my understanding at the moment, but it could change later if needed.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: This doesn't get end up in API, I wrote that as default behavior so that I can check in test and also down the road, if we decide to add endpoint or something, have some response will allow checking for error from this lambda.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: yes. I did look at your code was questioning myself too. I didn't read the document so what you are saying makes sense. I will make that change to be consistent.

sds_data_manager/lambda_code/SDSCode/path_helper.py
tech3371: I will follow-up about this offline and fix it upcoming PR

tech3371:
pre-commit.ci autofix

sdhoyt (approved):

maxinelasp (approved):

PR #230, opened 2024-01-24
Batch starter

laspsandoval:
# Batch Starter Updates

## Overview
Changed batch starter to work with current database setup and new architecture plans.

## New Files
None

## Deleted Files
None

## Updated Files
- batch_starter.py
   - added orm and now uses current database models
   - improved command to batch job
   - added ability to get filename from event and retrieve information from the filename
   - retrieves s3 bucket name from env variable
   - major reformatting
- downstream_dependents.json
   - make names lowercase
- instrument_lambdas.py
   - set s3 bucket name from env variable
- imap_api.py
   - changed arguments passed to match command in batch_starter.py
- test_batch_starter.py
   - updated tests for batch_starter.py including how mock database was setup

## Testing
- Verified that docker container ran properly with expected command: docker run test-image --instrument hit --level l1a --s3_uri 's3://sds-data-449431850278/imap/hit/l0/2024/01/imap_hit_l0_sci_20240101_20240104_v02-01.pkts' --dependency '[{"instrument": "hit", "level": "l0", "version": "v00-01"}]'

### DIFF
--- a/sds_data_manager/lambda_code/batch_starter.py
+++ b/sds_data_manager/lambda_code/batch_starter.py
@@ -4,14 +4,11 @@ import os
 import re
 from collections import defaultdict
 from datetime import datetime, timedelta
-from .SDSCode.database import models
-from .SDSCode.database.database import engine
-from sqlalchemy.orm import Session
-from sqlalchemy import func, or_
-from datetime import timedelta
 
 import boto3
 import psycopg2
+from sqlalchemy import func
+from sqlalchemy.orm import Session
 
 # Setup the logging
 logging.basicConfig(level=logging.INFO)
@@ -66,19 +63,24 @@ def query_instruments(db_session, version, process_dates, instruments):
     all_records = []
 
     for instrument in instruments:
-        instrument_name = instrument['instrument'].lower()
-        instrument_class = globals()[instrument_name.capitalize() + 'Table']
+        instrument_name = instrument["instrument"].lower()
+        instrument_class = globals()[instrument_name.capitalize() + "Table"]
 
         query = db_session.query(instrument_class).filter(
             instrument_class.version == version,
-            instrument_class.data_level == func.lower(instrument['level']),
-            instrument_class.ingestion_date.between(min(process_dates), max(process_dates) + timedelta(days=1))
+            instrument_class.data_level == func.lower(instrument["level"]),
+            instrument_class.ingestion_date.between(
+                min(process_dates), max(process_dates) + timedelta(days=1)
+            ),
         )
 
         records = query.all()
         for record in records:
             # convert the record to a dictionary, or handle as needed
-            record_dict = {column.name: getattr(record, column.name) for column in record.__table__.columns}
+            record_dict = {
+                column.name: getattr(record, column.name)
+                for column in record.__table__.columns
+            }
             all_records.append(record_dict)
 
     return all_records
@@ -108,7 +110,6 @@ def query_upstream_dependencies(cur, downstream_dependents):
 
     # Iterate over each key-value pair
     for instr, levels in data.items():
-
         # Check if all dependencies for this date are present in result
         result = query_instruments(
             cur, version, [record["date"]], upstream_dependencies
@@ -223,16 +224,17 @@ def get_downstream_dependents(instrument, level):
 
     return dependents
 
+
 def extract_components(filename):
     """Extracts components from filename"""
     pattern = (
-        r"^imap_"  
-        r"(?P<instrument>[^_]*)_"  
-        r"(?P<datalevel>[^_]*)_"  
-        r"(?P<descriptor>[^_]*)_"  
-        r"(?P<startdate>\d{8})_"  
-        r"(?P<enddate>\d{8})_"  
-        r"(?P<version>v\d{2}-\d{2})"  
+        r"^imap_"
+        r"(?P<instrument>[^_]*)_"
+        r"(?P<datalevel>[^_]*)_"
+        r"(?P<descriptor>[^_]*)_"
+        r"(?P<startdate>\d{8})_"
+        r"(?P<enddate>\d{8})_"
+        r"(?P<version>v\d{2}-\d{2})"
         r"\.cdf$"
     )
     match = re.match(pattern, filename)
@@ -270,12 +272,16 @@ def lambda_handler(event: dict, context):
     )
 
     with Session(database.engine) as session:
-        result = query_instruments(session, 1, [datetime(2023, 5, 31)], [{"instrument": "codicehi", "level": "l1b"}])
+        result = query_instruments(
+            session,
+            1,
+            [datetime(2023, 5, 31)],
+            [{"instrument": "codicehi", "level": "l1b"}],
+        )
         print(result)
 
     with db_connect(db_secret_arn) as conn:
         with conn.cursor() as cur:
-
             # TODO: query the version table here for latest version
             #  of each downstream_dependent. Add logic for reprocessing.
             # TODO: add universal spin table query for ENAs and GLOWS
--- a/sds_data_manager/lambda_code/batch_starter.py
+++ b/sds_data_manager/lambda_code/batch_starter.py
@@ -2,13 +2,14 @@ import json
 import logging
 import os
 import re
-from collections import defaultdict
-from datetime import datetime, timedelta
+from datetime import datetime
+from pathlib import Path
 
 import boto3
-import psycopg2
-from sqlalchemy import func
 from sqlalchemy.orm import Session
+
+from .SDSCode.database import database as db
+from .SDSCode.database import models
 
 # Setup the logging
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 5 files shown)


sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
     logger.info(f"Context: {context}")
 
-    # Event details (TBD)
-    instrument = event["detail"]["object"]["instrument"]
-    instrument_downstream = get_downstream_dependencies(instrument)
-    db_secret_arn = os.environ.get("SECRET_ARN")
+    # Event details:
+    filename = event["detail"]["object"]["key"]
+    components = extract_components(filename)
+    instrument = components["instrument"]
```
tech3371, line 349: Don't worry about it for this PR. In upcoming PR, we can replace this with path_helper.py class. That's in review too.

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
+        dependent["version"] = "v00-01"  # placeholder
+        dependent["start_date"] = start_date
+        dependent["end_date"] = end_date
 
-    return level, version, process_dates
+    return downstream_dependents
 
 
-def query_instruments(cur, version, process_dates, instruments):
+def find_dependencies(
```
tech3371: logically, I follow it. Later, we can simplify this search by making query to pre-processing table. We can modify this later post demo.

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
+    downstream_dependent_instrument,
+    downstream_dependent_inst_level,
+    downstream_dependent_version,
+    data,
+):
     """
-    Queries the database for instruments and retrieves their records.
+    Finds dependency information for each instrument.
```
tech3371, line 110: Can you add a comment that this function is looking for upstream dependency of current downstream dependency?

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
-                if dependency_found:
-                    upstream_dependencies.append({"instrument": instr, "level": level})
-
-        # Check if all dependencies for this date are present in result
-        result = query_instruments(
-            cur, version, [record["date"]], upstream_dependencies
-        )
-        all_dependencies_met = all_dependency_present(result, upstream_dependencies)
+        # If all dependencies are available, prepare the data for batch job.
+        if queried_instrument_dependencies == upstream_dependencies:
```
tech3371: should this be a length comparison?

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
-
-
-def prepare_data(instruments_to_process):
+    with filepath.open() as file:
+        data = json.load(file)
+
+    return data
+
+
+def prepare_data(filename, upstream_dependencies, s3_bucket):
```
tech3371, line 258: later, this function can be simplify or replaced by path_helper class.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -1,293 +1,231 @@
...
-            "date": "2023-05-31 14:45:00+03",
-            "ingested": "2023-06-02 14:45:00+11",
-            "mag_id": 4,
-            "codicelo_id": None,
-            "spice_id": None,
-            "parent_codicehi_id": 4,
-            "pointing_id": 1,
-        },
-    ]
+def _aws_credentials():
```
tech3371: `conftest.py` in this directory already takes care of this. You can remove this and call it in functions that needs it.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -1,293 +1,231 @@
...
+    assert data["codicehi"]["l0"][0]["level"] == "l1a"
 
-    # Since there are 2 instruments x 3 dates and one record to remove = 5
-    output = remove_ingested(records, instrument_downstream, process_dates)
 
-    assert len(output) == 5
+def test_find_dependencies():
+    "Tests find_dependencies function."
+    base_directory = Path(__file__).resolve()
+    base_path = base_directory.parents[2] / "sds_data_manager" / "lambda_code"
```
tech3371, line 141: interesting. didn't know you go two folders up this way. nice!

tech3371 (approved):
I add few minor comments and about the instrument names in dependency json as we talked in person. Besides that, I think it looks great for the demo! It was much easier to follow and understand. nice work!

tests/docker_images/imap_api.py
```diff
@@ -25,15 +26,30 @@ def _parse_args():
...
+    parser.add_argument(
+        "--level", type=str, required=True, help="Data processing level."
+    )
+    parser.add_argument(
+        "--s3_uri",
+        type=str,
+        required=True,
+        help="Full path to the file in the S3 bucket.",
+    )
+    parser.add_argument(
```
maxinelasp, line 47: Is this all dependencies, or only downstream dependencies?

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
+    version = upstream_dependency["version"]
+
+    record = (
+        session.query(models.FileCatalog)
+        .filter(
+            models.FileCatalog.instrument == instrument,
+            models.FileCatalog.data_level == level,
+            models.FileCatalog.version == version,
+            models.FileCatalog.start_date >= datetime.strptime(start_date, "%Y%m%d"),
+            models.FileCatalog.end_date <= datetime.strptime(end_date, "%Y%m%d"),
```
maxinelasp, line 55: Does it make sense to move the database query methods into a different file that you can them import? Then it can be used in other pieces of code.

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
+    data : dict
+        Dictionary containing dependency data.
 
     Returns
     -------
-    all_records : list of dict
-        A list of dictionaries where each dictionary corresponds to a record
-        from the database that matches the given criteria.
+    dependencies : list of dict
+        A list of dictionaries containing dependency instrument,
```
maxinelasp, line 126: Specify if this is upstream or downstream

maxinelasp: In general, I think it would be useful to get more info on what this actually looks like - what is the structure of the dict? How is it used? The dependencies are really confusing so being really clear will help, lol

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
-        # Map the column names to the records
-        records_dicts = [dict(zip(column_names, record)) for record in records]
+    for instr, levels in data.items():
+        for level, deps in levels.items():
+            if any(
+                dep["instrument"] == downstream_dependent_instrument
+                and dep["level"] == downstream_dependent_inst_level
+                for dep in deps
+            ):
+                dependencies.append({"instrument": instr, "level": level})
```
maxinelasp: So you are searching the dependencies provided in `data` to find if any of them have `downstream_dependent` as a dependency? Meaning this list is all the upstream dependencies of `downstream_dependent`?

maxinelasp: Could you rename `dependencies` to a more descriptive name?

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
-    return all_records
+    return dependencies
 
 
-def remove_ingested(records, inst_list, process_dates):
+def query_upstream_dependencies(session, downstream_dependents, data, s3_bucket):
     """
-    Identifies and returns a list of instruments
-    that have not been ingested for the specified dates.
+    Queries and checks if dependencies are available.
```
maxinelasp: This method is called `query_upstream_dependencies` but takes in `downstream_dependents`. Could you add some more info in the method description about _what_ dependencies are looked for?

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
+    instruments_to_process : list of dict
+        A list of dictionaries containing the filename and prepared data.
 
     """
 
-    output = []
+    instruments_to_process = []
+
+    # Iterate over each downstream dependent
+    for dependent in downstream_dependents:
```
maxinelasp, line 184: Talked about this with Tenzin - but are these truly downstream_dependents, or are they the current records you are trying to find dependency information on? Could you rename this to "current_records" or something to make it clear that this is a list of records you are trying to find upstream dependencies for? (If I'm interpreting correctly...)

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
+                session, upstream_dependency, start_date, end_date
+            )
+            if record is not None:
+                logging.info(
+                    f"Dependency found: {upstream_dependency['instrument']}, "
+                    f"{upstream_dependency['level']}, "
+                    f"{upstream_dependency['version']}"
+                )
+                # Add dependency to list of available dependencies.
+                queried_instrument_dependencies.append(
```
maxinelasp: This is duplicating `upstream_dependencies` - perhaps instead, you could use a boolean value `all_dependencies_available` that switches to false once a dependency is not found. Then you can save the space and comparison between `upstream_dependencies` and `queried_instrument_dependencies`. I also think it would be more clear.

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -1,433 +1,364 @@
...
 
-            if (
-                date_obj,
-                inst["instrument"].lower(),
-                inst["level"].lower(),
-            ) not in records_set:
-                output.append(
+        for upstream_dependency in upstream_dependencies:
+            # Check to see if each upstream dependency is available
+            record = query_instrument(
```
maxinelasp, line 199: Would it make sense to make this query into a request for all the upstream dependencies? If you could make one request for all of them, it would eliminate this whole for-loop, which would be nice. If you're planning on changing the database setup (which I talked about with Tenzin) then leaving it like this is also fine

sds_data_manager/lambda_code/batch_starter.py
maxinelasp: Agreed with Tenzin that this should be pulled out for other code to use!

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -442,59 +373,35 @@ def lambda_handler(event: dict, context):
...
-            if not uningested:
-                logger.info("No uningested downstream dependents found.")
-                return
-
-            # TODO: add universal spin table query for ENAs and GLOWS
-            # decide if we have sufficient upstream dependencies
-            downstream_instruments_to_process = query_upstream_dependencies(
-                cur, uningested, version
-            )
+    with Session(engine) as session:
```
maxinelasp, line 382: Nice, I like this section, the design works out to be very clear about when a batch job is started vs skipped!

maxinelasp (changes requested):
A few comments, mostly related to making the docs very clear about upstream/downstream dependencies

tests/docker_images/imap_api.py
laspsandoval: This is upstream dependencies only. And this is only a test script....not for the final processing. Only for the demo. The one that we will use more permanently is located at: imap_processing/examples. The cli.py in imap_processing probably also needs to be updated, but in another review.

sds_data_manager/lambda_code/batch_starter.py
laspsandoval: Yes. That is correct. We are making a list of upstream dependencies of the downstream dependent.

sds_data_manager/lambda_code/batch_starter.py
laspsandoval: Downstream dependents are the candidates for the batch job. They will have records in the status table when the batch_starter.py completes. I'll put a note that they are candidates for the batch job.

sds_data_manager/lambda_code/batch_starter.py
laspsandoval: Yes you are right. There is a better way to do this. I'll put it as a todo for when change the database setup.

sds_data_manager/lambda_code/batch_starter.py
laspsandoval: That would be a nice to have. I have to think about what other pieces of code would use these.

sds_data_manager/lambda_code/batch_starter.py
laspsandoval: I like your idea about dong a single query without a loop. I'll see if I can implement that.

maxinelasp (approved):
Nice, looks good!

PR #231, opened 2024-01-24
Added pre-processing dependency table

tech3371:
# Change Summary
This PR only creates pre-processing table and test that we are able to make query to it. This is setting up something but will be modified in upcoming PR.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -32,6 +32,7 @@ instruments = Enum(
 # data level enums for the file catalog table
 data_levels = Enum(
     "l0",
+    "l1",
     "l1a",
     "l1b",
     "l1c",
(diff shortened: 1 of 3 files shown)


sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -20,6 +20,48 @@
...
+            .filter(models.PreProcessingDependency.primary_instrument == instrument)
+            .filter(models.PreProcessingDependency.primary_data_level == data_level)
+            .filter(models.PreProcessingDependency.primary_descriptor == descriptor)
+            .filter(models.PreProcessingDependency.direction == direction)
+            .filter(models.PreProcessingDependency.relationship == relationship)
+            .all()
+        )
+        for result in results:
+            dependency.append(result)
+    return dependency
```
maxinelasp, line 61: Very clean! :D

maxinelasp (approved):
Nice, I like this!

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -51,6 +52,8 @@
 # status enums for the status tracking table
 statuses = Enum("INPROGRESS", "SUCCEEDED", "FAILED", name="status")
 
+dependency_directions = Enum("UPSTREAM", "DOWNSTREAM", name="dependency_direction")
```
bourque: It might be helpful to add a small comment of what exactly we mean by "upstream" and "downstream". Perhaps something like this: (Do I have this backwards?)

```suggestion
# "upstream" dependency means an instrument's processing depends on the existence of another instrument's data
# "downstream" dependency means that the instrument's data is used in another instrument's processing
dependency_directions = Enum("UPSTREAM", "DOWNSTREAM", name="dependency_direction")
```

bourque (approved):
Nice work, looks great!

sds_data_manager/lambda_code/SDSCode/database/models.py
sdhoyt: It looked like in your indexer PR that you started using SqlEnum. Should you use that here instead of Enum?

sdhoyt (approved):
Looks great!

PR #232, opened 2024-01-25
Various updates to database schema

bourque:
# Change Summary

## Overview
This PR makes some additions/changes to our database schema. I reckon that this might spawn in a larger discussion on the finer details of the schema. All opinions welcome!

## Updated Files

- `sds_data_manager/lambda_code/SDSCode/database/models.py`
    - Adding unique constraints
    - Adding new enum for file extensions
    - Renaming global variable to uppercase (I think this is conventional?)
    - Added some maximum lengths for some string columns
    - Added draft of `version` table


sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -98,32 +109,79 @@ class FileCatalog(Base):
...
 
     # TODO: determine cap for strings
     id = Column(Integer, Identity(start=1, increment=1), primary_key=True)
     file_path = Column(String, nullable=False)
-    instrument = Column(instruments, nullable=False)
-    data_level = Column(data_levels, nullable=False)
-    descriptor = Column(String, nullable=False)
+    instrument = Column(INSTRUMENTS, nullable=False)
+    data_level = Column(DATA_LEVELS, nullable=False)
+    descriptor = Column(String(20), nullable=False)
```
bourque, line 142: I'm thinking we don't want our `descriptors` to be incredibly long. I mostly arbitrarily chose `20` here, but i'm not really sure what this should be, if anything.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -98,32 +109,79 @@ class FileCatalog(Base):
...
     start_date = Column(DateTime, nullable=False)
     end_date = Column(DateTime, nullable=False)
-    version = Column(String, nullable=False)
-    extension = Column(String, nullable=False)
+    version = Column(String(6), nullable=False)
```
bourque: Chose `6` here for the 6 characters in `vXX-YY`

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -98,32 +109,79 @@ class FileCatalog(Base):
     """File catalog table"""
 
     __tablename__ = "file_catalog"
+    __table_args__ = UniqueConstraint(
```
tech3371: I looked up what this does and chatgpt said this
```
The UniqueConstraint in SQLAlchemy is used to define a constraint at the table level that ensures the uniqueness of data across multiple columns. When you apply a UniqueConstraint to a set of columns, SQLAlchmey and the underlying database ensure that no two rows of the table contain the same combination of values in those columns. This is particularly useful when you want to enforce uniqueness across a combination of columns, rather than just a single column.
```
That's great that it does that for us. Nice!

tech3371 (approved):
Nice! Thank you for doing this! LGTM

sds_data_manager/lambda_code/SDSCode/database/models.py
maxinelasp: What is the result of this if you try and add a duplicate? Does it throw an error? I like this constraint in the database, but it will be something we need to handle in indexer

sds_data_manager/lambda_code/SDSCode/database/models.py
maxinelasp: Maybe make it 7 or 8 just in case? I could see hitting 3 digit versions and it would be an annoying thing to fix

maxinelasp (approved):
I like the changes! One comment about the version len

sds_data_manager/lambda_code/SDSCode/database/models.py
bourque: Good point. Yes, I believe SQLAlchemy will throw an `IntegrityError`. I will check the indexer code to make sure it is handled properly.

PR #233, opened 2024-01-26
added batch job event handler to indexer lambda

tech3371:
# Change Summary
Added batch job event handler to indexer lambda

## Updated Files
   - Main changes to handle batch job event
- sds_data_manager/lambda_code/SDSCode/path_helper.py
   - Added TODO

## Testing
tests/lambda_endpoints/test_indexer_lambda.py

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/indexer.py
@@ -46,8 +46,8 @@ def get_file_creation_date(s3_uri):
     # time looks like this:
     # 2024-01-25 23:35:26+00:00
     # Formats the datetime object to a string with the format "%Y%m%d".
-    ingestion_data_str = file_creation_date.strftime("%Y%m%d")
-    return datetime.datetime.strptime(ingestion_data_str, "%Y%m%d")
+    # ingestion_data_str = file_creation_date.strftime("%Y%m%d")
+    return file_creation_date
 
 
 def get_dependency(instrument, data_level, descriptor, direction, relationship):
@@ -146,7 +146,7 @@ def s3_event_handler(event):
     try:
         science_file = ScienceFilepathManager(filename)
     except InvalidScienceFileError as e:
-        logger.info(str(e))
+        logger.error(str(e))
         return http_response(status_code=400, body=str(e))
 
     # setup a dictionary of metadata parameters to unpack in the
@@ -155,7 +155,7 @@ def s3_event_handler(event):
     metadata_params["file_path"] = s3_filepath
 
     if metadata_params["data_level"] != "l0":
-        logger.info("Invalid data level. Expected l0")
+        logger.error("Invalid data level. Expected l0")
         return http_response(status_code=400, body="Invalid data level")
 
     # Add data to the file catalog and status tables
@@ -179,7 +179,7 @@ def s3_event_handler(event):
         logger.info(f"Inserting {filename} into database")
         update_status_table(status_params)
     except Exception as e:
-        logger.info(str(e))
+        logger.error(str(e))
         return http_response(status_code=400, body=str(e))
 
     # Query to get foreign key id for catalog table
@@ -193,14 +193,14 @@ def s3_event_handler(event):
         status_tracking = session.execute(query).first()
 
     if status_tracking is None:
-        logger.info("No status tracking record found")
+        logger.error("No status tracking record found")
         return http_response(status_code=400, body="No status tracking record found")
 
     try:
         metadata_params["status_tracking_id"] = status_tracking.id
         update_file_catalog_table(metadata_params)
     except Exception as e:
-        logger.info(str(e))
+        logger.error(str(e))
         return http_response(status_code=400, body=str(e))
 
     return http_response(status_code=200, body="Success")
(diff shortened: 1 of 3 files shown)


sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -21,6 +21,35 @@
...
+    logger.info(f"key: {key}")
+
+    response = s3_client.head_object(Bucket=bucket_name, Key=key)
+    file_creation_date = response["LastModified"]
+
+    # time looks like this:
+    # 2024-01-25 23:35:26+00:00
+    # Formats the datetime object to a string with the format "%Y%m%d".
+    ingestion_data_str = file_creation_date.strftime("%Y%m%d")
+    return datetime.datetime.strptime(ingestion_data_str, "%Y%m%d")
```
greglucas: If we are using datetime objects in the DB, do we actually want to keep the entire datetime interval?

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -215,16 +245,94 @@ def batch_event_handler(event):
...
+    # TODO: post demo, revisit this and improve it
     if event["detail"]["status"] == "SUCCEEDED":
-        # TODO:
-        # Get filename and then
-        # write it to file catalog table
-        pass
+        # Frist write to status table and then
+        # write to file catalog with foreign key
+        # information
+        try:
```
greglucas, line 267: A couple of comments here.

* Should we move the try/except up to the very highest levels of these handlers and surround the entry-point into the function? Then we wouldn't need to try/except all the various blocks.
```python
def handler(event, context):
    try:
        return sub_handler(event, context)
    except Except as e:
        return http_response(status_code=400, body=str(e))
```

* You might want to move the if/elif inside the Session() calls, so you don't need to repeat quite as much boilerplate.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -215,16 +245,94 @@ def batch_event_handler(event):
...
+                #     "DetailType": "Processed File",
+                #     "Detail": {
+                #         "object": {
+                #               "key": filename
+                #         },
+                #     },
+                # }
+        except Exception as e:
+            logger.info(str(e))
+            return http_response(status_code=400, body=str(e))
```
greglucas: minor nit, but if you are catching an exception it should probably be logged at the error level to really catch our eyes.

tests/lambda_endpoints/test_indexer_lambda.py
```diff
@@ -128,26 +129,68 @@ def populate_db(test_engine):
...
+    # first create test bucket
+    s3_client.create_bucket(
+        Bucket="data-bucket",
+        CreateBucketConfiguration={"LocationConstraint": "us-west-2"},
+    )
+    # write file to s3
+    s3_client.put_object(
+        Bucket="data-bucket",
+        Key=(
+            "imap/swapi/l1/2023/01/" "imap_swapi_l1_sci-1m_20230724_20230724_v02-01.cdf"
```
greglucas:
```suggestion
            "imap/swapi/l1/2023/01/imap_swapi_l1_sci-1m_20230724_20230724_v02-01.cdf"
```

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -200,10 +229,11 @@ def batch_event_handler(event):
                     "123456789012.dkr.ecr.us-west-2.amazonaws.com/" "codice-repo:latest"
                 ),
                 "command": [
-                    "python",
                     (
-                        "imap_cli --instrument codice --level l1a "
-                        "--filename '<s3-filepath>'"
+                        "--instrument hit --level l1a "
+                        "--s3_uri 's3://bucket_name/imap/hit/l1a/2024/01/"
```
greglucas, line 234: Can we avoid `s3_uri`? I think it'd be nice if we only use the paths and let the bucket be a detail. I think that might also simplify your path parsing below too.

greglucas (approved):

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: I think that makes sense since it's important to keep time of ingestions. I won't reformat it.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: I wondered that myself but Batch command was passing `s3_uri` and I changed others to match that. It gonna need some refactoring. I will revisit that demo. We will need to change in most places.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: I will visit that after demo too.

sds_data_manager/lambda_code/SDSCode/indexer.py
laspsandoval: Noted.

tech3371:
I am going to merge this even though codecov percentage is low. places it's complaining will be refactor and go away after demo.

PR #234, opened 2024-01-26
Added custom PutEvent to Batch Starter Lambda

tech3371:
# Change Summary
Added custom PutEvent to batch starter lambda which sends event to EventBridge. And There is EventBridge rule on indexer lambda that listens for this event.

## New Files
- sds_data_manager/lambda_code/SDSCode/lambda_custom_events.py
   - Dataclass to store EventBridge event details
 

## Updated Files
- sds_data_manager/lambda_code/batch_starter.py
   - Added function to send custom event to EventBridge from lambda
   - mock events

## Testing

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -12,7 +12,6 @@ from .database import database as db
 from .database import models
 from .lambda_custom_events import IMAPLambdaPutEvent
 
-
 # Logger setup
 logger = logging.getLogger()
 logger.setLevel(logging.DEBUG)
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -10,7 +10,8 @@ from sqlalchemy.orm import Session
 
 from .database import database as db
 from .database import models
-from .lambda_custom_events import EventDetail, IMAPLambdaPutEvent
+from .lambda_custom_events import IMAPLambdaPutEvent
+
 
 # Logger setup
 logger = logging.getLogger()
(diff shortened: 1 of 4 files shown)


sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -338,6 +336,60 @@ def extract_components(filename: str):
...
+    command = command_parameters.split("--")
+    s3_uri = command[3].replace("s3_uri '", "").replace("' ", "")
+    dependency = command[-1].replace("dependency ", "")
+
+    # Create event["detail"] information
+    detail = EventDetail(
+        file_to_create=s3_uri,
+        status=models.Status.INPROGRESS.value,
+        dependency=dependency,
+    )
```
greglucas: I will just voice my slight preference for not using dataclasses everywhere and leaving this as a simple dictionary and avoiding the `asdict(EventDetail)` call in the other dataclass. I know I'm in the minority here though, so it is fine to keep it as-is too.
```suggestion
    detail = {
        "file_to_create": s3_uri,
        "status": models.Status.INPROGRESS.value,
        "dependency": dependency,
    }
```

greglucas (approved):

sds_data_manager/lambda_code/batch_starter.py
```diff
@@ -338,6 +336,60 @@ def extract_components(filename: str):
     return components
 
 
+def send_lambda_put_event(command_parameters):
+    """Sends custom PutEvent to EventBridge
```
laspsandoval: So glad you have an example here. My question is: I thought that this was the information that would be used to populate the StatusTracking table, which has the following columns. 

    file_to_create_path = Column(String, nullable=False)
    status = Column(statuses, nullable=False)
    job_definition = Column(String, nullable=False)
    ingestion_date = Column(DateTime, nullable=True)
    
We've had a lot of conversations so maybe that's changed. You might just need to get me up to speed on this.

laspsandoval:
Just a comment on the PUT event. We can discuss at the meeting. Looks good besides that.

sds_data_manager/lambda_code/batch_starter.py
tech3371: That's true. This event is suppose to send custom putevent from batch starter to indexer lambda to populate status table. But we send this information to initialized that table with `file_to_create`, `status` and initialize others a null. Then when batch job finishes we update table with new information and update status as well based on batch job event.

tech3371:
pre-commit.ci autofix

PR #235, opened 2024-01-26
FIX: Change to path params instead of query params in APIs

greglucas:
# Change Summary

## Overview
For the download and upload API we are switching to path params instead of query params. Query params are still used for the query route to search the database.

## Updated Files

* Changed `download_query_api` to `download_api` because there is a separate `query_api` file. This makes all the naming consistent in the lambdas, but shows up as a diff because there was already a `download_api` file which was only being used for testing.


sds_data_manager/lambda_code/SDSCode/download_api.py
```diff
@@ -1,30 +1,104 @@
...
+        corresponding status code in case of failure.
+    """
+    one_day = 86400
+    url_life = os.environ.get("URL_EXPIRE", one_day)
+    path_params = event.get("pathParameters", {}).get("proxy", None)
+    logger.debug("Parsing path parameters=[%s] from event=" "[%s]", path_params, event)
+    if not path_params:
+        response_body = (
+            "No file requested for download. Please provide a filename "
+            "in the path. Eg. /download/filename.pkts"
```
tech3371: Should we add this so that user know that we expect full path on s3 since that's how we check in the try/except.
```suggestion
            "in the path. Eg. /download/<filepath-if-exist>/filename.pkts"
```

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -50,25 +50,30 @@ def lambda_handler(event, context):
 
     :return: A pre-signed url where users can upload a data file to the SDS.
     """
-    logger.info(f"Event: {event}")
```
tech3371, line 53: I found event logs helpful when looking through logs.

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -50,25 +50,30 @@ def lambda_handler(event, context):
...
+    try:
+        science_file = ScienceFilepathManager(filename)
+    except InvalidScienceFileError as e:
+        logger.error(str(e))
+        return {"statusCode": 400, "body": str(e)}
 
     s3_key_path = science_file.construct_upload_path()
 
-    url = _generate_signed_upload_url(s3_key_path, tags=event["queryStringParameters"])
+    url = _generate_signed_upload_url(s3_key_path)
```
tech3371, line 76: Recently, I noticed that we(including myself) don't add as many logs for when everything goes as expected and that leads to almost empty logs on cloudwatch which makes it seems like nothing happened.

tech3371 (approved):
I had few minor comments. Otherwise, it looks great

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: It is included down below now in a debug log.
`logger.debug("Parsing path parameters=[%s] from event=" "[%s]", path_params, event)`

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: You want to include tags on the s3 object? The queryStringParameters should be what was sent in the pathParams now, which was logged on entry to the lambda I thought.

sds_data_manager/lambda_code/SDSCode/download_api.py
greglucas: Good point, I'll update this!

sds_data_manager/lambda_code/SDSCode/upload_api.py
tech3371: ah yes. Didn't see that

sds_data_manager/lambda_code/SDSCode/upload_api.py
tech3371: Sorry for the confusion. This comment was not meant for this line of code but as general comment.

sds_data_manager/lambda_code/SDSCode/download_query_api.py
bourque, line 1: Thanks for moving the logic to `download_api.py`. It was bugging me that we had both a `download_api` and `download_query_api` 😅

sds_data_manager/lambda_code/SDSCode/download_api.py
```diff
@@ -1,30 +1,104 @@
...
+
+    s3_client = boto3.client("s3")
+
+    # check if object exists
+    try:
+        s3_client.head_object(Bucket=bucket, Key=filepath)
+    except botocore.exceptions.ClientError as e:
+        if e.response["Error"]["Code"] == "404":
+            # object doesn't exist
+            return http_response(status_code=404, body="File not found in S3.")
```
bourque, line 90: I will need to add this `404` to the API docs (easy to do!). Should I just push a commit to this PR for that or would you prefer I open a separate PR?

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -50,25 +50,30 @@ def lambda_handler(event, context):
...
-
-    filename = event["queryStringParameters"]["filename"]
-
-    science_file = ScienceFilepathManager(filename)
-
-    if not science_file.is_valid:
-        return {"statusCode": 400, "body": science_file.error_message}
+    # TODO: Handle other filetypes other than science files
+    #      The current ScienceFilepathManager only accepts filenames, not the full path
+    filename = os.path.basename(path_params)
```
bourque, line 67: Do we anticipate adding support for filepaths before our demo next week? Just wondering so I can make sure the API documentation matches what we support. Currently the API docs specify that filepaths are valid.

bourque (commented):
This looks good and (mostly) matches what we have documented in the API documentation. I just have a few questions.

sds_data_manager/lambda_code/SDSCode/download_api.py
greglucas: This PR? Your docs were in imap_processing though right?

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: This is somewhat confusing to read at first. But the `SciencFilepathManager` only accepts the filenames, our API accepts the full path. This `basename` is here to strip off the front path part for now and why I put the TODO in. So, to summarize we currently accept full paths (or just filenames too), but only full paths to science file naming conventions if that makes sense.

sds_data_manager/lambda_code/SDSCode/download_api.py
bourque: Ah yes, I completely overlooked that this PR is in `sds-data-manager` 😅

sds_data_manager/lambda_code/SDSCode/upload_api.py
bourque: Yes, that makes sense. Thanks for clarifying!

PR #236, opened 2024-01-26
Add event rule to batch starter lambda

tech3371:
# Change Summary
- Added event rule to batch starter lambda
- Found some issue when testing the lambda. Included fix for that.


sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -2,19 +2,20 @@
...
-# Setup the logging
-logging.basicConfig(level=logging.INFO)
-logger = logging.getLogger(__name__)
+from .database import database as db
+from .database import models
 
+# Logger setup
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+logging.basicConfig(stream=sys.stdout, level=logging.DEBUG)
```
greglucas: Why are these set separately? I think we can get rid of the basicConfig and sys import and put this logger to debug.
```suggestion
logger.setLevel(logging.DEBUG)
```

greglucas (approved):

sds_data_manager/lambda_code/SDSCode/batch_starter.py
tech3371: That's how we had in indexer and it's been able to log. I recall batch job logging troublesome. I am testing it on AWS.

laspsandoval (approved):
Nice work.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
tech3371: Ok. That seems to work. It just produced long logs that I haven't seen before. I see INFO logs I usually see. 
Screen Shot 2024-01-29 at 9 29 16 AM

PR #239, opened 2024-01-29
Batch starter fix

tech3371:
# Change Summary
Fix for batch starter lambda timing out when submitting job. These are the fixes
* Added batch endpoint to VPC gateway interface.
* Added permission to the lambda to submit batch and send EventBridge event.


laspsandoval (approved):
Thanks for doing this.

PR #240, opened 2024-01-30
Rename s3 uri

tech3371:
# Change Summary
Rename to match database name:
* s3_uri to file_path
* file_to_create to file_to_create_path

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -221,7 +221,7 @@ def query_upstream_dependencies(session, downstream_dependents, data, s3_bucket)
                 f"imap_{instrument}_{level}_sci_{start_date}_{end_date}_{version}.cdf"
             )
 
-            prepared_data = prepare_data(filename, upstream_dependencies, s3_bucket)
+            prepared_data = prepare_data(filename, upstream_dependencies)
             instruments_to_process.append(
                 {"filename": filename, "prepared_data": prepared_data}
             )
@@ -253,7 +253,7 @@ def load_data(filepath: Path):
     return data
 
 
-def prepare_data(filename, upstream_dependencies, s3_bucket):
+def prepare_data(filename, upstream_dependencies):
     """
     Prepares data for batch job.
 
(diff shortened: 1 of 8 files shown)


sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -347,7 +345,7 @@ def send_lambda_put_event(command_parameters):
         "Source": "imap.lambda",
         "DetailType": "Job Started",
         "Detail": {
-            "file_to_create": "<s3_uri>",
+            "file_to_create_path": "<file_path>",
```
greglucas: I would put the path right after the file throughout here.
```suggestion
            "file_path_to_create": "<file_path>",
```

tests/lambda_endpoints/test_indexer_lambda.py
```diff
@@ -132,6 +132,13 @@ def populate_db(test_engine):
         yield session
 
 
+@pytest.fixture()
+def set_env():
+    # Set the environment variable
+    os.environ["S3_DATA_BUCKET"] = "data-bucket"
+    return ""
```
greglucas: We should set this with setenv:
https://docs.pytest.org/en/latest/reference/reference.html#pytest.MonkeyPatch.setenv
Do you want to put this in a conftest.py file for all tests?

```suggestion
@pytest.fixture(autouse=True)
def set_env(monkeypatch):
    # Set the environment variable
    monkeypatch.setenv("S3_DATA_BUCKET", "test-data-bucket")
```

greglucas (approved):
Looks good to me. Just two minor comments.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
tech3371: I was trying to match what we called in the database. Should I rename all to that?

tests/lambda_endpoints/test_indexer_lambda.py
tech3371: oh ok. Thank you!

sds_data_manager/lambda_code/SDSCode/batch_starter.py
greglucas: My main comment is just that we're changing `file` -> `file_path`, not appending `path` to the variables. I think it sounds better that way at least.

So, IMO, yes I think all of the places that were updated to `file_to_create_path` should be renamed to `file_path_to_create`.

tech3371: changed that now.

PR #241, opened 2024-01-30
Return list of dicts in query API

bourque:
# Change Summary

## Overview
This PR modifies the query API to return a list of dicts instead of a list of tuples, thus making the responses easier to read/parse. I also renamed the module (`queries.py`) to `query_api.py` to be consistent with the other data access API modules. 

## Updated Files

- `sds_data_manager/lambda_code/SDSCode/query_api.py`
   - Renamed from `queries.py`; converted results from list of tuples to list of dicts
- `pyproject.toml`
   - Ignoring line too long error in `test_query_api` -- Caught in a loophole where `ruff` simultaneously wants to format the dict so that each key is on its own line, but also complains that one of those lines is too long. I thought this was the best workaround.
-   `sds_data_manager/stacks/sds_api_manager_stack.py`
   - Updated `query_api` name 
- `tests/lambda_endpoints/test_query_api.py`
   - Updated tests to reflect changes

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/query_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/query_api.py
@@ -87,6 +87,14 @@ def lambda_handler(event, context):
     # Convert the search results (list of tuples) to a list of dicts
     search_results = [result._asdict() for result in search_results]
 
+    # Convert datetimes to string values of format 'YYYYMMDD'
+    # Also remove values that are not needed by users
+    for result in search_results:
+        result["start_date"] = result["start_date"].strftime("%Y%m%d")
+        result["end_date"] = result["end_date"].strftime("%Y%m%d")
+        del result["id"]
+        del result["status_tracking_id"]
+
     logger.info(
         "Found [%s] Query Search Results: %s", len(search_results), str(search_results)
     )
@@ -94,7 +102,7 @@ def lambda_handler(event, context):
     # Format the response
     response = {
         "statusCode": 200,
-        "body": json.dumps(str(search_results)),  # returns a list of tuples
+        "body": json.dumps(search_results),  # returns a list of tuples
         "headers": {
             "Content-Type": "application/json",
             "Access-Control-Allow-Origin": "*",  # Allow CORS
--- a/sds_data_manager/lambda_code/SDSCode/query_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/query_api.py
@@ -1,4 +1,5 @@
-# Standard
+"""Contains the lambda handler for the 'query' data access API"""
+
 import datetime
 import json
 import logging
@@ -81,19 +82,23 @@ def lambda_handler(event, context):
 
     engine = db.get_engine()
     with Session(engine) as session:
-        search_result = session.execute(query).all()
+        search_results = session.execute(query).all()
+
+    # Convert the search results (list of tuples) to a list of dicts
+    search_results = [result._asdict() for result in search_results]
 
     logger.info(
-        "Found [%s] Query Search Results: %s", len(search_result), str(search_result)
+        "Found [%s] Query Search Results: %s", len(search_results), str(search_results)
     )
 
     # Format the response
     response = {
         "statusCode": 200,
-        "body": json.dumps(str(search_result)),  # returns a list of tuples
+        "body": json.dumps(str(search_results)),  # returns a list of tuples
         "headers": {
             "Content-Type": "application/json",
             "Access-Control-Allow-Origin": "*",  # Allow CORS
         },
     }
+
     return response
--- a/tests/lambda_endpoints/test_query_api.py
+++ b/tests/lambda_endpoints/test_query_api.py
@@ -57,24 +57,29 @@ def setup_test_data(test_engine):
 @pytest.fixture()
 def expected_response():
     expected_response = json.dumps(
-        str(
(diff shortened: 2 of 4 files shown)


tests/lambda_endpoints/test_query_api.py
```diff
@@ -59,18 +59,18 @@ def expected_response():
...
-                    "raw",
-                    datetime.datetime(2025, 11, 7, 0, 0),
-                    datetime.datetime(2025, 11, 8, 0, 0),
-                    "v02-01",
-                    "pkts",
-                    1,
-                )
+                {
+                    "id": 1,
+                    "file_path": "test/file/path/imap_hit_l0_raw_20251107_20251108_v02-01.pkts",
```
greglucas: If it is just this link causing issues you can add an inline noqa as well if you find that more helpful so it doesn't ignore other lines within the file.
```suggestion
                    "file_path": "test/file/path/imap_hit_l0_raw_20251107_20251108_v02-01.pkts",  # noqa: E501
```

Could you add a quick test that asserts you can `json.loads(response["body"])` as well?

tests/lambda_endpoints/test_query_api.py
```diff
@@ -59,18 +59,18 @@ def expected_response():
     expected_response = json.dumps(
         str(
```
greglucas: I think this might be an issue with the `str(` call here that I was running into yesterday. Can you remove that and let json dump the objects directly to string?

```
>>> json.dumps([1, 2, 3])
'[1, 2, 3]'
>>> json.dumps(str([1, 2, 3]))
'"[1, 2, 3]"'
```

sds_data_manager/lambda_code/SDSCode/query_api.py
```diff
@@ -81,19 +82,23 @@ def lambda_handler(event, context):
 
     engine = db.get_engine()
     with Session(engine) as session:
-        search_result = session.execute(query).all()
+        search_results = session.execute(query).all()
+
+    # Convert the search results (list of tuples) to a list of dicts
+    search_results = [result._asdict() for result in search_results]
```
greglucas, line 88: Do we want to return everything from the database, or a limited subset of entries? I don't have a preference either way, so totally up to others if we want to trim even more columns/attributes off before returning to users.

greglucas (commented):
I think we want to remove the `str(xyz)` calls before dumping the json objects to string.

sds_data_manager/lambda_code/SDSCode/query_api.py
bourque: I think it would make sense to strip out the `id` and `status_tracking_id`, since those are more for our own bookkeeping instead of for users.

tests/lambda_endpoints/test_query_api.py
bourque: Unfortunately this wouldn't work with `datetime` objects:

```
 TypeError: Object of type datetime is not JSON serializable
```

greglucas: Hm... Can we change those datetime objects to strings somewhere along the path before returning? What do end users need, the start_date and end_date as YYYYMMDD string is all, or do we need the full datetime object printed out?

tests/lambda_endpoints/test_query_api.py
bourque: Thanks! I always forget about that handy `noqa` thing.

tests/lambda_endpoints/test_query_api.py
bourque: Thanks for the suggestion, I got things working this way.

greglucas (approved):
Looks great!

sds_data_manager/lambda_code/SDSCode/query_api.py
```diff
@@ -81,19 +82,31 @@ def lambda_handler(event, context):
...
+
+    # Convert datetimes to string values of format 'YYYYMMDD'
+    # Also remove values that are not needed by users
+    for result in search_results:
+        result["start_date"] = result["start_date"].strftime("%Y%m%d")
+        result["end_date"] = result["end_date"].strftime("%Y%m%d")
+        del result["id"]
+        del result["status_tracking_id"]
```
greglucas, line 96: This should be left as a follow-up PR, but out of curiosity would it make sense to:

1. store the start_date and end_date directly as strings? Do we need the datetime / does it make it handy to store that in the DB rather than just the raw strings? Above we even do conversions from the user input to dates, so there is kind of a double transformation going on here.
```python
            query = query.where(
                models.FileCatalog.start_date
                <= datetime.datetime.strptime(value, "%Y%m%d")
```
2. Have the SQL query not return `id` and `status_tracking_id` somehow?

bourque: Good questions, and agree that they warrant a follow-up discussion/PR.

Do we know if we or the users will ever care about a time component to the start/end date? In other words, will it ever be important to know what time(s) of day a particular file covers? 

If the answer is 'yes', then it makes sense to me to keep some sort of full `datetime` component stored in the database, and then have the APIs figure out how to show that information based on the use case.

bourque:
I'm going to merge this now so that I can prep for API demo for our meeting next week, but other reviewers feel free to comment here if you would like to see any other changes!

PR #242, opened 2024-01-30
updated imap_api

tech3371:
# Change Summary
Update imap_api so that we can test batch job for demo.


tests/docker_images/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Dockerfile that installs imap-data-access and its dependencies
+FROM public.ecr.aws/docker/library/python:3.10-slim
+USER root
+
+# Upgrade pip to the latest version
+RUN pip install --upgrade pip
+
+RUN pip install requests
```
greglucas, line 8: We shouldn't need either of these now.
```suggestion
```

tests/docker_images/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+# TODO: delete this section once imap-data-access is released
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y git
+RUN pip install git+https://github.com/IMAP-Science-Operations-Center/imap-data-access.git
+
+# Create the /mnt/spice directory
+WORKDIR /data
```
greglucas, line 16: I think that you might be able to set the `IMAP_DATA_DIR` environment variable here to make sure you are in `/data`, otherwise my guess is the file might actually be in `/data/data/imap` with the double nested "data".
```suggestion
WORKDIR /data
ENV IMAP_DATA_DIR /data
```

tests/docker_images/imap_api.py
greglucas, line 1: Should we call this `imap_processing_example.py`? And would you want to simulate going from l0 -> l1a rather than updating the descriptor?

greglucas (approved):
Minor comments that can be addressed in follow-up PRs as well. Looks good for now.

tests/docker_images/imap_api.py
tech3371: I was thinking that after the fact. I had chat with Laura briefly about that. Instead of just updating description, we also need to update filename to higher data level to continue like we would in real processing.

tests/docker_images/Dockerfile
tech3371: Even though this code doesn’t use it, the imap-data-access uses it. Will it fail on that? I will check that later too

tests/docker_images/Dockerfile
tech3371: I didn’t notice double data folder. Will check it.

tests/docker_images/Dockerfile
greglucas: `imap-data-access` shouldn't use requests. If it did the `pip install xyz` should pick that up and install the required dependencies as well.

tech3371:
closing this since we have real one working on `imap_processing` repo. I will open PR there.

PR #244, opened 2024-01-31
Prevent database duplicates

sdhoyt:
# Change Summary

## Overview
Added check in upload_api to prevent duplicate entries to the database

## New Dependencies
None

## New Files
None

## Deleted Files
None

## Updated Files
- upload_api
   - added check for an already existing file entry
- sds_api_manager_stack.py
   - added secret name to upload api lambda environment
   - gave upload api lambda needed vpc / security groups
   - gave upload api lambda permission to read rds

## Testing
Deployed and attempted a duplicate file upload

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/upload_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/upload_api.py
@@ -83,7 +83,6 @@ def lambda_handler(event, context):
             models.FileCatalog.file_path == s3_key_path
         )
         result = session.execute(query).first()
-        print(result)
         if result:
             response = {
                 "statusCode": 409,
--- a/sds_data_manager/lambda_code/SDSCode/upload_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/upload_api.py
@@ -85,7 +85,7 @@ def lambda_handler(event, context):
             models.FileCatalog.file_path == s3_key_path
         )
         result = session.execute(query).first()
-        # return a 409 response if a existing file is found
+        # return a 409 response if an existing file is found
         if result:
             response = {
                 "statusCode": 409,
--- a/sds_data_manager/lambda_code/SDSCode/upload_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/upload_api.py
@@ -4,6 +4,11 @@ import os
 
 import boto3
 from SDSCode.path_helper import InvalidScienceFileError, ScienceFilepathManager
+from sqlalchemy import select
+from sqlalchemy.orm import Session
+
+from .database import database as db
+from .database import models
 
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -73,6 +78,25 @@ def lambda_handler(event, context):
 
     s3_key_path = science_file.construct_upload_path()
 
+    # Check for already existing file in the database
+    with Session(db.get_engine()) as session:
+        # query and check for a matching file path
+        query = select(models.FileCatalog.__table__).where(
+            models.FileCatalog.file_path == s3_key_path
+        )
```
greglucas, line 86: Should this be using an exists query?
https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.exists

I'm out of my element on this, so since this works for now we can try and optimize the queries later.

I also thought that the unique constraint might automatically handle these for us too :-/ Maybe @bourque would know more.

Edit: I think I see now. The unique portion wouldn't come in until the indexer hits in the subsequent call, and we want to head it off right away.

greglucas (approved):

sds_data_manager/lambda_code/SDSCode/upload_api.py
sdhoyt: It seemed like in version 2 of sqlalchemy, they are moving more toward the select().where() syntax and away from the .query() syntax in version 1 (the documentation linked in version 1). It looks like you can append an exists() to where() (`select().where().exists()`) in version 2 though, so I'll give that a shot

sdhoyt: @bourque Have you used the exists() construct before? It's not working how I expected it to. I thought from the documentation I could do:

```python
query = exists().where(models.FileCatalog.file_path == s3_key_path)
session.execute(query)
```

but it's creating an exists construct instead of sql that I can execute

https://docs.sqlalchemy.org/en/20/core/selectable.html#sqlalchemy.sql.expression.exists

bourque: @sdhoyt Unfortunately I haven't used `exists()` before, I think your guess is as good as mine.

RE: unique constraints -- Correct, I think they are only applied when the insertion of data is executed.

bourque (approved):
Thanks for adding in that check!

sds_data_manager/lambda_code/SDSCode/upload_api.py
sdhoyt: Ok sounds good. I'll merge this in for now and look into how that works for later. Thanks for the reviews!

PR #245, opened 2024-01-31
Fixes from Integration tests

tech3371:
# Change Summary

fixes some integration testings

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -294,12 +294,27 @@ def prepare_data(filename, upstream_dependencies):
 
     # Prepare the final command
     # Pre-construct parts of the string
-    instrument_part = f"--instrument {instrument}"
-    level_part = f"--level {level}"
-    file_path = f"--file_path '{s3_base_path}{filename}'"
-    dependency_part = f"--dependency {upstream_dependencies}"
-
-    prepared_data = f"{instrument_part} {level_part} {file_path} {dependency_part}"
+    # NOTE: Batch job expects command like this:
+    # "Command": [
+    #     "--instrument",
+    #     "swe",
+    #     "--level",
+    #     "l1b",
+    #     "--file_path",
+    #     "imap/swe/l1b/2023/09/imap_swe_l1b_lveng-hk_20230927_20230927_v01-00.cdf",
+    #     "--dependency",
+    #     "[{'instrument': 'swe', 'level': 'l0', 'version': 'v00-01'}]"
+    #   ]
+    prepared_data = [
+        "--instrument",
+        instrument,
+        "--level",
+        level,
+        "--file_path",
+        s3_base_path + filename,
+        "--dependency",
+        f"{upstream_dependencies}",
+    ]
 
     return prepared_data
 
@@ -359,10 +374,17 @@ def send_lambda_put_event(command_parameters):
     command_parameters : str
         IMAP cli command input parameters.
         Example of input:
-            "--instrument codice
-            --level l1a
-            --file_path '<s3-filepath>'
-            --dependency 'list of dict'"
+            [
+            "--instrument",
+            "hit",
+            "--level",
+            "l1a",
+            "--file_path",
+            ("imap/hit/l1a/2024/01/"
+            "imap_hit_l1a_sci_20240101_20240102_v00-01.cdf"),
+            "--dependency",
+            "[{'instrument': 'hit', 'level': 'l0', 'version': 'v00-01'}]",
+        ]
     Returns
     -------
     dict
@@ -371,9 +393,8 @@ def send_lambda_put_event(command_parameters):
     event_client = boto3.client("events")
 
     # Get event inputs ready
-    command = command_parameters.split("--")
-    file_path = command[3].replace("file_path '", "").replace("' ", "")
-    dependency = command[-1].replace("dependency ", "")
+    file_path = command_parameters[5]
+    dependency = command_parameters[7]
 
     # Create event["detail"] information
     detail = {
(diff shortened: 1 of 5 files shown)


sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -450,16 +471,24 @@ def lambda_handler(event: dict, context):
...
-            batch_client.submit_job(
-                jobName=filename,
+        for downstream_data in downstream_instruments_to_process:
+            filename = downstream_data["filename"]
+
+            command = downstream_data["prepared_data"]
+            logger.info(f"Submitting job with this command - {command}")
+            # NOTE: The batch job name should contain only
+            # alphanumeric characters and hyphens.
+            logger.info(f"Job name: {instrument}-{level}-job")
```
greglucas: Does the job name need to be unique with an extra hash or anything added?

You might want to bring that string out and store it for logging here and setting it later.
```python
job_name = f"{instrument}-{level}-job"
logger.info("Job name: %s", job_name)
```

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -450,16 +471,24 @@ def lambda_handler(event: dict, context):
...
+            response = batch_client.submit_job(
+                jobName=f"{instrument}-{level}-job",
                 jobQueue=job_queue,
                 jobDefinition=job_definition,
                 containerOverrides={
-                    "command": [instrument["prepared_data"]],
+                    "command": downstream_data["prepared_data"],
```
greglucas, line 492: No need to update here, but it is still a bit confusing to me the downstream/upstream and causes me to think every time. It almost seems like we could possibly only use one (upstream), and then call this `new_file` or something instead of downstream to really separate out the confusion with both streams.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -295,18 +297,10 @@ def batch_event_handler(event):
...
-    #   'level l1a ',
-    #   "file_path 'path/filename'",
-    #   "dependency [{'instrument': 'hit', 'level': 'l0', 'version': 'v00-01'}]"
-    #  ]
-    command = event["detail"]["container"]["command"][0].split("--")
+    command = event["detail"]["container"]["command"]
 
     # Get event inputs ready
-    file_path = command[3].replace("file_path '", "").replace("'", "")
+    file_path = command[5]
```
greglucas, line 303: Would it be possible to add this to the event information as another field that could be looked up rather than needing to parse a command out?

sds_data_manager/stacks/indexer_lambda_stack.py
```diff
@@ -72,15 +72,15 @@ def __init__(
...
+                "S3_DATA_BUCKET": data_bucket.bucket_name,
                 "SECRET_NAME": db_secret_name,
             },
         )
 
         # events:PutEvents
         put_event_policy = iam.PolicyStatement(
             effect=iam.Effect.ALLOW,
-            actions=["events:PutEvents"],
+            actions=["events:PutEvents", "s3:*"],
```
greglucas, line 85: Why is the s3:* needed here? Might be worth a comment.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -227,12 +230,16 @@ def test_lambda_handler(test_file_catalog_simulation, batch_client, sts_client):
...
-        "imap_swapi_l1_sci-1m_20230724_20230724_v02-01.cdf' "
-        "--dependency [{'instrument': 'hit', 'level': 'l0', 'version': 'v00-01'}]"
-    )
+    input_command = [
+        "--instrument",
+        "hit",
+        "--level",
+        "l1a",
+        "--file_path",
+        ("imap/hit/l1a/2024/01/" "imap_hit_l1a_sci_20240101_20240102_v00-01.cdf"),
```
```suggestion
        "imap/hit/l1a/2024/01/imap_hit_l1a_sci_20240101_20240102_v00-01.cdf",
```

tests/lambda_endpoints/test_indexer_lambda.py
```diff
@@ -185,13 +185,17 @@ def test_batch_job_event(test_engine, write_to_s3, events_client, set_env):
                     "123456789012.dkr.ecr.us-west-2.amazonaws.com/" "codice-repo:latest"
                 ),
                 "command": [
```
greglucas, line 187: I'm not sure I follow this command and what it is doing.
I read it as: process the SWE instrument at l1b, but the filepath is for SWAPI l1, and the dependency is SWE l0.

greglucas (commented):
It looks like you're pulling things out of the event command to use in other places. I think it might make sense to put some of those directly into the event as named items because I can imagine the command is going to change over time.

laspsandoval (approved):
Thanks for the changes.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: we can't do that because this event is sent by AWS and we can't change that. But if we want, we can send event from within batch job but I think we decided not to do that since instrument may not do same in their L3 code.

tests/lambda_endpoints/test_indexer_lambda.py
tech3371: good catch. I was copy pasting lot yesterday with fix. It was remain from that. will fix it.

sds_data_manager/lambda_code/SDSCode/indexer.py
greglucas: Ahh, I see. That makes sense to me then that we need to parse this. We can try and think about other options later.

greglucas (approved):

tech3371:
I am going to merge this over codecov percentage being low because we are going to be making changes in the near future.

PR #247, opened 2024-02-02
MNT: Standardize logging across the package

greglucas:
# Change Summary

We should get an explicit logger in the module and set that log level rather than the global log level. If we set the global logger's level we get a lot of external boto debug logs that aren't of interest and pollute our logs.

* If we are logging from a caught exception log that at the error level
* Set INFO log level rather than debug. Debug should be for when we are actively debugging or want to go look into things later.


laspsandoval (approved):
That makes sense. Thanks for changing it.

bourque (approved):
LGTM

PR #253, opened 2024-02-13
Indexer lambda update post demo

### LINKED ISSUE #249: Indexer Lambda updates post February demo
indexer.py updates:
* Update s3 event rule to listen for all imap/ folder. Right now, we have the rule to allow only l0 data to go through and because of that, it caused some issues. If our l0 data file contained multiple apid(s), our processing code could produce more than one files which won't know known by batch starter lambda.
* Decouple StatusTracking and FileCatalog table now. Status table will track these information, instrument, level, upstream_file(?), and batch job informations. File catalog table will only store information about files that were processed and uploaded to s3 bucket.

tech3371:
# Change Summary
Changes listed in this issue - #249 

## Updated Files
   - moved ingestion date to catalog table
   - added few more values to status tracking table
   - updated indexer to write to file catalog when s3 event occurs
   - updated indexer to write to status table when lambda or batch event happens
   - refactored and remove many try/except lines and moved checks up higher like suggested in the past
- sds_data_manager/lambda_code/SDSCode/query_api.py
   - updated to reflect database changes
   - updated s3 rule to allow all science files
## Testing
- tests/lambda_endpoints/test_query_api.py

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -10,7 +10,6 @@ from sqlalchemy import (
     Boolean,
     Column,
     DateTime,
-    ForeignKey,
     Identity,
     Integer,
     String,
@@ -105,17 +104,19 @@ class StatusTracking(Base):
     __table_args__ = (
         UniqueConstraint(
             "id",
-            "file_path_to_create",
+            "input_data_file_path",
             "status",
             name="status_tracking_uc",
         ),
     )
 
     id = Column(Integer, Identity(start=1, increment=1), primary_key=True)
-    file_path_to_create = Column(String, nullable=False)
+    input_data_file_path = Column(String, nullable=False)
     status = Column(STATUSES, nullable=False)
     job_definition = Column(String)
-    ingestion_date = Column(DateTime)
+    job_log_stream_id = Column(String)
+    container_image = Column(String)
+    container_command = Column(String)
 
 
 class FileCatalog(Base):
@@ -144,9 +145,7 @@ class FileCatalog(Base):
     end_date = Column(DateTime, nullable=False)
     version = Column(String(8), nullable=False)
     extension = Column(EXTENSIONS, nullable=False)
-    status_tracking_id = Column(
-        Integer, ForeignKey("status_tracking.id"), nullable=False
-    )
+    ingestion_date = Column(DateTime)
 
 
 class PreProcessingDependency(Base):
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -113,6 +113,10 @@ class StatusTracking(Base):
     id = Column(Integer, Identity(start=1, increment=1), primary_key=True)
     input_data_file_path = Column(String, nullable=False)
     status = Column(STATUSES, nullable=False)
+    # TODO:
+    # Didn't make it required field yet. Revisit this
+    instrument = Column(INSTRUMENTS)
+    data_level = Column(DATA_LEVELS)
     job_definition = Column(String)
     job_log_stream_id = Column(String)
     container_image = Column(String)
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -115,6 +115,7 @@ class StatusTracking(Base):
     status = Column(STATUSES, nullable=False)
     # TODO:
     # Didn't make it required field yet. Revisit this
+    # post discussion
     instrument = Column(INSTRUMENTS)
     data_level = Column(DATA_LEVELS)
     job_definition = Column(String)
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -104,14 +104,14 @@ class StatusTracking(Base):
     __table_args__ = (
         UniqueConstraint(
             "id",
-            "input_data_file_path",
+            "file_path_to_create",
             "status",
             name="status_tracking_uc",
         ),
     )
 
     id = Column(Integer, Identity(start=1, increment=1), primary_key=True)
-    input_data_file_path = Column(String, nullable=False)
+    file_path_to_create = Column(String, nullable=False)
     status = Column(STATUSES, nullable=False)
     # TODO:
     # Didn't make it required field yet. Revisit this
--- a/sds_data_manager/lambda_code/SDSCode/indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/indexer.py
@@ -1,4 +1,3 @@
-import datetime
 import json
 import logging
 import os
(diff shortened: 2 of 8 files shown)


sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -186,72 +185,36 @@ def s3_event_handler(event):
...
-    try:
-        science_file = ScienceFilepathManager(filename)
-    except InvalidScienceFileError as e:
-        logger.error(str(e))
-        return http_response(status_code=400, body=str(e))
-
+    # Check if the file is a valid science file or not
+    # TODO: change these lines once filename validator
+    # is implemented on sds-data-access repo and released
+    science_file = ScienceFilepathManager(filename)
```
maxinelasp, line 191: Is now available in imap-data-access v0.3.0! If you bump the required version in pyproject.toml you can use that class instead.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -403,28 +335,42 @@ def custom_event_handler(event):
...
+event_handlers = {
+    "aws.s3": s3_event_handler,
+    "aws.batch": batch_event_handler,
+    "imap.lambda": custom_event_handler,
+}
+
+
+def handle_event(event, handler):
+    """Common event handling logic."""
+    try:
```
maxinelasp, line 365: Nice, I like this change

sds_data_manager/stacks/indexer_lambda_stack.py
```diff
@@ -128,7 +128,7 @@ def __init__(
                     "bucket": {"name": [data_bucket.bucket_name]},
                     "object": {
                         "key": [
-                            {"prefix": f"imap/{instrument}/l0/"}
+                            {"prefix": f"imap/{instrument}/"}
```
maxinelasp: Yay nice!

maxinelasp (commented):
Definitely an improvement, I'd suggest removing `ScienceFIlepathManager`, but it's cool if you want to leave that for another PR instead

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -186,72 +185,36 @@ def s3_event_handler(event):
...
-        # Query to get foreign key id for catalog table
-        query = select(models.StatusTracking.__table__).where(
-            models.StatusTracking.file_path_to_create == s3_filepath
-        )
+    ingestion_date_object = get_file_creation_date(s3_filepath)
 
-        status_tracking = session.execute(query).first()
+    metadata_params["ingestion_date"] = ingestion_date_object
+    update_file_catalog_table(metadata_params)
+    logger.info("Wrote data to file catalog table")
```
greglucas, line 212: `info()` calls should probably contain some more information about what is going on, like the filepath, ingestion date you created or something like that if you wanted to know that. This seems more like a debug level that we don't really care about in normal logs, but may want to see how it progresses if we increase the logging level.
```suggestion
    logger.debug("Wrote data to file catalog table")
```

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -186,72 +185,36 @@ def s3_event_handler(event):
...
-    except Exception as e:
-        logger.error(str(e))
-        return http_response(status_code=400, body=str(e))
-
-    # Send L0 event from this lambda for Batch starter
+    # Send event from this lambda for Batch starter
     # lambda
     send_event_from_indexer(filename)
-    return http_response(status_code=200, body="Success")
+    logger.info("S3 event handler complete")
```
```suggestion
    logger.debug("S3 event handler complete")
```

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -297,77 +260,46 @@ def batch_event_handler(event):
...
-                # update two fields with updated information
-                result.status = models.Status.FAILED
-                result.job_definition = event["detail"]["jobDefinition"]
-                session.commit()
-        except Exception as e:
-            logger.error(str(e))
-            return http_response(status_code=400, body=str(e))
+        if result is None:
+            logger.info(f"No record found for {file_path_to_create}")
+            logger.info(f"Creating new record for {file_path_to_create}")
```
```suggestion
            logger.info(f"No existing record found, creating new record for {file_path_to_create}")
```

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -297,77 +260,46 @@ def batch_event_handler(event):
...
+        )
 
-                # update two fields with updated information
-                result.status = models.Status.FAILED
-                result.job_definition = event["detail"]["jobDefinition"]
-                session.commit()
-        except Exception as e:
-            logger.error(str(e))
-            return http_response(status_code=400, body=str(e))
+        if result is None:
```
greglucas, line 283: If you try to insert into the db, does it fail and return the object that is already there because it wasn't unique? I'm wondering if you could try and insert the object into the database right away and then just update whatever object comes out of it rather than having to query() to find the object?

sds_data_manager/stacks/indexer_lambda_stack.py
greglucas: The rule names above this should be updated since it isn't just l0 anymore. Also, it doesn't appear that `instrument` is used for anything inside here. Would it make sense to just have a single prefix that is listening on `imap/`

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -144,9 +150,7 @@ class FileCatalog(Base):
     end_date = Column(DateTime, nullable=False)
     version = Column(String(8), nullable=False)
     extension = Column(EXTENSIONS, nullable=False)
-    status_tracking_id = Column(
```
greglucas, line 147: I guess this means that there is no link between the FileCatalog and StatusTracking anymore? I _think_ that makes sense, but just wanted to verify we don't need that for any other reason.

greglucas (approved):
Nice consolidation! Only minor comments that can be ignored/pushed off for future work too.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: Yes. Not that I can think of.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: Thank you! I will do that in follow-up PR.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: yes. I tried to avoid query and it will fail to update.

sds_data_manager/stacks/indexer_lambda_stack.py
tech3371: we could make it `imap/` but I was wondering if we will have calibrations or spice inside `imap/` too.

greglucas: Good point! Calibrations might also be within the instrument level though, so would be caught by this. I wonder if that means we need a lot more prefixes to distinguish between all of that, or if we should do a general prefix `imap/` and then do the sorting/filtering within the lambda that gets triggered...

Fine to leave it as-is for now.

tech3371: true. we have already captured `TODO` in the indexer lambda's `s3_event_handler()`. I could change this rule to be `imap/` and then we can handle that in the indexer lambda. That sounds good to me.

PR #255, opened 2024-02-15
update to use imap-data-access file validator

### LINKED ISSUE #251: Filename validator update
path_helper.py updates:
- improve init functions of ScienceFilepathManager.py. Maxine has already started this work. moving this function to sds-data-access repo or package.

indexer.py and upload_api.py updates:
- Use and verify that indexer and upload API lambda is able to use sds-data-access's filepath validator.

tech3371:
# Change Summary
main changed to use imap-data-access file validator now and deleted old path helper.


sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -202,7 +202,7 @@ def s3_event_handler(event):
     #     "extension": self.extension,
     #     "ingestion_date": date_object,
     # }
-    metadata_params = science_file.get_file_metadata_params()
+    metadata_params = science_file.extract_filename_components(filename)
```
greglucas: I think this is a static method and should take as input the science_file you created?
```suggestion
    metadata_params = SciencFilePath.extract_filename_components(science_file)
```

Should we change "metadata_params" to "file_params" or something more descriptive? I'm not sure it is metadata. Can save this for a follow-up small PR.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -337,7 +337,8 @@ def custom_event_handler(event):
     filename = os.path.basename(file_path_to_create)
     logger.info(f"Attempting to insert {filename} into database")
 
-    ScienceFilepathManager(filename)
+    # Check if the file is a valid science file or not
+    ScienceFilePath(filename)
```
greglucas: Should we call the validate routine directly, or use the output of the ScienceFilePath to construct the full `file_path_to_create` below?

I think you might be able to pass in the full filepath to the class as well and not have to call `basename` above.

greglucas (approved):

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: we still only take filename with latest changes. Seems like we can take a Path object but I don't know if we get filename from path or not. That's why I didn't change that.

But I agree. I can use construct file path to get `file_path_to_create`

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: I can change that to be `file_params`.

pyproject.toml
```diff
@@ -68,6 +68,7 @@ requests-aws4auth = "^1.2.3"
 psycopg2-binary = "^2.9.9"
 urllib3 = "<2.0.0"
 SQLAlchemy = "<=3.0.0"
+imap-data-access = ">=0.3.1"
 
 # Used for installing with pip
 [tool.poetry.extras]
```
maxinelasp, line 74: Is this intentionally deleted?

sds_data_manager/lambda_code/SDSCode/indexer.py
maxinelasp: It should actually probably be `ScienceFilePath.extract_filename_components(filename)` and you remove the `science_file` variable entirely. Since you're just extracting the filename components, you don't need to use the rest of the class.

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -67,11 +67,11 @@ def lambda_handler(event, context):
             ),
         }
     # TODO: Handle other filetypes other than science files
-    #      The current ScienceFilepathManager only accepts filenames, not the full path
+    #      The current ScienceFilePath only accepts filenames, not the full path
     filename = os.path.basename(path_params)
```
maxinelasp, line 71: Would it be useful to let ScienceFilePath take a filename or a full path? That might be worth adding at some point, since the imap-data-access CLI uses the same pattern of accepting either a filename or a full path...

maxinelasp (approved):
Looks good to me, I think my only suggested change is to use only the static method `ScienceFilePath.extract_components(filename)` instead of making an instance. Besides that, looks like a pretty simple replacement, which is what I was hoping for!

pyproject.toml
tech3371: It's not deleted. It's highlighted in weird way.

sds_data_manager/lambda_code/SDSCode/indexer.py
tech3371: I thought of doing that but was hesitant. will change to do that.

sds_data_manager/lambda_code/SDSCode/upload_api.py
tech3371: yeah, I don't mind either way.

pyproject.toml
maxinelasp: Oh, odd 🤷

PR #261, opened 2024-02-27
added few instruments to list

tech3371:
# Change Summary
very small PR just to add some instruments to the list


bourque (approved):
LGTM. I agree we should import the `instrument_list` from `imap-data-acess`, but you marked it as `TODO` so I think this could be merged as it stands.

tech3371:
> LGTM. I agree we should import the `instrument_list` from `imap-data-acess`, but you marked it as `TODO` so I think this could be merged as it stands.

Yes. I wanted to do that now but since instrument names are going to get combine with new filename convention, we will need to update `imap-data-access` and release those new changes before I can update this.

PR #262, opened 2024-03-12
Infrastructure changes to match filename convention updates

tech3371:
# Change Summary
Changes to these lambda to use new filename convention. Main changes was removing end_data and captured one TODO:
1. Batch starter Lambda
2. Indexer Lambda
3. Query API Lambda

Other major updates:
1. Update imap-data-access update
2. Database update
3. Replace instrument_list to use valid instrument list from imap-data-access in stackbuilder.py
4. Commented out ECR push permission because it was causing error with limits

## Testing
Update tests for above changes

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -85,13 +85,20 @@ def query_instrument(session, upstream_dependency, start_date, version):
     instrument = upstream_dependency["instrument"]
     data_level = upstream_dependency["data_level"]
 
+    # TODO: narrow down the query using end_date.
+    # This will give ability to query range of time.
+    # Eg. when we are query 3 months of data to create
+    # 3 months map, end_date will help narrow search.
+    # When we have the end_date, we can query
+    # using table.start_date >= start_date and
+    # table.end_date <= end_date.
     record = (
         session.query(models.FileCatalog)
         .filter(
             models.FileCatalog.instrument == instrument,
             models.FileCatalog.data_level == data_level,
             models.FileCatalog.version == version,
-            models.FileCatalog.start_date >= datetime.strptime(start_date, "%Y%m%d"),
+            models.FileCatalog.start_date == datetime.strptime(start_date, "%Y%m%d"),
         )
         .first()
     )
@@ -271,7 +278,7 @@ def prepare_data(instrument, data_level, start_date, version, upstream_dependenc
     #             'start_date': '20231212',
     #             'version': 'v001',
     #         }]""",
-    #    "--repointing", "None | int",
+    #    "--repointing", 1,
     #     "--upload-to-sdc"
     # ]
     prepared_data = [
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -62,8 +62,8 @@ def get_dependency(instrument, data_level, descriptor, direction, relationship):
     return dependency
 
 
-def query_instrument(session, upstream_dependency, start_date, end_date, version):
-    """Append information to downstream dependents.
+def query_instrument(session, upstream_dependency, start_date, version):
+    """Query to database to get the first FileCatalog record.
 
     Parameters
     ----------
@@ -73,8 +73,6 @@ def query_instrument(session, upstream_dependency, start_date, end_date, version
         Dictionary of upstream dependency.
     start_date : str
         Start date of the event data.
-    end_date : str
-        End date of the event data.
     version : str
         Version of the event data.
 
@@ -94,7 +92,6 @@ def query_instrument(session, upstream_dependency, start_date, end_date, version
             models.FileCatalog.data_level == data_level,
             models.FileCatalog.version == version,
             models.FileCatalog.start_date >= datetime.strptime(start_date, "%Y%m%d"),
-            models.FileCatalog.end_date <= datetime.strptime(end_date, "%Y%m%d"),
         )
         .first()
     )
@@ -132,11 +129,9 @@ def query_downstream_dependencies(session, filename_components):
         dependent["version"] = filename_components["version"]  # placeholder
 
         # TODO: add repointing table query if dependent is ENA or GLOWS
-        #  Use start_date and end_date to query repointing table.
-        # Use pointing start_time and end_time in place of start_date and end_date.
+        # Use start_date to query repointing table.
         # Add pointing number to dependent.
         dependent["start_date"] = filename_components["start_date"]
-        dependent["end_date"] = filename_components["end_date"]
 
     return downstream_dependents
 
@@ -165,10 +160,9 @@ def query_upstream_dependencies(session, downstream_dependents):
     for dependent in downstream_dependents:
         instrument = dependent["instrument"]
         data_level = dependent["data_level"]
-        version = dependent["version"]
-        start_date = dependent["start_date"]
-        end_date = dependent["end_date"]
         descriptor = dependent["descriptor"]
+        start_date = dependent["start_date"]
+        version = dependent["version"]
 
         # For each downstream dependent, find its upstream dependencies
         upstream_dependencies = get_dependency(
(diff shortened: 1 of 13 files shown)


sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -66,6 +66,12 @@
 # instrument's processing
 DEPENDENCY_DIRECTIONS = SqlEnum("UPSTREAM", "DOWNSTREAM", name="dependency_direction")
 
+# 'hard' dependency means that the dependent instrument's processing cannot
+# proceed without the primary instrument's data. 'soft' dependency means that
+# the dependent instrument's processing can proceed without the primary
+# instrument's data. It's nice to have but not necessary.
```
bourque, line 72: Thanks for adding this! Very helpful

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -270,25 +255,23 @@ def prepare_data(
...
-    #     "--end-date", "20231212",
     #     "--version", "v00-01",
     #     "--dependency", """[
     #         {
     #             'instrument': 'swe',
     #             'data_level': 'l0',
     #             'descriptor': 'lveng-hk',
     #             'start_date': '20231212',
-    #             'end_date': '20231212',
     #             'version': 'v01-00',
```
bourque: Perhaps outside the scope of this PR, but the version format examples here now need to be updated.

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -225,14 +224,12 @@ def batch_event_handler(event):
                     "--instrument", "swapi",
                     "--level", "l1",
                     "--start-date", "20230724",
-                    "--end-date", "20230724",
                     "--version", "v02-01",
```
bourque: some more outdated version examples in this file

sds_data_manager/utils/stackbuilder.py
```diff
@@ -120,13 +121,11 @@ def build_sds(
...
 
-    # TODO: import this from imap-data-access
-    instrument_list = ["codice", "swe", "mag"]  # etc
-
     lambda_code_directory = Path(__file__).parent.parent / "lambda_code"
     lambda_code_directory_str = str(lambda_code_directory.resolve())
 
-    for instrument in instrument_list:
+    # This valid instrument list is from imap-data-access package
+    for instrument in imap_data_access.VALID_INSTRUMENTS:
```
bourque, line 128: Nice!

tests/lambda_endpoints/test_indexer_lambda.py
```diff
@@ -27,12 +27,12 @@ def write_to_s3(s3_client):
     # write file to s3
     s3_client.put_object(
         Bucket="test-data-bucket",
-        Key=("imap/swapi/l1/2023/01/imap_swapi_l1_sci-1m_20230724_20230724_v02-01.cdf"),
+        Key=("imap/swapi/l1/2023/01/imap_swapi_l1_sci-1m_20230724_v001.cdf"),
```
bourque:
```suggestion
        Key=("imap/swapi/l1/2023/01/imap_swapi_l1_sci-1min_20230724_v001.cdf"),
```

Did we decide to be more explicit with our time descriptors?

tests/lambda_endpoints/test_indexer_lambda.py
```diff
@@ -253,7 +247,7 @@ def test_unknown_event(test_engine):
 
 def test_send_lambda_put_event(events_client):
     """Test the ``send_event_from_indexer`` function."""
-    filename = "imap_swapi_l1_sci-1m_20230724_20230724_v02-01.cdf"
+    filename = "imap_swapi_l1_sci-1m_20230724_v001.cdf"
```
bourque:
```suggestion
    filename = "imap_swapi_l1_sci-1min_20230724_v001.cdf"
```

bourque (approved):
Thanks for making these updates! I just have a few non-blocking suggestions.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
tech3371: Yes. Good catch! Will update comments.

tests/lambda_endpoints/test_indexer_lambda.py
tech3371: I didn't hear any push back from SWAPI when I suggested this to them. I updated it.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -94,7 +92,6 @@ def query_instrument(session, upstream_dependency, start_date, end_date, version
             models.FileCatalog.data_level == data_level,
             models.FileCatalog.version == version,
             models.FileCatalog.start_date >= datetime.strptime(start_date, "%Y%m%d"),
```
greglucas: Do we need to keep the end_date and put that into the start_date query here.
```suggestion
            models.FileCatalog.start_date >= datetime.strptime(start_date, "%Y%m%d"),
            models.FileCatalog.start_date <= datetime.strptime(end_date, "%Y%m%d"),
```
I'd be somewhat worried about a reprocessing finding lots of matches for files greater than some date without a restriction criteria on the upper end.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -270,25 +255,23 @@ def prepare_data(
...
     #         {
     #             'instrument': 'mag',
     #             'data_level': 'l0',
     #             'descriptor': 'lveng-hk',
     #             'start_date': '20231212',
-    #             'end_date': '20231212',
-    #             'version': 'v00-01',
+    #             'version': 'v001',
     #         }]""",
+    #    "--repointing", "None | int",
```
greglucas: The other ones are explicit examples, so maybe make this one explicit too? I'm not sure `--repointing None` would actually work...
```suggestion
    #    "--repointing", 1,
```

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -209,8 +213,11 @@ class Version(Base):
     id = Column(Integer, Identity(start=1, increment=1), primary_key=True)
     instrument = Column(INSTRUMENTS, nullable=False)
     data_level = Column(DATA_LEVELS, nullable=False)
+    # TODO: determine cap for strings based on what software version
```
greglucas, line 216: I think we would be removing software version now? All of that is wrapped up in the single version scheme now, whereas before it was `software-data` and we could parse them out.

sds_data_manager/lambda_code/SDSCode/query_api.py
```diff
@@ -60,6 +66,8 @@ def lambda_handler(event, context):
                     "Access-Control-Allow-Origin": "*",  # Allow CORS
                 },
             }
+            logger.debug(f"valid_parameters: {valid_parameters}")
+            logger.debug(f"Invalid query parameter: {param}")
```
greglucas:
```suggestion
            logger.debug(f"Received an invalid query parameter [{param}], valid options are: {valid_parameters}")
```

sds_data_manager/lambda_code/SDSCode/batch_starter.py
tech3371: Yes. I like the point of narrowing down time but we don't have end_date now in the filename. should we construct end_date to be start_date + 24 hour?

greglucas: Maybe for now that is good, or just `start_date == ` instead? So we only get the start_date we explicitly want?

We will have to think later about what happens when I need all 3-months worth of daily files to create my 3-month map.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: can we remove this in future PR when we incorporate version changes? This will remind us to decide on we should track software version.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
tech3371: ok. I will add TODO there and change comparison to `start_date == `

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: Yes, we can certainly leave it for later. My concern is only that it says `nullable=False`, but I don't see how we'd be getting it now... Are we not using this table, so it just doesn't matter right now?

tech3371: yes. We don't use it yet. It's there as place holder for future use.

greglucas: OK, makes sense then!

greglucas (approved):

PR #264, opened 2024-04-13
ENH: Enable SPICE file uploads to the SDC

greglucas:
# Change Summary

## Overview
This takes in any SPICE suffix in the filename, tests to see if there is already a file with that name (returns 409 if so), and then adds that file to the flat namespace `imap/spice/filename`. There is a bit of additional refactoring of the upload code to re-use the pre-signed url and responses in the SPICE section.

Tested by creating a `.ck` file suffix and uploading that, which put the file in s3 in the place expected. Ran it again and got a 409 conflict response.
```bash
touch /tmp/test-spice-file.ck
IMAP_DATA_DIR=/private/tmp imap-data-access upload /tmp/test-spice-file.ck
```

This brought up the fact that we require SPICE files to also be within the `IMAP_DATA_DIR` (from within `imap-data-access`), do we want to enforce that for uploads too? Add an `IMAP_SPICE_DIR`, or just allow the wild west for file uploads?
This can be fixed in follow-up PRs, but this one got me to think about it.


sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -69,14 +85,48 @@ def lambda_handler(event, context):
...
+                "No filename given for the upload. Please provide a filename "
                 "in the path. Eg. /upload/path/to/file/filename.pkts"
             ),
         }
     # TODO: Handle other filetypes other than science files
     #      The current ScienceFilePath only accepts filenames, not the full path
     filename = os.path.basename(path_params)
+
+    # Supported SPICE extensions, must be a tuple for the endswith() check
+    spice_extensions = (".tsc", ".bsp", ".tf", ".tls", ".ck", ".tpc")
```
tmplummer: I would think that these extensions are/will be defined elsewhere. It would be good to centralize on a single definition source.

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -16,29 +17,44 @@
...
         ExpiresIn=3600,
     )
 
-    return url
+    if url is None:
+        return {
+            "statusCode": 400,
+            "body": json.dumps(
+                "A pre-signed URL could not be generated. Please ensure that the "
+                "file name matches mission file naming conventions."
```
tech3371: Do you think it will help to import filename convention from `imap_data_access` and send that in this message? I will make a suggestion edit

tech3371: Please feel free to reword this
```suggestion
                "A pre-signed URL could not be generated. Please ensure that the "
                f"file name matches mission file naming convention, {FILENAME_CONVENTION} or SPICE data file name has these extensions - {some_list_var}"
```

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -69,14 +85,48 @@ def lambda_handler(event, context):
...
+        return {
+            "statusCode": 409,
+            "body": json.dumps(f"{filepath} already exists."),
+            "headers": {
+                "Content-Type": "application/json",
+                "Access-Control-Allow-Origin": "*",
+            },
+        }
+    except botocore.exceptions.ClientError:
+        pass
```
tech3371: Did you mean to add some log here?

tech3371 (commented):
couple minor questions. Otherwise it looks good to me! Thank you for doing this!

tmplummer (commented):
I don't think that this PR needs to be held up waiting for feedback and a decision on the SDC organization of SPICE kernels.

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: Our canonical data area for defining these things is: https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/476c3b96e6d4aacc8fed04624ea31774bc2f996c/imap_data_access/__init__.py

We can update that package later once they are clarified and then bump the required version here and replace this. Or are you confident enough in these right now that we should just add them into the other package right now?

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: No, this means a file wasn't found, but nothing else we should do. I added a quick comment for this pass that we just keep continuing.

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: 👍 I added the filename convention and that it should be a valid SPICE file.

sds_data_manager/lambda_code/SDSCode/upload_api.py
tmplummer: From the NAIF documentation, this is the complete list of kernel types with recommended (I would argue _required_) extensions.

>  NAIF recommends kernel names use only lower case letters. NAIF further recommends one follows the conventions established for kernel name extensions, shown below.

            .bc    binary CK
            .bds   binary DSK
            .bes   binary Sequence Component EK
            .bpc   binary PCK
            .bsp   binary SPK
            .tf    text FK
            .ti    text IK
            .tls   text LSK
            .tm    text meta-kernel (FURNSH kernel)
            .tpc   text PCK
            .tsc   text SCLK
 
Both of your proposed methods for addressing this comment are fine by me.

tmplummer:
> Tested by creating a `.ck` file suffix and uploading that, which put the file in s3 in the place expected. Ran it again and got a 409 conflict response.

Is this the only test coverage for this code? I poked around in the repo a bit and found a `test_download_api.py` file but no apparent coverage for the upload api.

sds_data_manager/lambda_code/SDSCode/upload_api.py
tmplummer: Note the discrepancy in the CK extension. It should be '.bc' not '.ck'

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/upload_api.py
laspsandoval: Agreed. It should be .bc. For some reason the attitude example kernel was sent with the .ck extension so it may change.

laspsandoval (approved):
Just a comment about the attitude kernel extension. Nice work!

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: Awesome, thanks! I'm going to update `imap-data-access` with @subagonsouth's list and I agree that we should call that definitive and not allow others. I will move this to draft for now until I'm able to update the other package and push out a new release.

greglucas:
>> Tested by creating a .ck file suffix and uploading that, which put the file in s3 in the place expected. Ran it again and got a 409 conflict response.

> Is this the only test coverage for this code? I poked around in the repo a bit and found a test_download_api.py file but no apparent coverage for the upload api.

Our test coverage of infrastructure is definitely lacking. Thank you for bringing this up. I will update this because I agree with you that we should do better when we can and I don't think it will be too hard to add here.

greglucas:
@subagonsouth thanks for bringing up the lack of testing. I took some time and added that here now. Unfortunately, that blew things up a bit in scope as our tests were somewhat stepping on each other. So I ended up refactoring a bunch of minor things that were needed to get the upload tests working due to monkeypatching.

@tech3371 and @laspsandoval I dismissed your reviews because there are significant changes since you last reviewed.

The most controversial change is that I have removed the database check for existing file from upload_api. Now we are checking the s3 bucket if there is an object in the data bucket or not. Who should be the source of truth, the data bucket or the database table? I find this to be easier (avoids a database connection/query), but it is a change from what we were doing.

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -5,40 +5,75 @@
...
-            "Key": key_path,
+            "Bucket": BUCKET_NAME,
+            "Key": s3_key_path,
             "Metadata": tags or dict(),
         },
         ExpiresIn=3600,
     )
 
-    return url
+    if url is None:
```
tmplummer: If I understand this correctly, `boto.client.generate_presigned_url` returns `None` if the `s3_key_path` does not exist? Is the `s3_key_path` generated based on filename? I guess what I am driving at is that it seems like checking that the upload filename matches the mission filename convention would be done much earlier and a definitive message could be returned stating that in fact, "The filename does not match the mission file naming convention."

tmplummer (approved):
It is great to see the improved testing.

As for the source of truth for if a file exists, it seems to me that the S3 bucket is the ultimate truth and the DB should be designed never be out of sync with what is in S3. That is my naive opinion with no background on the SDC design to back it up.

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: Everything is checked before this point and returns error messages/codes in the SPICE / Science routines, so it should be good by this point.

In fact, I am not sure that this function can return None... @tech3371 do you know why this was added/needed?
I think it would raise an Exception, and then in the documentation it shows that exception returns a None from a wrapped function, so perhaps that is a bit of an indirection and we don't need it.

https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-presigned-urls.html

I would be happy to remove it, but left things the way they were before for now.

greglucas: Here is the implementation. I'm not sure if `request.url` can be None or not 🤷 
https://github.com/boto/botocore/blob/d85b98d44c3b9ed8661270db60e0b70de85c650d/botocore/signers.py#L323

tech3371: I can't remember why it was added. We can remove this

sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -69,55 +103,52 @@ def lambda_handler(event, context):
...
+        return _spice_file_upload(filename)
+    except imap_data_access.SPICEFilePath.InvalidSPICEFileError:
+        # Not a good SPICE filename, continue on and try science file type next
+        pass
+
+    # Upload for science files, which will catch other filetype errors
+    return _science_file_upload(filename)
+
+
+def _spice_file_upload(filename):
```
tech3371: A quick preference request. Can we move all of these functions above `lambda_handler` function? I got used to having `lambda_handler` function at the bottom and everything that it calls above it.

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: I have removed this in the latest commit.

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: Yes, no problem, I moved that down to the bottom of the module now. After the previous refactor, the spice and science functions were largely the same, so I actually removed those two helper functions entirely and inlined the code into `lambda_handler`.

tech3371 (approved):
Looks great! Thank you for refactoring and updating lot of duplicate code.

PR #265, opened 2024-04-16
Add ECS to I-ALiRT Architecture

### LINKED ISSUE #226: I-ALiRT: Create ECS for POC
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Adds ECS to I-ALiRT Architecture for container 1. **The next step will be to add container 2.** I tried to design the architecture to handle the two containers. I am working on being able to deploy resources for both, but this PR is for a single container. 

For this setup we are using an Application Load Balancer and following best practices shown here: https://docs.aws.amazon.com/AmazonECS/latest/bestpracticesguide/networking-inbound.html

## New Files
- __init__.py
   - Add ports list for container 1,2
 
## Updated Files
- ialirt_processing_stack.py
   - Add ecs security group
   - Add security group for application load balancer
   - Add an ecs service and cluster 
   - Add load balancer 
   - Add autoscaling for container

Note: I put docstrings throughout the code to describe each of the bullet points.

## Testing
   - After deploying the IalirtProcessing stack you can test that the container is running as expected on different ports. Also note, that an output url will be shown after deployment that you can copy and paste in your browser. You should see the text "Hola Mundo"

**Important**: You must deploy the ECR and push the Docker image in tests/test-data/ialirt_ec2 before deployment of IalirtProcessing.


sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +48,204 @@ def __init__(
...
+            container_name, ports, self.ecs_security_group
+        )
+        # Add load balancer for each container
+        load_balancer = self.add_load_balancer(
+            container_name, ports, ecs_service, self.load_balancer_security_group
+        )
+        # Add autoscaling for each container
+        self.add_autoscaling(container_name, ecs_cluster, load_balancer, ports)
+
+    def add_ecs_security_group(self, ports):
```
tech3371: Can you rename this function? On first glance, it seems like we are adding ports to security group and I was wondering where we were creating the security group which I found out this function does.
```suggestion
    def create_ecs_security_group(self, ports):
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +48,204 @@ def __init__(
...
-            "systemctl enable docker",
-            "$(aws ecr get-login --no-include-email --region us-west-2 | bash)",
-            f"docker pull {self.repo.repository_uri}:latest",
-            f"docker run --rm -d -p 8080:8080 {self.repo.repository_uri}:latest",
+        # Defines the type and port number for the
+        # allowed traffic to the Application Load Balancer
+        # and the EC2 instances.
+        ports = IALIRT_PORTS_TO_ALLOW_CONTAINER_1
+
+        # Add single security group in which
```
tech3371:
```suggestion
        # Create single security group in which
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +48,204 @@ def __init__(
...
+        for port in ports:
+            # Add ingress rule for each port
+            ecs_security_group.add_ingress_rule(
+                peer=ec2.Peer.any_ipv4(),
+                connection=ec2.Port.tcp(port),
+                description="Allow inbound traffic on TCP port",
+            )
+        return ecs_security_group
+
+    def add_load_balancer_security_group(self, ports):
```
tech3371: same here and in other places too. Lot of the functions that had `add` seems to be creating resources.
```suggestion
    def create_load_balancer_security_group(self, ports):
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +48,204 @@ def __init__(
...
 
-        # Create an IAM role for the EC2 instance
-        # - Read-only access to AWS ECR
-        # - Basic instance management via AWS Systems Manager
-        # Note: the Systems Manager provides a secure way for
-        # users to interact and access the EC2 during development.
-        ec2_role = iam.Role(
+        return load_balancer_security_group
+
+    def add_compute_resources(self, container_name, ports, ecs_security_group):
```
tech3371: Can you move security creating code into this function since it's not reused or another way to reduce code tracing?

sds_data_manager/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""sds-data-manager processing package.
+
+This package contains the IMAP architecture.
+"""
+
+# Ports configuration for various containers within the IALiRT system
+IALIRT_PORTS_TO_ALLOW_CONTAINER_1 = [8080, 8081]
+IALIRT_PORTS_TO_ALLOW_CONTAINER_2 = [8082, 8083]
```
tech3371: looks like this is not used yet. remove it for now?

sds_data_manager/stacks/ialirt_processing_stack.py
laspsandoval: I haven't done it yet, but I plan on adding another container and placing both containers into a single security group. I will add it below this line.

self.add_container_resources("Container1", IALIRT_PORTS_TO_ALLOW_CONTAINER_1)

So the way I have it setup I need the security creating code to be separate from the add_compute_resources.

sds_data_manager/__init__.py
greglucas, line 1: Can we put the allowed ports into the file that needs them directly rather than the indirection to this global location?

What about just `IALIRT_PORTS` for a name? I don't think we are going to have different ports between the two containers, one is a backup for the other so should be basically identical between the implementations I'd think.

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +49,206 @@ def __init__(
...
-            f"docker pull {self.repo.repository_uri}:latest",
-            f"docker run --rm -d -p 8080:8080 {self.repo.repository_uri}:latest",
+        # Defines the type and port number for the
+        # allowed traffic to the Application Load Balancer
+        # and the EC2 instances.
+        ports = IALIRT_PORTS_TO_ALLOW_CONTAINER_1
+
+        # Create single security group in which
+        # both containers will reside
+        self.ecs_security_group = self.create_ecs_security_group(ports)
```
greglucas: I think it would be easier to follow if you explicitly mutate self within the method and remove the return statement, the same for the `create_load_balancer` below.
```suggestion
        self.create_ecs_security_group(ports)
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +49,206 @@ def __init__(
...
+
+        # Add a single security group in which
+        # both application load balancers will reside
+        self.load_balancer_security_group = self.create_load_balancer_security_group(
+            ports
+        )
+
+        # Initialize resources for each container
+        self.add_container_resources("Container1", IALIRT_PORTS_TO_ALLOW_CONTAINER_1)
```
greglucas: Here you go two levels deep with methods. Can you move the 3 container resources method calls up into this layer?

```python
self.add_compute_resources()
self.add_load_balancer()
self.add_autoscaling()
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +49,206 @@ def __init__(
...
+        # Add load balancer for each container
+        load_balancer = self.add_load_balancer(
+            container_name, ports, ecs_service, self.load_balancer_security_group
+        )
+        # Add autoscaling for each container
+        self.add_autoscaling(container_name, ecs_cluster, load_balancer, ports)
+
+    def create_ecs_security_group(self, ports):
+        """Create and return a security group for containers."""
+        ecs_security_group = ec2.SecurityGroup(
```
greglucas:
```suggestion
        self.ecs_security_group = ec2.SecurityGroup(
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +49,206 @@ def __init__(
...
+            self,
+            "IalirtEcsSecurityGroup",
+            vpc=self.vpc,
+            description="Security group for Ialirt",
+            allow_all_outbound=True,
         )
 
-        # Security Group for the EC2 Instance
-        security_group = ec2.SecurityGroup(
+        for port in ports:
```
greglucas:
```suggestion
        for port in IALIRT_PORTS:
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +49,206 @@ def __init__(
...
+            # Add ingress rule for each port
+            ecs_security_group.add_ingress_rule(
+                peer=ec2.Peer.any_ipv4(),
+                connection=ec2.Port.tcp(port),
+                description="Allow inbound traffic on TCP port",
```
greglucas:
```suggestion
                description=f"Allow inbound traffic on TCP port {port}",
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +49,206 @@ def __init__(
...
+            )
+        return ecs_security_group
+
+    def create_load_balancer_security_group(self, ports):
+        """Create and return a security group for load balancers."""
```
greglucas, line 96: This looks like an identical security group to the ECS security group, can you re-use that SG here?

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +49,206 @@ def __init__(
...
-            ec2.Peer.ipv4("128.138.131.0/24"),
-            ec2.Port.tcp(8080),
-            "Allow inbound traffic on TCP port 8080",
+        # Allow inbound traffic from a specific port and
+        # any ipv4 address.
+        for port in ports:
+            load_balancer_security_group.add_ingress_rule(
+                peer=ec2.Peer.any_ipv4(),
+                connection=ec2.Port.tcp(port),
+                description=f"Allow inbound traffic on " f"TCP port {port}",
```
greglucas:
```suggestion
                description=f"Allow inbound traffic on TCP port {port}",
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +49,206 @@ def __init__(
...
+            )
+
+        # Allow all outbound traffic.
```
greglucas: Why do we want to allow all outbound? Don't we want the inbound and outbound rules to be the same?

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +49,206 @@ def __init__(
...
+        task_definition = ecs.Ec2TaskDefinition(
+            self, f"IalirtTaskDef{container_name}", network_mode=ecs.NetworkMode.AWS_VPC
+        )
+
+        # Adds a container to the ECS task definition
+        # Logging is configured to use AWS CloudWatch Logs.
+        container = task_definition.add_container(
+            f"IalirtContainer{container_name}",
+            image=ecs.ContainerImage.from_ecr_repository(self.repo, "latest"),
+            memory_limit_mib=512,
```
greglucas, line 148: Add a quick link to what these values mean?
```suggestion
            # Allowable values: 
            # https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_ecs.TaskDefinition.html#cpu
            memory_limit_mib=512,
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -46,78 +49,206 @@ def __init__(
...
+            f"IalirtService{container_name}",
+            cluster=ecs_cluster,
+            task_definition=task_definition,
+            security_groups=[ecs_security_group],
+            desired_count=1,
         )
 
-        # Create an EC2 instance
-        ec2.Instance(
+        return ecs_service, ecs_cluster
```
greglucas: Remove these if setting within this function: `self.ecs_service = ...`, `self.ecs_cluster = ...`
```suggestion
```

tests/docker_images/Dockerfile.testing
```diff
@@ -7,6 +7,7 @@ RUN pip install --upgrade pip
 
 ARG DEBIAN_FRONTEND=noninteractive
 RUN pip install requests
+RUN pip install imap_data_access
```
greglucas: Are you still using `requests` within the container example, or just using imap_data_access which has all of its own dependencies already?

If you do need requests, then they should both be on the same line for a single Docker layer, not separate RUN commands, but if you don't need it, just remove that line.
```suggestion
RUN pip install imap_data_access
```

sds_data_manager/__init__.py
laspsandoval: From the IMAP SOC I-ALiRT ICD:
Each station will be assigned two ports to allow for backup receivers to run simultaneously.

laspsandoval: Ok. So we will listen to the backup port, but most likely there will be no data coming in unless there is a problem with the first. Makes sense.

sds_data_manager/stacks/ialirt_processing_stack.py
laspsandoval: It allows the load balancer to easily send a response back the client and I couldn't think of a reason to restrict outbound. I'll change it to be more restrictive and we could always change it back if needed.

tests/docker_images/Dockerfile.testing
laspsandoval: I need it for test_ialirt_processing_stack.py

bmcclellan-cu (approved):
If this works I don't see an issue. However, based on the conversation during the AWS tag-up meeting, if I was tasked with this work, I would go with a network load balancer first, and then only if that did not meet my requirements (and I couldn't get it working any other way) would I go application load balancer. Just based on the way the various GS send via ports, I'd look at routing via the Layer 4 network protocol opposed to Layer 7 application level, based on the architecture and everything I've heard to-date.

I hope this helps some!

laspsandoval:
Added the NLB as suggested by Brian. Confirm that everything is up and running.

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -36,107 +42,206 @@ def __init__(
             VPC into which to put the resources that require networking.
         repo : ecr.Repository
             ECR repository containing the Docker image.
-        instance_type : str
-            The EC2 instance type.
+        processing_name : str
+            Name of the processing stack.
+        ialirt_ports : list[int]
+            List of ports to for incoming traffic.
```
greglucas:
```suggestion
            List of ports to listen on for incoming traffic.
```

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -36,107 +42,206 @@ def __init__(
...
+        container_port : int
+            Port to be used by the container.
```
greglucas, line 50: Why is this one only a single `int` whereas above you have `list[int]`. I thought we wanted a one-to-one mapping of ports, not a `many-to-one` relationship?

Below I thought we could just say `for port in ialirt_ports: container_port=port, host_port=port`

sds_data_manager/utils/stackbuilder.py
```diff
@@ -187,16 +187,21 @@ def build_sds(
...
-        "IalirtProcessing",
-        env=env,
-        vpc=networking.vpc,
-        repo=ialirt_ecr.container_repo,
-    )
+    # All traffic to I-ALiRT is directed to listed container ports
+    ialirt_ports = {"Primary": [8080, 8081], "Secondary": [80]}
+    container_ports = {"Primary": 8080, "Secondary": 80}
+
+    for port in ialirt_ports:
```
greglucas: This is the string that is coming out, not the ports. The ports are in your lookup below.
```suggestion
    for primary_or_secondary in ialirt_ports:
```

greglucas (commented):
I find it a bit odd that we have to deploy two separate stacks for i-alirt now. Since it is a unit of deployment I think we'd want just one stack if possible. Are you able to move the `for primary/secondary:` loop into the stack itself to create the load balancers with separate names there? I am wondering if there are name conflicts within the stack that are causing the issues and we just need to add some more of your f-string additions to the resource names to create multiples of the resources.

sds_data_manager/stacks/ialirt_processing_stack.py
laspsandoval: We want all of the traffic from the primary ports to be directed to the first container. And then all of the traffic from the secondary ports to be directed to the second container. The Dockerfile only exposes one port:

Make port 8080 available to the world outside this container
EXPOSE 8080

Therefore any of the primary ports will be directed to port 8080. It is probably possible to expose the Docker container to other ports and have a one-to-one ratio, but then every time you add another port you have to create a new Docker image.

laspsandoval:
> I find it a bit odd that we have to deploy two separate stacks for i-alirt now. Since it is a unit of deployment I think we'd want just one stack if possible. Are you able to move the `for primary/secondary:` loop into the stack itself to create the load balancers with separate names there? I am wondering if there are name conflicts within the stack that are causing the issues and we just need to add some more of your f-string additions to the resource names to create multiples of the resources.

I thought you might say that. I like it because if we need to change the cdk code for any reason then we don't have to bring the entire system down. We could deploy the primary stack first, for example, and maintain functionality. Everything is now redundant.

greglucas (approved):

PR #272, opened 2024-05-17
added dependency for ultra and a starting yaml

laspsandoval:
# Change Summary

## Overview
Getting the yaml started that we had discussed at the last IMAP SDC meeting. Similar to what we had before with the dependency json. @subagonsouth and @maxinelasp if you have something else in mind feel free to let me know and I'll delete this one. I also added Ultra l0 -> l1a dependency/dependent info to dependency_config.py in case we want to continue to go that route.

## New Files
- dependency.yml
   - getting the yaml started that we had discussed at the last IMAP SDC meeting. 

## Updated Files
- dependency_config.py
   - Added Ultra l0 -> l1a dependencies and dependents.


tmplummer (commented):
This is outside my area of expertise, so I don't have much to say. Maybe I was confused if I said something in a meeting related to this.

greglucas:
I thought that the `yaml` configuration files were going to be used as a templating engine essentially over in the `imap_processing` and used for metadata, not the dependency graphs. I'm not sure I am following the yaml vs code portion here. Is the added yaml file used anywhere?

laspsandoval:
> I thought that the `yaml` configuration files were going to be used as a templating engine essentially over in the `imap_processing` and used for metadata, not the dependency graphs. I'm not sure I am following the yaml vs code portion here. Is the added yaml file used anywhere?

It is not currently used anywhere. The idea here is that dependency_config.py is going to be very large so in order to reduce the potential for human error we probably need to have a better way. I spoke with Tenzin about this and for now I'm creating a separate dependency_config.py for ultra. Eventually we will need to revisit how it is all setup.

sds_data_manager/lambda_code/SDSCode/dependency_config_ultra.py
```diff
@@ -0,0 +1,180 @@
...
+code in imap_processing repo.
+"""
+
+from .database.models import PreProcessingDependency
+
+downstream_dependents = [
+    PreProcessingDependency(
+        primary_instrument="ultra",
+        primary_data_level="l0",
+        primary_descriptor="45raw",
```
tech3371: I wonder if we should replace `45raw` and `90raw` with just `raw`. This is something I wondered myself on what is best way for this from POC perspective. They may just give us `imap_ultra_l0_raw_yyyymmdd_v001.pkts` which will contain all of ultra 45 and 90 datas. Then it will kick off all of 45 and 90's downstream data product as you have it.

tech3371 (commented):
It looks great! Nice work! I have one minor comment about the l0 data descriptor. Otherwise it looks good to me!

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/dependency_config_ultra.py
laspsandoval: Good point.

PR #273, opened 2024-05-22
SWE Dependency Update

tech3371:
# Change Summary

## Overview
* Update SWE dependency
* Fix import error


laspsandoval (approved):
Thanks for fixing the import error.

PR #274, opened 2024-05-22
Ultra L1b dependency update

laspsandoval:
# Change Summary

## Overview
Adding dependencies for L1b

## Updated Files
- dependency_config_ultra.py
   - Adding new dependencies for L1b


sds_data_manager/lambda_code/SDSCode/dependency_config_ultra.py
```diff
@@ -16,80 +16,142 @@
     PreProcessingDependency(
         primary_instrument="ultra",
         primary_data_level="l0",
-        primary_descriptor="raw",
+        primary_descriptor="45raw",
```
tech3371:
```suggestion
        primary_descriptor="raw",
```

sds_data_manager/lambda_code/SDSCode/dependency_config_ultra.py
```diff
@@ -16,80 +16,142 @@
...
         dependent_instrument="ultra",
         dependent_data_level="l1a",
-        dependent_descriptor="45events",
+        dependent_descriptor="45all",
```
tech3371, line 53:
```suggestion
        dependent_descriptor="all",
```

sds_data_manager/lambda_code/SDSCode/dependency_config_ultra.py
```diff
@@ -16,80 +16,142 @@
...
+        dependent_data_level="l1b",
+        dependent_descriptor="45annotated-de",
         relationship="HARD",
         direction="DOWNSTREAM",
     ),
     PreProcessingDependency(
         primary_instrument="ultra",
         primary_data_level="l0",
-        primary_descriptor="raw",
+        primary_descriptor="90raw",
```
tech3371: With above changes, you don’t need this dependency. Don’t forget to make same change to upstream as well for this.  
```suggestion

```

tech3371 (commented):
One comment. Otherwise others look good

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/dependency_config_ultra.py
laspsandoval: Wouldn't this make it to where if only packets from Ultra 45 arrive then it would kick off Ultra 45 and Ultra 90 l1a processing?

tech3371:
I couldn't find your comment. But to answer your question
>Wouldn't this make it to where if only packets from Ultra 45 arrive then it would kick off Ultra 45 and Ultra 90 l1a processing?

Even if your l0 data file only exists Ultra 45 data, in your processing code, you group your data by apid and then process each apid at a time. That should take care of not kicking Ultra 90 since it should have different apid, right? MAG has an example

PR #275, opened 2024-05-24
Added SWAPI dependency

tech3371:
# Change Summary

## Overview
* Added SWAPI dependency
* Tested in AWS


sds_data_manager/lambda_code/SDSCode/dependency_config.py
```diff
@@ -23,6 +23,16 @@
         relationship="HARD",
         direction="DOWNSTREAM",
     ),
+    PreProcessingDependency(
+        primary_instrument="swapi",
+        primary_data_level="l0",
+        primary_descriptor="raw",
+        dependent_instrument="swapi",
+        dependent_data_level="l1",
+        dependent_descriptor="sci",
```
greglucas: Should this be `sci-1min`?

sds_data_manager/lambda_code/SDSCode/dependency_config.py
```diff
@@ -23,6 +23,16 @@
...
+        relationship="HARD",
+        direction="DOWNSTREAM",
```
greglucas, line 34: Just a comment and nothing to change now, but I find the `dependent_descriptor` to be a bit confusing still. It took me a bit to figure out that the dependent depends on the direction the dependency is going. Here you've added both `raw` and `sci` as dependent and the differentiator is that there is another "direction" keyword argument.

Two thoughts for the future:
Make an upstream/downstream class that could be used to clarify directly in the name:
```python
class UpstreamDependency(PreProcessingDependency):
    def __init__(self, *args, **kwargs):
        self.direction="upstream"
        super().__init__(self, *args, **kwargs)
```

Use `child` and `parent` to infer direction instead possibly. Is there a direct 1-1 mapping of these, so every time you add in a downstream you have to add in an upstream with the opposite values? It seems like we might be able to automate the upstream dependency creation by reversing the tree if so without needing to duplicate these entries.

greglucas (approved):

sds_data_manager/lambda_code/SDSCode/dependency_config.py
tech3371: I would love to chat on possibilities of improvement to this. I feel like there was reason to keep upstream and downstream. Since right now, most of dependency is one-to-one, it seems like we can reserve it. But when we add-on inter-instrument dependency, it might change. But again, I would love to hear ideas on how to improve dependency tree and dependency association.

sds_data_manager/lambda_code/SDSCode/dependency_config.py
tech3371: Good catch. It should have been `all` instead.

PR #276, opened 2024-05-29
Added CoDICE dependencies

bourque:
This PR adds the CoDICE L0, L1a, and L1b dependencies


sds_data_manager/lambda_code/SDSCode/dependency_config_codice.py
```diff
@@ -0,0 +1,383 @@
...
+        dependent_data_level="l1a",
+        dependent_descriptor="all",
+        relationship="HARD",
+        direction="DOWNSTREAM",
+    ),
+    # l1a to l1b
+    PreProcessingDependency(
+        primary_instrument="codice",
+        primary_data_level="l1a",
+        primary_descriptor="hskp",
```
tech3371, line 191: It's ok right now. Something we can bring up during nomenclature discussion in the future. Should we use `hk` or `hskp`?

tech3371 (approved):
It looks good to me. Nice

sds_data_manager/lambda_code/SDSCode/dependency_config_codice.py
bourque: Yeah, I agree, something we should discuss. I know that CoDICE uses `hskp` all throughout their algorithm document ...

bourque:
Just noting here that I need to change this to have a one-to-one mapping of L0 APIDs/descriptors to L1a products

laspsandoval (approved):
Nice work.

tech3371 (commented):
looks good to me. Post SIT-3, we can figure out how we want data from POC and then change how we want to trigger l1a process based on outcome from that discussion.

PR #278, opened 2024-05-30
Typo fix in dependency config file for Ultra

laspsandoval:
# Change Summary

## Overview
Typo fix in dependency config file for Ultra

## Updated Files
- dependency_config_ultra.py
   - fixed typo


greglucas (approved):
Would it be possible to write this as:
```python
upstream_dependents = []
for dep in downstream_dependents:
    upstream_dep = PreProcessingDependency(
        primary_instrument=dep.dependent_instrument,
        primary_data_level=dep.dependent_data_level,
        primary_descriptor=dep.dependent_descriptor,
        dependent_instrument=dep.primary_instrument
        dependent_data_level=dep.primary_data_level
        dependent_descriptor=dep.primary_descriptor,
        direction="UPSTREAM",
    )
    upstream_dependents.append(upstream_dep)
```

Essentially reversing the directions. If so, perhaps we should add a method to `PreProcessingDependency` that does this for us automatically?

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -190,6 +190,19 @@ class PreProcessingDependency(Base):
     relationship = Column(DEPENDENCY_RELATIONSHIPS, nullable=False)
     direction = Column(DEPENDENCY_DIRECTIONS, nullable=False)
 
+    def reverse_direction(self):
```
tech3371, line 194: I think it's ok to do this for this SIT but it won't be true after that when we add in inter-instrument dependency. Can you note that somewhere as reminder?

sds_data_manager/lambda_code/SDSCode/dependency_config_ultra.py
```diff
@@ -230,219 +230,7 @@
...
-        dependent_instrument="ultra",
-        dependent_data_level="l1b",
-        dependent_descriptor="90badtimes",
-        relationship="HARD",
-        direction="DOWNSTREAM",
-    ),
-]
+upstream_dependents = []
+
+for dep in downstream_dependents:
```
tech3371, line 236: may be add note here about what I was mentioning in first comment?

tech3371 (approved):
I commented on one concern I have of this change. But I like the reasons why we want to  do this.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: I thought of an idea for this case. I will share that after SDWG meeting.

tech3371: But I don't mean to not merge this. You can merge this because I will need to make same changes for other instruments. I can do that after this PR is merged.

PR #280, opened 2024-05-31
combine all dependency into one file

tech3371:
# Change Summary

## Overview
- added reserve logics and combined all dependency into one file
- add IMAP-Hi dependency
- update sds-data-access version


sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -72,16 +60,16 @@ def lambda_handler(event, context):
         # Create tables
         engine = db.get_engine()
         Base.metadata.create_all(engine)
-        # Write dependencies to pre-processing dependency table
+        # Get dependencies for pre-processing dependency table
+        reversed_downstream_dependents = []
+        for dep in downstream_dependents:
+            if dep.reserve_direction is True:
+                reversed_downstream_dependents.append(dep.reverse_direction())
```
greglucas: Should the "reversed" dependencies be added to upstream dependencies in the dependency_config module instead of here?
```suggestion
        # Get dependencies for pre-processing dependency table
        reversed_downstream_dependents = []
        for dep in downstream_dependents:
            if dep.reverse_direction:
                reversed_downstream_dependents.append(dep.reverse_direction())
```

greglucas (commented):
* Typo throughout with "reserve" -> "reversed"
* If it is not able to be used for inter-instrument dependencies can you try to reverse everything and have a check within your reverse function that says "is primary_instrument == "dependent_instrument" and return nothing if so.

I am still not following why inter-instrument dependencies would be different and not reversed? I guess I'm not seeing why the dependencies can't be straight edges and they have to be directional edges. I'm likely missing something in the implementation since I haven't followed along what is being queried elsewhere. But, here is my quick confusion:

If `hit_l1b` needs `mag_l2` and `hit_l1a` to process, then couldn't it also be useful information to reverse that and say `mag_l2` is used for `hit_l1b` processing. Why would we not want that reversed case as well?

sds_data_manager/lambda_code/SDSCode/create_schema.py
tech3371: Yeah, that makes sense.

tech3371:
>I am still not following why inter-instrument dependencies would be different and not reversed?
@greglucas yeah, you may be right. I may be thinking about case of `HARD` and `SOFT` relationship. Because some instrument doesn't required some instrument data to kickoff their processing. It's nice to have. In that case, we could still have upstream and downstream relationship reversed but the `relationship` will determine if something should be kicked off or not.

sds_data_manager/lambda_code/SDSCode/dependency_config.py
```diff
@@ -1,58 +1,505 @@
...
         dependent_descriptor="all",
         relationship="HARD",
         direction="DOWNSTREAM",
+        reverse_direction=True,
     ),
     PreProcessingDependency(
-        primary_instrument="swe",
+        primary_instrument="hi",
+        primary_data_level="l1a",
+        primary_descriptor="45sensor-histogram",
```
tmplummer, line 388: I assume that the `primary_descriptor` must match filename descriptors. Right now, this string is hard coded here and in the `imap_processing` repo global attributes definition for the file and must match. It seems like we need to decide on a single source of truth for where these strings are defined. This is not for this PR, but a ticket should be written to capture this.

tmplummer (approved):
I just had one comment for some follow on work that this PR helped me identify. Thanks for this work. I would like a high level overview of how these dependencies are used at some point.

sds_data_manager/lambda_code/SDSCode/dependency_config.py
tech3371: I totally agree. I have been looking at CDF attrs' `logical_source` to match these but would be nice to have source of truth for descriptor as well. I wonder if we can use imap-data-access. But I will create a ticket to capture this.

sds_data_manager/lambda_code/SDSCode/dependency_config.py
```diff
@@ -153,188 +548,258 @@
...
+        direction="DOWNSTREAM",
     ),
 ]
+
+upstream_dependents = []
+
+# Get dependencies for pre-processing dependency table
+for dep in downstream_dependents:
+    if dep.reverse_direction:
+        upstream_dependents.append(dep.reverse_direction())
```
greglucas:
```suggestion
    upstream_dependents.append(dep.reverse_direction())
all_dependents = downstream_dependents + upstream_dependents
```

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -7,18 +7,6 @@
 from SDSCode.database import database as db
 from SDSCode.database.models import Base
 from SDSCode.dependency_config import downstream_dependents, upstream_dependents
```
greglucas:
```suggestion
from SDSCode.dependency_config import all_dependents
```

sds_data_manager/lambda_code/SDSCode/dependency_config.py
```diff
@@ -801,7 +801,6 @@
 
 # Get dependencies for pre-processing dependency table
 for dep in downstream_dependents:
-    if dep.reverse_direction:
-        upstream_dependents.append(dep.reverse_direction())
+    upstream_dependents.append(dep.reverse_direction())
```
greglucas: Minor nit, but you could write this as a one-line list comprehension as well.
```suggestion
    upstream_dependents = [dep.reverse_direction() for dep in downstream_dependents]
```

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -130,6 +132,74 @@ def sts_client():
...
+def test_reverse_direction():
+    """Test PreProcessingDependency reverse_direction method."""
+    # Test that they have the same length
+    assert len(downstream_dependents) == len(upstream_dependents)
+    # Check that first downstream dependent is the reverse of the
+    # first upstream dependent
+    first_reversed_dependent = upstream_dependents[0]
+    assert (
+        downstream_dependents[0].relationship == first_reversed_dependent.relationship
+    )
```
greglucas, line 144: Does `==` not work on the objects? Then you wouldn't have to compare every attribute.

```suggestion
    assert (
        downstream_dependents[0] == upstream_dependents[0].reverse_direction()
    )
```

greglucas (approved):

tests/lambda_endpoints/test_batch_starter.py
tech3371: Yeah, == didn't work. It requires to compare attribute.

sds_data_manager/lambda_code/SDSCode/dependency_config.py
tech3371: for sure!

PR #283, opened 2024-06-03
Updated Ultra Dependency and Added one Lo dependency

tech3371:
# Change Summary

## Overview
- Update ultra dependency after discussing with Laura
- Added one Lo dependency. TODO: add remaining


sds_data_manager/lambda_code/SDSCode/dependency_config.py
```diff
@@ -628,173 +638,84 @@
...
-        primary_descriptor="90sensor-badtimes",
+        primary_descriptor="45sensor-cullingmask",
         dependent_instrument="ultra",
         dependent_data_level="l1c",
-        dependent_descriptor="90all",
+        dependent_descriptor="45sensor-histogram",
         relationship="HARD",
         direction="DOWNSTREAM",
     ),
+    # TODO: Add Ultra-90 products
```
greglucas, line 697: Might want to write a function to do this automatically for you.

greglucas (approved):
I think we should talk with the teams again about whether these are individual data products, or if they are variables within a single data product. The names to me sound like you'd want them to all be in a single data product. I'd want my badtimes, cullingmask, de, ... all to be together in a single file load. Might just be a personal preference though. Are all the other ENA instruments similarly going to have many of these datasets as well?

sds_data_manager/lambda_code/SDSCode/dependency_config.py
tech3371: totally

tech3371:
>I think we should talk with the teams again about whether these are individual data products, or if they are variables within a single data product. The names to me sound like you'd want them to all be in a single data product. I'd want my badtimes, cullingmask, de, ... all to be together in a single file load. Might just be a personal preference though. Are all the other ENA instruments similarly going to have many of these datasets as well?

Laura and I have been saying the same thing. She is going to try to talk to them through example.

sds_data_manager/lambda_code/SDSCode/dependency_config.py
```diff
@@ -628,173 +638,84 @@
...
-        primary_descriptor="45sensor-de",
-        dependent_instrument="ultra",
-        dependent_data_level="l1c",
-        dependent_descriptor="45all",
+        dependent_descriptor="45sensor-extendedspin",
         relationship="HARD",
         direction="DOWNSTREAM",
     ),
+    # TODO: ask difference of following two products
     PreProcessingDependency(
```
laspsandoval: The way it works now we only need l1a sensor rates to kick off l1b extended spin. Kind of like your question about pset. There isn't a reason to kickoff another batch job since cullingmask and badtimes are the exact same thing as extended spin, but with quality flags applied for filtering.

laspsandoval: So lines 645-665 can be deleted.

laspsandoval (approved):
Just one change for Ultra.

sds_data_manager/lambda_code/SDSCode/dependency_config.py
tech3371: cool. That makes sense. Will remove that!

PR #289, opened 2024-06-04
Adding IMAP-Lo dependency

tech3371:
# Change Summary

## Overview
Adding IMAP-Lo dependency

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/dependency_config.csv
+++ b/sds_data_manager/lambda_code/SDSCode/dependency_config.csv
@@ -67,10 +67,27 @@ hit, l1a, hk, hit, l1b, hk, HARD, DOWNSTREAM
 idex, l0, raw, idex, l1, sci, HARD, DOWNSTREAM
 
 # <---- LO Dependencies ---->
-# TODO: add LO dependencies
-lo, l0, raw, lo, l1a, all, HARD, DOWNSTREAM
+lo, l0, de, lo, l1a, de, HARD, DOWNSTREAM
+lo, l0, histogram, lo, l1a, histogram, HARD, DOWNSTREAM
+lo, l0, star, lo, l1a, star, HARD, DOWNSTREAM
+lo, l0, spin, lo, l1a, spin, HARD, DOWNSTREAM
+lo, l0, hk, lo, l1a, hk, HARD, DOWNSTREAM
 
-lo, l1a, histogram, lo, l1b, histogram, HARD, DOWNSTREAM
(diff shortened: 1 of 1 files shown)


sdhoyt (approved):
I think this looks good. A lot of this is up in the air right now because we reorganized products in both L1B and L1C and removed some things in L1B, so the dependencies are a little blurry to me right now. Colin is working on updating the algorithm document to reflect those changes. I think this is good for SIT-3 though. Thanks for doing this!

Will it be a problem to leave the badtimes and goodtimes as tbd? I still have no idea how either of those are going to be made yet.

tech3371:
> I think this looks good. A lot of this is up in the air right now because we reorganized products in both L1B and L1C and removed some things in L1B, so the dependencies are a little blurry to me right now. Colin is working on updating the algorithm document to reflect those changes. I think this is good for SIT-3 though. Thanks for doing this!
> 
> Will it be a problem to leave the badtimes and goodtimes as tbd? I still have no idea how either of those are going to be made yet.

I totally understand. We can iterate on this as you need. To you last question, it won't be triggered since we won't be producing data product with descriptor `tbd`. I am mainly using that as place holder for future.

sds_data_manager/lambda_code/SDSCode/dependency_config.csv
```diff
@@ -67,10 +67,27 @@ hit, l1a, hk, hit, l1b, hk, HARD, DOWNSTREAM
...
-lo, l1a, histogram, lo, l1b, histogram, HARD, DOWNSTREAM
+lo, l0, de, lo, l1a, de, HARD, DOWNSTREAM
+lo, l0, histogram, lo, l1a, histogram, HARD, DOWNSTREAM
+lo, l0, star, lo, l1a, star, HARD, DOWNSTREAM
+lo, l0, spin, lo, l1a, spin, HARD, DOWNSTREAM
+lo, l0, hk, lo, l1a, hk, HARD, DOWNSTREAM
+
+lo, l1a, de, lo, l1b, de, HARD, DOWNSTREAM
+lo, l1a, spin, lo, l1b, de, HARD, DOWNSTREAM
+lo, l1a, tbd, lo, l1b, badtimes, HARD, DOWNSTREAM
```
greglucas: It might be good to make this all caps `TBD`, or a different word `TODO` and something that will stick out more to us in the editors? Otherwise it looks like `tbd` might be a valid name at first glance here.
```suggestion
lo, l1a, TODO, lo, l1b, badtimes, HARD, DOWNSTREAM
```

greglucas (approved):
Looks correct at first glance

PR #290, opened 2024-06-06
Change extendedspin to extspin for Ultra

laspsandoval:
# Change Summary

## Updated Files
- dependency_config.csv
   - name change


greglucas (commented):
Is the reason for this the length of the filename in the database, or is this what the instrument team wants it to be called? If it is the DB length issue, then I think we should fix the root cause rather than adjust the names of our variables to fit the arbitrary limit. It looks like codice will also run into issues with long descriptor names `lo-counters-aggregated`

greglucas (changes requested):
OK, I missed the descriptor length requirement before, but just saw that now. I think we need to update this instead: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/fe1a758bf98cbb46f7a2e673d05b2bec56959561/sds_data_manager/lambda_code/SDSCode/database/models.py#L155

laspsandoval:
I'll close without merging.

PR #291, opened 2024-06-06
removed char limit on descriptor

tech3371:
# Change Summary

## Overview
removed char limit on descriptor


bourque (approved):

laspsandoval (approved):
That's where it is! Thanks for finding it.

PR #292, opened 2024-06-11
minor update to lo dependency and fixed Batch Starter Lambda query to FileCatalog table

tech3371:
# Change Summary

## Overview
Minor Lo's dependency update

### DIFF
--- a/tests/lambda_endpoints/test_batch_starter.py
+++ b/tests/lambda_endpoints/test_batch_starter.py
@@ -66,7 +66,7 @@ def test_file_catalog_simulation(test_engine):
             file_path="/path/to/file",
             instrument="ultra",
             data_level="l2",
-            descriptor="science",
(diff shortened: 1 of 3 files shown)


tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -235,6 +235,7 @@ def test_query_instrument(test_file_catalog_simulation):
         "instrument": "ultra",
         "data_level": "l2",
         "version": "v001",
+        "descriptor": "science",
```
```suggestion
        "descriptor": "sci",
```

greglucas (approved):

tests/lambda_endpoints/test_batch_starter.py
tech3371: First I thought same too but she had the descriptor as `science` in above line. I will change above line to match this `sci`.

PR #295, opened 2024-06-14
Automate Docker processing deployments

greglucas:
# Change Summary

## Overview

This will automate the build and push of dockerfiles up to our ECR repositories. The idea is that it is a semi-manual procedure where we will go to "dispatch workflow" and put in our instrument name we want to deploy.

🤞 With the automated credentials, we have to sort of wait and see I think. I tried adding a branch directly to this upstream repo and that wasn't enough. I think we need to merge it into "dev" so that it is vetted before being approved as a workflow.

### DIFF
--- a/.github/workflows/deploy_containers.yml
+++ b/.github/workflows/deploy_containers.yml
@@ -1,0 +1,44 @@
+name: Docker image build and publish
+on:
+  workflow_dispatch:
+    inputs:
+      instrument_name:
+        description: 'Instrument name'
+        required: true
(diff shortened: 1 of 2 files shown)


bourque (approved):
Very neat!

tmplummer (approved):
You put that together super-fast ⚡. Thanks!

greglucas:
My guess is this won't be enough, but let me merge and try because this doesn't impact anything else.

PR #296, opened 2024-06-14
Fix for out of memory error for batch jobs

maxinelasp:
# Change Summary
Minor change, already tested with GLOWS but not deployed elsewhere.


pyproject.toml
```diff
@@ -67,7 +68,6 @@ requests-aws4auth = "^1.2.3"
 psycopg2-binary = "^2.9.9"
 urllib3 = "<2.0.0"
 SQLAlchemy = "<=3.0.0"
-imap-data-access = ">=0.6.1"
```
greglucas, line 70: We need this here still I think because our Lambda's use it to validate filenames. We could also add it to the cdk group in addition to this, but this one I think needs to stay.

https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/42f10d50366133070b4c19110acf6f1ecf66c2ae/sds_data_manager/lambda_code/SDSCode/upload_api.py#L9

bourque (approved):

pyproject.toml
maxinelasp: ooh yes you're right, I forgot that ONLY that dependency group was output in requirements.txt. Let me fix it

greglucas (approved):

PR #297, opened 2024-06-14
Automate development Docker processing deployments

greglucas:
# Change Summary

## Overview

Now we should be able to point to any git url from forks or the development branch by default. Hopefully making it easier to deploy dev work as well as production.

### DIFF
--- a/.github/workflows/deploy_dev_containers.yml
+++ b/.github/workflows/deploy_dev_containers.yml
@@ -1,0 +1,51 @@
+name: (dev) Docker image build and publish
+on:
+  workflow_dispatch:
+    inputs:
+      instrument_name:
+        description: 'Instrument name'
+        required: true
+        type: string
+      git_url:
+        description: 'IMAP Processing Git URL to build from'
+        required: true
+        type: string
+        default: 'https://github.com/IMAP-Science-Operations-Center/imap_processing.git@dev'
+
+# concurrency required to avoid lock contention during ECR provisioning
+concurrency: ci-${{ github.repository }}-${{ inputs.instrument_name }}-dev-docker-pipeline
+
+jobs:
+  deploy_containers:
+    runs-on: ubuntu-latest
+    # These permissions are needed to interact with GitHub's OIDC Token endpoint.
+    permissions:
+      id-token: write
+      contents: read
+
+    steps:
+      - uses: actions/checkout@v3
+
+      # https://github.com/aws-actions/configure-aws-credentials
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: arn:aws:iam::449431850278:role/GitHubDeploy
+          aws-region: us-west-2
+
+      - name: Login to Amazon ECR
+        id: login-ecr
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_dev_containers.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Build, tag, and push docker image to Amazon ECR
+        env:
+          REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          REPOSITORY: ${{ inputs.instrument_name}}
+          IMAGE_TAG: latest
+        run: |
+          docker build -t $REGISTRY/$REPOSITORY-repo:$IMAGE_TAG \
+            --build-arg GIT_URL=${{ inputs.git_url }} \
+            -f dockerfiles/Dockerfile.dev .
+          docker push $REGISTRY/$REPOSITORY-repo:$IMAGE_TAG
```
tech3371, line 51: Glad you are adding `-repo` here.

.github/workflows/deploy_dev_containers.yml
```diff
@@ -0,0 +1,51 @@
...
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v2
+        with:
+         mask-password: "true" # see: https://github.com/aws-actions/amazon-ecr-login#docker-credentials
+
+      - name: Build, tag, and push docker image to Amazon ECR
+        env:
+          REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          REPOSITORY: ${{ inputs.instrument_name}}
+          IMAGE_TAG: latest
```
tech3371, line 46: Not for this PR, but just for future discussion. I think we had discussion about pushing tag with some version that we can revert back to if needed. Right now, we only push latest which replaces last one and all repo history doesn't have any tag. That could be a concern if we want to revert back to some version. If we make more frequent release later, I wonder if we can automate like you have here in the prod Dockerfile to push using both latest and latest `imap_processing` released tag?

tech3371 (approved):
This is gonna help speed up so much and make build/push process much more uniform. Thank you!

tech3371:
Oh. I just realized that this PR is in `sds-data-manager` and not in `imap_processing`. Is that what we intended? Just curious

.github/workflows/deploy_dev_containers.yml
greglucas: I definitely agree with this! I didn't want to mess with it for now, but we should move to this version of things. In fact, maybe the workflow should be different and we add another workflow job that would just move tags from "dev"/"latest" -> "prod" when we verify it is working as expected rather than rebuilding the prod container.

greglucas:
> Oh. I just realized that this PR is in sds-data-manager and not in imap_processing. Is that what we intended? Just curious

I think so because we are deploying to our infrastructure. We are pointing to anywhere we want to build the container from, but the container will ultimately live in our infrastructure area. (It is also easier/more secure to only allow one area to access AWS credentials)

.github/workflows/deploy_dev_containers.yml
tech3371: I captured a note about this which I will create GitHub issues afterwards.

PR #301, opened 2024-06-18
Duplicate job check

tech3371:
# Change Summary

## Overview
- duplicate job checks. As a result updates both `imap_cli` command input

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -368,7 +368,7 @@ def lambda_handler(event: dict, context):
 
             if job_already_exist:
                 logger.info(
-                    f"Job was already started for {downstream_data['instrument']}, "
+                    f"Job record found for {downstream_data['instrument']}, "
                     f"{downstream_data['data_level']}, "
                     f"{downstream_data['descriptor']}, "
                     f"{downstream_data['start_date']}, "
@@ -377,11 +377,29 @@ def lambda_handler(event: dict, context):
                 continue
 
             logger.info(
-                f"Job not in progress for {downstream_data['instrument']}, "
+                f"Job record not found for {downstream_data['instrument']}, "
                 f"{downstream_data['data_level']}, "
                 f"{downstream_data['descriptor']}, "
                 f"{downstream_data['start_date']}, "
                 f"{downstream_data['version']}"
+            )
+
+            # Write to status tracking table with initial values
+            status_params = {
+                "status": models.Status.INPROGRESS,
+                "instrument": downstream_data["instrument"],
+                "data_level": downstream_data["data_level"],
+                "descriptor": downstream_data["descriptor"],
+                "start_date": datetime.strptime(
+                    downstream_data["start_date"], "%Y%m%d"
+                ),
+                "version": downstream_data["version"],
+            }
+
+            update_status_table(status_params)
+
+            logger.info(
+                f"Wrote job in progress to status tracking table - {status_params}"
             )
 
             # FYI, these are the keys the upstream_dependencies should contain:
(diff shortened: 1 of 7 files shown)


sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -311,48 +323,101 @@ def prepare_data(instrument, data_level, start_date, version, upstream_dependenc
     return prepared_data
 
 
-def send_lambda_put_event(command_parameters):
+def check_duplicate_job(
```
greglucas: Could we change this to an explicit name about what we are doing?
```suggestion
def is_job_in_status_table(
```

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -437,24 +504,47 @@ def lambda_handler(event: dict, context):
...
             # NOTE: The batch job name should contain only
             # alphanumeric characters and hyphens.
-            level_to_process = command[3]
+            level_to_process = downstream_data["data_level"]
             # Eg. "codice-l1a-job"
-            job_name = f"{instrument}-{level_to_process}-job"
+            job_name = (
+                f"{downstream_data['instrument']}-{level_to_process}-"
+                f"{downstream_data['descriptor']}-job"
+            )
```
greglucas, line 437: I'm just thinking of this now but could we update the status table based on the job name? Parse the information needed out of this string. Add in the `start_date` and `version` and I think we'd be there.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -437,24 +505,64 @@ def lambda_handler(event: dict, context):
...
+                "start_date": datetime.strptime(
+                    downstream_data["start_date"], "%Y%m%d"
+                ),
+                "version": downstream_data["version"],
+            }
+
+            update_status_table(status_params)
+
+            logger.info(
+                f"Wrote job in prgoress to status tracking table - {status_params}"
```
maxinelasp:
```suggestion
                f"Wrote job in progress to status tracking table - {status_params}"
```

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -437,24 +505,64 @@ def lambda_handler(event: dict, context):
...
+            #     'descriptor': 'lveng-hk',
+            #     'start_date': '20231212',
+            #     'version': 'v001',
+            # },
+            batch_command = [
+                "--instrument",
+                downstream_data["instrument"],
+                "--data-level",
+                downstream_data["data_level"],
+                "--descriptor",
```
maxinelasp, line 418: Does "descriptor" now always include info from query_upstream_dependencies, meaning that the dependencies table will provide descriptor for all levels and instruments because it's a required field there? Does that mean the descriptor will be "all" if that's the descriptor in the dependencies table?

maxinelasp (approved):
Looks good to me, if you're confident it will work for every instrument. From my understanding it won't affect anything since the CLI will just drop descriptor unless it's used in processing...

sds_data_manager/lambda_code/SDSCode/batch_starter.py
tech3371: yes. 'all' descriptor would be used to kick of certain level which will produce multiple files from within processing code.

We had to add 'descriptor' because without that, status tracking table will only have instrument, level, startdate, version and etc. which was causing row to be not unique initially and during lookup also.

maxinelasp: Yeah, I understand the issue. That sounds good to me!

PR #303, opened 2024-06-20
MNT: Sort query return results when returning from the API

greglucas:
# Change Summary

## Overview

This adds a sort to our query responses from the database table. We are currently sorting by "id" is all, so it depends on when files got inserted. The filename is a good key to sort by and includes the logical ordering I think we want.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/query_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/query_api.py
@@ -89,6 +89,11 @@ def lambda_handler(event, context):
         else:
             query = query.where(getattr(models.FileCatalog, param) == value)
 
+    # We want to order the query returns by the filename
+    # This will implicitly sort by: instrument, data level, descriptor, start_date, ...
+    # Default for the table is by the ascending id so by insertion order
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/query_api.py
```diff
@@ -89,6 +89,10 @@ def lambda_handler(event, context):
         else:
             query = query.where(getattr(models.FileCatalog, param) == value)
 
+    # We want to order the query returns by the filename
+    # This will implicitly sort by: instrument, data level, descriptor, start_date, ...
+    query = query.order_by(models.FileCatalog.file_path)
```
tech3371, line 95: What is the default sorting order? I was wondering that during the test. can we add comment that it's that by default?

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/query_api.py
greglucas: Yep, added a sentence for that.

bourque (approved):
LGTM!

PR #305, opened 2024-06-24
Add S3 Mount to Containers for I-ALiRT

### LINKED ISSUE #227: I-ALiRT: Create and mount S3 bucket to ECS
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Mounts s3 to two separate containers running in ECS.

## New Files
- ialirt_bucket_stack.py
   - Create ialirt bucket

## Updated Files
- stackbuilder.py
   - Addition of s3 bucket stack
- ialirt_processing_stack.py
   - Add IAM role and policy for S3 access
   - Add s3 bucket to environment
- ialirt_ec2_primary/Dockerfile and ialirt_ec2_primary/test_app.py (same for secondary)
   - Install dependencies and mountpoint-s3
   - Command to mount S3 and run the application sequentially

## Testing
   - Must deploy IalirtProcessingPrimary and IalirtProcessingSecondary stacks first.


sds_data_manager/stacks/ialirt_bucket_stack.py
```diff
@@ -0,0 +1,47 @@
...
+        """
+        super().__init__(scope, construct_id, env=env, **kwargs)
+        # Get the current account number so we can use it in the bucket names
+        account = env.account
+
+        # This is the S3 bucket used to mount to the container.
+        self.ialirt_bucket = s3.Bucket(
+            self,
+            "IAlirtBucket",
+            bucket_name=f"sds-data-{account}",
```
greglucas: This is re-using an already existing bucket name, so I don't think this will work?
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/d6f2d7c67ed7f1acb3d11b9f7ff77030ba108b2a/sds_data_manager/stacks/data_bucket_stack.py#L42

Do you want a new bucket, or do you want to use the same data bucket we already have and just use a different subfolder in that one?

sds_data_manager/stacks/ialirt_bucket_stack.py
greglucas, line 1: I don't like having a stack with one resource. This should just be added to our current i-alirt stack directly IMO.

sds_data_manager/stacks/ialirt_processing_stack.py
```diff
@@ -148,6 +172,8 @@ def add_compute_resources(self, processing_name):
             memory_limit_mib=512,
             cpu=256,
             logging=ecs.LogDrivers.aws_logs(stream_prefix=f"Ialirt{processing_name}"),
+            environment={"S3_BUCKET": self.s3_bucket_name},
+            privileged=True,
```
greglucas, line 178: Why do we need pivileged access? Can we make a note about this if it is required, or remove it if it was just for testing.

sds_data_manager/utils/stackbuilder.py
```diff
@@ -187,6 +188,11 @@ def build_sds(
         instrument_name="IalirtEcr",
     )
 
+    # I-ALiRT IOIS S3 bucket
+    ialirt_bucket = ialirt_bucket_stack.IAlirtBucketStack(
+        scope=scope, construct_id="IAlirtBucket", env=env
+    )
+
```
greglucas, line 195:
```suggestion
```

sds_data_manager/utils/stackbuilder.py
```diff
@@ -201,6 +207,7 @@ def build_sds(
             processing_name=primary_or_secondary,
             ialirt_ports=ialirt_ports[primary_or_secondary],
             container_port=container_ports[primary_or_secondary],
+            ialirt_bucket=ialirt_bucket.ialirt_bucket,
```
greglucas, line 210: Create the bucket within this stack.
```suggestion
```

tests/test-data/ialirt_ec2_primary/Dockerfile
```diff
@@ -13,8 +21,24 @@ WORKDIR /app
...
 # Make port 8080 available to the world outside this container
 EXPOSE 8080
 
-# Run test_app.py when the container launches
-CMD ["python", "test_app.py"]
+# Set the AWS region
+ENV AWS_REGION us-west-2
+
+# Command to mount S3 and run the application sequentially
+CMD ["sh", "-c", "mount-s3 sds-data-301233867300 /mnt/s3 && python test_app.py"]
```
greglucas: Is there a way we can mount the bucket for the users? This is wrapping up two commands in one. Our command (mount s3) and ops-software's command (test_app.py). (including it in fstab or some other daemon service possibly)

tests/test-data/ialirt_ec2_primary/Dockerfile
```diff
@@ -13,8 +21,24 @@ WORKDIR /app
...
+
+# Install dependencies and mountpoint-s3
+RUN apt-get update && \
+    apt-get install -y sudo wget fuse apt-utils && \
+    wget https://s3.amazonaws.com/mountpoint-s3-release/latest/x86_64/mount-s3.deb && \
+    sudo apt-get install -y ./mount-s3.deb && \
+    rm ./mount-s3.deb
+
+# Create the /mnt/s3 directory
+RUN mkdir -p /mnt/s3
```
greglucas: Remove an image layer here.
```suggestion
    rm ./mount-s3.deb && \
    # Create the /mnt/s3 directory
    mkdir -p /mnt/s3
```

tests/test-data/ialirt_ec2_secondary/Dockerfile
greglucas, line 1: I personally don't like that we have ec2_primary and ec2_secondary folders. This seems like duplication. I think we should be using identical containers and scripts between them, and then having the instance spit out its id (or something else unique), or instantiating the same container with a different option instead. This will prevent them from getting out of sync.

sds_data_manager/stacks/ialirt_bucket_stack.py
laspsandoval: The primary and secondary stacks will both use the same bucket. So the bucket must be created external to IalirtProcessing{primary_or_secondary}.

bryan-harter (approved):
Looks good to me!

PR #307, opened 2024-06-28
MNT: Pass around database session rather than creating new ones

greglucas:
# Change Summary

## Overview

We can refactor the database session handling to always begin with the lambda handler and then pass the session into all of the querying functions. This speeds up the database access by keeping the same connection alive for all queries rather than requesting a new one each time. This also makes the tests easier to mock and inject our test sessions by patching the module-level session factory.

## Testing

Rather than using a fixture to populate the databases, I switched it over to explicitly call the function from within the test so that we are only populating what needs to be populated and isolating the tests. I think this makes it more explicit what is happening, but I could be convinced to leave the setup/populate functions as a fixture if people prefer that style.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/database/database.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/database.py
@@ -2,9 +2,11 @@
 
 import json
 import os
+from contextlib import contextmanager
 
 import boto3
 from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
 
 
 def get_engine():
@@ -23,3 +25,15 @@ def get_engine():
     db_uri = f'postgresql://{db_config["username"]}:{db_config["password"]}@{db_config["host"]}:{db_config["port"]}/{db_config["dbname"]}'
 
     return create_engine(db_uri)
+
+
+@contextmanager
+class Session:
+    """Create session from engine.
+
+    Setting it up this way allows us to more easily mock this behavior in tests.
+    """
+
+    def __call__(self):
+        """Make a session from the engine."""
+        yield sessionmaker(get_engine())()
--- a/tests/lambda_endpoints/conftest.py
+++ b/tests/lambda_endpoints/conftest.py
@@ -4,10 +4,10 @@ from unittest.mock import patch
 
 import boto3
 import pytest
-from moto import mock_events, mock_s3
+from moto import mock_batch, mock_events, mock_s3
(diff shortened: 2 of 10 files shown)


sds_data_manager/lambda_code/SDSCode/database/database.py
```diff
@@ -23,3 +25,15 @@ def get_engine():
     db_uri = f'postgresql://{db_config["username"]}:{db_config["password"]}@{db_config["host"]}:{db_config["port"]}/{db_config["dbname"]}'
 
     return create_engine(db_uri)
+
+
+@contextmanager
+class Session:
+    """Create session from engine.
+
+    Setting it up this way allows us to more easily mock this behavior in tests.
```
tech3371, line 34: Thank you! I find these notes helpful because I know test setup can be tricky and it gives background of why we did this way.

tests/lambda_endpoints/conftest.py
```diff
@@ -77,17 +84,29 @@ def events_client():
...
-    """Create an in-memory SQLite database engine."""
-    with patch.object(db, "get_engine") as mock_engine:
-        engine = create_engine("sqlite:///:memory:")
-        mock_engine.return_value = engine
+def session():
+    """Create a test postgres database engine."""
+    from sds_data_manager.lambda_code.SDSCode.database import database as db
+
+    with patch.object(db, "Session") as mock_session:
+        connection = "sqlite:///:memory:"
```
tech3371, line 95: Does this setup postgres database engine?

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -32,37 +25,20 @@
...
+def _populate_dependency_table(session):
     """Add test data to database."""
-    # Setup: Add records to the database
-    with Session(db.get_engine()) as session:
-        session.add_all(all_dependents)
-        session.commit()
-        yield session
-        session.rollback()
-        session.close()
+    # We need to deepcopy these, otherwise there is an old
```
tech3371, line 30: helpful to know!

tech3371 (approved):
looks good to me!

tests/lambda_endpoints/conftest.py
greglucas: No, that is happening in the subsequent PRs. This is still only sqlite.

maxinelasp (approved):

PR #308, opened 2024-06-28
DB unique constraint

greglucas:
# Change Summary

## Overview
We were running into race conditions when adding multiple items to the database. This will prevent that by raising an integrity error when someone tries to commit a duplicate entry. It is a partial index where we want to keep track of multiple FAILED entries, but we only want one INPROGRESS or COMPLETED job to be in the table at a time.

There were a few updates required on this because of the switch to testing with postgres from sqllite. Partial unique indexes aren't supported in sqlite, and postgres is timezone aware with datetimes (sqlite apparently just stores strings of dates under the hood). This meant that our current tests were failing when switching over to a postgres backend, so I updated all tests so that they should pass with either postgres or sqllite.

This builds upon #307, so only the final commit has the updates that need to be reviewed for this PR.

## New Dependencies
* Adds pytest-postgresql for testing with a postgresql database.
* Bumps psycopg to 3+ (which is now named just psycopg)


greglucas:
Postgres is available on the ubuntu runners by default
https://github.com/actions/runner-images/?tab=readme-ov-file#package-managers-usage
which is nice, so we are getting the test cases we want with no skips.
```
============ 64 passed, 1 xfailed, 2 warnings in 133.40s (0:02:13) =============
```

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -133,6 +129,27 @@ class StatusTracking(Base):
     container_command = Column(String)
     processing_time = Column(Integer)
 
+    __table_args__ = (
+        # Partial unique index to ensure only one INPROGRESS or COMPLETED for a record
+        # We do want to allow multiple FAILED records
+        # NOTE: This does not work with sqllite (testing) DBs, only postgres
+        Index(
+            "idx_unique_status",
```
tech3371, line 137: Just curious why you had this name different than others. Is it used somewhere?

tests/lambda_endpoints/conftest.py
```diff
@@ -84,15 +86,31 @@ def events_client():
...
+POSTGRES_AVAILABLE = (
+    subprocess.run("which psql", shell=True, check=False).returncode == 0
+)
+
+if POSTGRES_AVAILABLE:
+
+    @pytest.fixture()
+    def connection(postgresql):
+        """Use a postgres connection string."""
+        return f"postgresql+psycopg://{postgresql.info.user}:@{postgresql.info.host}:{postgresql.info.port}/{postgresql.info.dbname}"
```
tech3371, line 99: Are these, user and host and port and so on, something we need to set locally for test or does it know it if we have `psql`?

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -404,3 +408,74 @@ def test_is_job_in_status_table(session):
         version="v001",
     )
     assert not result
+
+
+@pytest.mark.skipif(
+    not POSTGRES_AVAILABLE, reason="Only postgres supports partial unique indexes."
+)
```
tech3371, line 415: Thank you for handling this case!

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -404,3 +408,74 @@ def test_is_job_in_status_table(session):
...
+        status=models.Status.FAILED,
+        instrument="lo",
+        data_level="l1b",
+        descriptor="de",
+        start_date=datetime(2010, 1, 1),
+        version="v001",
+    )
+    session.add(record)
+    session.commit()
+    assert session.query(StatusTracking).count() == 5
```
tech3371, line 481: nice tests!

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: There is already a primary key index `id = Column(Integer, Identity(start=1, increment=1), primary_key=True)` on the database. This is a new (second) index which I've called `idx_unique_status`, so a new name for this index specifically and I'm indicating it is an index with `idx`.

tests/lambda_endpoints/conftest.py
greglucas: These are available from the `pytest-postgresql` test fixture called on entry to this fixture `connection(postgresql)` gives us that postgresql information for us (it spins up a DB in the background and spits out the specific connection information for us).

This won't get used if we don't have `psql` installed.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: Ah ok. Thank you for clarifying that. Makes sense now!

tests/lambda_endpoints/conftest.py
tech3371: good to know. Thank you!

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -396,7 +406,7 @@ def lambda_handler(event: dict, context):
                 "version": downstream_data["version"],
             }
 
-            update_status_table(status_params)
+            update_status_table(session, status_params)
```
maxinelasp: So is the basic idea that this call will throw an error if the session already exists? Does that error have enough information about why this job failed?

sds_data_manager/lambda_code/SDSCode/database_handler.py
```diff
@@ -29,58 +29,55 @@ def update_status_table(status_params):
...
+        if result is None:
+            logger.info(
+                "No existing record found, creating"
+                f" new record for {instrument}, {data_level},"
+                f"{descriptor}, {start_date}, {version}"
+            )
+            session.add(models.StatusTracking(**status_params))
+            session.commit()
+        else:
+            logger.info(f"Updating existing record, {result.__dict__}, with batch info")
```
maxinelasp: It seems like if the record is found, it doesn't attempt to write the instrument, version, etc, meaning we won't hit the unique constraint if the record already exists in the db...?

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -445,10 +400,82 @@ def test_is_job_in_status_table(populate_status_tracking_table):
...
         version="v001",
     )
     assert not result
+
+
+@pytest.mark.skipif(
+    not POSTGRES_AVAILABLE, reason="Only postgres supports partial unique indexes."
+)
+# Loop over all combinations of status attempts that should fail
+@pytest.mark.parametrize(
```
maxinelasp, line 417: This is a great test for the unique constraint, can you add a test for `update_status_table` that also checks for failed duplicates?

maxinelasp (changes requested):
I'm a little confused about if this would actually exercise the unique constraint in normal processing, could you clarify for me?

maxinelasp:
Is the change to make unique constraints work split across multiple reviews? Could you explain what each review entails? I see in #309 that you add some code that would hit the unique constraint.

greglucas:
@maxinelasp, sorry for this confusion with the multiple PRs and commits! To actually use the unique constraints I had to refactor the code even more because we were simultaneously updating and writing in the same function. So I pushed that portion off to the following PR #309. Here we should only focus on the second commit: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/308/commits/df22c8d26a98c286ad47118c3154624be369d613
The first commit is from #307 which is where I did a separate refactor to pass a single session around to all DB work rather than getting new sessions each time.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
greglucas: This one is associated with the first commit where we are passing a single session around rather than grabbing a new session each time we make a request to the status table. This avoids the latency associated with requesting a new connection to the engine.

sds_data_manager/lambda_code/SDSCode/database_handler.py
greglucas: Agreed, you are asking great questions here! This comes in the next PR.

tests/lambda_endpoints/test_batch_starter.py
greglucas: Yes, comes in the next PR I believe?

maxinelasp (approved):
Changing to an approval, will change my review on the last PR to be request changes to reflect the additional testing around batch starter.

PR #309, opened 2024-06-28
Db refactor

### LINKED ISSUE #306: Refactor of batch job kickoff and tracking process
There is actually an ERROR message just a little further down one of the Ultra logs.
```
[ERROR] KeyError: 'job_definition'
  File "/var/task/SDSCode/batch_starter.py", line 399, in lambda_handler
    update_status_table(status_params)
  File "/var/task/SDSCode/database_handler.py", line 60, in update_status_table
    result.job_definition = status_params["job_definition"]
```
This is coming from within update_status_table
[https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/d6f2d7c67e[…]108b2a/sds_data_manager/lambda_code/SDSCode/database_handler.py](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/d6f2d7c67ed7f1acb3d11b9f7ff77030ba108b2a/sds_data_manager/lambda_code/SDSCode/database_handler.py#L47-L64)
and we are actually hitting the first path and writing the record in the first log. Then going down the other path in the second log. But then, we are getting a Python exception thrown here because the job_definition key doesn't exist, and that exception kills the lambda and therefore we don't actually get the second batch job somewhat luckily :slightly_smiling_face: So that is why there wasn't two batch jobs kicked off, even though theoretically we should have.
All of this being said, I think we should do multiple changes to address this:
* Reduce the number of DB connections we are making and pass the session around to our other functions more often. So request a DB connection right away in the lambda handler and then pass it through. (Right now we are making a session in batch_starter and a second one in update_status_table, and that added a pretty significant delay of ~0.5s)
* Update our Unique Constraint on the database to include all of the specific properties that we expect to be unique for each job so we can't get duplicates (remove ID). [https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/d6f2d7c67e[…]a108b2a/sds_data_manager/lambda_code/SDSCode/database/models.py](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/d6f2d7c67ed7f1acb3d11b9f7ff77030ba108b2a/sds_data_manager/lambda_code/SDSCode/database/models.py#L113-L118)
* With the unique constraint in place, switch to a try/except around writing to the DB. A write will be guaranteed to fail if a record already exists and we can catch that exception and know a job already exists. This will avoid the time delay between query / write.
* Move to a queue-based job starter system that 
[@Maxine](https://lasp.slack.com/team/UC2209B41)
 is working on.

I think all of the first 3 bullets are relatively minor updates and we should probably add them in regardless of the queue system as a backup and best practice, unless the queue system removes the DB work entirely Maxine?
[database_handler.py](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/d6f2d7c67ed7f1acb3d11b9f7ff77030ba108b2a/sds_data_manager/lambda_code/SDSCode/database_handler.py)
<https://github.com/[IMAP-Science-Operations-Center/sds-data-manager](https://github.com/IMAP-Science-Operations-Center/sds-data-manager)|IMAP-Science-Operations-Center/sds-data-manager>IMAP-Science-Operations-Center/sds-data-manager | Added by [GitHub](https://lasp.slack.com/services/B02RT0NPJNL)
[models.py](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/d6f2d7c67ed7f1acb3d11b9f7ff77030ba108b2a/sds_data_manager/lambda_code/SDSCode/database/models.py)
<https://github.com/[IMAP-Science-Operations-Center/sds-data-manager](https://github.com/IMAP-Science-Operations-Center/sds-data-manager)|IMAP-Science-Operations-Center/sds-data-manager>IMAP-Science-Operations-Center/sds-data-manager | Added by [GitHub](https://lasp.slack.com/services/B02RT0NPJNL)

greglucas:
# Change Summary

## Overview

This is a major refactor of the batch job submission and updating. I have renamed the table to "Processing Job Table" and removed all references to "status tracking".

I find this logic more serial and easier to follow, but please let me know if you find this harder to follow or anything isn't clear. I tried to avoid the upstream/downstream terminologies when used together and instead switched to "potential jobs" for the downstream case. So we iterate through potential jobs. For each potential job we look "upstream" to see if the required files are available. If any are missing, abort. If all are available, try to submit the record. The record submission will fail if it isn't unique, so we just abort and log that there was already a record for that case.

I also added the unique ID from the table to the job-name so that we can directly get the proper record to update in the database. Before we were just getting the first, so not guaranteed to hit the right job to update if we had multiple failures or attempts.

This builds upon #308 with the final two commits here being relevant for review. Refactoring the batch starter and the indexer functions.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -7,20 +7,23 @@ from datetime import datetime
 import boto3
 from imap_data_access import ScienceFilePath
 from sqlalchemy import select
+from sqlalchemy.exc import IntegrityError
 
 from .database import database as db
 from .database import models
-from .database_handler import update_status_table
 
 # Logger setup
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
+# Create a batch client
+BATCH_CLIENT = boto3.client("batch", region_name="us-west-2")
 
-def get_dependency(
+
+def get_dependencies(
     session, instrument, data_level, descriptor, direction, relationship
 ):
-    """Make query to dependency table to get dependency.
+    """Make a query to the dependency table to get all dependencies.
 
     Parameters
     ----------
@@ -40,10 +43,10 @@ def get_dependency(
 
     Returns
     -------
-    dependency : list
+    dependencies : list
         List of dictionary containing the dependency information.
     """
-    dependency = []
+    dependencies = []
 
     query = select(models.PreProcessingDependency.__table__).where(
         models.PreProcessingDependency.primary_instrument == instrument,
@@ -55,25 +58,29 @@ def get_dependency(
     results = session.execute(query).all()
 
     for result in results:
-        dependency.append(
+        dependencies.append(
             {
                 "instrument": result.dependent_instrument,
                 "data_level": result.dependent_data_level,
                 "descriptor": result.dependent_descriptor,
             }
         )
-    return dependency
+    return dependencies
 
 
-def query_instrument(session, upstream_dependency, start_date, version):
+def get_file(session, instrument, data_level, descriptor, start_date, version):
     """Query to database to get the first FileCatalog record.
 
     Parameters
     ----------
     session : orm session
         Database session.
-    upstream_dependency : dict
-        Dictionary of upstream dependency.
+    instrument : str
+        Instrument name.
+    data_level : str
+        Data level.
+    descriptor : str
+        Data descriptor.
     start_date : str
         Start date of the event data.
     version : str
@@ -85,10 +92,6 @@ def query_instrument(session, upstream_dependency, start_date, version):
         The first FileCatalog record matching the query criteria.
         None is returned if no record matches the criteria.
     """
-    instrument = upstream_dependency["instrument"]
-    data_level = upstream_dependency["data_level"]
-    descriptor = upstream_dependency["descriptor"]
-
     # TODO: narrow down the query using end_date.
     # This will give ability to query range of time.
     # Eg. when we are query 3 months of data to create
@@ -101,9 +104,9 @@ def query_instrument(session, upstream_dependency, start_date, version):
         .filter(
             models.FileCatalog.instrument == instrument,
             models.FileCatalog.data_level == data_level,
-            models.FileCatalog.version == version,
             models.FileCatalog.descriptor == descriptor,
             models.FileCatalog.start_date == datetime.strptime(start_date, "%Y%m%d"),
+            models.FileCatalog.version == version,
         )
         .first()
     )
@@ -111,7 +114,7 @@ def query_instrument(session, upstream_dependency, start_date, version):
     return record
 
 
-def query_downstream_dependencies(session, filename_components):
+def get_downstream_dependencies(session, filename_components):
     """Get information of downstream dependents.
 
     Parameters
@@ -127,7 +130,7 @@ def query_downstream_dependencies(session, filename_components):
         Dictionary containing components with dates and versions appended.
     """
     # Get downstream dependency data
-    downstream_dependents = get_dependency(
+    downstream_dependents = get_dependencies(
         session=session,
         instrument=filename_components["instrument"],
         data_level=filename_components["data_level"],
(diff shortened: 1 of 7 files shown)


sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -290,158 +185,188 @@ def is_job_in_status_table(
...
-                jobName=job_name,
-                jobQueue=job_queue,
-                jobDefinition=job_definition,
-                containerOverrides={
-                    "command": batch_command,
-                },
-            )
-            logger.info(f"Submitted job - {response}")
+        for job in potential_jobs:
+            submit_job(session, batch_client, job)
```
tech3371: can you add note here that you check for upstream availability before submitting job? The name made it seem like it's submitting job which it is but I didn't realized the check for upstream was in there as well.

Also this function name is same as what Batch uses to submit job. That could be why I got confused.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -290,158 +185,188 @@ def is_job_in_status_table(
...
+        descriptor=descriptor,
+        direction="UPSTREAM",
+        relationship="HARD",
+    )
+
+    for upstream_dependency in upstream_dependencies:
+        instrument = upstream_dependency["instrument"]
+        data_level = upstream_dependency["data_level"]
+        descriptor = upstream_dependency["descriptor"]
+        # Check to see if each upstream dependency file is available
```
tech3371, line 260: As I am thinking about it more, the version of upstream could be different if upstream dependent is different instrument. Because instrument could increment version at different cadence. We can refactor that in the future work. For the moment, can you add TODO here to look up latest version or some specific version?

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -290,158 +185,188 @@ def is_job_in_status_table(
...
+        record = get_file(
+            session, instrument, data_level, descriptor, start_date, version
+        )
+        if record is None:
+            logger.info(
+                f"Dependency not found: {instrument}, "
+                f"{data_level}, "
+                f"{descriptor}, "
+                f"{version}"
```
tech3371: can we log start date as well?

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -290,158 +185,188 @@ def is_job_in_status_table(
...
+            )
+            return False  # Exit the loop early as we already found a missing dependency
+
+        # Add additional information to the upstream dependency
+        # TODO: Should we actually be using the record's information?
+        #      We are currently using our upstream dependency, but it seems like we
+        #      might want to use the file record itself as our source of truth.
```
tech3371: This comment feels outdated but the concept may be same as what I mentioned in previous comment.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -290,158 +185,188 @@ def is_job_in_status_table(
...
+        additional_info = {
+            "start_date": start_date,
+            "version": version,
+        }
+        upstream_dependency.update(additional_info)
+
```
tech3371, line 287: Can we add info log saying that it found all dependency needed? It was helpful to trace what's going on in processing.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -290,158 +185,188 @@ def is_job_in_status_table(
...
+        "--version",
+        version,
+        "--dependency",
+        f"{upstream_dependencies}",
+        "--upload-to-sdc",
+    ]
+
+    # NOTE: The batch job name should contain only alphanumeric characters and hyphens
+    # Eg. "codice-l1a-job"
+    job_name = f"{instrument}-{data_level}-" f"{descriptor}-job-{processing_job.id}"
```
tech3371:
```suggestion
    job_name = f"{instrument}-{data_level}-{descriptor}-job-{processing_job.id}"
```

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -248,36 +250,26 @@ def batch_event_handler(event):
...
-        "version": version,
-        "job_definition": event["detail"]["jobDefinition"],
-        "job_log_stream_id": event["detail"]["container"]["logStreamName"],
-        "container_image": event["detail"]["container"]["image"],
-        "container_command": " ".join(command),
-    }
-
-    update_status_table(status_params)
+    # We injected our table ID into the job name
+    job_id = event["detail"]["jobName"].split("-")[-1]
```
tech3371, line 261: I first missed that we are adding table ID into the job name. Can you add comment in batch starter lambda that this is added which will used later to look up and update with batch information?

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -248,36 +250,26 @@ def batch_event_handler(event):
...
+        job = session.get(models.ProcessingJob, job_id)
+        # Make the updates
+        job.status = job_status
+        job.job_definition = event["detail"]["jobDefinition"]
+        job.job_log_stream_id = event["detail"]["container"]["logStreamName"]
+        job.container_image = event["detail"]["container"]["image"]
+        job.container_command = " ".join(event["detail"]["container"]["command"])
+        session.commit()
 
     return http_response(status_code=200, body="Success")
```
tech3371, line 274: I remember we had discussion about refactoring this `batch_event_handler` into it's own lambda. Should we save it for future PR?

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -1,68 +1,47 @@
...
+def _populate_dependency_table(session):
     """Add test data to database."""
-    # Setup: Add records to the database
-    with Session(db.get_engine()) as session:
-        session.add_all(all_dependents)
-        session.commit()
-        yield session
-        session.rollback()
-        session.close()
+    # We need to deepcopy these, otherwise there is an old
```
tech3371: nice comments!

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -137,46 +116,25 @@ def test_file_catalog_simulation(test_engine):
...
-        session.commit()
-
-    return session
+    session.add_all(test_records)
+    session.commit()
 
 
-@pytest.fixture()
-def populate_status_tracking_table(test_engine):
+def _populate_status_tracking_table(session):
```
tech3371: should we rename this test function to match new table name?

tech3371 (commented):
Really nice improvement and so detailed work. Thank you so much!

sds_data_manager/lambda_code/SDSCode/indexer.py
greglucas: Yes, I would prefer to do it later because that involves updating the Lambda event listeners. Right now, this PR is completely separate from all AWS work.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
greglucas: Yes, good idea. I changed it to `try_to_submit_job`, does that help or do you have a different suggestion?

sds_data_manager/lambda_code/SDSCode/batch_starter.py
greglucas: :+1: Yes, I updated the comment and made this more explicit by calling these `upstream_instrument`, ... to add the explicit upstream. Then setting `upstream_version = version` to make it clear we are getting the upstream version from the current job's version, and adding a note about that.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
greglucas: Yep, done.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
greglucas: Yeah, this is definitely confusing right now and it is related to your other comment. I moved this logic up and updated the comments to make it more explicit that we should be changing these in the future. Right now we are using the input-job's start date and version number.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
greglucas: Yes, done.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
greglucas: Thanks! Updated.

sds_data_manager/lambda_code/SDSCode/indexer.py
greglucas: Yes, I added a comment about this. Putting more information into the job name is what I think will help us avoid having to parse the command and get information out of it. If we can be explicit about our job names and add details there that should be easier than string splitting and messing with our CLI commands as well so our CLI doesn't need to be updated in the future, and we could just update our job naming convention.

tests/lambda_endpoints/test_batch_starter.py
greglucas: Yes, thanks! I also updated other old status tracking comments.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
tech3371: That sounds good to me

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -288,160 +183,197 @@ def is_job_in_status_table(
...
+        upstream_data_level = upstream_dependency["data_level"]
+        upstream_descriptor = upstream_dependency["descriptor"]
+
+        # Check to see if each upstream dependency file is available
+        # TODO: Update start_date / version request to be more specific
+        #       Currently we are using the same as the job product, but the versions
+        #       may not match exactly if one dependency updates before another
+        upstream_start_date = start_date
+        upstream_version = version
+        upstream_dependency.update(
```
maxinelasp, line 266: I like these changes and it seems like it will make it easier to accommodate weird complicated dependencies

maxinelasp (approved):
This, combined with #308, looks good to me

maxinelasp (changes requested):
Per comments on #308 , I'd prefer to see some batch_starter tests, or at least that you tested manually and validated that the unique constraint does prevent processing in this case.

tech3371 (approved):
Thank you for tackling this! I couldn't do it in time for SIT-3 and I know this is not easy.

maxinelasp (approved):
Looks good, thanks for doing this!

PR #311, opened 2024-07-02
Refactor Lambda to use Lambda Layer

tech3371:
# Change Summary

## Overview
Refactor Lambda code to use layer instead. This will allow code visibility on Lambda and allow easy update and testing.

Summary of some things I tried in the process of updating this:
I tried to create generic lambda layer which could be used as input to other stacks. But it requires to create stack output value using cdk.cfnOutput and then pass that value and use it to import layer stack's information into stack that uses it. With this, it requires to add dependency between layer stack and stacks that uses it. Also, it will mean that we need to be careful of what we call cdk.cfnOutput key as it's used by other stacks. I was concerned that it could cause room for error and too many dependency could introduce complication in updating some stack and could become roadblocks to making improvement.

We can revisit this if we find better solution in the near future. This could be first step to this refactoring.


tech3371:
Going to refactor some more.

PR #314, opened 2024-07-16
I-ALiRT: Added DynamoDB Streams for Packet Ingest

laspsandoval:
# Change Summary

## Overview
I-ALiRT: Added DynamoDB Streams for Packet Ingest

## New Files
- ialirt_database_stack.py
   - creation of DynamoDB streams
- ialirt_ingest_lambda_stack.py
   - contains Eventbridge which kicks off lambda
   - defines environment for lambda that ingests data to DynamoDB streams
- ialirt_ingest.py
   - lambda that ingests data to DynamoDB streams
 - test_ialirt_database_stack.py
   - test ialirt_database_stack
- test_ialirt_ingest_lambda.py
   - test ialirt ingest lambda

## Updated Files
- stackbuilder.py
   - ability to deploy ialirt_ingest_lambda_stack and ialirt_database_stack

## Testing
* Verified that upon placing a file in s3 that the DynamoDB becomes populated as expected in AWS.

## Diagram
https://lasp.colorado.edu/galaxy/display/IMAP/I-ALiRT+SDC+Architecture+Diagram


sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -0,0 +1,52 @@
...
+            "sct_vtcw_reset#sct_vtcw": "0#2025-07-11T12:34:56Z",
+            "packet_blob": b"binary_data_string",
+        }
+
+        table.put_item(Item=item)
+        logger.info("Successfully wrote item to DynamoDB: %s", item)
+
+    except Exception as e:
+        logger.error("Error processing event: %s", str(e))
+        raise
```
greglucas: You can remove this entire try/except block since it doesn't add anything and you're re-raising. The log will contain the raised exception message.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -0,0 +1,52 @@
...
+
+        s3_filepath = event["detail"]["object"]["key"]
+        filename = os.path.basename(s3_filepath)
+        logger.info("Retrieved filename: %s", filename)
+
+        # TODO: item is temporary and will be replaced with actual packet data.
+        item = {
+            "sct_vtcw_reset#sct_vtcw": "0#2025-07-11T12:34:56Z",
+            "packet_blob": b"binary_data_string",
+        }
```
greglucas: Maybe I'm not following this, but, is the plan to put all the items into the database *and* s3? I thought we were going to put the packet into the database directly, but if you are triggering off an s3 event then that means the data is already in s3 so do you want to just leave the items there and put the s3 key into the DB?

Right now you're putting the temporary string into the database, what about reading the incoming even filepath and putting the read binary content into the DB so at least it isn't hard-coded?

sds_data_manager/stacks/ialirt_database_stack.py
```diff
@@ -0,0 +1,52 @@
...
+        super().__init__(scope, construct_id, **kwargs)
+
+        self.packet_data_table = ddb.Table(
+            self,
+            "PacketDataTable",
+            table_name="imap-packetdata-table",
+            # Change to RemovalPolicy.RETAIN to keep the table after stack deletion.
+            removal_policy=RemovalPolicy.DESTROY,
+            # Restore data to any point in time within the last 35 days.
+            point_in_time_recovery=False,
```
greglucas: I would think we'd want both of these things enabled?

sds_data_manager/stacks/ialirt_database_stack.py
```diff
@@ -0,0 +1,52 @@
...
+            # Partition key (PK) = spacecraft time ugps.
+            partition_key=ddb.Attribute(
+                name="sct_vtcw_reset#sct_vtcw",
+                type=ddb.AttributeType.STRING,
+            ),
```
greglucas: Do we need to stick to ops-software shorthand names or can we be more descriptive in our names (what is ugps? We don't have GPS on IMAP, so is it equivalent to an earth received time, or MET, something else...). 
Also, do we want this as the PK alone, or do you want to split this into `sct_vtcw_reset` as the PK, and `sct_vtcw` as the SK?

```suggestion
            # Partition key (PK) = spacecraft time ugps.
            partition_key=ddb.Attribute(
                name="spacecraft_time",  # MET?
                type=ddb.AttributeType.STRING,
            ),
```

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
```diff
@@ -0,0 +1,124 @@
...
+                    "dynamodb:BatchGetItem",
+                    "dynamodb:DescribeStream",
+                    "dynamodb:DescribeTable",
+                    "dynamodb:GetItem",
+                    "dynamodb:ListStreams",
+                    "dynamodb:ListTables",
+                    "dynamodb:Query",
+                    "dynamodb:Scan",
+                    "dynamodb:UpdateItem",
+                    "dynamodb:DeleteItem",
```
greglucas: Might be easier to add a DynamoDB managed policy in the list above?

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
```diff
@@ -0,0 +1,124 @@
...
+                    "s3:GetObject",
+                ],
+                resources=[
+                    packet_data_table.table_arn,
+                    f"{ialirt_bucket.bucket_arn}/*",
+                ],
+            )
+        )
+
+        ialirt_ingest_lambda = lambda_alpha_.PythonFunction(
```
greglucas, line 114: I don't think you are using any Python dependencies, so this should just be a standard `aws_lambda.Function`.

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
```diff
@@ -0,0 +1,124 @@
...
+            self,
+            id="IalirtIngestLambda",
+            function_name="ialirt-ingest",
+            entry=str(
+                pathlib.Path(__file__).parent.joinpath("..", "lambda_code").resolve()
+            ),
+            index="IAlirtCode/ialirt_ingest.py",
+            handler="lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_9,
```
greglucas: Please bump this to 3_12, we don't want to have to upgrade off an old runtime soon after launch.

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+
+        packet_data_table.grant_read_write_data(ialirt_ingest_lambda)
+
+        # The resource is deleted when the stack is deleted.
+        ialirt_ingest_lambda.apply_removal_policy(cdk.RemovalPolicy.DESTROY)
+
+        # Event that triggers Lambda:
+        # Arrival of packets in s3 bucket.
+        ialirt_data_arrival_rule = events.Rule(
```
greglucas: I thought there was potentially also an option for you to send ops-software a python script that would send data to dynamodb directly that they could invoke from their tools as a subprocess. Is every packet going to be saved as an individual file in this scenario, or is there going to be a pool of packets in a single s3 file that you'll need to parse later?

Living within their container as `write_to_dynamodb.py`:
```python
import boto3
table = boto3.resource('dynamodb').Table(TABLE_NAME)
table.put_item(
       Item={
            'spacecraft_time': time,
            "packet": packet_data_from_command_line
        }
    )
```

sds_data_manager/utils/stackbuilder.py
```diff
@@ -18,6 +18,8 @@
     ecr_stack,
     efs_stack,
     ialirt_bucket_stack,
+    ialirt_database_stack,
+    ialirt_ingest_lambda_stack,
```
greglucas: These are coupled together, could we turn this into a single stack?

I still would prefer one "ialirt" stack I think since it is a "deployable unit" all together IMO.

tests/infrastructure/test_ialirt_database_stack.py
```diff
@@ -0,0 +1,62 @@
...
+            KeySchema=[
+                # Partition key
+                {"AttributeName": "sct_vtcw_reset#sct_vtcw", "KeyType": "HASH"},
+            ],
+            AttributeDefinitions=[
+                {"AttributeName": "sct_vtcw_reset#sct_vtcw", "AttributeType": "S"},
+            ],
+            BillingMode="PAY_PER_REQUEST",
+        )
+        yield dynamodb
```
greglucas: can you yield the table itself here? You are connecting to the table in each of the other tests below this.

```suggestion
        yield table
```

tests/infrastructure/test_ialirt_database_stack.py
```diff
@@ -0,0 +1,62 @@
...
+    return item
+
+
+def test_query_by_sct_vtcw(dynamodb, populate_table):
+    """Test to query irregular packet length."""
+    table = dynamodb.Table(TABLE_NAME)
+
+    response = table.query(
+        KeyConditionExpression=Key("sct_vtcw_reset#sct_vtcw").eq(
+            "0#2025-07-12T12:34:57Z"
```
greglucas: Since this is in UTC, I'm guessing this is Earth Received Time, or is this already SPICE corrected by ops-software to get the packet MET into UTC?

tests/lambda_endpoints/test_ialirt_ingest.py
```diff
@@ -0,0 +1,55 @@
...
+
+import boto3
+import pytest
+from moto import mock_dynamodb
+
+from sds_data_manager.lambda_code.IAlirtCode.ialirt_ingest import lambda_handler
+
+
+@pytest.fixture()
+def dynamodb_table():
```
greglucas: Do we need to mock this again here? I think we can put it into a `conftest.py` to share.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: The s3 bucket is mounted to the running container, which was what was requested by Steve Monk and Evan Brookens. See Slack message for database layout.

sds_data_manager/stacks/ialirt_database_stack.py
laspsandoval: I'll put a TODO to change for production.

tests/infrastructure/test_ialirt_database_stack.py
laspsandoval: sct_vtcw = spacecraft time ugps.

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
laspsandoval: We talked about this a while back and Steve Monk and Evan Brookens said they wanted to mount to s3, which was what the previous pr was about. Each file in s3 contains multiple packets.

sds_data_manager/stacks/ialirt_database_stack.py
laspsandoval: Ah, that's true. It would be MET. I can make that change. Let's discuss database layout in Slack. I could see us going back and forth for that.

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
greglucas: Chatted about this on a call and I was mistaken about the mapping here between s3 and dynamodb.

s3 will have a 1-minute pkts file written to it containing multiple packets. This function will then loop over those packets and extract them into DynamoDB, which will have one entry per packet.

Ignore this suggestion.

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
laspsandoval: I will use space-packet-parser.

sds_data_manager/utils/stackbuilder.py
bryan-harter: An ialirt stack also makes sense to me.

sds_data_manager/stacks/ialirt_database_stack.py
laspsandoval: We want to have High Cardinality: break up the partition key in as small pieces as possible.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -0,0 +1,47 @@
...
+    dynamodb = boto3.resource("dynamodb")
+    table = dynamodb.Table(table_name)
+
+    s3_filepath = event["detail"]["object"]["key"]
+    filename = os.path.basename(s3_filepath)
+    logger.info("Retrieved filename: %s", filename)
+
+    # TODO: item is temporary and will be replaced with actual packet data.
+    item = {
+        "reset_number#met": "0#2025-07-11T12:34:56Z",
```
greglucas:
```suggestion
        "reset_number#met": "0#123",
```

sds_data_manager/stacks/ialirt_database_stack.py
```diff
@@ -0,0 +1,54 @@
...
+            table_name="imap-packetdata-table",
+            # Change to RemovalPolicy.RETAIN to keep the table after stack deletion.
+            # TODO: change to RETAIN in production.
+            removal_policy=RemovalPolicy.DESTROY,
+            # Restore data to any point in time within the last 35 days.
+            # TODO: change to True in production.
+            point_in_time_recovery=False,
+            # Partition key (PK) = spacecraft time ugps.
+            partition_key=ddb.Attribute(
+                name="sct_vtcw_reset#sct_vtcw",
```
greglucas: I think this needs to match your Lambda insert function above.

```suggestion
                name="reset_number#met",
```

I also think that including "reset" in the key might be hard to include in your queries and what happens when a reset does happen, will your processing code have to do a scan over all items to make sure we are still on reset 0, and where do you get that information from in the CCSDS packet? I think all you'll have access to is the SCLK / MET.

sds_data_manager/stacks/ialirt_database_stack.py
```diff
@@ -0,0 +1,54 @@
...
+                type=ddb.AttributeType.STRING,
+            ),
+            # Enable DynamoDB streams for real-time processing
+            stream=ddb.StreamViewType.NEW_IMAGE,
```
greglucas: Can this be added on/updated later? We aren't using the stream yet per @bryan-harter's comment.

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
```diff
@@ -0,0 +1,159 @@
...
+            managed_policies=[
+                iam.ManagedPolicy.from_aws_managed_policy_name(
+                    "service-role/AWSLambdaBasicExecutionRole"
+                )
+            ],
+        )
+
+        lambda_role.add_managed_policy(
+            iam.ManagedPolicy.from_aws_managed_policy_name("AmazonDynamoDBFullAccess")
+        )
```
greglucas: Move this into the managed policies above.
```suggestion
                )
            iam.ManagedPolicy.from_aws_managed_policy_name("AmazonDynamoDBFullAccess"),
            ],
        )
```

tests/infrastructure/test_ialirt_ingest_lambda_stack.py
```diff
@@ -0,0 +1,33 @@
...
+import pytest
+from boto3.dynamodb.conditions import Key
+
+
+@pytest.fixture()
+def populate_table(table):
+    """Populate DynamoDB table."""
+    items = [
+        {
+            "reset_number#met": "0#2025-07-11T12:34:56Z",
```
greglucas: MET should be an integer value
```suggestion
            "reset_number#met": "0#123",
```

tests/lambda_endpoints/conftest.py
greglucas, line 1: If you move this file up one more level to the general `tests/` directory, then this table can be shared amongst the lambda and infrastructure tests.

sds_data_manager/stacks/ialirt_database_stack.py
laspsandoval: From conversations with Steve Monk a reset would be a big deal. For now I plan on assigning a value in the lambda to a reset_number = 0. If a reset does occur we will be notified via email and I would have to manually go in, change the reset_number in the lambda, and redeploy.

sds_data_manager/stacks/ialirt_database_stack.py
laspsandoval: It looks like I have options for using streams, so I'll save this for another PR.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -0,0 +1,47 @@
...
+    dynamodb = boto3.resource("dynamodb")
+    table = dynamodb.Table(table_name)
+
+    s3_filepath = event["detail"]["object"]["key"]
+    filename = os.path.basename(s3_filepath)
+    logger.info("Retrieved filename: %s", filename)
+
+    # TODO: item is temporary and will be replaced with actual packet data.
+    item = {
+        "reset_number#met": "0#123",
```
greglucas: I don't quite understand how you'll get the reset number when writing to the DB since it isn't in the CCSDS packet, or am I mistaken here. I think you can only use MET. Maybe ops-software gets more information out of the "Frame" that wraps the packets which we won't have?

sds_data_manager/stacks/ialirt_database_stack.py
```diff
@@ -0,0 +1,55 @@
...
+            self,
+            "PacketDataTable",
+            table_name="imap-packetdata-table",
+            # Change to RemovalPolicy.RETAIN to keep the table after stack deletion.
+            # TODO: change to RETAIN in production.
+            removal_policy=RemovalPolicy.DESTROY,
+            # Restore data to any point in time within the last 35 days.
+            # TODO: change to True in production.
+            point_in_time_recovery=False,
+            # Partition key (PK) = spacecraft time ugps.
```
greglucas:
```suggestion
            # Partition key (PK) = Mission Elapsed Time (MET).
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
greglucas: To maybe illustrate why we wouldn't want to include the reset number is that you won't want to do two queries in the future for data if we have a reset. Because you wouldn't know what `0#` or `1#` or `2#`, ... to prepend to the query. DynamoDB is more focused on a good key for the retrieval than the initial storage.

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Create Event Rule
+        self.create_event_rule(ialirt_bucket, self.ialirt_ingest_lambda)
+
+    def create_dynamodb_table(self) -> aws_dynamodb.Table:
+        """Create and return the DynamoDB table."""
+        table = ddb.Table(
+            self,
+            "PacketDataTable",
+            table_name="imap-packetdata-table",
```
greglucas:
```suggestion
            "IalirtPacketDataTable",
            table_name="ialirt-packetdata-table",
```

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
```diff
@@ -0,0 +1,158 @@
...
+
+        # Create Lambda Function
+        self.ialirt_ingest_lambda = self.create_lambda_function(
+            ialirt_bucket, self.packet_data_table
+        )
+
+        # Create Event Rule
+        self.create_event_rule(ialirt_bucket, self.ialirt_ingest_lambda)
+
+    def create_dynamodb_table(self) -> aws_dynamodb.Table:
```
greglucas, line 56: Why do we have two Tables being created? We have this one here, but we also have the one in `ialirt_database_stack` As far as I can tell the setup looks identical.

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
```diff
@@ -0,0 +1,158 @@
...
+        ialirt_data_arrival_rule = events.Rule(
+            self,
+            "IalirtDataArrival",
+            rule_name="ialirt-data-arrival",
+            event_pattern=events.EventPattern(
+                source=["aws.s3"],
+                detail_type=["Object Created"],
+                detail={
+                    "bucket": {"name": [ialirt_bucket.bucket_name]},
+                    "object": {"key": [{"prefix": "packets/"}]},
```
greglucas, line 155: Do we want to mimic the main IMAP bucket naming?
```suggestion
                    "object": {"key": [{"prefix": "imap/ialirt/l0/"}]},
```

tests/lambda_endpoints/test_ialirt_ingest.py
```diff
@@ -0,0 +1,52 @@
...
+    for item in items:
+        table.put_item(Item=item)
+
+    return item
+
+
+def test_lambda_handler(table):
+    """Test the lambda_handler function."""
+    # Mock event data
+    event = {"detail": {"object": {"key": "path/to/s3/object/file.txt"}}}
```
greglucas: Shouldn't this be the path to `packets/` currently? I guess this means it isn't testing the event triggering itself here.

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
laspsandoval: Yes. I meant to delete that.

sds_data_manager/stacks/ialirt_ingest_lambda_stack.py
laspsandoval: Our s3 bucket is named ialirt-<acct#>. It seems odd to add imap/ialirt as a directory since everything in the bucket will be ialirt. I think packets is fine. It is what Steve Monk said he wanted and I don't think it will confuse anyone.

tests/lambda_endpoints/test_ialirt_ingest.py
laspsandoval: When I change the TODO in lambda_handler so that I am actually getting the packet data I will update this test. For now I changed to name for the directory.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: The reason for the reset_number is that we need to have a unique partition key. In the event that the spacecraft clock is reset then we could have non-unique mets, which would be a big problem. So my initial plan was to change the value of the reset number and redeploy if that happened, but that wouldn't be an automated approach so I think you are right. This need more consideration.

It might be better to put in a sort key of mission date. If the spacecraft had a reset then the combination of partition key and sort key would still be unique. 

I'll take out reset# and put in the mission date as the sort key.

bryan-harter (approved):
This all looks good to me!

PR #315, opened 2024-07-19
Sqs processing

maxinelasp:
# Change Summary
Added a queue in front of batch_processing per our discussion about eliminating database race conditions. This did involve an update to batch_starter to change the events getting passed in, but beyond that there are minimal impacts.

## Overview

Rather than eventbridge events directly triggering batch_starter, the events instead go into a queue. 

This is a FIFO queue with multiple message groups. Right now, the message groups are divided by instrument, but they could be further divided. Within the group, they are handled in a FIFO manner, meaning they will be processed in order and processing won't proceed until each event is complete. However, multiple groups are processed in parallel. 

I created an eventbridge rule for each instrument that sorts the events into the queue message groups. This required adding an additional piece of information to each event - instrument name. 

Then, the target for the batch_starter is switched from eventbridge to lambda and everything else is handled.

Notes:
- SQS by default has some batching properties, so I updated batch_starter to loop through each event in the batch
- I looked into having one eventbridge rule for the queue, but I couldn't figure out how to extract the message detail of the event
- Adding additional message groups will create increased parallelism and therefore increase processing speed. Due to the batching, I don't think this will be necessary. 
- We could turn off batching but I like it in this case, it keeps the lambdas warm

## New Files

- sqs_stack.py
   - Creates SQS queue and eventbridge rules

## Updated Files

- BatchStarterStack
   - changed source target from eventbridge to SQS
- batch_starter.py
   - Changed input to a list of SQS events

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -362,6 +362,7 @@ def try_to_submit_job(session, job_info):
 
 def lambda_handler(event: dict, context):
     """Lambda handler."""
+    # TODO: UPdate event to be an eventsourcemapping from SQS
     logger.info(f"Event: {event}")
     logger.info(f"Context: {context}")
 
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -1,5 +1,6 @@
 """Functions for supporting the batch starter component of the architecture."""
 
+import json
 import logging
 import os
 from datetime import datetime
@@ -360,21 +361,27 @@ def try_to_submit_job(session, job_info):
     logger.info(f"Submitted job {job_name} with this command: {batch_command}")
 
 
-def lambda_handler(event: dict, context):
+def lambda_handler(events: dict, context):
     """Lambda handler."""
-    # TODO: UPdate event to be an eventsourcemapping from SQS
-    logger.info(f"Event: {event}")
+    logger.info(f"Event: {events}")
     logger.info(f"Context: {context}")
 
-    # Event details:
-    filename = event["detail"]["object"]["key"]
-    components = ScienceFilePath.extract_filename_components(filename)
-    logger.info(f"Initial event parsed filename: {components}")
+    for event in events["Records"]:
+        # Event details:
+        logger.info(f"Individual event: {event}")
+        body = json.loads(event["body"])
 
-    with db.Session() as session:
-        # Potential jobs are the instruments that depend on the current file.
-        potential_jobs = get_downstream_dependencies(session, components)
-        logger.info(f"Potential jobs found [{len(potential_jobs)}]: {potential_jobs}")
+        filename = body["detail"]["object"]["key"]
+        logger.info(f"Retrieved filename: {filename}")
+        components = ScienceFilePath.extract_filename_components(filename)
+        logger.info(f"Initial event parsed filename: {components}")
 
-        for job in potential_jobs:
-            try_to_submit_job(session, job)
+        with db.Session() as session:
+            # Potential jobs are the instruments that depend on the current file.
+            potential_jobs = get_downstream_dependencies(session, components)
+            logger.info(
+                f"Potential jobs found [{len(potential_jobs)}]: {potential_jobs}"
+            )
+
+            for job in potential_jobs:
+                try_to_submit_job(session, job)
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -363,20 +363,21 @@ def try_to_submit_job(session, job_info):
 
 def lambda_handler(events: dict, context):
     """Lambda handler."""
-    logger.info(f"Event: {events}")
+    logger.info(f"Events: {events}")
     logger.info(f"Context: {context}")
 
-    for event in events["Records"]:
-        # Event details:
-        logger.info(f"Individual event: {event}")
-        body = json.loads(event["body"])
+    with db.Session() as session:
+        # Since the SQS events can be batched together, we need to loop through
+        # each event. In this loop, "event" represents one file landing.
+        for event in events["Records"]:
+            # Event details:
+            logger.info(f"Individual event: {event}")
+            body = json.loads(event["body"])
 
-        filename = body["detail"]["object"]["key"]
-        logger.info(f"Retrieved filename: {filename}")
-        components = ScienceFilePath.extract_filename_components(filename)
-        logger.info(f"Initial event parsed filename: {components}")
+            filename = body["detail"]["object"]["key"]
+            logger.info(f"Retrieved filename: {filename}")
+            components = ScienceFilePath.extract_filename_components(filename)
 
-        with db.Session() as session:
             # Potential jobs are the instruments that depend on the current file.
             potential_jobs = get_downstream_dependencies(session, components)
             logger.info(
--- a/sds_data_manager/stacks/indexer_lambda_stack.py
+++ b/sds_data_manager/stacks/indexer_lambda_stack.py
@@ -122,6 +122,7 @@ class IndexerLambda(Stack):
             ),
         )
 
+        # ADD SQS HERE
         # Uses batch job status
         # to update status in the database and
         # update ingested time if status was success
--- a/sds_data_manager/stacks/indexer_lambda_stack.py
+++ b/sds_data_manager/stacks/indexer_lambda_stack.py
@@ -122,7 +122,6 @@ class IndexerLambda(Stack):
             ),
         )
 
-        # ADD SQS HERE
         # Uses batch job status
         # to update status in the database and
         # update ingested time if status was success
--- a/sds_data_manager/stacks/instrument_lambdas.py
+++ b/sds_data_manager/stacks/instrument_lambdas.py
@@ -9,6 +9,8 @@ from aws_cdk import aws_lambda as lambda_
 from aws_cdk import aws_lambda_python_alpha as lambda_alpha
 from aws_cdk import aws_s3 as s3
 from aws_cdk import aws_secretsmanager as secrets
+from aws_cdk import aws_sqs as sqs
+from aws_cdk.aws_lambda_event_sources import SqsEventSource
 from constructs import Construct
 
 from sds_data_manager.stacks.database_stack import SdpDatabase
@@ -27,6 +29,7 @@ class BatchStarterLambda(Stack):
         rds_security_group: ec2.SecurityGroup,
         subnets: ec2.SubnetSelection,
         vpc: ec2.Vpc,
+        sqs_queue: sqs.Queue,
         **kwargs,
     ):
         """BatchStarterLambda Constructor.
@@ -49,6 +52,8 @@ class BatchStarterLambda(Stack):
             RDS subnet selection.
         vpc : ec2.Vpc
             VPC into which to put the resources that require networking.
+        sqs_queue: sqs.Queue
+            A FIFO queue to trigger the lambda with.
         kwargs : dict
             Keyword arguments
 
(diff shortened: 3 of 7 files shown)


sds_data_manager/stacks/sqs_stack.py
```diff
@@ -0,0 +1,75 @@
+"""Stack for creating instrument SQS queues and attaching batch_starter as a target."""
+
+from aws_cdk import Duration, Stack, aws_sqs
+from aws_cdk import aws_events as events
+from aws_cdk import aws_events_targets as targets
+from constructs import Construct
+
+
+class SqsStack(Stack):
```
laspsandoval, line 9: I vaguely remember that it is best practice to use a dead letter queue for SQS. Are you planning on adding that in?

sds_data_manager/stacks/sqs_stack.py
```diff
@@ -0,0 +1,75 @@
...
+        # This needs to be a FIFO queue to enforce ordering
+        self.instrument_queue = aws_sqs.Queue(
+            self,
+            "FileArrivalQueue",
+            queue_name="file_arrival_queue.fifo",
+            fifo=True,
+            encryption=aws_sqs.QueueEncryption.UNENCRYPTED,
+            # This timeout determines how long the queue waits for processing. It must
+            # be longer than the timeout of the lambda.
+            visibility_timeout=Duration.seconds(300),
```
laspsandoval, line 53: Add similar comment in the lambda. What would happen if someone accidentally set the lambda to be greater than 300s? Would the stack fail to deploy or go unnoticed?

sds_data_manager/stacks/sqs_stack.py
```diff
@@ -0,0 +1,75 @@
...
+            # This is required. It removes messages with identical content. Since
+            # the event includes a filename each event should be totally unique.
+            content_based_deduplication=True,
+        )
+
+        for instrument_name in instrument_names:
```
laspsandoval, line 63: How will this work for SPICE? Maybe we aren't there yet, but something to think about for next PR.

laspsandoval: When we are reprocessing will SQS FIFO slow us way down if we don't have the date as part of the event Rule?

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -360,20 +361,27 @@ def try_to_submit_job(session, job_info):
...
-        potential_jobs = get_downstream_dependencies(session, components)
-        logger.info(f"Potential jobs found [{len(potential_jobs)}]: {potential_jobs}")
+        filename = body["detail"]["object"]["key"]
+        logger.info(f"Retrieved filename: {filename}")
+        components = ScienceFilePath.extract_filename_components(filename)
+        logger.info(f"Initial event parsed filename: {components}")
 
-        for job in potential_jobs:
-            try_to_submit_job(session, job)
+        with db.Session() as session:
```
greglucas: Maybe bring this outside of the event for-loop too so you're only grabbing one session for each Lambda invocation?

sds_data_manager/stacks/indexer_lambda_stack.py
```diff
@@ -122,6 +122,7 @@ def __init__(
             ),
         )
 
+        # ADD SQS HERE
```
greglucas: Do you still intend to add something here, or unneeded now?

laspsandoval:
Great start. I would like to discuss how it works only using instrument name and reprocessing.

sds_data_manager/stacks/sqs_stack.py
```diff
@@ -0,0 +1,75 @@
...
+            Parent construct.
+        construct_id : str
+            A unique string identifier for this construct.
+        instrument_names : list[str]
+            A list of all instrument names
+        kwargs : dict
+            Keyword arguments
+        """
+        super().__init__(scope, construct_id, **kwargs)
+        # This needs to be a FIFO queue to enforce ordering
```
greglucas, line 44: Out of curiosity, do we need to enforce ordering? I think our checks say "if all dependencies are met continue on", which is independent on which incoming file triggered the job.

sds_data_manager/stacks/sqs_stack.py
greglucas, line 53: I think this is a re-do timer, not the initial timer? i.e. we aren't waiting 5 minutes here after a file arrives are we? Suggest rewording comment. Also the 30 second default seems like it should be reasonable as well and not have to go down to 5 minutes here or is our database kickoff lambda slow? (I figured it was a few seconds)

laspsandoval (commented):
Some comments mostly for me to talk it out with you.

greglucas:
I'm still not convinced that this is absolutely necessary and to me it still adds complexity to the system with another resource to keep track of. But, it does look fine, and no major complaints from me.

sds_data_manager/stacks/sqs_stack.py
maxinelasp: I wasn't planning on it, but I am happy to if we would like that extra bit of resiliency/monitoring.

sds_data_manager/stacks/sqs_stack.py
maxinelasp: The stack does fail to deploy

sds_data_manager/stacks/sqs_stack.py
maxinelasp: It will slow processing down. I would be surprised if it ended up causing issues, but we can also change settings to reduce that issue (for example by batching processing together to eliminate lambda startup times) or add additional queues

sds_data_manager/stacks/indexer_lambda_stack.py
maxinelasp: Oops, old comment, I will remove

sds_data_manager/stacks/sqs_stack.py
maxinelasp: We don't need to enforce ordering from an abstract perspective. However, in order to enforce concurrency requirements, AWS requires a FIFO queue. So, in order to enforce the one queue per processing step rule it does need to be FIFO.

sds_data_manager/stacks/sqs_stack.py
maxinelasp: No, we aren't waiting 5 minutes for a file arrives. It's a timeout for how long the queue will wait before moving the event to the dead letter queue, I believe. I made it long just to accommodate any changes in the future to that lambda - it only comes into play if processing is stalled for some reason, so I made it longer just in case. In most cases, processing takes a second or two and the events are removed from the queue that quickly.

sds_data_manager/stacks/sqs_stack.py
maxinelasp: Ultimately this is not going to be a bottleneck for processing because the actual processing steps still run in parallel (with a few seconds delayed starting) and they will take much longer than triggering jobs will.

sds_data_manager/stacks/instrument_lambdas.py
```diff
@@ -102,20 +105,8 @@ def __init__(
...
-                detail={
-                    "object": {"key": [{"exists": True}]},
-                },
-            ),
-        )
-
-        event_from_indexer_lambda.add_target(
-            targets.LambdaFunction(self.instrument_lambda)
-        )
+        # Point the provided instrument queue to trigger the lambda. This will
```
tech3371: I think I am little confused with this comment. Can you clarify it? Thank you!

sds_data_manager/utils/stackbuilder.py
```diff
@@ -148,6 +149,14 @@ def build_sds(
             env=env,
         )
 
+    # Create SQS pipeline for each instrument and add it to instrument_sqs
```
tech3371, line 152: Maxine, can you check if I added architecture changes from this PR in our architecture diagram correctly? Appreciate it!
https://lasp.colorado.edu/galaxy/display/IMAP/IMAP+EventBridge+Rules

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -360,20 +361,27 @@ def try_to_submit_job(session, job_info):
     logger.info(f"Submitted job {job_name} with this command: {batch_command}")
 
 
-def lambda_handler(event: dict, context):
+def lambda_handler(events: dict, context):
     """Lambda handler."""
-    logger.info(f"Event: {event}")
+    logger.info(f"Event: {events}")
```
tech3371:
```suggestion
    logger.info(f"Events: {events}")
```

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -360,20 +361,27 @@ def try_to_submit_job(session, job_info):
...
     logger.info(f"Context: {context}")
 
-    # Event details:
-    filename = event["detail"]["object"]["key"]
-    components = ScienceFilePath.extract_filename_components(filename)
-    logger.info(f"Initial event parsed filename: {components}")
+    for event in events["Records"]:
```
tech3371: can you add this comment from your PR description here? I think it will be helpful later.
```
SQS by default has some batching properties, so I updated batch_starter to loop through each event in the batch
```

tech3371 (commented):
couple minor questions. Otherwise, it looks good to me.

sds_data_manager/stacks/instrument_lambdas.py
maxinelasp: Ah yes, this is slightly outdated. I will fix it, thanks for pointing it out!

sds_data_manager/utils/stackbuilder.py
maxinelasp: Currently, the lambda putevent used by the indexer lambda template only has filename and instrument, not data_level or data_product. I'm not sure if that's the same product as the one in your document. The diagram looks good to me!

sds_data_manager/lambda_code/SDSCode/batch_starter.py
maxinelasp: Sure, will do

sds_data_manager/utils/stackbuilder.py
tech3371: Yeah, that put event was triggering batch starter lambda before and only passed filename. Now with your changes, it has instrument name as well and also SQS between that putevent and batch starter lambda. That's what I understood from this PR.

Cool, thank you for confirming!

maxinelasp:
@tech3371 @laspsandoval I addressed your comments and added a DLQ, could I get another review please? 😁

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -360,20 +361,28 @@ def try_to_submit_job(session, job_info):
...
-    # Event details:
-    filename = event["detail"]["object"]["key"]
-    components = ScienceFilePath.extract_filename_components(filename)
-    logger.info(f"Initial event parsed filename: {components}")
-
     with db.Session() as session:
-        # Potential jobs are the instruments that depend on the current file.
-        potential_jobs = get_downstream_dependencies(session, components)
-        logger.info(f"Potential jobs found [{len(potential_jobs)}]: {potential_jobs}")
+        # Since the SQS events can be batched together, we need to loop through
```
tech3371, line 370: thanks!

sds_data_manager/stacks/instrument_lambdas.py
```diff
@@ -102,20 +105,8 @@ def __init__(
...
-            ),
-        )
-
-        event_from_indexer_lambda.add_target(
-            targets.LambdaFunction(self.instrument_lambda)
-        )
+        # This sets up the lambda to be triggered by the SQS queue. Since this is a FIFO
+        # queue, each instrument will have messages processed in order. However,
+        # different instruments will be processed in parallel, with multiple instances
+        # of the batch_starter lambda.
```
tech3371, line 111: we can discuss about this outside of this PR. I feel like we can start with one event rule that hits SQS target and then add on as we need.

tech3371 (approved):
LGTM!

PR #323, opened 2024-08-06
Share Lambda layer for all lambda(s)

### LINKED ISSUE #321: Create Lambda layer
# Change Summary

## Overview
Refactor Lambda code to use layer instead. This will allow code visibility on Lambda and allow easy update and testing.

Summary of some things I tried in the process of updating this:
I tried to create generic lambda layer which could be used as input to other stacks. But it requires to create stack output value using cdk.cfnOutput and then pass that value and use it to import layer stack's information into stack that uses it. With this, it requires to add dependency between layer stack and stacks that uses it. Also, it will mean that we need to be careful of what we call cdk.cfnOutput key as it's used by other stacks. I was concerned that it could cause room for error and too many dependency could introduce complication in updating some stack and could become roadblocks to making improvement.

We can revisit this if we find better solution in the near future. This could be first step to this refactoring.

tech3371:
# Change Summary

## Overview
Create Lambda layer that could be shared with lambda(s)

## New Dependencies
Adds new dependency for all lambda to deploy lambda layer stacks first. This logic is already added in code and we don't need to do anything manually.

## New Files
- sds_data_manager/stacks/lambda_layer_stack.py
   - This creates new lambda layer that will be shared 

## Updated Files
Updates these files to use new lambda layer.
- .pre-commit-config.yaml
- lambda_layer/python/requirements.txt
- sds_data_manager/lambda_code/SDSCode/query_api.py
- sds_data_manager/stacks/create_schema_stack.py
- sds_data_manager/stacks/instrument_lambdas.py
- sds_data_manager/stacks/sds_api_manager_stack.py

## Testing
Tested in AWS that everything still works as before.


sds_data_manager/stacks/lambda_layer_stack.py
```diff
@@ -0,0 +1,58 @@
...
+                    ),
+                ],
+            ),
+        )
+
+        layer = lambda_.LayerVersion(
+            self,
+            id=f"{id}-Layer",
+            code=code_bundle,
+            compatible_runtimes=[lambda_.Runtime.PYTHON_3_12],
```
laspsandoval, line 53: I think spiceypy is compatible with versions of python <=3.11.

laspsandoval (commented):
I would actually like to start using this asap. I think that I create a layer for the msopck program. I'm assuming that that would go in the lambda_layer_stack?

laspsandoval:
Conceptually I understand that lambda layers makes it easier to share an environment between lambdas. Although I know this, can you explain how this is better than using our current system of a requirements.txt?

tech3371:
> Conceptually I understand that lambda layers makes it easier to share an environment between lambdas. Although I know this, can you explain how this is better than using our current system of a requirements.txt?

We are still using `requirements.txt` still to build lambda layer. With this, we abstract out lambda layer code and use that across instead of building it each time. Also, our current system was using docker image which reduces code visibility and lot of things are happening under the hood. With this new change, it allows us to understand and control little more than before. That's what I think.

sds_data_manager/stacks/lambda_layer_stack.py
tech3371: I have not thought much to make this layer stack generic, would you like me to make this compatible_runtimes as input parameter?

greglucas: I think that might make sense, but I also don't think we should do it here, but rather when we add the thing that requires <=3.11
What if we don't actually need the lower versions, then we've added options that aren't necessary.

greglucas (approved):
This is a nice step and also nice to limit the number of Docker builds. It is not currently possible to share Layers across the stacks in a nice way, so this is the best we can do until we refactor the Application stack structure IMO.

sds_data_manager/stacks/lambda_layer_stack.py
tech3371: ok. I will not change in this PR but we can change it when we add `spiceypy` and update accordingly.

tech3371:
@laspsandoval Do you mind if I merge this? Or do you have any change request at the moment?

laspsandoval:
> @laspsandoval Do you mind if I merge this? Or do you have any change request at the moment?

Go ahead and merge.

PR #325, opened 2024-08-14
Pointing frame lambda

laspsandoval:
# Change Summary

## Overview
Provides algorithm that mocks Nick Duttons pointing frame (DPS frame) kernel generation. Note that I have TODOs that will be included in the next tickets.

## New Files
- pointing_frame_handler.py
   - writes a new C-kernel file containing the averaged pointing data
- changes in the environment: poetry.lock, requirements.txt, pyproject.toml

## Testing
- test_pointing_frame_handler.py
   - tests pointing_frame_handler.py
- imap_science_0001.tf 
   - pointing frame kernel used for test data
-  imap_sclk_0000.tsc
   - spacecraft clock kernel used for test data
- imap_wkcp.tf 
   - spacecraft frame kernel used for test data


pyproject.toml
```diff
@@ -65,7 +66,8 @@ requests = "^2.28.2"
 psycopg = "^3.1"
 psycopg2-binary = "^2.9.9"
 SQLAlchemy = "<=3.0.0"
-imap-data-access = ">=0.7.0"
+imap-data-access = ">=0.8.0"
+spiceypy = "^6.0.0"
```
greglucas, line 70:
```suggestion
```

pyproject.toml
```diff
@@ -54,6 +54,7 @@ pytest-postgresql = "^6.0.0"
 moto = "^4.1.3"
 boto3 = "^1.26.78"
 SQLAlchemy = "<=3.0.0"
+planetmapper = "^1.11.2"
```
greglucas, line 57: This looks like it is only used to download kernels, so I don't think warrants a new dependency. We should be able to use standard library routines to do that if needed.
```suggestion
```

sds_data_manager/lambda_code/SDSCode/pointing_frame_handler.py
```diff
@@ -0,0 +1,248 @@
...
+
+import numpy as np
+import spiceypy as spice
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+# Each spin is 15 seconds. We want 10 quaternions per spin.
+# duration / # samples (nominally 15/10 = 1.5 seconds)
+STEP = 1.5
```
greglucas, line 32: Is 10 an arbitrary number?

`STEP` is a somewhat ambiguous name as to what it means. I'd suggest bringing this inside the function below since it is only used in one place.

sds_data_manager/lambda_code/SDSCode/pointing_frame_handler.py
```diff
@@ -0,0 +1,248 @@
...
+    id_imap_spacecraft = spice.gipool("FRAME_IMAP_SPACECRAFT", 0, 1)
+
+    # TODO: Queried pointing start and stop times here.
+
+    # Get the coverage window
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.ckcov
+    cover = spice.ckcov(ck_kernel, int(id_imap_spacecraft), True, "SEGMENT", 0, "TDB")
+    # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.wnfetd
+    et_start, et_end = spice.wnfetd(cover, 0)
+    et_times = np.arange(et_start, et_end, STEP)
```
greglucas, line 63: `arange` can produce sometimes odd results when used with a float input. I'd suggest using `linspace` instead.

From the docs: https://numpy.org/doc/stable/reference/generated/numpy.arange.html
> When using a non-integer step, such as 0.1, it is often better to use [numpy.linspace](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html#numpy.linspace).

```suggestion
    et_times = np.linspace(et_start, et_end, (et_end - et_start) // 10)
```

sds_data_manager/lambda_code/SDSCode/pointing_frame_handler.py
```diff
@@ -0,0 +1,248 @@
...
+    for tdb in et_times:
+        # Rotation matrix from IMAP spacecraft frame to ECLIPJ2000.
+        # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.pxform
+        body_rots = spice.pxform("IMAP_SPACECRAFT", "ECLIPJ2000", tdb)
+        # Convert rotation matrix to quaternion.
+        # https://spiceypy.readthedocs.io/en/main/documentation.html#spiceypy.spiceypy.m2q
+        body_quat = spice.m2q(body_rots)
+        body_quats.append(body_quat)
+        # z-axis of the ECLIPJ2000 frame.
+        z_eclip_time.append(body_rots[:, 2])
```
greglucas, line 96: These aren't used anywhere below? Only the single instance of them, so I don't think we need to keep the entire list?
```suggestion
```

sds_data_manager/lambda_code/SDSCode/pointing_frame_handler.py
```diff
@@ -0,0 +1,248 @@
...
+    # that has the most influence.
+    # The eigenvector corresponding to the largest
+    # eigenvalue points in the direction that has the most
+    # combined rotation influence.
+    eigvals, eigvecs = np.linalg.eig(aggregate)
+    # q0: The scalar part of the quaternion.
+    # q1, q2, q3: The vector part of the quaternion.
+    q_avg = eigvecs[:, np.argmax(eigvals)]
+
+    return q_avg, z_eclip_time
```
greglucas, line 122: Unused in the calling function?
```suggestion
    return q_avg
```

sds_data_manager/lambda_code/SDSCode/pointing_frame_handler.py
```diff
@@ -0,0 +1,248 @@
...
+    z_avg = spice.q2m(list(q_avg))[:, 2]
+    # y_avg is perpendicular to both z_avg and the standard Z-axis.
+    y_avg = np.cross(z_avg, [0, 0, 1])
+    # x_avg is perpendicular to y_avg and z_avg.
+    x_avg = np.cross(y_avg, z_avg)
+
+    # Construct the rotation matrix from x_avg, y_avg, z_avg
+    rotation_matrix = np.array([x_avg, y_avg, z_avg])
+
+    return rotation_matrix, z_avg
```
greglucas, line 155:
```suggestion
    return rotation_matrix
```

greglucas (commented):
I left some initial comments, but want to bring a bigger question up as I might not be up to speed on all of the past decisions. This seems like 2 things going on to me here. 1) how to create the new frame kernel, and 2) how to use that within the infrastructure. Right now those are blended together in this single PR, but I'd like to propose separating them out. (1) could be in `imap_processing` so that all of our SPICE helper routines etc. are in a single location and we load kernels, and save kernels using the same exact utilities. (2) we can make a lambda/batch/whatever that depends on `imap_processing` if needed and calls this routine. I personally like the separation of concerns so we don't force infrastructure decisions based on processing, or vice-versa.

I think that decision will impact future reviews, so didn't review everything in-depth.

laspsandoval:
@greglucas I thought about your question for a while too. Let's talk at Monday's meeting if that is ok. I re-read your comment and probably you are right. It would be easy to move it to imap_processing, but let's discuss.

sds_data_manager/lambda_code/SDSCode/pointing_frame_handler.py
laspsandoval: The return of z_avg was used for plotting in the tests.

sds_data_manager/lambda_code/SDSCode/pointing_frame_handler.py
laspsandoval: This was used in the tests.

sds_data_manager/lambda_code/SDSCode/pointing_frame_handler.py
laspsandoval: 10 is what Nick Dutton recommended. I can ask him.

sds_data_manager/lambda_code/SDSCode/pointing_frame_handler.py
laspsandoval: Odd. I got an error trying to use a non-integer step and linspace.

greglucas: Yes, it isn't a non-integer step then, it is the integer number of items you want. The code I suggested provides an integer with the `//` operator.

PR #326, opened 2024-08-16
Stack to construct

greglucas:
# Change Summary

## Overview

This is a major refactor of our AWS infrastructure. It is best practice to focus on lifecycle for Stacks and deployment updates to avoid circular dependencies. We were running into a few circular dependencies because we had Stack A depending on Stack B, which would then want to try and update Stack A again. Putting everything into the same deployment unit makes sure that these updates get processed and deployed in the correct order.

There is a lot of thrash just changing names from "abc_stack" to "abc_construct", might be advantageous to avoid the "_specific_name" in the future...

We have 3 primary stacks now, with our previous fine-grained Stacks turning into Constructs.
1. Networking Stack (sets up VPC)
2. SDC Stack (sets up all resources that the SDC uses/needs)
3. I-ALiRT Stack (sets up all resources for the I-ALiRT processing)

The easiest way to review this might be to step through by commits. I tried to break the process up into small incremental commits. Although there were two other PRs that were merged during this process and it was easier to merge them in than rebase through all the commits.

### Testing

I have deployed this all into a separate dev account and it deploys successfully and works as expected from an initial cut at depositing a file and querying the DB for that file.


sds_data_manager/utils/stackbuilder.py
```diff
@@ -3,30 +3,30 @@
 from pathlib import Path
```
laspsandoval, line 3: Minor comment about naming. Keep uniform conventions. For example we have:
IAlirtBucketConstruct
IalirtProcessingPrimary

So keep or remove Construct part of the name for the Constructs.

sds_data_manager/utils/stackbuilder.py
```diff
@@ -262,7 +269,7 @@ def build_backup(scope: App, env: Environment, source_account: str):
 
     """
     # This is the S3 bucket used by upload_api_lambda
-    backup_bucket_stack.BackupBucket(
+    backup_bucket_construct.BackupBucket(
         scope,
```
laspsandoval: For all the other constructs the scope is equal to the name of the stack. Why not this one?

sds_data_manager/utils/stackbuilder.py
```diff
@@ -47,112 +47,135 @@ def build_sds(
...
     api.deliver_to_sns(monitoring.sns_topic_notifications)
 
+    # create Code asset and Layer for Lambda(s)
+    layer_code_directory = (
+        Path(__file__).parent.parent.parent / "lambda_layer/python"
+    ).resolve()
+    lambda_code_directory = Path(__file__).parent.parent / "lambda_code"
+
+    lambda_code = lambda_.Code.from_asset(str(lambda_code_directory))
+    db_lambda_layer = lambda_layer_construct.LambdaLayerConstruct(
```
laspsandoval, line 138: Why do we only have a single lambda layer and it's named for the database?

laspsandoval (commented):
I left a few comments. I'll take some more time to try it out. Thanks!

sds_data_manager/utils/stackbuilder.py
greglucas: Yeah, this was a bit all over the place, and I didn't do a good job of consolidating. My preference would be to drop the `_construct` everywhere, see this comment:
> There is a lot of thrash just changing names from "abc_stack" to "abc_construct", might be advantageous to avoid the "_specific_name" in the future...

If you all don't mind the churn in this single PR I can do it now, but I was trying to make this a little more manageable to review.

laspsandoval: No worries to do it in another PR.

sds_data_manager/utils/stackbuilder.py
greglucas: Good catch, this needs a Stack at some point, so I added one above it here to follow the same style as the others.

sds_data_manager/utils/stackbuilder.py
greglucas: We want to build the dependencies only once (@tech3371 's previous PR), and attach that to all the Lambda Functions that need it. Here this is the database requirements layer. For instance, if you add "imap_processing" as a requirement for the SPICE kernel work, that could be a separate layer and only attached to the Lambda Functions that need that specific piece.

laspsandoval: Thanks for the explanation. I'm still getting intuition on how this works.

laspsandoval (commented):
I thought about this over the weekend. For I-ALiRT the ECR needs to be deployed before anything else. Afterwards an image needs to be pushed to it. And then the ECS can be deployed. If you deploy the ECS without having an image in the ECR then it times out. Probably everything worked for you since we already had an ECR and image in the dev account so no update was required.

I know there is probably a better way to do this, but for now that is how it works. So that could cause issues with the way you have currently set things up.

sds_data_manager/constructs/data_bucket_construct.py
```diff
@@ -1,14 +1,14 @@
 """Configure the data bucket stack."""
```
tech3371:
```suggestion
"""Configure the data bucket construct."""
```

sds_data_manager/constructs/database_construct.py
```diff
@@ -0,0 +1,156 @@
+"""Configure the database stack."""
```
tech3371:
```suggestion
"""Configure the database construct."""
```

sds_data_manager/constructs/domain_construct.py
```diff
@@ -1,12 +1,11 @@
 """Configure the domain stack."""
```
tech3371:
```suggestion
"""Configure the domain construct."""
```

sds_data_manager/constructs/efs_construct.py
```diff
@@ -1,10 +1,8 @@
 """Configure the EFS stack."""
```
tech3371:
```suggestion
"""Configure the EFS construct."""
```

sds_data_manager/constructs/ialirt_bucket_construct.py
```diff
@@ -1,13 +1,13 @@
 """Configure the IAlirt bucket stack."""
```
tech3371:
```suggestion
"""Configure the IAlirt bucket construct."""
```

sds_data_manager/constructs/indexer_lambda_construct.py
```diff
@@ -1,9 +1,6 @@
 """Configure the indexer lambda stack."""
```
tech3371:
```suggestion
"""Configure the indexer lambda construct."""
```

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -93,7 +86,7 @@ def __init__(
             vpc_subnets=subnets,
             security_groups=[rds_security_group],
             allow_public_subnet=True,
-            layers=batch_starter_layers,
+            layers=layers,
```
tech3371, line 89: good catch!

sds_data_manager/constructs/lambda_layer_construct.py
```diff
@@ -46,13 +45,9 @@ def __init__(
             ),
         )
 
-        layer = lambda_.LayerVersion(
+        self.layer = lambda_.LayerVersion(
```
tech3371, line 48: I am glad construct removes output and import steps for layer and EFS.

sds_data_manager/constructs/sds_api_manager_construct.py
```diff
@@ -1,24 +1,22 @@
 """Configure the SDS API Manager stack."""
```
tech3371:
```suggestion
"""Configure the SDS API Manager construct."""
```

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -13,56 +15,20 @@
...
-        with db.Session() as session:
-            session.add_all(all_dependents)
-            session.commit()
-        send_response(event, context, "SUCCESS")
-    except Exception as e:
-        logger.error(e)
-        send_response(event, context, "FAILED")
+    # NOTE: If we run this when trying to delete the stack, it fails because
+    #       the RDS isn't available, so skip it here.
+    if event.get("RequestType") != "Create":
```
tech3371, line 25: nice find!

greglucas:
Per the conversation earlier today, I have removed the I-ALiRT stack from the deployment in the most recent commit with a note to review and add it back in later once the updates have been made.

tech3371 (approved):
Looks great to me! Thank you for doing this!

bmcclellan-cu (approved):
Looks good to me! I didn't test the deployment, but if you ran this code today then I'd say it's good.

greglucas:
> Looks good to me! I didn't test the deployment, but if you ran this code today then I'd say it's good.

@bmcclellan-cu, I have just pushed up the latest changes that I deployed from today. The most recent couple of commits include the work from Jackson for the HostedZone certificate.

tech3371 (approved):
We have verified that it works as it should. So looks good to me!

PR #330, opened 2024-09-12
FIX: Add region to the s3 presigned urls

greglucas:
# Change Summary

When initially deploying the bucket and then trying to upload a file, I was getting 307 redirects which can apparently happen if the region isn't specified in the initial PUT request. This happens because the default presigned URL returns a global s3 endpoint rather than a region specific one.

This updates the code to use s3v4 signature for our presined URLs, which contain the region information in them. We also connect boto3 to the region our bucket is in for this purpose.

NOTE: We may be able to update `imap-data-access` to catch the 307 redirects and do that for the users, but the reason it fails is because the HTTP spec says you should not redirect a PUT to 307 for a user automatically, and the Python standard library therefore doesn't follow that redirect.

To me, this seemed like a good approach to try for now to return region-specific URLs to the users.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/download_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/download_api.py
@@ -45,6 +45,10 @@ def lambda_handler(event, context):
     error. Otherwise, it returns pre-signed s3 url that user can use to download
     data from s3.
 
+    To avoid any 307 redirects we use s3v4 signing method.
+    This method includes the region in the URL, so when the user uploads a file,
+    the URL will point directly to the correct regional S3 endpoint.
+
     Parameters
     ----------
     event : dict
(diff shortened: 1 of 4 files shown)


sds_data_manager/lambda_code/SDSCode/upload_api.py
```diff
@@ -12,7 +12,14 @@
 logger.setLevel(logging.INFO)
 
 BUCKET_NAME = os.getenv("S3_BUCKET")
-S3_CLIENT = boto3.client("s3")
+REGION = os.getenv("REGION")
+# The default presigned url signature does not include the region information
+# within the signature and we should be hitting the actual s3 region endpoint
+# to avoid any 307 redirects. (Generally only an issue on newly created buckets
```
laspsandoval, line 23: Can you add to this docstring so I remember in the future?

"to avoid any 307 redirects we use s3v4 signing method. This method includes the region in the URL, so when the user uploads a file, the URL will point directly to the correct regional S3 endpoint. "

sds_data_manager/lambda_code/SDSCode/download_api.py
```diff
@@ -74,9 +74,14 @@ def lambda_handler(event, context):
         return http_response(status_code=400, body=response_body)
 
     bucket = os.getenv("S3_BUCKET")
+    region = os.getenv("REGION")
     filepath = path_params
 
-    s3_client = boto3.client("s3")
+    s3_client = boto3.client(
```
laspsandoval, line 88: Add same docstring here as what is in upload_api.py

laspsandoval (approved):
Looks good.

tech3371 (approved):
Interesting find! Looks good to me

sds_data_manager/lambda_code/SDSCode/upload_api.py
greglucas: Yep, done.

PR #331, opened 2024-09-12
MNT: Use direct inheritance for simple resource like LambdaLayer

greglucas:
# Change Summary

Rather than subclassing Construct and then calling the `LambdaLayer` Construct within the initializer, we can actually just subclass the `LambdaLayer` directly here and do whatever else we need to (bundling the code through Docker) before initializing the AWS's Construct.

### DIFF
--- a/sds_data_manager/constructs/lambda_layer_construct.py
+++ b/sds_data_manager/constructs/lambda_layer_construct.py
@@ -2,14 +2,17 @@
 
 import aws_cdk as cdk
 from aws_cdk import aws_lambda as lambda_
-from constructs import Construct
 
 
-class LambdaLayerConstruct(Construct):
-    """Lambda Layer Construct."""
+class IMAPLambdaLayer(lambda_.LayerVersion):
+    """Lambda Layer."""
 
     def __init__(
-        self, scope: Construct, id: str, layer_dependencies_dir: str, **kwargs
+        self,
+        id: str,
+        layer_dependencies_dir: str,
+        runtime=lambda_.Runtime.PYTHON_3_12,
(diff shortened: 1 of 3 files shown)


sds_data_manager/constructs/lambda_layer_construct.py
```diff
@@ -45,9 +46,6 @@ def __init__(
             ),
         )
 
-        self.layer = lambda_.LayerVersion(
-            self,
-            id=f"{id}-Layer",
-            code=code_bundle,
-            compatible_runtimes=[lambda_.Runtime.PYTHON_3_12],
+        super().__init__(
+            id=f"{id}-Layer", code=code_bundle, compatible_runtimes=[runtime], **kwargs
```
tech3371, line 50: Makes sense on very high level. I remember you mentioning this in last refactor meeting. Good to see it implemented and working.

In the future, if someone wants to do similar thing, were you looking at this [doc](https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_lambda/LayerVersion.html) and then setting everything following `scope` as needed and then return this `super().__init__()` with what we would pass in `LayerVersion` construct?

tech3371 (approved):

sds_data_manager/constructs/lambda_layer_construct.py
greglucas: Yes, this is basically just setting default values for us and then calling the `LayerVersion.__init__()` with whatever we want to pass it. The benefit is that this can then be passed directly to the other CDK constructs/functions that reequire Layers rather than having to do `my_construct.layer`

PR #332, opened 2024-09-12
MNT: Refactor batch processing

greglucas:
# Change Summary

## Overview

- Use AWS L2 constructs instead of L1
  - It was really verbose with the direct CloudFormation calls and we can utilize the higher-level L2 constructs instead
- Call it ProcessingJobs instead of AWS specific tools like "batch"
- Use only one compute environment / job queue, but still individual job definitions for each specific container / repo
- Add L3 processing job definitions


sds_data_manager/constructs/processing_construct.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        super().__init__(scope, construct_id, **kwargs)
+
+        # Create compute environment
+        compute_environment = batch.FargateComputeEnvironment(
+            self,
+            "ProcessingComputeEnvironment-spot",
+            vpc=vpc,
+            vpc_subnets=ec2.SubnetSelection(subnet_type=ec2.SubnetType.PUBLIC),
+            spot=True,
```
tech3371, line 54: looks like old one had roles. I think it will need add some service roles to be able to execute.

tech3371: Nice that it allows to use spot instance though.

sds_data_manager/constructs/processing_construct.py
```diff
@@ -0,0 +1,103 @@
...
+                cpu=1,
+                environment={"IMAP_SPICE_DIR": "/mnt/spice"},
+                volumes=self.volumes,
+                # TODO: Do we need to explicitly specify architecture and OS family?
+                #       We are building containers in GitHub Actions and need to
+                #       make sure these are aligned.
+                # fargate_cpu_architecture=ecs.CpuArchitecture.ARM64,
+                # fargate_operating_system_family=ecs.OperatingSystemFamily.LINUX
+            ),
+        )
```
tech3371, line 109: Here, we are missing batch job role as well. I am pretty sure it needs it as well.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -340,16 +339,9 @@ def try_to_submit_job(session, job_info):
...
-    job_definition = (
-        f"arn:aws:batch:{region}:{account}:job-definition/"
-        f"fargate-batch-job-definition{instrument}"
-    )
-    job_queue = (
-        f"arn:aws:batch:{region}:{account}:job-queue/"
-        f"{instrument}-fargate-batch-job-queue"
-    )
+    step = "-l3" if data_level >= "l3" else ""
+    job_definition = f"ProcessingJob-{instrument}-{step}"
```
tech3371: will this result in `ProcessingJob-swe--l3` for l3 data product?

Thank you for adding l3 stuff in. Huge help!

sds_data_manager/utils/stackbuilder.py
```diff
@@ -194,23 +195,21 @@ def build_sds(
...
-            data_bucket=data_bucket.data_bucket,
-            repo=ecr.container_repo,
-            db_secret_name=db_secret_name,
-            efs_instance=efs_instance,
+    processing_volumes = [
+        batch.EfsVolume(
+            name=f"{efs_instance.volume_name}-ECS-mount",
+            access_point_id=efs_instance.spice_access_point.access_point_id,
+            file_system=efs_instance.efs,
+            container_path="/mnt/spice",
```
tech3371, line 210: can you also add mountPoints? I believe that mounts efs onto batch volume.

tech3371 (commented):
Really nice refactor! Can you test if it works, especially batch job and EFS volume mount?

sds_data_manager/constructs/processing_construct.py
laspsandoval: @greglucas are you using higher level constructs for the roles? I didn't know you could do that until now. Nice!

sds_data_manager/utils/stackbuilder.py
```diff
@@ -194,23 +195,21 @@ def build_sds(
...
-            scope=sdc_stack,
-            construct_id=f"{instrument}BatchJob",
-            vpc=networking.vpc,
-            processing_step_name=instrument,
-            data_bucket=data_bucket.data_bucket,
-            repo=ecr.container_repo,
-            db_secret_name=db_secret_name,
-            efs_instance=efs_instance,
+    processing_volumes = [
+        batch.EfsVolume(
```
laspsandoval, line 206: Will this still be able to work in the same way with the example in imap_processing/examples/imap_processing/Dockerfile.efs and/or Dockerfile.processing?

sds_data_manager/constructs/processing_construct.py
```diff
@@ -0,0 +1,103 @@
...
+    def add_job(self, job_name: str):
+        """Create an ECR repo and a job definition for the given job.
+
+        Parameters
+        ----------
+        job_name : str
+            Name of the job for which to create the job definition.
+        """
+        # Create a registry for each job definition (swe-repo)
+        container_repo = ecr.Repository(
```
laspsandoval, line 80: Oh I like how you solve the instrument/level vs only instrument ecr problem here.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -2,7 +2,6 @@
```
laspsandoval, line 2: I'm trying to think if anything here would be any different for L3. I'm just brainstorming here:
Would we need to pass in the binning mechanism here (abrate vs HEALPIX etc) in the command?

laspsandoval (commented):
I think it all looks in order. Address the comments and then I think it will be good to go.

sds_data_manager/constructs/processing_construct.py
greglucas: Yes, it creates some minimal roles that are necessary. So everything works as expected after my last couple of commits. (Needed to assign_public_ip to the fargate job to be able to pull from the ECR, which I added as a comment)

sds_data_manager/constructs/processing_construct.py
greglucas: This works in my dev environment. What role/action do you think we're missing?

sds_data_manager/lambda_code/SDSCode/batch_starter.py
greglucas: I think that is a good question, and we would need to update the command sent to the container. Where that command comes from is TBD though if we do need something else.

sds_data_manager/utils/stackbuilder.py
greglucas: I'm not sure what you mean? It is working with the containers we have currently. We are mounting the EFS to `/mnt/spice` via the container_path below.

sds_data_manager/utils/stackbuilder.py
greglucas: This is the mount point. Inside the container this EFS drive is mounted at `/mnt/spice`

laspsandoval (approved):
Ok. Approved.

sds_data_manager/constructs/processing_construct.py
tech3371: Like these permission for the container:
- service-role/AmazonECSTaskExecutionRolePolicy
- permission to pull images from ECR

Like these permission for the batch role:
- iam.ManagedPolicy.from_aws_managed_policy_name("AmazonS3FullAccess"),
- iam.ManagedPolicy.from_aws_managed_policy_name(
                    "AmazonElasticFileSystemFullAccess"
                ),

tech3371: We discussed why those permission are not needed on the call.

tech3371 (approved):
Looks good to me!

PR #333, opened 2024-09-12
ENH: Add a stack for website deployments

greglucas:
# Change Summary

This adds all of the resources required for a frontend website as a separate stack. It relies on the hosted_zone stack to get the root hosted_zone and certificate. It creates an s3 bucket to host the website and corresponding IAM user, group, and policy.

This pulls in the content from SWx TREC's CDK CloudFront Stack.

## Testing

There is some routing logic that goes on via CloudFront Functions, that is tested in a subprocess and has a separate javascript test associated with it.


bourque (approved):
Thanks again for talking through this with me today!

PR #336, opened 2024-09-16
Harrison 1692 - SWYN/TLAY - added L3 SWAPI dependency configuration

pleasant-menlo:
# Change Summary

## Overview
I line was added within the dependency configuration that will enable testing of the SWAPI L3 processor within AWS.
 
## Updated Files
- _sds_data_manager/lambda_code/SDSCode/dependency_config.csv_
  - add entry that sets the test L2 file as a dependency to the SWAPI L3 processor

Ted and Sebastian from Menlo Innovations.


sds_data_manager/lambda_code/SDSCode/dependency_config.csv
```diff
@@ -110,6 +110,10 @@ mag, l1b, burst-magi, mag, l1c, norm-magi, HARD, DOWNSTREAM
 
 swapi, l0, raw, swapi, l1, all, HARD, DOWNSTREAM
 
+# <---- TEST SWAPI L3 Dependencies ---->
+
+swapi, l2, fake-swapi-l2, swapi, l3a, all, HARD, DOWNSTREAM
```
maxinelasp: Rather than the descriptor `fake-swapi-l2` could we use the generic `sci` descriptor? And relatedly, could we use `sci` rather than `all` for the L3A descriptor, unless `all` is a meaningful descriptor?

maxinelasp (approved):
Overall looks good. This will need to be deployed, but then it should trigger processing when a L2 swapi file arrives. I want to point out that since there is no L1->L2 connection for SWAPI, L1 and L0 files will not trigger L3A processing, so only directly uploaded L2 files will start processing. 

I just finished some documentation on the dependency management, for more info: https://imap-processing.readthedocs.io/en/latest/data-access/data-dependency.html

maxinelasp:
I would also like to share our github and review standards documentation: https://imap-processing.readthedocs.io/en/latest/development-guide/style-guide/review-standards.html

This PR does match with those requirements, just sharing so you have all the info!

sds_data_manager/lambda_code/SDSCode/dependency_config.csv
tech3371: I believe this will work. We will need to deploy this change to the database but otherwise, it looks good.

I think you guys know this already but to make sure we all are on same page, with your current change, it will work if you upload intermediate L2 file with filename, `imap_swapi_l2_fake-swapi-l2_v001.cdf`. Then it will kickoff all `l3a` products processing if that's what you intended. Once you test the mechanism of this dependency files, the actual processing will need to have something like this like @maxinelasp was saying. You can open another PR once you and SWAPI have defined those names.

```suggestion
swapi, l2, sci, swapi, l3a, all, HARD, DOWNSTREAM
```

or

```suggestion
swapi, l2, sci, swapi, l3a, <product-1-name>, HARD, DOWNSTREAM
swapi, l2, sci, swapi, l3a, <product-2-name>, HARD, DOWNSTREAM
swapi, l3a, <product-1-name>, swapi, l3b, <product-1-name>, HARD, DOWNSTREAM
...
```

tech3371 (approved):

PR #337, opened 2024-09-16
Changes to DynamoDB to accomodate query patterns

### LINKED ISSUE #736: I-ALiRT: HIT DynamoDB Design
### I-ALiRT: HIT DynamoDB Design - I think this will be for all instruments in the same table.

Scratch everything below the following diagram:
Screenshot 2024-08-16 at 11 05 25 AM

### Proposed:
partition key: data_product_name (string)
sort key: met
attributes: ['data_product_value', 'ingest_time', 'instrument_met', 'status'] *maybe instrument_met if it is different than sc met.

example data_product_name : hit_l1a_trig_00
example ingest_time : "2021-01-01T00:00:00Z"

Global Secondary Index:
partition key: data_product_name
sort key: ingest_time

Global Secondary Index:
partition key: status
sort key: data_product_name

Global Secondary Index:
partition key: data_product_name
sort key: instrument_met * maybe

### Example query patterns:

response = table.query(
KeyConditionExpression=Key('data_product_name').begins_with('hit_l1atrigger')&
Key('met').between(123,125))

laspsandoval:
# Change Summary

## Overview
The scientists may need to query the ingest db so they can product intermediate data products. Therefore the query patterns have changed.

Before I had the db table setup this way:

partition key: met
sort key: ingest date

But to query the table you have to do this:

``` response = table.query( KeyConditionExpression=dynamodb.conditions.Key('met').eq(partition_key_value) & dynamodb.conditions.Key('ingest_date').between(start_ingest_date, end_ingest_date) ) ```

So you can't query a range because the met and ingest date are 1:1 ratio. There is 1 met/s so just to get a days worth of data a user would need to query the db  86400 times. 

This was all fine if we didn't have to do anything with this table except store the binary blobs, but now it will possibly be used by the instrument teams via an api. They will likely want to query a range of mets and/or ingest dates. 

> Also, the smallest partition that DynamoDB can do is 10 GB. So I couldn't just use a single partition (like apid) because after ~8 years we would run out of space.

Update: Greg suggested reverting to apid. If you run over that, AWS will start combining the partition key and sort key into the hash and create a new partition for you. So it isn't a hard-stop on the 10GB limit.

## Updated Files

   - Add GSI and update partition key
- ialirt_ingest.py
   - Changed example of what will be ingested using the lambda.

## Testing
test_ialirt_ingest_lambda_construct.py
   - test ialirt_ingest_lambda_construct.py.
test_ialirt_ingest.py
   - test ialirt_ingest.py.
conftest.py
   - Updated test DynamoDB.

### DIFF
--- a/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
+++ b/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
@@ -67,14 +67,16 @@ class IalirtIngestLambda(Construct):
                 name="met",
                 type=ddb.AttributeType.NUMBER,
             ),
-            # Sort key (SK) = Ingest Time (ISO).
-            sort_key=ddb.Attribute(
-                name="ingest_time",
-                type=ddb.AttributeType.STRING,
-            ),
-            # Define the read and write capacity units.
-            # TODO: change to provisioned capacity mode in production.
             billing_mode=ddb.BillingMode.PAY_PER_REQUEST,  # On-Demand capacity mode.
+        )
+
+        # Add a GSI for ingest time.
+        table.add_global_secondary_index(
+            index_name="ingest_time",
+            partition_key=ddb.Attribute(
+                name="ingest_time", type=ddb.AttributeType.STRING
+            ),
+            projection_type=ddb.ProjectionType.ALL
         )
         return table
 
--- a/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
+++ b/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
@@ -62,21 +62,34 @@ class IalirtIngestLambda(Construct):
             # Restore data to any point in time within the last 35 days.
             # TODO: change to True in production.
             point_in_time_recovery=False,
-            # Partition key (PK) = Mission Elapsed Time (MET).
+            # Partition key (PK) = ingest year (YYYY).
             partition_key=ddb.Attribute(
+                name="ingest_year",
+                type=ddb.AttributeType.NUMBER,
+            ),
+            # Sort key (SK) = Mission Elapsed Time (MET).
+            sort_key=ddb.Attribute(
                 name="met",
                 type=ddb.AttributeType.NUMBER,
             ),
+            # Define the read and write capacity units.
+            # TODO: change to provisioned capacity mode in production.
             billing_mode=ddb.BillingMode.PAY_PER_REQUEST,  # On-Demand capacity mode.
         )
 
         # Add a GSI for ingest time.
         table.add_global_secondary_index(
-            index_name="ingest_time",
+            index_name="ingest_date",
+            # Partition key (PK) = ingest year (YYYY).
             partition_key=ddb.Attribute(
-                name="ingest_time", type=ddb.AttributeType.STRING
+                name="ingest_year", type=ddb.AttributeType.NUMBER
             ),
-            projection_type=ddb.ProjectionType.ALL
+            # Sort key (SK) = Ingest Time (ISO).
+            sort_key=ddb.Attribute(
+                name="ingest_date",
+                type=ddb.AttributeType.STRING,
+            ),
+            projection_type=ddb.ProjectionType.ALL,
         )
         return table
 
--- a/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
+++ b/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
@@ -64,7 +64,7 @@ class IalirtIngestLambda(Construct):
             point_in_time_recovery=False,
             # Partition key (PK) = ingest year (YYYY).
             partition_key=ddb.Attribute(
-                name="ingest_year",
+                name="apid",
                 type=ddb.AttributeType.NUMBER,
             ),
             # Sort key (SK) = Mission Elapsed Time (MET).
@@ -81,9 +81,7 @@ class IalirtIngestLambda(Construct):
         table.add_global_secondary_index(
             index_name="ingest_date",
             # Partition key (PK) = ingest year (YYYY).
-            partition_key=ddb.Attribute(
-                name="ingest_year", type=ddb.AttributeType.NUMBER
-            ),
+            partition_key=ddb.Attribute(name="apid", type=ddb.AttributeType.NUMBER),
             # Sort key (SK) = Ingest Time (ISO).
             sort_key=ddb.Attribute(
                 name="ingest_date",
--- a/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
+++ b/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
@@ -79,12 +79,12 @@ class IalirtIngestLambda(Construct):
 
         # Add a GSI for ingest time.
         table.add_global_secondary_index(
-            index_name="ingest_date",
+            index_name="ingest_time",
             # Partition key (PK) = ingest year (YYYY).
             partition_key=ddb.Attribute(name="apid", type=ddb.AttributeType.NUMBER),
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: tests/conftest.py [hunk 1]
```diff
@@ -19,13 +19,27 @@ def table():
             TableName="imap-data-table",
             KeySchema=[
                 # Partition key
-                {"AttributeName": "met", "KeyType": "HASH"},
+                {"AttributeName": "ingest_year", "KeyType": "HASH"},
```
maxinelasp: This isn't a good partition key, since people will almost always be using the current year and therefore there will be a lot of load on one partition and none on the rest. 

This may or may not even really matter based on your access patterns and data size, just wanted to call it out. 
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/bp-partition-key-uniform-load.html

Maybe using a hash on the time instead would make sense?

sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
```diff
@@ -62,20 +62,35 @@ def create_dynamodb_table(self) -> aws_dynamodb.Table:
             # Restore data to any point in time within the last 35 days.
             # TODO: change to True in production.
             point_in_time_recovery=False,
-            # Partition key (PK) = Mission Elapsed Time (MET).
+            # Partition key (PK) = ingest year (YYYY).
```
maxinelasp: Why are you making this change? Just want some backstory

maxinelasp (commented):
It would be good to get some more info on the motivation behind this change, because I'm not really sure why it's being made

laspsandoval:
> It would be good to get some more info on the motivation behind this change, because I'm not really sure why it's being made

@maxinelasp agreed. I added more info to the Overview above. Hopefully that helps.Let me know if you have any questions.

tests/conftest.py
laspsandoval: Ideally a single partition (like apid) would have been better. Greg and I talked about this. A single DynamoDB partition can store 10 GB of data. We will generate ~1.25 GB/yr. So we would run out of space in 8 years. So I had to break it down. Year seemed to be the least cumbersome way to do that. I could do it by day, but that would require more queries. Time would result in the same problem that I had earlier which is you can't query a range which is likely what the instrument teams will want. I'm open to other ideas though.

sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
laspsandoval: Apologies. I added more info in the Overview above.

tests/conftest.py
greglucas: FYI, I read about this and a single partition does have a 10GB limit internal to AWS, but if you run over that, AWS will start combining the partition key and sort key into the hash and create a **new** partition for you. So it isn't a hard-stop on the 10GB limit.

My suggestion would be to have `apid` be the Partition Key, and `met` be the Sort Key like you initially proposed. Time series queries are much better done along the sort key, and I think it is totally fine to have the table have a single partition key.

laspsandoval: Oh interesting. Let me read the link you sent: https://aws.amazon.com/blogs/database/effective-data-sorting-with-amazon-dynamodb/

laspsandoval: Ok. Updated. Thanks for that info Greg.

maxinelasp: That sounds good to me, thank you both for your consideration!

maxinelasp (approved):

sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
```diff
@@ -62,20 +62,33 @@ def create_dynamodb_table(self) -> aws_dynamodb.Table:
...
-                type=ddb.AttributeType.STRING,
+                name="met",
+                type=ddb.AttributeType.NUMBER,
             ),
             # Define the read and write capacity units.
             # TODO: change to provisioned capacity mode in production.
             billing_mode=ddb.BillingMode.PAY_PER_REQUEST,  # On-Demand capacity mode.
         )
+
+        # Add a GSI for ingest time.
```
greglucas, line 80: How important is this GSI? Are these just one-off searches once/year kind of thing, or are we regularly going to be querying for ingest time?

I'm curious if we could simply do the comparisons against the attributes of the main table instead of creating a new index. Our ingest time is also constrained to probably a +/- 10 minute buffer, so the ingest time probably doesn't get you a lot of great additional information to search by and you could do a query against +/- 10 minutes on the main index items even.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -39,8 +39,9 @@ def lambda_handler(event, context):
 
     # TODO: item is temporary and will be replaced with actual packet data.
     item = {
+        "apid": 478,
         "met": 123,
-        "ingest_time": "2021-01-01T00:00:00Z",
+        "ingest_date": "2021-01-01T00:00:00Z",
```
greglucas: I'd leave it as "time" if you are going to include the time component, or if you want date, remove the "T00:00:00Z"

tests/infrastructure/test_ialirt_ingest_lambda_construct.py
```diff
@@ -9,26 +9,55 @@ def populate_table(table):
...
 
+    response = table.query(KeyConditionExpression=Key("apid").eq(478))
+
+    items = response["Items"]
+
+    for item in range(len(items)):
+        assert items[item]["apid"] == expected_items[item]["apid"]
+        assert items[item]["met"] == expected_items[item]["met"]
+        assert items[item]["ingest_date"] == expected_items[item]["ingest_date"]
+        assert items[item]["packet_blob"] == expected_items[item]["packet_blob"]
```
greglucas: Can you compare the entire `items == expected_items` or each individual item? This will help if you add/remove attributes in the future and not needing to update everything here too.
```suggestion
        assert items[item] == expected_items[item]
```

tests/infrastructure/test_ialirt_ingest_lambda_construct.py
```diff
@@ -9,26 +9,55 @@ def populate_table(table):
...
+
+    response = table.query(
+        KeyConditionExpression=Key("apid").eq(478) & Key("met").between(100, 123)
+    )
+    items = response["Items"]
+    assert len(items) == 1
+    assert items[0]["met"] == expected_items[0]["met"]
```
greglucas, line 46:
```suggestion
    assert items == expected_items
```

tests/lambda_endpoints/test_ialirt_ingest.py
```diff
@@ -11,20 +10,22 @@ def populate_table(table):
...
         {
+            "apid": 478,
             "met": 124,
-            "ingest_time": "2021-01-01T00:00:01Z",
+            "ingest_date": "2021-02-01T00:00:00Z",
             "packet_blob": b"binary_data_string",
         },
     ]
     for item in items:
         table.put_item(Item=item)
```
greglucas, line 26: No need to fix this here, but there is a batch_write_item that I think will do multiple inserts at once.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb/client/batch_write_item.html

greglucas (commented):
I like this table design, just a minor question about the GSI's purpose. I also think that indexes can generally be added later pretty easily.

sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
laspsandoval: Good points. I'm still gathering information. I'll talk with Eric Christian today (hopefully) which may lead for this to change at some point.

PR #338, opened 2024-09-17
SWAPI L2 dependency

tech3371:
# Change Summary

## Overview
SWAPI L2 dependency


greglucas (approved):
Does this need to be `all`?

tech3371:
> Does this need to be `all`?

No because it should be 1-1 from l1 onward. HK only gets processed up until L1. That's why we used `all` there.

PR #342, opened 2024-09-20
ENH: Add a DB synchronizer for our file database

### LINKED ISSUE #254: BUG - redeploying database loses file catalog for existing files
### ## Description of the issue

If we redeploy or delete lines out of the dev database, we lose information on existing files in S3 and can no longer query them. 

1. 2. 3.

1. Delete a row from file_catalog
2. Attempt to query existing file
3. see it fails

### Code Snippet:

```python
Code
```

The file catalog and the S3 buckets should always be aligned

We can delete lines out of the database and have no way to re-index the files.

### Additional notes

This should never happen in production. We won't be deleting database rows or S3 files. However, if they do get out of sync, that will be very, very bad. It would be good to have some kind of system or tool for fixing that problem quickly if it ever does occur. 

### Affected areas (code, data, or process)

indexer lambda, S3 buckets, API, RDS database

### Suggested fix?

Perhaps a syncing lambda we can run manually to align the bucket and the RDS database. Or just delete the files out of S3 every time.

greglucas:
# Change Summary

## Overview

This will keep s3 and our DB in sync with s3 being the source of truth. Iterate through all items in s3 to get the keys, then do a scan of the File Table to get all items in that table and compare the two sets of items against one another. Then iterate through adding / removing entries from the database table as necessary.

## Notes

* s3 is the source of truth, so the database is the only thing getting updated
* Do we want to run this on a schedule (currently daily), or just have the lambda there for when we need it?

closes https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/254

### DIFF
--- a/sds_data_manager/constructs/database_construct.py
+++ b/sds_data_manager/constructs/database_construct.py
@@ -155,3 +155,44 @@ class SdpDatabase(Construct):
         # Add an explicit dependency on the RDS instance because we need the secret
         # populated with the DB credentials before we can create the schema.
         db_custom_resource.node.add_dependency(db)
+
+    def add_synchronizer(self, code, layers, bucket_name, vpc):
+        """Add synchronizer lambda.
+
+        This is a separate lambda function that is used to synchronize the database
+        with the s3 bucket. It is currently implemented requiring a manual trigger
+        by the user. If there are discrepancies between the database and the s3 bucket,
+        we want a user to take action explicitly rather than automatically syncing
+        at night currently. This can be revisited in the future.
+
+        Parameters
+        ----------
+        code : lambda_.Code
+            Lambda code bundle to create the initial DB schema.
+        layers : list
+            List of Lambda layers to attach to the lambda function.
+        bucket_name : str
+            S3 bucket name.
+        vpc : ec2.Vpc
+            Virtual private cloud that this lambda should be placed in.
+        """
+        lambda_.Function(
+            self,
+            id="db-synchronizer",
+            function_name="db-synchronizer",
+            code=code,
+            handler="SDSCode.database.synchronizer.lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_12,
+            timeout=cdk.Duration.seconds(180),
+            memory_size=2048,
+            allow_public_subnet=True,
+            vpc=vpc,
+            vpc_subnets=self.rds_subnet_selection,
+            security_groups=[self.rds_security_group],
+            environment={
+                "S3_BUCKET": bucket_name,
+                "SECRET_NAME": self.secret_name,
+            },
+            layers=layers,
+            architecture=lambda_.Architecture.ARM_64,
+        )
--- a/sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
@@ -1,0 +1,105 @@
+"""Synchronize database with S3 bucket.
+
+This script compares the contents of an S3 bucket with a database table and
+updates the database with any missing files or removes entries for deleted
+files.
+"""
+
+import logging
+import os
+from datetime import datetime
+
+import boto3
+import imap_data_access
+from sqlalchemy import delete, select
+
+from . import database as db
+from . import models
+
+logger = logging.getLogger(__name__)
+
+
+def lambda_handler(event, context):
+    """Entry point to the database synchronizer lambda.
+
+    Parameters
+    ----------
+    event : dict
+        The JSON formatted document with the data required for the
+        lambda function to process
+    context : LambdaContext
+        This object provides methods and properties that provide
+        information about the invocation, function,
+        and runtime environment.
+
+    """
+    logger.info("Synchronizing database with S3 bucket")
+
+    # S3 and database configuration
+    client = boto3.client("s3")
+    bucket = os.getenv("S3_BUCKET")
+    # Paginate through S3 objects (needed because we likely have more than 1000 items)
+    # TODO: Do we want to limit the scope of these query comparisons?
+    #       We may run into performance issues if we have a large number of files.
+    #       Could put an outer loop over instrument + level if needed.
+    paginator = client.get_paginator("list_objects_v2")
+    prefix = "imap/"
+    s3_files_dict = {}
+    for page in paginator.paginate(Bucket=bucket, Prefix=prefix):
+        if "Contents" in page:
+            s3_files_dict.update(
+                {obj["Key"]: obj["LastModified"] for obj in page["Contents"]}
+            )
+
+    s3_files = set(s3_files_dict.keys())
+
+    # Fetch database entries
+    with db.Session() as session:
+        with session.begin():
+            query = select(models.FileCatalog.file_path)
+            search_results = session.execute(query).all()
+
+        # result is a one-element tuple, so we need to extract the filepath
+        db_files = set([result[0] for result in search_results])
+
+        # Find discrepancies
+        s3_only_files = set(s3_files) - db_files
+        db_only_files = db_files - set(s3_files)
+
+        if len(s3_files) == 0 and len(db_files) == 0:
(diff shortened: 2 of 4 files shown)


sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
```diff
@@ -0,0 +1,104 @@
...
+
+            file_params["file_path"] = filename
+            file_params["ingestion_date"] = s3_files_dict[filename]
+
+            session.add(models.FileCatalog(**file_params))
+
+        # Remove database entries for files that were deleted from s3
+        delete_statement = delete(models.FileCatalog).where(
+            models.FileCatalog.file_path.in_(db_only_files)
+        )
```
tech3371: Code looks great but I have general concern. What cases are we thinking that will cause this issue? Should we have a human intervention before we delete these records. May be send a SNS or something to our dev to look into this.

maxinelasp: I think having some kind of email notification on this would be great. It would be a good way of tracking if science files were deleted. We could send an event to eventbridge that can send an email.

sds_data_manager/constructs/database_construct.py
```diff
@@ -155,3 +156,50 @@ def __init__(
...
+            security_groups=[self.rds_security_group],
+            environment={
+                "S3_BUCKET": bucket_name,
+                "SECRET_NAME": self.secret_name,
+            },
+            layers=layers,
+            architecture=lambda_.Architecture.ARM_64,
+        )
+
+        # Every day at 3 UTC
```
maxinelasp: It would be great to trigger this immediately prior a chunk of processing in addition to the once a day (like when a batch of L0 files arrives)

sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
```diff
@@ -0,0 +1,104 @@
...
+import imap_data_access
+from sqlalchemy import delete, select
+
+from . import database as db
+from . import models
+
+logger = logging.getLogger(__name__)
+
+
+def lambda_handler(event, context):
```
maxinelasp, line 22: Do you think it would make sense to add a dependency db sync step here? So not just synchronizing the S3 files and the db, but also the dependency file and db?

sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
```diff
@@ -0,0 +1,104 @@
...
+        and runtime environment.
+
+    """
+    logger.info("Synchronizing database with S3 bucket")
+
+    # S3 and database configuration
+    client = boto3.client("s3")
+    bucket = os.getenv("S3_BUCKET")
+    # Paginate through S3 objects (needed because we likely have more than 1000 items)
+    # TODO: Do we want to limit the scope of these query comparisons?
```
maxinelasp, line 42: Could we just go through recently modified files? Would that track deleted files...? Is there some S3 functionality we can leverage there?

maxinelasp (approved):
I like this change, although I worry it will slow down a lot when we get loads of files. It's possible that our data set is not large enough to cause issues in AWS though. I do think that's something we can address if it becomes an issue, as well.

sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
greglucas: I think there is some concern over doing this automatically, so I'm thinking about removing the automation part of this and just adding the lambda for us to trigger manually for now.

We can set up automation later if needed.

sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
greglucas: That is a good idea! Lets do it as a follow-up but a dependency sync would also be useful I think.

sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
greglucas: This is a really good idea. A quick search didn't show me anything obvious (let me know if I missed it), so I think we do have to iterate through all files.

This did get me to thinking though that our current events only do things when files arrive, I think we could add an event handler for when files get deleted too, which would help with the synchronization thing too.

maxinelasp: Yeah, that makes a lot of sense. I think if we make indexer a little simpler, remove all unique constraints except file path, and add in all events (add, delete, modify) then I can feel confident enough in that being accurate that we can then run the slow synchronizer only as needed.

sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
```diff
@@ -0,0 +1,104 @@
...
+            # delete mission key from metadata params
+            file_params.pop("mission")
+            file_params["start_date"] = datetime.strptime(
+                file_params.pop("start_date"), "%Y%m%d"
+            )
+
+            file_params["file_path"] = filename
+            file_params["ingestion_date"] = s3_files_dict[filename]
+
+            session.add(models.FileCatalog(**file_params))
```
laspsandoval, line 96: An idea that you might consider. I read this was possible so you might try it out.

records_to_add = []
for filename in s3_only_files:
...
    records_to_add.append(file_params)

session.bulk_insert_mappings(models.FileCatalog, records_to_add)

laspsandoval: Similarly you could do a batch delete.

sds_data_manager/constructs/database_construct.py
```diff
@@ -155,3 +156,50 @@ def __init__(
...
+            Virtual private cloud that this lambda should be placed in.
+        """
+        synchronizer_lambda = lambda_.Function(
+            self,
+            id="db-synchronizer",
+            function_name="db-synchronizer",
+            code=code,
+            handler="SDSCode.database.synchronizer.lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_12,
+            timeout=cdk.Duration.seconds(180),
```
laspsandoval, line 186: Is this enough time for what you are doing? Should it be set to the max (15 min)?

sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
```diff
@@ -0,0 +1,104 @@
...
+                filename.split("/")[-1]
+            )
+
+            # delete mission key from metadata params
+            file_params.pop("mission")
+            file_params["start_date"] = datetime.strptime(
+                file_params.pop("start_date"), "%Y%m%d"
+            )
+
+            file_params["file_path"] = filename
```
laspsandoval, line 94: An idea that you might consider. I read this was possible so you might try it out.

records_to_add = []
for filename in s3_only_files:
...
records_to_add.append(file_params)

session.bulk_insert_mappings(models.FileCatalog, records_to_add)

Similarly you could do a batch delete.

laspsandoval (approved):
Makes sense to me. I put a couple of ideas/suggestions, but you don't have to take them.

sds_data_manager/constructs/database_construct.py
greglucas: I sure hope we can synchronizer within 3 minutes. If it is taking 15 minutes to go through we are going to have trouble I think. I'm always a fan of starting smaller and running up against the limit rather than making it too large initially.

sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
greglucas: Good idea, I am not sure what the difference is between `session.add(x), session.add(y), session.commit()` vs `session.add_all([x, y]), session.commit()` because I thought everything happens during the commit stage, but your idea seems better 👍 . I changed this to `session.add_all(records_to_add)` because I don't think we have a mapping here but rather the instances themselves in our list.

PR #344, opened 2024-09-20
MNT/ENH: Separate database tables for different file types

### LINKED ISSUE #343: Create a way to track multiple file types
### Summary

Currently we have a single `FileCatalog` DB Table:
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/44774864a373b8f6019581288539587bcff8ae65/sds_data_manager/lambda_code/SDSCode/database/models.py#L148

This has rigidly defined schemas for Science Files currently, but we also need to keep track of SPICE and Ancillary files. We either need a way to put those into the current database table by creating optional fields, or make separate tables. It is likely easier to separate these out to separate tables and then do a join along file_path if we do need to track "all s3 files", or simply do a paginated search of s3 itself.

### Requirements

The more specific the better.
  - [ ] Rename FileCatalog to `ScienceFiles`
  - [ ] Create a table for SPICE `SPICEFiles`
  - [ ] Create a table for ancillary files `AncillaryFiles`

### Related tickets

_No response_

greglucas:
The current FileCatalog has categories that don't make sense for other types of files. To handle this, we can separate out the different types of files into different tables with different required columns.

closes #343

## Notes

* Will fail until https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/89 is merged in `imap-data-access`
* The `file_path` is the unique key in the database for all file type tables
* Rename `FileCatalog` to `ScienceFiles`
* Get valid instruments and data levels from `imap-data-access` which is our canonical location for storing this information

## Questions

Is there additional info we want stored in the SPICE or Ancillary DB? Currently:
```
    file_path = Column(String, nullable=False, primary_key=True, unique=True)
    start_date = Column(DateTime, nullable=True)
    end_date = Column(DateTime, nullable=True)
    extension = Column(String, nullable=False)
```


sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -24,36 +25,13 @@
...
-    "l1cb",
-    "l1d",
-    "l2",
-    "l2pre",
-    "l3",
-    "l3a",
-    "l3b",
-    "l3c",
-    "l3d",
+    *imap_data_access.VALID_DATALEVELS,
```
tech3371, line 34: Thank you for doing these changes!

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -174,6 +140,30 @@ class FileCatalog(Base):
...
 
+class SPICEFiles(Base):
+    """SPICE files table."""
+
+    __tablename__ = "spice_files"
+
+    file_path = Column(String, nullable=False, primary_key=True, unique=True)
+    start_date = Column(DateTime, nullable=True)
+    end_date = Column(DateTime, nullable=True)
+    extension = Column(String, nullable=False)
```
tech3371, line 152: Can you add a version column? I believe all SPICE files has version associated with it.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -174,6 +140,30 @@ class FileCatalog(Base):
...
+
+class AncillaryFiles(Base):
+    """Ancillary files table."""
+
+    __tablename__ = "ancillary_files"
+
+    file_path = Column(String, nullable=False, primary_key=True, unique=True)
+    start_date = Column(DateTime, nullable=True)
+    end_date = Column(DateTime, nullable=True)
+    extension = Column(String, nullable=False)
```
tech3371, line 165: here too. Looks like we are requesting version on ancillary files as well [here](https://imap-processing.readthedocs.io/en/latest/data-access/calibration-files.html)

sds_data_manager/lambda_code/SDSCode/database_handler.py
```diff
@@ -24,7 +24,7 @@ def update_file_catalog_table(session, metadata_params):
     """
     try:
         # Add data to the file catalog table
-        session.add(models.FileCatalog(**metadata_params))
+        session.add(models.ScienceFiles(**metadata_params))
```
tech3371: Right now, it's only called by indexer lambda but we may need to add some logics in the future if this function is called by others too to make sure we update we update the correct table. Or change this function name to be `update_science_catalog_table` or something. Just a suggestions. You are better with names.

sds_data_manager/lambda_code/SDSCode/query_api.py
```diff
@@ -37,15 +37,15 @@ def lambda_handler(event, context):
     query_params = event["queryStringParameters"]
 
     # select the file catalog for the query
```
tech3371:
```suggestion
    # select the table for the query
```

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: Yes, I added version to both SPICE and Ancillary now. The SPICE one I left as optional because I think there may be some SPICE files without versions like `de430.bsp` or the main ones that NAIF gives us.

sds_data_manager/lambda_code/SDSCode/database_handler.py
greglucas: I actually inlined the `update_file_catalog` since it was only a one-line function in that extra file. I agree though that we will need to follow up with more handling of ancillary and spice file uploads later.

sds_data_manager/lambda_code/SDSCode/query_api.py
greglucas: Made a lot of text updates like this with a global find and replace, thanks for finding this!

sds_data_manager/lambda_code/SDSCode/indexer.py
```diff
@@ -179,9 +178,9 @@ def s3_event_handler(event):
     ingestion_date_object = get_file_creation_date(s3_filepath)
 
     file_params["ingestion_date"] = ingestion_date_object
-    with db.Session() as session:
-        update_file_catalog_table(session, file_params)
-    logger.info("Wrote data to file catalog table")
+    with db.Session() as session, session.begin():
+        session.add(models.ScienceFiles(**file_params))
```
tech3371, line 182: that's better!

tech3371: Can you add TODO in here that we will need to call different table based on which files we ingest?

tech3371 (approved):
Thank you!

tests/lambda_endpoints/conftest.py
```diff
@@ -36,38 +36,18 @@ def science_file():
...
-def s3_client(science_file, spice_file):
+def s3_client():
     """Mock S3 Client, so we don't need network requests."""
     with mock_s3():
         s3_client = boto3.client("s3", region_name="us-east-1")
 
         s3_client.create_bucket(
             Bucket=BUCKET_NAME,
         )
-        result = s3_client.list_buckets()
```
maxinelasp, line 51: Why was this test removed?

greglucas: This is actually a fixture and not a test, so shouldn't be asserting things, the dependent tests will fail if this isn't set up properly.

But, the real reason I removed it was that this was actually creating an s3 bucket **and** populating it with items, and I'm saying we don't want the population of items here (because those items aren't added to the DB properly) and we'll populate it later for each test that needs it.

maxinelasp: Sounds good, thanks for explaining!z

laspsandoval (commented):
Looks good. I didn't see anything that needed to be changed.

sds_data_manager/lambda_code/SDSCode/indexer.py
greglucas: I made an issue which I prefer to adding TODOs in the code: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/350

greglucas:
This is failing because we need a new release of `imap-data-access` that includes: https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/89

PR #345, opened 2024-09-20
Update EventBridge Rule to match new spice path in data bucket

tech3371:
# Change Summary

## Overview
Updated Event Bridge rule to trigger on any event in `spice/` directory on s3 data bucket.

## New Dependencies
Update to match this new update on `imap-data-access`

## Updated Files
- sds_data_manager/constructs/efs_construct.py
   - update rule


maxinelasp (approved):
Good catch

PR #347, opened 2024-09-23
HIT - Use Raw L0 File as Single L1A Dependency

### LINKED ISSUE #346: Update HIT L0 Dependency Table
### Summary

Need to update the HIT L0 dependency table to run all L1A processing from single raw L0 file:

HIT currently runs L1A housekeeping and science processing from separate dependencies:
```
hit, l0, sci, hit, l1a, sci, HARD, DOWNSTREAM
hit, l0, hk, hit, l1a, hk, HARD, DOWNSTREAM
```

These should be combined so all L1A is triggered from the L0 raw file:
```
hit, l0, raw, hit, l1a, all, HARD, DOWNSTREAM
```

sdhoyt:
# Change Summary

## Overview
This PR updates the HIT dependency table to use a single raw ccsds as the depdency for all L1A processing instead of splitting it up by housekeeping and science

Closes #346

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/dependency_config.csv
+++ b/sds_data_manager/lambda_code/SDSCode/dependency_config.csv
@@ -59,8 +59,7 @@ hi, l1b, 45sensor-de, hi, l1c, 45sensor-pset, HARD, DOWNSTREAM
 
 # <---- HIT Dependencies ---->
 
-hit, l0, sci, hit, l1a, sci, HARD, DOWNSTREAM
-hit, l0, hk, hit, l1a, hk, HARD, DOWNSTREAM
+hit, l0, raw, hit, l1a, all, HARD, DOWNSTREAM
 
 hit, l1a, sci, hit, l1b, sci, HARD, DOWNSTREAM
 hit, l1a, hk, hit, l1b, hk, HARD, DOWNSTREAM


tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/dependency_config.csv
```diff
@@ -59,8 +59,7 @@ hi, l1b, 45sensor-de, hi, l1c, 45sensor-pset, HARD, DOWNSTREAM
 
 # <---- HIT Dependencies ---->
 
-hit, l0, sci, hit, l1a, sci, HARD, DOWNSTREAM
-hit, l0, hk, hit, l1a, hk, HARD, DOWNSTREAM
+hit, l0, raw, hit, l1a, all, HARD, DOWNSTREAM
```
vmartinez-cu, line 62: With this change, will cli be able to create the hk product without the science product while that code is still under development?

vmartinez-cu (approved):

sds_data_manager/lambda_code/SDSCode/dependency_config.csv
sdhoyt: The HIT L0 with a `raw` descriptor will now trigger the L1A processing like the `hk` and `sci` files did. So it can have one or both inside the L0. Since the HIT sample data is the same date (I think), but are split up by `hk` and `sci`, you'll just need to keep track on your end which is which and you won't be able to have both in the SDC at the same time

Issue #299, opened 2024-06-17
BUG - Dependency manager database creator lambda running on destroy


maxinelasp:
### Description of the issue

When the destroy step to remove the CreateSchemaStack runs, it triggers the lambda to create the database tables.  This results in duplicate entries into the database.

Workaround: Destroy the stack and then drop the tables, this way you won't end up with duplicates. 

### Steps to reproduce the issue

It's not always consistent. 

1. Drop the tables in datagrip
2. Run CDK destroy on CreateSchemaStack
3. See that the tables have been recreated

### Expected behavior (What should happen)

The destroy should either destroy the tables or leave them deleted

### Actual behavior (What does happen)

The destroy runs the create tables lambda

### Additional notes, affected areas, and suggested fixes

Suggested fixes:
- Add a check in the lambda for database non-existence
- Make it so the lambda drops any existing tables as a first step 
- Make it so the lambda checks to see if the db entries exist before creating them
- generally, making the lambda idempotent is probably the best way to fix this

More info in [this slack thread](https://lasp.slack.com/archives/C04L7EUMJ8G/p1718402531277479)

maxinelasp:
This lambda should be idempotent. It also does not run on deploy which is wrong. The main issue right now is that if a dependency is removed from the file, the table is not updated. There is a constraint that prevents duplicates.

maxinelasp:
This needs to be updated in create_schema.py

tech3371:
Do you think this PR takes care of this bug? https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/351

maxinelasp:
To add a note - we have decided to move dependency management away from the database and into a lambda, which will render this bug fix unnecessary.

PR #349, opened 2024-09-24
Fixing unique constraint for dependency table

### LINKED ISSUE #348: BUG - Batch starter will submit two identical dependencies
### Description of the issue

Due to our issues with the dependency management table, sometimes a job will retrieve the same dependent file twice. Batch starter will then send both dependencies through, when really it should only send it once. 

Add a duplicate line to the dependency management db table. Attempt to run batch starter. See that there are duplicate dependencies. 

The dependencies should be unique

_No response_

### Code Snippet:

```python
Code
```

#299 

To fix: 
- update unique constraint to not include ID in database
- Update batch_starter to remove duplicate files

maxinelasp:
# Change Summary
This adds a check within batch_starter to remove duplicate dependencies returned from sciencefiles/filecatalog before moving along, due to changes in #344 that I didn't want to conflict with.

This fixes part of but not all of #299 and completely fixes #348. The MAG team noticed that processing didn't run because the dependency was incorrect, so I wanted to get this fix in quickly. 

I did also open another issue in imap processing related to this bug: https://github.com/IMAP-Science-Operations-Center/imap_processing/issues/870

### DIFF
--- a/tests/lambda_endpoints/test_batch_starter.py
+++ b/tests/lambda_endpoints/test_batch_starter.py
@@ -238,6 +238,15 @@ def test_pre_processing_dependency(session):
     assert downstream_dependency[0]["descriptor"] == "norm-mago"
 
 
+def test_duplicate_dependencies(session):
+    """Tests the unique constraint in the dependency table."""
+    _populate_dependency_table(session)
+
+    with pytest.raises(IntegrityError):
+        # Duplicate dependencies should raise an error
+        _populate_dependency_table(session)
+
+
 def test_get_file(session):
     """Tests the get_file function."""
     _populate_file_catalog(session)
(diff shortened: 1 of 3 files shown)


vmartinez-cu (approved):
LGTM

tech3371 (approved):
Thank you for making this change!

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -238,6 +238,27 @@ def test_pre_processing_dependency(session):
     assert downstream_dependency[0]["descriptor"] == "norm-mago"
 
 
+def test_duplicate_dependencies(session):
+    """Tests the unique constraint in the dependency table."""
+    _populate_dependency_table(session)
+    _populate_dependency_table(session)
```
greglucas, line 244: Just a note for future selves that this might cause an issue if/when we make a unique constraint on the table. I think this is still a good test and we'll just need to verify that this second call raises the expected error.

greglucas (approved):

tests/lambda_endpoints/test_batch_starter.py
maxinelasp: Agreed, but this test does mean that we need to understand what the consequences are if this method is called twice, which is good enough for me right now!

PR #351, opened 2024-09-26
Added way to trigger create schema lambda automatically when dependency config is updated

tech3371:
# Change Summary

## Overview
Added way to trigger create schema lambda and also added checks from adding duplicate records.

## New Dependencies
When we update the dependency config, we also need to update the `last_dependency_update": "09/26/2024"` in database_construct.py

## Updated Files
- sds_data_manager/constructs/database_construct.py
   - Updates to trigger create schema lambda automatically at CDK deploy
- sds_data_manager/lambda_code/SDSCode/create_schema.py
   - checks to avoid adding duplicate record to dependency table
- sds_data_manager/lambda_code/SDSCode/dependency_config.csv
   - Added SWE L2 to the config

## Testing
Tested that this changes works in AWS

### DIFF
--- a/sds_data_manager/constructs/database_construct.py
+++ b/sds_data_manager/constructs/database_construct.py
@@ -149,9 +149,21 @@ class SdpDatabase(Construct):
         res_provider = cr.Provider(
             self, "crProvider", on_event_handler=schema_create_lambda
         )
+
+        # Setting 'properties' provides a mechanism to re-trigger the
+        # `schema_create_lambda` during CDK deployment after initial deployment.
+        # This ensures that the schema is updated automatically whenever
+        # there are changes in the dependency config. Before, the Lambda
+        # function was only triggered on stack creation or deletion,
+        # but not during updates. Since the service token remains unchanged,
+        # the 'Update' event wouldn't trigger the `db_custom_resource`.
         db_custom_resource = CustomResource(
-            self, "CustomResource-DB-Schema", service_token=res_provider.service_token
+            self,
+            "CustomResource-DB-Schema",
+            service_token=res_provider.service_token,
+            properties={"last_dependency_update": "09/26/2024"},
         )
+
         # Add an explicit dependency on the RDS instance because we need the secret
         # populated with the DB credentials before we can create the schema.
         db_custom_resource.node.add_dependency(db)
--- a/sds_data_manager/lambda_code/SDSCode/create_schema.py
+++ b/sds_data_manager/lambda_code/SDSCode/create_schema.py
@@ -7,6 +7,7 @@ https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.custom_resources/README.ht
 import logging
 
 from SDSCode.database import database as db
+from SDSCode.database import models
 from SDSCode.database.models import Base
 from SDSCode.dependency_config import all_dependents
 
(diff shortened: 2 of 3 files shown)


sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -17,18 +18,48 @@
...
-        session.commit()
+        if event.get("RequestType") == "Create":
+            logger.info("Creating RDS tables")
+            # Create tables
+            Base.metadata.create_all(db.get_engine())
+            session.add_all(all_dependents)
+
+        elif event.get("RequestType") == "Update":
+            for dependent in all_dependents:
+                existing_record = (
```
greglucas, line 38: Can we make a unique index on the database to avoid duplicates? It seems like we should be able to do a large insert/update of all of the items. If there are entries already this would just "update" that entry. Maybe we need to have a better index set up to handle this instead of iterating through items.

greglucas (commented):
I don't love this because it still requires manual intervention to update that date string.

What about separating these things out into separate functions?

1. In this create_schema lambda, that is literally all we do.
```python
        if event.get("RequestType") == "Create":
            logger.info("Creating RDS tables")
            # Create tables
            Base.metadata.create_all(db.get_engine())
```
2. We make another Lambda that is called something like `update_dependency_table`, which updates the dependencies. That specific dependency update could be triggered as a step after the layer creation (making the lambda depend on the lambda layer itself I think) because the layer includes information about this table in it, I think every time we update the csv file we create a new layer.

sds_data_manager/lambda_code/SDSCode/create_schema.py
```diff
@@ -17,18 +18,48 @@
...
+                        == dependent.dependent_data_level,
+                        models.PreProcessingDependency.dependent_descriptor
+                        == dependent.dependent_descriptor,
+                        models.PreProcessingDependency.relationship
+                        == dependent.relationship,
+                        models.PreProcessingDependency.direction == dependent.direction,
+                    )
+                    .first()
+                )
+                if existing_record is None:
```
maxinelasp, line 59: This doesn't remove records that were deleted from the file, so it still doesn't totally solve the ticket. I think there should be a unique constraint on the table that prevents duplicates, but I honestly lost track of if that unique constraint still exists or ever existed?

sds_data_manager/constructs/database_construct.py
```diff
@@ -149,9 +149,21 @@ def __init__(
...
+        # there are changes in the dependency config. Before, the Lambda
+        # function was only triggered on stack creation or deletion,
+        # but not during updates. Since the service token remains unchanged,
+        # the 'Update' event wouldn't trigger the `db_custom_resource`.
         db_custom_resource = CustomResource(
-            self, "CustomResource-DB-Schema", service_token=res_provider.service_token
+            self,
+            "CustomResource-DB-Schema",
+            service_token=res_provider.service_token,
+            properties={"last_dependency_update": "09/26/2024"},
```
maxinelasp, line 164: Is this piece what creates the "update" event that is used in create_schema?

maxinelasp (changes requested):
It still doesn't accomodate the case where a dependency is deleted from dependency_config. I'd probably prefer always just deleting all rows and re-creating them, it's simpler and more reliable. But, if you add the delete step, then I think this will fix the issues we've been seeing

maxinelasp:
Related PR: #352

tech3371:
Pause reviewing this PR as this PR may become irrelevant due to changes in this PR: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/352

sds_data_manager/constructs/database_construct.py
tech3371: yes. That learned something new.

tech3371:
This PR is irrelevant now with new design and changes.

PR #352, opened 2024-09-27
MNT: Change the dependency database table into a local lookup

greglucas:
# Change Summary

## Overview

This changes the `PreProcessingDependency` table into an in-memory "database" that is stored as nested dictionaries of "nodes". This makes it simpler to keep track of (IMO) and makes sure we are always in sync when deploying code changes because they are read into Python objects directly, thus avoiding any duplicates and other synchronization issues for a table that gets deployed/updated. The csv dependency file is still the source of truth and what populates this dictionary lookup.

@maxinelasp I'd like to hear from you especially on this since you've said that we need a database, but I'm still not sure what we need the capabilities of a SQL search for yet, thus the proposal here to try and simplify things.

My vision for this is that this will tell us what the data product mapping is (product A -> product B). Then we can go and query the Science / SPICE / Ancillary file tables for the necessary products and versions to see if those things exist. So this is basically a lookup table mapping of initial items to check. We still use the DB File Tables to query for actual items.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/batch_starter.py
@@ -9,6 +9,7 @@ from imap_data_access import ScienceFilePath
 from sqlalchemy import select
 from sqlalchemy.exc import IntegrityError
 
+from . import dependency_config
 from .database import database as db
 from .database import models
 
@@ -20,24 +21,20 @@ logger.setLevel(logging.INFO)
 BATCH_CLIENT = boto3.client("batch", region_name="us-west-2")
 
 
-def get_dependencies(
-    session, instrument, data_level, descriptor, direction, relationship
-):
-    """Make a query to the dependency table to get all dependencies.
+def get_dependencies(node, direction, relationship):
+    """Lookup the dependencies for the given ``node``.
+
+    A ``node`` is an identifier of the data product, which can be an
+    (instrument, data_level, descriptor) tuple, SPICE file identifiers,
+    or ancillary data file identifiers.
 
     Parameters
     ----------
-    session : orm session
-        Database session.
-    instrument : str
-        Primary instrument that we are looking for its dependency.
-    data_level : str
-        Primary data level.
-    descriptor : str
-        Primary data descriptor.
-    direction: str
+    node : tuple
+        Quantities that uniquely identify a data product.
+    direction : str
         Whether it's UPSTREAM or DOWNSTREAM dependency.
-    relationship: str
+    relationship : str
         Whether it's HARD or SOFT dependency.
         HARD means it's required and SOFT means it's nice to have.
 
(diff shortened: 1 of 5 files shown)


maxinelasp:
Major drawback - updating this file without a dependency will require re-deploys of all the lambdas. In dev, we redeploy these a fair amount, but in prod, it is riskier and more annoying. 

We also can't assume that the person deploying knows what services use this table. That means we will need to redeploy everything just in case, so we can update everything that uses it. 

It also means anything not using sds-data-manager can't access this table, or has to access it in a roundabout way, which may or may not be a problem. 

I agree that having a file and a database that syncs up to that file is not great and is causing us issues. I am open to a compromise, like maybe if the .csv file is maintained in the S3 bucket and that acts as the source of truth? Then that piece could be updated without having to redeploy everything.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
```diff
@@ -20,24 +21,20 @@
...
-    session : orm session
-        Database session.
-    instrument : str
-        Primary instrument that we are looking for its dependency.
-    data_level : str
-        Primary data level.
-    descriptor : str
-        Primary data descriptor.
-    direction: str
+    node : tuple
```
tech3371, line 33: can we use another name for this?

sds_data_manager/lambda_code/SDSCode/dependency_config.py
```diff
@@ -41,9 +50,11 @@
             raise ValueError(f"Each dependency must have 8 items\nCurrent line: {line}")
 
         logger.info(contents)
-        dependency = PreProcessingDependency(**{h: c for h, c in zip(header, contents)})
-        downstream_dependents.append(dependency)
-
-upstream_dependents = [dep.reverse_direction() for dep in downstream_dependents]
-
-all_dependents = downstream_dependents + upstream_dependents
+        # Instrument, data level, descriptor
```
tech3371, line 53: can you write an example of what it will look like in the comment for reference later?

tech3371:
I am onboard about moving this dependency tracking away from DB because it's causing more complication to keep it sync with latest changes. I don't know yet how we are going deploy things to production but I hope we don't update production right away like we do to dev from PR merge.

A question about s3 option, we still need to trigger s3 upload automatically to upload most recent files, right? I am wondering how we want to do that or will that still add similar complication like I have on my other PR. I feel like this is a big change and it may be easier to discuss on Slack or hop on zoom when Greg gets back.

sds_data_manager/lambda_code/SDSCode/batch_starter.py
laspsandoval: Agreed. "identifier"?

laspsandoval (commented):
Agreed w/ both Maxine and Tenzin :) Here is my take on it:

I like the idea of eliminating the DB table since it is causing complications. So I think this is a good start as a PR. There is a possibility that early in the mission (1 year in) the ancillary dependencies will dramatically change and I think we should be prepared to handle that without (as Maxine said) always redeploying. It is hard for me to predict whether or not s3 will be needed until we define how our ancillary table will be designed and used. Some things that I think would be good to clarify:

Will the ancillary db tables track lookup tables? 
What are the specific needs of each instrument? 
Will we be able to just simply change the version and use latest or do we forsee new dependencies? 
How will batch_starter.py use the ancillary db tables?

From my perspective to keep the ball moving we should approve and start working on the ancillary tables and modify to use s3 (as suggested by Maxine) if we find that it is needed.

tech3371 (approved):
This PR is good first steps and we will need to create new PR to refactor into its own lambda if that's the route we want to go. LGTM

PR #374, opened 2024-10-30
I-ALiRT: Add a way to delete stack despite using autoscaling

### LINKED ISSUE #375: I-ALiRT: Add a way to delete stack despite using autoscaling
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Previously you had to manually go in and delete the autoscaling groups before you destroyed the Stack. This allows the stack to be destroyed without that manual process.

## Updated Files
- ialirt_processing_construct
   - update the AsgCapacityProvider so that (1) the ECS can terminate ASG even if EC2 is running tasks and (2) avoid potential issues with automated scaling during cdk destroy.

## Testing
This now destroys the stack: cdk destroy --context account_name="lcs" IalirtStack

### DIFF
--- a/sds_data_manager/constructs/ialirt_processing_construct.py
+++ b/sds_data_manager/constructs/ialirt_processing_construct.py
@@ -270,6 +270,8 @@ class IalirtProcessing(Construct):
             self,
             f"AsgCapacityProvider{processing_name}",
             auto_scaling_group=auto_scaling_group,
+            enable_managed_termination_protection=False,
+            enable_managed_scaling=False,
         )
 
         self.ecs_cluster.add_asg_capacity_provider(capacity_provider)


sds_data_manager/constructs/ialirt_processing_construct.py
```diff
@@ -266,6 +266,8 @@ def add_autoscaling(self, processing_name):
             self,
             f"AsgCapacityProvider{processing_name}",
             auto_scaling_group=auto_scaling_group,
+            enable_managed_termination_protection=False,
+            enable_managed_scaling=False,
```
maxinelasp, line 274: So does this mean there is no more auto-scaling? I don't know if that's necessary or anything for i-ALiRT, but if we're switching to manual scaling only that will require some kind of close monitoring for the first few months to make sure we're hitting the 5 minute requirement.

laspsandoval: Totally agree. But this part of the script (desired_count=1) will make it so that a new EC2 instance will be launched by the ASG if one is deleted, and ECS will reschedule the task on the new instance.

        self.ecs_service = ecs.Ec2Service(
            self,
            f"IalirtService{processing_name}",
            cluster=self.ecs_cluster,
            task_definition=task_definition,
            security_groups=[self.ecs_security_group],
            desired_count=1,
            vpc_subnets=ec2.SubnetSelection(
                subnet_type=ec2.SubnetType.PRIVATE_WITH_EGRESS
            ),
        )
        
   Just to make certain I tested it. Attached are screenshots.

laspsandoval:
Screenshot 2024-11-08 at 2 09 26 PM
Screenshot 2024-11-08 at 2 10 10 PM
Screenshot 2024-11-08 at 2 11 46 PM

sds_data_manager/constructs/ialirt_processing_construct.py
maxinelasp: That's good, but does I-ALiRT ever need more than one EC2 instance? With autoscaling off, it will never launch new ones even if the existing one(s) are maxed out on CPU or something. So, you would be limited to one EC2 even if more could be needed. That's my main concern.

maxinelasp (approved):

PR #383, opened 2024-11-05
Update HIT dependencies for L1B and L2

vmartinez-cu:
Updated HIT dependencies for L1B and L2 data products

## Updated Files
- dependency_config.csv
   - updated HIT dependencies
- test_batch_starter.py
   - update HIT parameters


tech3371 (approved):
LGTM

PR #404, opened 2024-11-22
Instructions for SSH for OpsSW/IT

### LINKED ISSUE #382: I-ALiRT: ssh capabilities
### Summary

After talking with Melissa Mantey we realized that the ssh capability needs to come from within the container. Therefore the instructions that I wrote were not needed and the work needs to come from OpsSW. I'm keeping the branch around just in case we realize we need something from it.

### LINKED ISSUE #1113: I-ALiRT: Get ssh keys and IP address of EC2 for IT

laspsandoval:
# Change Summary

## Overview
OpsSW needs to use rsync to retrieve raw record files being created in their container running in an ECS task on an EC2 instance. Each EC2 instance is located in a security group and private subnet. They can only be accessed by the Network Load Balancer. This is best practice for production. So I had to consider a few options for how OpsSW can access their raw record files.

**1. Make the private subnet -> public subnet and then allow traffic from a specific LASP IP range and port.**

**Advantages**
* Simplest solution. Requires almost no changes.
* Direct SSH access without additional configuration steps.
* Familiar process: Uses standard, well-understood SSH mechanisms.

**Disadvantages**
* Compromises security: Exposing the subnet as public increases the attack surface and goes against production best practices. (this was the deciding factor)

**2. Use Services Systems Manager (SSM) to create a session that would allow ssh access.**

**Advantages**
* Enhanced security: No public IP exposure; access is tunneled through private SSM sessions.
* For Long-Lived Sessions: SSM is better as it supports up to 12 hours for a single session, with a more flexible idle timeout.
* Free
* Simple implementation
* Can be monitored using CloudTrail or CloudWatch

**Disadvantages**
* Requires an IAM user or role: OpsSW or IT needs an AWS IAM user or role with appropriate permissions (e.g., AmazonSSMFullAccess).
* Timeout after 24 hours (non-idle); 20 min (idle)

**3. Use EC2 Instance Connect Endpoint (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/connect-with-ec2-instance-connect-endpoint.html)**

**Advantages**
* Enhanced security: Keeps instances in private subnets without requiring a public IP.
* Only incurs charges for the endpoint's network usage.
* Can be monitored using CloudWatch/CloudTrail
* No .pem File Required

**Disadvantages**
* Requires AWS IAM permissions for users who need SSH access.
* Requires creating an interface endpoint in the VPC, which adds some complexity to the infrastructure setup (changes in cdk code)
* For Quick Temporary Access: EC2 Instance Connect is simpler for short-term access, but the 1-hour TCP connection limit and 60-second key validity require repeated setup for longer tasks.

Based on the advantages/disadvantages list I selected Option 2. I will need to make certain that the SSM timeout is acceptable with OpsSW/IT. **UPDATE**: Melissa Mantey thinks the timeout will not be an issue for rsync. I think we will move forward with the plan as described in this ticket unless any other objections/issues arise.

## New Files
- ialirt-ssh-access.rst
   - readthedocs description of using SSM

## Updated Files
   - associated keypair to EC2
- index.rst
   - update for readthedocs

## Testing
After deploying the code I went through the steps and verified they all work.

### DIFF
--- a/docs/source/cdk/ialirt-ssh-access.rst
+++ b/docs/source/cdk/ialirt-ssh-access.rst
@@ -1,0 +1,66 @@
+I-ALiRT IT SSH Access
+==================
+
+Create IAM Role for IT
+~~~~~~~~~
+SDC will have already created key-pairs and associated them with the EC2 instances. Note each keypair is required to be restricted::
+
+    chmod 600 <keypair-name.pem>
+
+SDC will create the IAM user, attach the AmazonSSMFullAccess Policy, and securely provide access keys for the user::
+
+    aws iam create-user --user-name <user>
+    aws iam attach-user-policy --user-name <user> --policy-arn arn:aws:iam::aws:policy/AmazonSSMFullAccess
+    aws iam create-access-key --user-name <user>
+
+IT should then follow the directions in Section "Existing User" on this page:
+:ref:`aws-setup`
+
+Setting Up IT SSH Key Pair
+~~~~~~~~~
+
+IT will generate a SSH key pair on their machine::
+
+    ssh-keygen -t rsa -b 2048 -f ~/.ssh/ialirt_key
+
+This command will create two files: a private key (rsa_name) and a public key (rsa_name.pub). IT will keep the private key secure and send the SDC the public key. The SDC will start a Services Systems Manager (SSM) session::
+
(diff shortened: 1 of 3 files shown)


laspsandoval:
**Still discussing this with OpsSW so please do not review this change!**

laspsandoval:
Decided this was not needed.

PR #408, opened 2024-11-25
Add GLOWS L2 as a dependency for L3a GLOWS processing

pleasant-menlo:
# Change Summary

## Overview

Added L2 as a dependency for L3a GLOWS to enable testing of the L3a processor within AWS.

The GLOWS algorithm document says that the descriptor for L2 and L3a histogram should contain the pointing number (i.e. histogram-00001 for the first pointing). Is it possible to set up this dependency in such a way that when any GLOWs L2 file of the form `histogram-XXXXX` is uploaded our L3a processor will be run?

Thanks,
Sebastian and Peter

## Updated Files

- updated dependency_config.csv
   - added a line that will trigger GLOWS l3a processing when an L2 file is uploaded


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
```diff
@@ -46,6 +46,8 @@ glows, l0, raw, glows, l1a, all, HARD, DOWNSTREAM
 glows, l1a, de, glows, l1b, de, HARD, DOWNSTREAM
 glows, l1a, hist, glows, l1b, hist, HARD, DOWNSTREAM
 
+glows, l2, hist, glows, l3a, histogram, HARD, DOWNSTREAM
```
maxinelasp: Could we match descriptors? I.e. everyone uses `hist` or everyone uses `histogram`? I don't have a strong preference either direction, but I would strongly prefer to match.

maxinelasp:
To answer the question in your description, the pointing number is a separate piece of information from the descriptor, and is not used for dependency resolution. So right now, a file arriving that looks like `imap_glows_l2_hist_20241212-0001_v001.cdf` would trigger processing for L3.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
pleasant-menlo: That sounds good to us as well. We will change ours to `hist` to match.

maxinelasp (approved):

bourque (approved):

tech3371 (approved):

tech3371:
I forgot that we need to merge Menlo's PR also.

Issue #396, opened 2024-11-12
BUG - Fix postgresql connection in pytest for batch starter lambda's test


tech3371:
See #395 for more details

Issue #394, opened 2024-11-12
BUG -psycopg error from pytests


tech3371:
### Description of the issue
We are unable to run pytest locally to test anything because of issue listed in the ticket. The logic in the file was added to tests test_batch_starter.py's test_duplicate_job(). I undid that change since it's more important to test other tests. If we want to continue testing batch starter's duplicate job, we will need to revisit. I will create issue to capture that.

Because of new changes in `tests/lambda_endpoints/conftest.py` that checks which database server to run, it causes below issue. I undid lot of new changes in that file related to the issue. We will need to revisit and figure out how we want to resolve that.

### Steps to reproduce the issue

run `pytest` command from sds-data-manager root path.

### Expected behavior (What should happen)

run pytest command without issue

### Actual behavior (What does happen)

/Users/tech3371/Library/Caches/pypoetry/virtualenvs/sds-data-manager-Be1fnlCS-py3.10/lib/python3.10/site-packages/_pytest/config/__init__.py:318: PluggyTeardownRaisedWarning: A plugin raised an exception during an old-style hookwrapper teardown.
Plugin: helpconfig, Hook: pytest_cmdline_parse
ImportError: no pq wrapper available.
Attempts made:
- couldn't import psycopg 'c' implementation: No module named 'psycopg_c'
- couldn't import psycopg 'binary' implementation: No module named 'psycopg_binary'
[...]

PR #409, opened 2024-11-26
BUG: Removing pytest-postgresql temporarily

tech3371:
# Change Summary

## Overview
This is related to BUG - https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/394. I thought I fixed it in that PR but found out today that I need to remove postgresql library in pytests as well.

We will need to fix this bug in https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/396.

This is blocking developing tests for upcoming infrastructure work.

## Updated Files
- pyproject.toml
   - removed pytest-postgresql from pytest dependency list until it will be fixed in future work.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -967,20 +967,6 @@ files = [
 ]
 
 [[package]]
-name = "mirakuru"
-version = "2.5.2"
-description = "Process executor (not only) for tests."
-optional = false
-python-versions = ">=3.8"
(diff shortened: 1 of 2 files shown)


tmplummer (approved):
I don't know all of the details, but you have a future ticket to work this out so I'm happy.

PR #412, opened 2024-12-04
I-ALiRT: Domain Name Update

laspsandoval:
# Change Summary

## Overview
I-ALiRT requires an API for querying and downloading logs. I was hoping to simply use the same domain name. However, when I was deploying I had the error:

**api.dev.imap-mission.com already exists in stack arn:aws:cloudformation:us-west-2:449431850278:stack/SDCStack/3699d440-7217-11ef-9abe-06fae9ea8027**

This small change just allows me to create a new domain for ialirt.

## Updated Files
- stackbuilder.py
   - Updated to allow "Ialirt" string to be passed to api_gateway_construct.ApiGateway
- api_gateway_construct.py
   - Allows prefix to be added for I-ALiRT.

## Testing
Deployed both SDC and I-ALiRT Stacks.

### DIFF
--- a/sds_data_manager/constructs/api_gateway_construct.py
+++ b/sds_data_manager/constructs/api_gateway_construct.py
@@ -4,8 +4,10 @@ Sets up api gateway, creates routes, and creates methods that are linked to the
 lambda function.
 
 An example of the format of the url: https://api.prod.imap-mission.com/query
-https://ialirtapi.prod.imap-mission.com/ialirt-log-query
+https://ialirt.prod.imap-mission.com/ialirt-log-query
 """
+
+from typing import Optional
 
 from aws_cdk import Duration, aws_sns
 from aws_cdk import aws_apigateway as apigw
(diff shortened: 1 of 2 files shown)


sds_data_manager/constructs/api_gateway_construct.py
```diff
@@ -61,26 +66,26 @@ def __init__(
         if domain_construct is not None:
             custom_domain = apigw.DomainName(
                 self,
-                "RestAPI-DomainName",
-                domain_name=f"api.{domain_construct.domain_name}",
+                f"{self.prefix}RestAPI-DomainName",
+                domain_name=f"{self.prefix.lower()}api.{domain_construct.domain_name}",
```
maxinelasp: I prefer `ialirt.prod.imap-mission.com` to `ialirtapi.prod.imap-mission.com`. It may also make sense to use `ialirt.imap-mission.com` and `dev.ialirt.imap-mission.com`. `alirtapi` is not easy to type or read. 

Could also do something like `api.ialirt.imap-mission.com`...?

sds_data_manager/utils/stackbuilder.py
```diff
@@ -281,6 +281,7 @@ def build_sds(
         construct_id="IAlirtApiGateway",
         domain_construct=domain,
         certificate=root_certificate,
+        ialirt_prefix="Ialirt",
```
maxinelasp: Could we make it `ialirt` or `IAlirt` to match?

maxinelasp (approved):
Looks good, thanks for making those changes

PR #417, opened 2024-12-11
I-ALiRT: Combine tools NLB and ECS Cluster

### LINKED ISSUE #389: I-ALiRT: Look into combining resources (NLB and ECS)
### Summary

We have redundant resources for packet ingest (NLB and ECS). Previously we did this so we would have completely redundant stacks that could be deployed separately while maintain functionality. This was important because ECS does not automatically update when a new image is pushed. However, the following command can be used to update the Service.

aws ecs update-service \
    --cluster IalirtCluster \
    --service IalirtServicePrimary \
    --force-new-deployment \
    --deployment-configuration maximumPercent=200,minimumHealthyPercent=100 > /dev/null

Example:
aws ecs update-service --cluster IalirtStack-IalirtProcessingIalirtCluster239B5309-KRdDdZpnvLuN --service IalirtStack-IalirtProcessingIalirtServicePrimaryService78BD5E27-nJxaEvhNxhtD --force-new-deployment --deployment-configuration maximumPercent=200,minimumHealthyPercent=100 > /dev/null

laspsandoval:
# Change Summary

## Overview
We had redundant resources for packet ingest (NLB, ECS Cluster, Security Groups, ASG). Previously we did this so we would have completely redundant constructs (previously they were stacks) that could be deployed separately while maintain functionality. This was important because ECS does not automatically update when a new image is pushed. However, the following command can be used to update the Service.

aws ecs update-service
--cluster IalirtCluster
--service IalirtServicePrimary
--force-new-deployment
--deployment-configuration maximumPercent=200,minimumHealthyPercent=100 > /dev/null

Example:
aws ecs update-service --cluster IalirtStack-IalirtProcessingIalirtCluster239B5309-KRdDdZpnvLuN --service IalirtStack-IalirtProcessingIalirtServicePrimaryService78BD5E27-nJxaEvhNxhtD --force-new-deployment --deployment-configuration maximumPercent=200,minimumHealthyPercent=100 > /dev/null

This PR removes the redundant tools while maintaining the same functionality as it did previously.

## Updated Files
- stackbuilder.py
   - Removes for loop to create the two redundant  systems
   - Remove redundant security groups (ECS security group and Load Balancer security group)
   - Switched to a single Network Load Balancer and ECS Cluster
   - Switched to a single Autoscaling Group
   - Increased size of EC2 instance being used

## Testing
- Confirmed that the code deployed and had same functionality as previously
- Changed test container in Nexus and then implemented the example command shown above. Verified that the task restarted seamlessly by visiting:

http://ialirt-ialir-9tlmgrxkquge-848bbb993d4f9ede.elb.us-west-2.amazonaws.com:1234

and seeing: "Hello from Port 1234 - update!" in my browser window

### DIFF
--- a/sds_data_manager/constructs/ialirt_processing_construct.py
+++ b/sds_data_manager/constructs/ialirt_processing_construct.py
@@ -86,7 +86,7 @@ class IalirtProcessing(Construct):
         )
 
         # Only allow traffic from the NLB security group
-        for port in self.ports:
+        for port in self.primary_ports + self.secondary_ports:
             self.ecs_security_group.add_ingress_rule(
                 peer=ec2.Peer.security_group_id(
                     self.load_balancer_security_group.security_group_id
@@ -108,7 +108,7 @@ class IalirtProcessing(Construct):
         # Allow inbound and outbound traffic from a specific port and IP.
         # IPs: LASP IP, BlueNet (tlm relay)
         ip_ranges = ["128.138.131.0/24", "198.118.1.14/32"]
-        for port in self.ports:
+        for port in self.primary_ports + self.secondary_ports:
             for ip_range in ip_ranges:
                 self.load_balancer_security_group.add_ingress_rule(
                     # TODO: allow IP addresses from partners
@@ -189,7 +189,7 @@ class IalirtProcessing(Construct):
         # Network Load Balancers (NLB).
         primary_task_definition = ecs.Ec2TaskDefinition(
             self,
-            f"IalirtTaskDef",
+            f"IalirtTaskDefPrimary",
             network_mode=ecs.NetworkMode.AWS_VPC,
             task_role=task_role,
             execution_role=execution_role,
@@ -197,7 +197,7 @@ class IalirtProcessing(Construct):
 
         secondary_task_definition = ecs.Ec2TaskDefinition(
             self,
-            f"IalirtTaskDef",
+            f"IalirtTaskDefSecondary",
             network_mode=ecs.NetworkMode.AWS_VPC,
             task_role=task_role,
             execution_role=execution_role,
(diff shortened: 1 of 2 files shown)


sds_data_manager/constructs/ialirt_processing_construct.py
```diff
@@ -26,8 +26,8 @@ def __init__(
         scope: Construct,
         construct_id: str,
         vpc: ec2.Vpc,
-        processing_name: str,
```
torimarbois, line 29: Question for understanding: Why are you able/needing to remove the `processing_name` parameter?

sds_data_manager/constructs/ialirt_processing_construct.py
```diff
@@ -184,28 +183,54 @@ def add_compute_resources(self, processing_name):
...
-        task_definition = ecs.Ec2TaskDefinition(
+        primary_task_definition = ecs.Ec2TaskDefinition(
+            self,
+            "IalirtTaskDefPrimary",
+            network_mode=ecs.NetworkMode.AWS_VPC,
+            task_role=task_role,
+            execution_role=execution_role,
+        )
+
+        secondary_task_definition = ecs.Ec2TaskDefinition(
```
torimarbois, line 196: Can you help me understand this primary vs. secondary stuff (including the ports)? I'm guessing it has to do with being 'backups' for each other while the service is updated.

sds_data_manager/constructs/ialirt_processing_construct.py
```diff
@@ -215,43 +240,63 @@ def add_compute_resources(self, processing_name):
...
-            container.add_port_mappings(port_mapping)
+            primary_container.add_port_mappings(port_mapping)
+
+        for port in self.secondary_ports:
+            port_mapping = ecs.PortMapping(
+                container_port=port,
+                host_port=port,
+                protocol=ecs.Protocol.TCP,
+            )
+            secondary_container.add_port_mappings(port_mapping)
```
torimarbois, line 259: Could this be combined into one for loop? Or would that not be worth it for the loss in clarity..?

torimarbois (approved):
This LGTM, but as I am not an expert in AWS stacks, you'll probably still want to get another set of more-experienced eyes on it.

laspsandoval:
Roz reviewed the PR and had the following comments ideas:
Improved infrastructure as code patterns
Look into sticky session to combine ECS sessions

I will create tickets for each of these great ideas. Probably somewhat out of scope for this PR though. But besides that she agreed with the changes made.

sds_data_manager/constructs/ialirt_processing_construct.py
laspsandoval: Before the name was defined in stackbuilder and now the names are defined in ialirt_processing_construct.py That's really the only reason.

sds_data_manager/constructs/ialirt_processing_construct.py
laspsandoval: Yes. We will have a primary and secondary container running at all times. The data from one set of ports will be directed to the primary container and the data from the other set of ports will be directed to the secondary container.

sds_data_manager/constructs/ialirt_processing_construct.py
laspsandoval: Thanks for your comment. For me it is easier to read this way. I think I'll leave it for clarity like you said.

PR #420, opened 2024-12-17
Adding monitoring to send clean logging emails

### LINKED ISSUE #438: Make batch job emails filterable
### Summary

We want to be able to filter by the batch job emails for instrument and level to reduce email spam to instrument teams/menlo. 

To do this, update the monitoring lambda to include messageattributes for instrument and level, and then update the subscriptions to add a filter by message attributes. 

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sns/client/publish.html

### Requirements

The more specific the better.
  - [ ] Add message attributes to SNS failure notifications
  - [ ] Filter messages for key parties (ie Menlo, MAG)

### Related tickets

_No response_

maxinelasp:
# Change Summary
Added a lambda for formatting output when batch jobs fail. 

In the future we should probably make SNS topics for each failure... This is also a place where we can add retry logic. 

Created because the MAG team kept emailing me asking why their jobs were failing. 
## Overview

- Moved the SNS failure topic and eventbridge event into a new construct
- Creates a lambda function to retrieve logs from the batch job and format everything
- This lambda sits between where the failure eventbridge rule and the SNS topic met. Nothing else is updated.

Email I received:
```
Batch Job Failed!

    Job Name: mag-l1a-all-job-160
    Job ID: 28b69f08-6a84-4e14-aef2-dbd739e632d4
    Status Reason: Essential container in task exited

    Logs (Last 10 lines):
      File "/usr/local/lib/python3.12/site-packages/imap_processing/cdf/utils.py", line 138, in write_cdf    xarray_to_cdf(  File "/usr/local/lib/python3.12/site-packages/cdflib/xarray/xarray_to_cdf.py", line 1034, in xarray_to_cdf    depend_0_vars, time_varying_dimensions = _epoch_checker(dataset, dim_vars, terminate_on_warning)                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^  File "/usr/local/lib/python3.12/site-packages/cdflib/xarray/xarray_to_cdf.py", line 456, in _epoch_checker    _warn_or_except(  File "/usr/local/lib/python3.12/site-packages/cdflib/xarray/xarray_to_cdf.py", line 91, in _warn_or_except    raise ISTPError(message)cdflib.xarray.xarray_to_cdf.ISTPError: Variable epoch was determined to be an ISTP 'Epoch' variable, but it is not monotonically increasing.

```

### DIFF
--- a/sds_data_manager/constructs/indexer_lambda_construct.py
+++ b/sds_data_manager/constructs/indexer_lambda_construct.py
@@ -46,9 +46,6 @@ class IndexerLambda(Construct):
             The RDS security group
         data_bucket : obj
             The data bucket
-        sns_topic : aws_sns.Topic
-            SNS Topic for sending notifications so that external
-            resources can subscribe to for alerts.
         layers : list
             List of Lambda layers cdk.cdfnOutput names
         kwargs : dict
--- a/sds_data_manager/constructs/indexer_lambda_construct.py
+++ b/sds_data_manager/constructs/indexer_lambda_construct.py
@@ -23,7 +23,6 @@ class IndexerLambda(Construct):
         vpc_subnets,
         rds_security_group,
         data_bucket,
-        sns_topic,
         layers: list,
         **kwargs,
     ) -> None:
@@ -134,20 +133,6 @@ class IndexerLambda(Construct):
             ),
         )
 
-        # Uses batch job status of failure
-        # to trigger a sns topic
-        batch_job_failure_rule = events.Rule(
-            self,
-            "batchJobFailure",
-            rule_name="batch-job-failure",
-            event_pattern=events.EventPattern(
-                source=["aws.batch"],
-                detail_type=["Batch Job State Change"],
-                detail={"status": ["FAILED"]},
-            ),
-        )
-
         # Add the Lambda function as the target for the rules
         imap_data_arrival_rule.add_target(targets.LambdaFunction(indexer_lambda))
         batch_job_status_rule.add_target(targets.LambdaFunction(indexer_lambda))
-        batch_job_failure_rule.add_target(targets.SnsTopic(sns_topic))
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
@@ -1,0 +1,70 @@
+"""Lambda function to send a formatted SNS notification when a Batch job fails."""
+
+import os
+
+import boto3
+
+sns_client = boto3.client("sns")
+logs_client = boto3.client("logs")
+
+
+def lambda_handler(event, context):
+    """Lambda handler to send an SNS notification when a Batch job fails.
+
+    Lambda will format the message and retrieve logging from the failed job, before
+    sending a message to the notification service (SNS topic defined by the environment
+    variable "SNS_ARN").
+
+    Parameters
+    ----------
+    event : dict
+        The JSON formatted document with the data required for the
+        lambda function to process. Source event is from AWS Batch.
+    context : obj
+        The context object for the lambda function
+    """
+    # Extract relevant details from the event
+    detail = event.get("detail", {})
+
+    job_name = detail.get("jobName", "Unknown")
+    job_id = detail.get("jobId", "Unknown")
+    log_stream_name = (
+        detail.get("attempts", [{}])[0].get("container", {}).get("logStreamName", None)
+    )
+
+    status_reason = detail.get("statusReason", "No reason provided")
+
+    # Fetch logs if logStreamName is available
+    logs = []
+    if log_stream_name:
+        log_group_name = "/aws/batch/job"
+        try:
+            response = logs_client.get_log_events(
+                logGroupName=log_group_name, logStreamName=log_stream_name, limit=10
+            )
+            logs = [event["message"] for event in response.get("events", [])]
+        except Exception as e:
+            logs.append(f"Could not fetch logs: {e!s}")
+
+    # Format email message
+    formatted_message = f"""
+    Batch Job Failed!
+
+    Job Name: {job_name}
+    Job ID: {job_id}
+    Status Reason: {status_reason}
+
+    Logs (Last 10 lines):
+    {''.join(logs) if logs else 'No logs available'}
+    """
+
+    print(f"Formatted Message: {formatted_message}")
+
+    # Send the formatted message to the SNS topic
+    sns_client.publish(
+        TopicArn=os.environ["SNS_TOPIC_ARN"],
+        Subject=f"Batch Job Failure: {job_name}",
+        Message=formatted_message,
+    )
+
+    return {"statusCode": 200, "body": "Notification sent"}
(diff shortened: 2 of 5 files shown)


sds_data_manager/constructs/indexer_lambda_construct.py
```diff
@@ -23,7 +23,6 @@ def __init__(
         vpc_subnets,
         rds_security_group,
         data_bucket,
-        sns_topic,
```
tech3371, line 26: can we remove in doc string as well?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Parameters
+    ----------
+    event : dict
+        The JSON formatted document with the data required for the
+        lambda function to process. Source event is from AWS Batch.
+    context : obj
+        The context object for the lambda function
+    """
+    # Extract relevant details from the event
```
tech3371, line 37: can we add a logger and log event?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
```diff
@@ -0,0 +1,70 @@
...
+    detail = event.get("detail", {})
+
+    job_name = detail.get("jobName", "Unknown")
+    job_id = detail.get("jobId", "Unknown")
```
tech3371: we should always get jobName or jobId from batch job event. So can we remove the unknown options and handle error instead and return something else?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
```diff
@@ -0,0 +1,70 @@
+"""Lambda function to send a formatted SNS notification when a Batch job fails."""
+
+import os
+
+import boto3
+
+sns_client = boto3.client("sns")
+logs_client = boto3.client("logs")
```
tech3371: ca you make these upper case to indicate it's global variable?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
```diff
@@ -0,0 +1,70 @@
...
+    formatted_message = f"""
+    Batch Job Failed!
+
+    Job Name: {job_name}
+    Job ID: {job_id}
+    Status Reason: {status_reason}
+
+    Logs (Last 10 lines):
+    {''.join(logs) if logs else 'No logs available'}
+    """
```
tech3371, line 71: nice!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
```diff
@@ -0,0 +1,70 @@
...
+    print(f"Formatted Message: {formatted_message}")
+
+    # Send the formatted message to the SNS topic
+    sns_client.publish(
+        TopicArn=os.environ["SNS_TOPIC_ARN"],
+        Subject=f"Batch Job Failure: {job_name}",
+        Message=formatted_message,
+    )
+
+    return {"statusCode": 200, "body": "Notification sent"}
```
tech3371, line 82: Good that you are returning values like API lambda because I think it will become an API lambda. We discussed about way to display these logs and job status on IMAP website. I can see this lambda being repurpose for that. Do you think we should move it into API now or later?

tech3371 (changes requested):

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
maxinelasp: Shouldn't we still send an email if the event is "batch job failed" even if we don't know the job name or job id? How would you handle the error instead?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
maxinelasp: I think we should leave it as is now. If we want to set something up, I would prefer to have it subscribe to the SNS topic. I also think the correct way to get status is from the status tracker table, and this should only be used for notifications or handling failures, rather than making requests.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
tech3371: For sure. I will try my best. I will add some clarification as well.

What I meant above is batch job event will be in this format where it **should** always have 'jobName' or 'jobId' or 'attempts' or 'container' and so on keys. I took out below example batch job event json from `indexer.py` lambda code. It's missing some other fields because I only kept information that indexer cares about. But batch event job will always has fixed keys.
```
{
        "detail-type": "Batch Job State Change",
        "source": "aws.batch",
        "detail": {
            "jobArn": (
                "arn:aws:batch:us-west-2:012345678910:"
                "job/26242c7e-3d49-4e41-9387-74fcaf9630bb"
            ),
            "jobName": "swe-l0-job",
            "jobId": "26242c7e-3d49-4e41-9387-74fcaf9630bb",
            "jobQueue": (
                "arn:aws:batch:us-west-2:012345678910:"
                "job-queue/swe-fargate-batch-job-queue"
            ),
            "status": "FAILED",
            "statusReason": "some error message",
            "jobDefinition": (
                "arn:aws:batch:us-west-2:012345678910:"
                "job-definition/fargate-batch-job-definitionswe:1"
            ),
            "container": {
                "image": (
                    "123456789012.dkr.ecr.us-west-2.amazonaws.com/" "swapi-repo:latest"
                ),
                "command": [
                    "--instrument", "swapi",
                    "--data-level", "l1",
                    "--descriptor", "sci",
                    "--start-date", "20230724",
                    "--version", "v001",
                    "--dependency", \"""[
                        {
                            'instrument': 'swapi',
                            'level': 'l0',
                            'start_date': 20230724,
                            'version': 'v001'
                        }
                    ]\""",
                    "--upload-to-sdc",
                ],
                "logStreamName": (
                    "fargate-batch-job-definitionswe/default/"
                    "8a2b784c7bd342f69ea5dac3adaed26f"
                ),
            },
        }
    }
```

Given above reason, we should never receive batch job event json that didn't have `jobName` or `jobId` and so on. If we do, then it's a bug and we need to correct our event source or lambda trigger source and update accordingly. I hope that helps add some context I missed to add last time.

tech3371: To your question about how we should handle error, I would check event source. If it's not 'aws.batch', then raise error and shouldn't send message to the SNS. That's my thought.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
tech3371: That's ok with me. I am just thinking about future work. We can revisit then and figure out.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/monitoring.py
maxinelasp: Ok, that makes sense! I'll add that case.

tech3371 (approved):
Thank you for making those updates! LGTM

Issue #377, opened 2024-10-31
I-ALiRT: Create new lambda within ialirt_api_manager_construct


laspsandoval:
### Description:
 Add new lambda function, route and role policy to `ialirt_api_manager_construct.py` within `sds-data-manager`. This will be for a new endpoint, `catalog`, which will be used for sharing out the pointing schedules to ground stations.

torimarbois:
- reformat the code in imap-processing to be able to use it in the lambda
- import the reformatted code, then write the code for the lambda

torimarbois:
Before this happens, Laura needs to sync up with Jenny Knuth about what she'll need for the website.

PR #444, opened 2025-01-21
I-ALiRT: Initial setup of catalog API lambda

torimarbois:
# Change Summary

## Overview
This is the first MR in the series of changes needed to set up the API endpoint that will be used to distribute our ephemeris pointing schedules. The ticket is [here](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/377). 
These changes are the skeleton of the new `catalog` endpoint, which will eventually return a list of file names and links to them in s3. 

## New Files

- IAlirtCode/ialirt_catalog_api.py
   - This will house the definition of the lambda function that will retrieve the files from s3 and return them from the API call. Right now, this file is fairly empty.

## Updated Files

   - Adds a lambda function, role policy and route to the I-ALiRT API Gateway structure for the new `catalog` endpoint.

## Testing

I've deployed these changes to the IalirtStack in our `dev` account, and queried the new endpoint using Postman. The message that was returned was "Missing Authentication Token", which makes sense considering there isn't any s3 client code in the lambda definition yet.

### DIFF
--- a/sds_data_manager/constructs/ialirt_api_manager_construct.py
+++ b/sds_data_manager/constructs/ialirt_api_manager_construct.py
@@ -110,3 +110,29 @@ class IalirtApiManager(Construct):
             lambda_function=download_api,
             use_path_params=True,
         )
+
+        # catalog API lambda
+        catalog_api = lambda_.Function(
+            self,
+            id="IAlirtCatalogAPILambda",
+            function_name="ialirt-catalog-api-handler",
+            code=code,
+            handler="IAlirtCode.ialirt_catalog_api.lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_12,
+            timeout=cdk.Duration.minutes(1),
+            environment={
+                "S3_BUCKET": data_bucket.bucket_name,
+                "REGION": env.region,
+            },
+            layers=layers,
+            architecture=lambda_.Architecture.ARM_64,
+        )
+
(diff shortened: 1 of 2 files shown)


laspsandoval:
For testing, it should return an empty list from Postman and not the "Missing Authentication Token" I'm pretty sure.

sds_data_manager/constructs/ialirt_api_manager_construct.py
```diff
@@ -110,3 +110,29 @@ def __init__(
             lambda_function=download_api,
             use_path_params=True,
         )
+
+        # catalog API lambda
+        catalog_api = lambda_.Function(
```
laspsandoval, line 115: This looks ok I think.

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
```diff
@@ -0,0 +1,31 @@
+"""Define lambda to support the catalog API."""
+
+import logging
+
+# Logger setup
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+def lambda_handler(event, context):
```
laspsandoval, line 10: I think this is ok for this review, but you will need to add a test for this lambda.

laspsandoval (approved):
Looks ok to start.

torimarbois:
pre-commit.ci autofix

PR #449, opened 2025-02-03
I-ALiRT: Remove NLB

### LINKED ISSUE #450: I-ALiRT: Remove NLB #1
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
The NLB can only forward to 5 targets for an ECS service. This is the first of two Pull Requests to remove the NLB. This ticket allows a direct connection from the incoming datastreams to the ECS. However it does not account for EC2 instances changing their IP addresses. That will be addressed in the second PR.

## Updated Files
   - Removed Network Load Balancer Security Group
   - ECS Security Group now has the properties of the Network Load Balancer Security group.
   - Changed the Network Mode to "HOST". Before it was using "AWS_VPC" because that was the only allowable Network Mode for Network Load Balancers. 
   - In AWS_VPC network mode, ECS tasks are assigned their own ENI with private IPs, and the EC2 instances are not automatically assigned public IPs for these tasks. This means that tasks cannot be reached via a public IP and require external routing (e.g., an NLB) to expose them—which is not what we want.
   - In "HOST" you can access the container using the EC2 public IP that is automatically assigned. I chose this mode because I want the ECS tasks to automatically inherit the EC2 instance’s Elastic IP so that they always use a publicly accessible IP address.
   - In the next PR I will assign static EIPs to the EC2 instances and I will need the ECS tasks to be able to use those.
   - ECS service runs directly on the EC2 instances, so it automatically uses the security group and subnet of the Auto Scaling Group (ASG). Therefore, I removed those from the ECS service parameters and added them to the ASG.
   - Set the EC2 instance desired, min, and max capacity to correct settings for ASG.
   
## Testing
**Test 1:**
Verify that I can connect using the IP address for the EC2 instance while CONNECTED to the VPN.
`http://<EC2 Public IP>:7526`
Result Test 1:
Hello from Port 7526!
**Test 2:**
Verify that I cannot connect using the IP address for the EC2 instance while DISCONNECTED from the LASP VPN.
Result Test 2:
Timeout

### DIFF
--- a/sds_data_manager/constructs/ialirt_processing_construct.py
+++ b/sds_data_manager/constructs/ialirt_processing_construct.py
@@ -1,7 +1,6 @@
-"""Configure the i-alirt processing stack.
-"""
+"""Configure the i-alirt processing stack."""
 
-from aws_cdk import CfnOutput, Duration, RemovalPolicy
+from aws_cdk import RemovalPolicy
 from aws_cdk import aws_autoscaling as autoscaling
 from aws_cdk import aws_ec2 as ec2
 from aws_cdk import aws_ecs as ecs
@@ -77,7 +76,7 @@ class IalirtProcessing(Construct):
                 self.ecs_security_group.add_ingress_rule(
                     # TODO: allow IP addresses from partners
                     peer=ec2.Peer.ipv4(ip_range),
-                    connection=ec2.Port.all_tcp(),
+                    connection=ec2.Port.tcp(port),
                     description=f"Allow inbound traffic on TCP port {port}",
                 )
 
@@ -157,7 +156,7 @@ class IalirtProcessing(Construct):
 
         # Adds a container to the ECS task definition
         # Logging is configured to use AWS CloudWatch Logs.
-        container = task_definition.add_container(
+        task_definition.add_container(
             "IalirtContainer",
             image=ecs.ContainerImage.from_registry(
                 "lasp-registry.colorado.edu/ialirt/ialirt:test",
@@ -202,7 +201,7 @@ class IalirtProcessing(Construct):
             vpc=self.vpc,
             desired_capacity=1,
             min_capacity=1,
-            max_capacity=2, # Allow one extra instance during updates
+            max_capacity=2,  # Allow one extra instance during updates
             vpc_subnets=ec2.SubnetSelection(
                 subnet_type=ec2.SubnetType.PUBLIC,
             ),
--- a/sds_data_manager/constructs/ialirt_processing_construct.py
+++ b/sds_data_manager/constructs/ialirt_processing_construct.py
@@ -1,17 +1,10 @@
 """Configure the i-alirt processing stack.
-
-This is the module containing the general stack to be built for computation of
-I-ALiRT algorithms. It was built using best practices as shown here:
-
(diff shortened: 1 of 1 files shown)


sds_data_manager/constructs/ialirt_processing_construct.py
```diff
@@ -179,19 +146,17 @@ def add_compute_resources(self):
         )
 
         # Specifies the networking mode as AWS_VPC.
-        # ECS tasks in AWS_VPC mode can be registered with
-        # Network Load Balancers (NLB).
         task_definition = ecs.Ec2TaskDefinition(
             self,
             "IalirtTaskDef",
-            network_mode=ecs.NetworkMode.AWS_VPC,
+            network_mode=ecs.NetworkMode.HOST,
```
greglucas, line 156: Can you add a quick comment about this change and why it needs to be HOST vs AWS_VPC?

sds_data_manager/constructs/ialirt_processing_construct.py
```diff
@@ -249,11 +198,14 @@ def add_autoscaling(self):
             ),
             machine_image=ecs.EcsOptimizedImage.amazon_linux2(),
             vpc=self.vpc,
-            desired_capacity=2,
+            desired_capacity=1,
+            min_capacity=1,
+            max_capacity=2,  # Allow one extra instance during updates
```
greglucas, line 207: I think Steve mentioned something about 2 EC2 instances would be OK for them. Is that desired by ops-software or they only want one and if it goes down then 🤷

sds_data_manager/constructs/ialirt_processing_construct.py
```diff
@@ -277,87 +229,3 @@ def add_autoscaling(self):
...
-                    self.ecs_service.load_balancer_target(
-                        container_name="IalirtContainer",
-                        container_port=port,
-                    )
-                ],
-                # Configures health checks for the target group
-                # to ensure traffic is routed only to healthy ECS tasks.
-                # Port 7568 is a dummy port used by IOIS to check the
-                # health of the container.
-                health_check=elbv2.HealthCheck(
```
greglucas, line 346: Do we need a health check on the tasks/EC2 itself now rather than the load balancer?

greglucas (approved):
Looks good 👍 Just a few comments/questions that might help to add in for future selves on why you chose to do certain things.

sds_data_manager/constructs/ialirt_processing_construct.py
laspsandoval: Good point. ECS has a health check already. If the task crashes then ECS will automatically restart a new one, but the NLB would "ping" a container port to verify that there was a connection. I think we are ok without that functionality though.

torimarbois (approved):

sds_data_manager/constructs/ialirt_processing_construct.py
laspsandoval: Steve only wants a single task to be running so we only need a single instance. I had the desired_capacity=2 before because it meant that an instance would be ready to go in case there was an unhealthy task. But I think that most likely we will really only need the second instance for updates (one instance is running while the other starts up). So I think this should be ok.

PR #456, opened 2025-02-06
I-ALiRT - Remove NLB 2

### LINKED ISSUE #451: I-ALiRT - Remove NLB 2
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Changes to replace NLB. Everything worked before, but only as long as the EC2 instance wasn't replaced by AutoScaling. This PR addresses that issue. Now the EC2 instance will always have the same static EIP.

## New Files
- ialirt_eip.py
   - Lambda that is kicked off as a result of the EventBridge Rules. This assigns a static EIP to the EC2 instance

## Updated Files
- stackbuilder.py
   - pass in eip secret name
   - Created autoscaling lifecycle hook that triggers when an EC2 instance is launching and allows up to 5 minutes for the Elastic IP (EIP) assignment.
   - Created an EventBridge rule that triggers the lambda either by the lifecycle hook or upon deployment (when EC2 instance transitions into running state)
- ialirt-setup.rst
   - Description of creating and saving static eip
   - New description of switching out containers. 

## Testing
test_ialirt_eip.py - Tests ialirt_eip.py

I also tested different situations:
1. Tested how long it took to switch out a container (my test image and OpsSW test image) - 1 minute
2. Looked to see if it was possible to have two running tasks at a time (it's not with the HOST Network). This might make us revert to the primary/secondary container scenerio.
3. Tested that EC2 instance has assigned public IP upon deployment
4. Tested that EC2 instance has assigned public IP upon deleting the EC2 (and autoscaling brining it back up)

### DIFF
--- a/docs/source/cdk/ialirt-setup.rst
+++ b/docs/source/cdk/ialirt-setup.rst
@@ -8,6 +8,10 @@ Ensure you have a secret in AWS Secrets Manager with a username and password for
 
     aws secretsmanager create-secret --name nexus-credentials --description "Credentials for Nexus Docker registry" --secret-string '{"username":"your-username", "password":"your-password"}'
 
+Also ensure you have created a static elastic ip that will be assigned to the EC2::
+
+    aws ec2 allocate-address --domain vpc
+    aws secretsmanager create-secret --name allocation-credentials --description "Credentials for Static EIP" --secret-string '{"eip_allocation_id":"<eip_allocation_id>"}'
 Image Versioning
 ~~~~~~~~~
 We will rely on semantic versioning for the images MAJOR.MINOR (e.g., 1.0).
@@ -44,6 +48,7 @@ We will have a versioned image and latest image in the Nexus repo. The versioned
 
     docker push docker-registry.pdmz.lasp.colorado.edu/ialirt/ialirt:X.Y.Z
     docker push docker-registry.pdmz.lasp.colorado.edu/ialirt/ialirt:latest
+
 #. Images may be viewed on the Nexus website: https://artifacts.pdmz.lasp.colorado.edu
 #. To verify that the latest image and the most recent version image are the same, run the following and compare the image IDs::
 
--- a/docs/source/cdk/ialirt-setup.rst
+++ b/docs/source/cdk/ialirt-setup.rst
@@ -58,5 +58,6 @@ ECS Recognition of a New Image
 ~~~~~~~~~~~~~
 To have ECS recognize a new image the cdk must be redeployed::
 
-    aws ecs update-service --cluster <cluster name> --service <service name> --force-new-deployment --deployment-configuration maximumPercent=200,minimumHealthyPercent=0
+    aws ecs update-service --cluster <cluster name> --service <service name> --force-new-deployment --deployment-configuration maximumPercent=100,minimumHealthyPercent=0
 
+The reason that we can only have a single task running is that the same ports would be in use by the old task, and as a result, the new task will fail to start because it wouldn't be able bind to the required ports.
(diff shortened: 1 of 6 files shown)


docs/source/cdk/ialirt-setup.rst
```diff
@@ -8,6 +8,10 @@ Ensure you have a secret in AWS Secrets Manager with a username and password for
 
     aws secretsmanager create-secret --name nexus-credentials --description "Credentials for Nexus Docker registry" --secret-string '{"username":"your-username", "password":"your-password"}'
 
+Also ensure you have created a static elastic ip that will be assigned to the EC2::
+
+    aws ec2 allocate-address --domain vpc
+    aws secretsmanager create-secret --name allocation-credentials --description "Credentials for Static EIP" --secret-string '{"eip_allocation_id":"<eip_allocation_id>"}'
```
greglucas: Are we able to do this within the CDK, or within your lambda automatically?

It looks like we can use boto3 calls to do this: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/allocate_address.html
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/describe_addresses.html

I'm wondering if you can add some logic into the lambda that says "do we have an allocated IP address with X name, and if not allocate one now" Then we shouldn't need any secrets manager things at all I don't think? You could just be querying the addresses we already have associated/allocated.

sds_data_manager/constructs/ialirt_processing_construct.py
```diff
@@ -189,8 +204,90 @@ def add_compute_resources(self):
...
+                iam.ManagedPolicy.from_aws_managed_policy_name("AutoScalingFullAccess"),
+            ],
+        )
+
+        eip_lambda = lambda_alpha_.PythonFunction(
+            self,
+            id="IalirtAssignEipLambda",
+            function_name="ialirt-eip",
+            entry=str(
+                pathlib.Path(__file__).parent.joinpath("..", "lambda_code").resolve()
```
greglucas: Be consisten in choice of parent traversal. If you are already using parent, might as well use it twice, or put "../.." in the joinpath.
```suggestion
                pathlib.Path(__file__).parent.parent.joinpath("lambda_code").resolve()
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_eip.py
```diff
@@ -0,0 +1,108 @@
...
+
+    """
+    logger.info("Received event: %s", json.dumps(event, indent=2))
+
+    eip_allocation_id = get_eip_allocation_id()
+
+    if eip_allocation_id:
+        logger.info("Available Elastic IPs: %s", eip_allocation_id)
+    else:
+        logger.warning("No available Elastic IPs found.")
```
greglucas: Turn this into a function that creates one if it doesn't exist, and uses the specific one you want if it does. Related to my earlier comment.

docs/source/cdk/ialirt-setup.rst
laspsandoval: Nice idea!

sds_data_manager/lambda_code/IAlirtCode/ialirt_eip.py
```diff
@@ -0,0 +1,115 @@
+"""Test the I-Alirt EIP lambda function."""
```
torimarbois, line 1: Little nitpick: this doc string uses the word 'test', but this file isn't really testing the EIP code, right? This file is running that code.

torimarbois (approved):
LGTM, one comment about a misleading doc string!

PR #460, opened 2025-02-11
I-ALIRT - Dynamodb api

### LINKED ISSUE #428: I-ALiRT: Create an API for Final DynamoDB
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
This creates an API for the Algorithm (FInal) DynamoDB.

## New Files
- ialirt_db_query_api.py
   - Creates a way to query the DynamoDB
   - To see the types of queries please refer to test_ialirt_db_query_api.py

## Updated Files
   - Adds the handler and route
- stackbuilder.py
   - Rearranging and passing in dynamodb table information
   - Rename table

## Testing
- test_ialirt_db_query_api.py
   - Tests ialirt_db_query_api.py

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api.py
@@ -1,0 +1,34 @@
+import os
+import json
+import boto3
+from boto3.dynamodb.conditions import Key, Attr
+
+
+def lambda_handler(event, context):
+    table_name = os.environ["TABLE_NAME"]
+    region = os.environ.get("AWS_DEFAULT_REGION", "us-west-2")
+    dynamodb = boto3.resource("dynamodb", region_name=region)
+    table = dynamodb.Table(table_name)
+
+    params = event["queryStringParameters"]
+    apid = int(params["apid"])
+    met_start = int(params["met_start"])
+    met_end = int(params["met_end"])
+
+    key_expr = Key("apid").eq(apid) & Key("met").between(met_start, met_end)
+    query_kwargs = {"KeyConditionExpression": key_expr}
+
+    if "product_name" in params:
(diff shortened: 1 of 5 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api.py
```diff
@@ -0,0 +1,155 @@
...
+            else params[f"{time_key}_end"]
+        )
+
+        key_expr &= Key(time_key).between(start_value, end_value)
+
+        if time_key == "insert_time":
+            query_kwargs["IndexName"] = "insert_time"
+
+    elif "met_start" in params or "insert_time_start" in params:
+        time_key = "met" if "met_start" in params else "insert_time"
```
maxinelasp, line 109: There's an unhandled case here where I user includes, for example, "met_start" and "insert_time_end", where the code will just query by start and ignore the end time. Not sure if you want to log something in that case, but I wanted to call it out

maxinelasp (approved):
One comment you may want to address, otherwise looks good!

sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api.py
laspsandoval: Thanks for the comment! I think that should be taken care of here. I added a test and verified that it took care of it.

    if any(param.startswith("met") for param in params) and any(
        param.startswith("insert_time") for param in params
    ):
        return {
            "statusCode": 400,
            "body": json.dumps(
                {"message": "Cannot query both MET and insert_time in the same request"}
            ),
            "headers": {
                "Content-Type": "application/json",
                "Access-Control-Allow-Origin": "*",
            },
        }

Issue #436, opened 2025-01-13
I-ALiRT: Test new lambda endpoint


torimarbois:
Follows https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/435.

Add a test for the new API endpoint in `tests/lambda_endpoints`. 

Also, decide if code for this endpoint should be added to the `ialirt-data-access` repo.

laspsandoval:
Done

Issue #435, opened 2025-01-13
I-ALiRT: Add 'catalog' API endpoint


torimarbois:
Follows https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/377. 

Create the new catalog API endpoint, deploy and test with Postman. This will likely require adding a file to the `IAlirtCode` directory.

PR #461, opened 2025-02-11
I-ALiRT: Catalog endpoint

torimarbois:
# Change Summary

## Overview
These changes correspond to the I-ALiRT project tickets #435 and #436. I've filled out the lambda function that supports the catalog endpoint, which is designed to return JSON (as requested by the web team) containing the pointing schedule files for a ground station over a specified coverage window.

## Updated Files

- pyproject.toml
   - I had to install numpy, which required an updated python version.
   - To be honest, I don't feel confident about these updates needing to be repo-wide and would appreciate feedback specifically around this and the poetry.lock changes.
- poetry.lock
   - See the above point!
- ialirt_catalog_api.py
   - Flushes out the details of the lambda handler function. 
   - Queries the s3 bucket using `list_catalog_v2()` and then generates presigned urls using an s3_client function.
   - Wraps final results in a json/dictionary design, as requested by the web team, who will be using this endpoint to display pointing schedules on the I-ALiRT website
- test_ialirt_catalog_api.py
   - Tests the expanded lambda handler function.

## Testing

All new code changes are covered by new unit tests in `test_ialirt_catalog_api.py`.

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
@@ -1,31 +1,127 @@
-"""Define lambda to support the catalog API."""
+"""Define the lambda function that supports the catalog API."""
 
+import json
 import logging
+import os
+from datetime import datetime, timedelta
+
+import boto3
+import botocore
 
 # Logger setup
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 
-def lambda_handler(event, context):
+def lambda_handler(event, context) -> dict:
     """Entry point to the catalog API lambda.
 
-    Blank function for initial setup.
+    Takes in a date range and ground station, and queries the s3 bucket that holds
+    the pointing schedules for ground stations part of the I-ALiRT network.
+
+    If files exist for the requested date range, the response body returns all file
+    versions for every date, including the file name and its last date modified.
 
     Parameters
     ----------
     event : dict
         The JSON formatted document with the data required for the
-        lambda function to process
+        lambda function to process, i.e., date range and ground station
+        Ex:
+            event = {
+                "start_date": "2025-02-06",  # inclusive
+                "end_date": "2025-02-07",  # exclusive
+                "station": "station1",
+            }
     context : LambdaContext
         This object provides methods and properties that provide
         information about the invocation, function,
         and runtime environment.
+
+    Returns
+    -------
+    dict
+        A dictionary containing headers, status code, and body, designed to be returned
+        as an API response.
+        Ex:
+        {
+            'statusCode': 200,
+            'headers': {
+                'Content-Type': 'application/json'
+            },
+            'body': '{
+                "2025-02-06":
+                    {"file_names":
+                        ["20250206_station1_01.txt", "20250206_station1_02.txt"],
+                     "dates_modified":
+                        ["2025-02-12 18:37", "2025-02-12 18:37"]}
+            }'
+        }
     """
-    logger.info(f"Event: {event}")
-    logger.info(f"Context: {context}")
+    logger.info("Received event: " + json.dumps(event, indent=2))
+
+    day = datetime.strptime(event["start_date"], "%Y-%m-%d")
+    end_date = datetime.strptime(event["end_date"], "%Y-%m-%d")
+
+    if not day < end_date:
+        response_body = (
+            "The start date of the requested time frame is not before the "
+            "end date. Please supply a valid time range."
+        )
+        return {
+            "statusCode": 400,
+            "headers": {"Content-Type": "application/json"},
+            "body": response_body,
+        }
+
+    if end_date - day > timedelta(days=30):
+        response_body = (
+            "The end date of the requested time frame must not be more "
+            "than 30 days after the start date. Please supply a valid time"
+            " range."
+        )
+        return {
+            "statusCode": 400,
+            "headers": {"Content-Type": "application/json"},
+            "body": response_body,
+        }
+
+    bucket = os.getenv("S3_BUCKET")
+    region = os.getenv("REGION")
+
+    s3_client = boto3.client(
+        "s3",
+        region_name=region,
+        config=botocore.client.Config(signature_version="s3v4"),
+    )
+
+    response_body = {}
+    while day < end_date:
+        day_path = f"pointing_schedules/{event['station']}/{day.strftime('%Y%m%d')}/"
+        files = s3_client.list_objects_v2(Bucket=bucket, Prefix=day_path)
+        if "Contents" not in files.keys():
+            response_body = (
+                f"There are not files associated with the provided date {day}. "
+                f"Please supply a valid time range that is covered by existing files."
+            )
+            return {
+                "statusCode": 404,
+                "headers": {"Content-Type": "application/json"},
+                "body": response_body,
+            }
+        file_names = []
+        dates_modified = []
+        for file in files["Contents"]:
+            file_names.append(file["Key"].rsplit("/", 1)[1])
+            dates_modified.append(file["LastModified"].strftime("%Y-%m-%d %H:%M"))
+        response_body[f"{day}"] = {
+            "file_names": file_names,
+            "dates_modified": dates_modified,
+        }
+        day += timedelta(days=1)
 
     return {
         "statusCode": 200,
         "headers": {"Content-Type": "application/json"},
+        "body": json.dumps(response_body),
     }
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
@@ -64,26 +64,20 @@ def lambda_handler(event, context) -> dict:
     end_date = datetime.strptime(event["end_date"], "%Y-%m-%d")
 
     if not day < end_date:
-        response_body = (
-            "The start date of the requested time frame is not before the "
-            "end date. Please supply a valid time range."
-        )
         return {
             "statusCode": 400,
             "headers": {"Content-Type": "application/json"},
-            "body": response_body,
+            "body": "The start date of the requested time frame is not before the "
+            "end date. Please supply a valid time range.",
         }
 
     if end_date - day > timedelta(days=30):
-        response_body = (
-            "The end date of the requested time frame must not be more "
-            "than 30 days after the start date. Please supply a valid time"
-            " range."
-        )
         return {
             "statusCode": 400,
             "headers": {"Content-Type": "application/json"},
-            "body": response_body,
+            "body": "The end date of the requested time frame must not be more "
+            "than 30 days after the start date. Please supply a valid time"
+            " range.",
         }
 
     bucket = os.getenv("S3_BUCKET")
@@ -100,14 +94,12 @@ def lambda_handler(event, context) -> dict:
         day_path = f"pointing_schedules/{event['station']}/{day.strftime('%Y%m%d')}/"
         files = s3_client.list_objects_v2(Bucket=bucket, Prefix=day_path)
         if "Contents" not in files.keys():
-            response_body = (
-                f"There are not files associated with the provided date {day}. "
-                f"Please supply a valid time range that is covered by existing files."
-            )
             return {
                 "statusCode": 404,
                 "headers": {"Content-Type": "application/json"},
-                "body": response_body,
+                "body": "There are not files associated with the provided date {day}. "
+                "Please supply a valid time range that is covered by existing "
+                "files.",
             }
         file_names = []
         dates_modified = []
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
```diff
@@ -7,25 +13,83 @@
...
+    s3_client = boto3.client(
+        "s3",
+        region_name=region,
+        config=botocore.client.Config(signature_version="s3v4"),
+    )
+
+    response_body = {}
+    for day in np.arange(event['start_date'], event['end_date'], dtype='datetime64[D]'):
+        day_path = (f"pointing_schedules/{event['station']}/" +
+                    day.item().strftime('%Y%m%d') + "/")
```
greglucas: I think you can avoid numpy here by using the datetime module. Something like:
```
day = datetime.strptime(event["stard_date"], "%Y%m%d")
end_date = datetime.strptime(event["end_date"], "%Y%m%d")
# Maybe put some logic in to limit the number of days allowed?
while day < end_date:
    day_path = ... day.strfime()
    day += datetime.timedelta(days=1)
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
```diff
@@ -7,25 +13,83 @@
...
+        files = s3_client.list_objects_v2(Bucket=bucket, Prefix=day_path)
```
greglucas, line 95: An alternative idea to this loop over listing objects per day would be to get all of your objects in one `list_objects_v2()` call and then filter the list of objects returned after that.

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
```diff
@@ -7,25 +13,83 @@
...
+                "headers": {"Content-Type": "application/json"},
+                "body": json.dumps(response_body)
+            }
+        file_names = []
+        file_urls = []
+        dates_modified = []
+        for file in files['Contents']:
+            pre_signed_url = s3_client.generate_presigned_url(
+                "get_object",
+                Params={"Bucket": bucket, "Key": file['Key']}, ExpiresIn=604800
```
greglucas: This is a long expires time, suggest changing it to like 10-15 minutes or so. A user can re-request the data if needed.

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
```diff
@@ -7,25 +13,83 @@
...
+            )
+            file_names.append(file['Key'].rsplit('/', 1)[1])
+            file_urls.append(pre_signed_url)
+            dates_modified.append(file['LastModified'].strftime('%Y-%m-%d %H:%M'))
```
greglucas: Out of curiosity, do they need the modified times returned in the API?

Is there a way to repurpose our current API handlers so that a user can request a specific filename and then be redirected to the file_url? So this endpoint would theoretically just return a list of filenames that the user could then go and retrieve later. Our API could send a redirect with a `Location` header in it for the presigned url, but wouldn't require the end-users to know about the details in the background.

tests/lambda_endpoints/test_ialirt_catalog_api.py
```diff
@@ -0,0 +1,53 @@
...
+        Key=test_file2,
+        Body=b"Hello world 2",
+    )
+    event = {
+        "start_date": "2025-02-06",
+        "end_date": "2025-02-07",
+        "station": "station1",
+    }
+    response = ialirt_catalog_api.lambda_handler(event=event, context=None)
+    assert response["statusCode"] == 200
```
laspsandoval, line 71: Could you give an example of what the response would look like? I was thinking it would be something like this?

[
  {
    "file_name": "20250206_station1.txt",
    "url": <url>
  },
  {
    "file_name": "20250207_station1.txt",
    "url": <url>
  }
]

pyproject.toml
```diff
@@ -28,12 +28,13 @@ classifiers = [
...
 
 [tool.poetry.dependencies]
-python = ">=3.9,<4"
+python = ">=3.10,<4"
 
 # Optional dependencies - install with `poetry install -E docs`
 sphinx = {version="^7.1.0", optional=true}
 myst-parser = {version="^2.0.0", optional=true}
 pydata-sphinx-theme = {version="^0.13.3", optional=true}
+numpy = "^2.2.2"
```
laspsandoval: Per Greg's comment remove numpy

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
```diff
@@ -7,25 +13,83 @@
...
+    -------
+    dict
+        A dictionary containing headers, status code, and body, designed to be returned
+        as an API response.
     """
-    logger.info(f"Event: {event}")
-    logger.info(f"Context: {context}")
+    logger.info("Received event: " + json.dumps(event, indent=2))
+
+    if not event['start_date'] < event['end_date']:
```
laspsandoval: Could we use only the start_date if we wanted? So it would retrieve only the most recent files from a given date? If so, add that capability, if not add error handling.

laspsandoval (commented):
Nice code! Thanks for doing this. I had a few comments in addition to what Greg said.

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
torimarbois: I paired with Jenny, who is going to be the main user of this endpoint as she builds it into the I-ALiRT website, and got some clarity around this -

Yes, we want to return the modified times when we return the file names, as they'll be displayed together.

And yes to your second point - Jenny and I agree that a better way to approach this is to break the current logic into two steps - step one is use the catalog endpoint to get a list of file names & dates modified, and step two is use the existing download endpoint to get a presigned url for a specific file.

We'll need to supply Jenny with an example url to query the catalog endpoint, and an example url to query the download endpoint within the I-ALiRT API Manager.

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
torimarbois: To do that, I would have to pass in a list of prefixes (dates), which would require doing the same iterating that the loop is already doing, so I'd prefer to keep the call to `list_objects_v2` within the loop logic and call it for every day. And every day will have multiple versions of the scheduling file that should be returned.

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
torimarbois: I'm removing the URL-generation logic of this endpoint, so I won't need to address this comment with a change. 👍

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
torimarbois: I'm not quite sure what you meant here, but I added error handling around the length of the date range a call can request (no more than 30 days). Let me know if that doesn't cover what you were thinking.

pyproject.toml
```diff
@@ -28,7 +28,7 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.9,<4"
+python = ">=3.10,<4"
```
maxinelasp: If you need numpy, then numpy should be added somewhere here too. I generally like enforcing higher Python versions on everyone, but 3.9 is still supported. I think you should ask in #imap-sdc-dev since this affects all the users of this repo. I'm inclined to say it's fine since it's pretty much limited to the SDC and I don't think anyone there is using 3.9 or would refuse to upgrade, but it's a little iffy.

maxinelasp (commented):
Overall looks good to me, I don't want to block merging, but I am a little concerned about the Python version bump and I just added a comment about that. But, I don't see where you're using numpy, so I'm not sure why that version change is required?

pyproject.toml
torimarbois: @maxinelasp good catch. This was changed because when I first created the PR, I was adding numpy as a dependency, but I've removed it so I can bump the Python back down.

torimarbois:
pre-commit.ci autofix

torimarbois:
pre-commit.ci autofix

maxinelasp (approved):
LGTM, although maybe you want an approval from Laura too? But overall looks good!

lambda_layer/python/requirements.txt
greglucas, line 1: Are you able to revert this and the poetry.lock file updates? I worry that we'll rebuild our layers unnecessarily here and we were recently having issues with that portion, so best to avoid if we can.

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
greglucas: 👍 to keeping what you have for now. But to elaborate on my suggestion, I think you could do a `Prefix="pointing_schedules/{event['station']}/"` (avoid the date) to get all items from that one station in one query (or a few if too many items), then filter that locally. It is often preferable to limit the number of queries you're making and get more items back in a single batch. But I think what you've got here looks good too since you aren't slamming too many queries.

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
```diff
@@ -1,31 +1,129 @@
...
+        "s3",
+        region_name=region,
+        config=botocore.client.Config(signature_version="s3v4"),
+    )
+
+    response_body = {}
+    while day < end_date:
+        day_path = (
+            f"pointing_schedules/{event['station']}/" + day.strftime("%Y%m%d") + "/"
+        )
```
greglucas:
```suggestion
        day_path = f"pointing_schedules/{event['station']}/{day.strftime('%Y%m%d')}/"
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
```diff
@@ -1,31 +1,129 @@
...
+                "statusCode": 404,
+                "headers": {"Content-Type": "application/json"},
+                "body": response_body,
+            }
+        file_names = []
+        dates_modified = []
+        for file in files["Contents"]:
+            file_names.append(file["Key"].rsplit("/", 1)[1])
+            dates_modified.append(file["LastModified"].strftime("%Y-%m-%d %H:%M"))
+        response_body[f"{day}"] = {
```
greglucas, line 109: What is `response_body` here? Above it is all strings, but here you are indexing into it like a dictionary lookup and then setting another dictionary.

```suggestion
        response_body = {
```

Somehow this is passing the tests still though, so I must be missing something...

tests/lambda_endpoints/test_ialirt_catalog_api.py
greglucas: I agree with this. I think you could expand the assert right below this to be exactly what you want in the response, or at least assert on those file names both or only one being provided based on the date ranges queries for.

```
assert test_file1 in response["body"]
assert test_file2 in response["body"]
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
greglucas, line 1: Nice tests, thank you for testing a bunch of the negative conditions and error statements!

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
torimarbois: I see what you're saying. At some point, there will be hundreds of files in a station folder (multiple versions for each day of a pointing schedule), so the approach you are suggesting might get out of hand fairly quickly in terms of the size of the response to the `list_objects` query. However, we could mitigate that by archiving or deleting objects in the bucket 30 days after the date a file corresponds to has passed.
I think it's good to stick with the currently committed approach now, but if at some point we have performance issues, we can circle back to this and discuss the best way to improve.

lambda_layer/python/requirements.txt
torimarbois: Sure, I think that would make sense. Although I'm unsure of the best way to revert poetry lock files (I suppose I could just throw out the changes I've committed for the related files), do you have suggestions?

greglucas: I'd remove the files from the commit: https://stackoverflow.com/questions/12481639/remove-file-from-latest-commit

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
torimarbois: The body of the response is a dictionary, where key = day and value = another dictionary containing file names and dates modified. There's an example of an expected response within the doc string of this function.

laspsandoval (approved):
Looks good. I'm looking forward to you incorporating it into ialirt-data-access.

greglucas (approved):

maxinelasp (approved):

PR #463, opened 2025-02-24
Upload lambda changes for Ancillary Support

daralynnrhode:
# Changes for upload to support Ancillary files

## Overview

changes needed for the upload lambda to support ancillary files. Changes will need to be made in `imap-data-access` before this can be merged in and functional. 

Testing will fail until data-access is updated. It is pulling form `imap_data-access` and `AncillaryFilePath` needs to be included in that, but it currently is not. 

## New Files

- 

## Deleted Files

- 

## Updated Files

- upload lambda


daralynnrhode:
The underlying error is ` AttributeError: module 'imap_data_access' has no attribute 'AncillaryFilePath'`. I thought I knew where I needed to make changes, but turns out those changes are already merged into the repo :\ Does anyone have any advice?

sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
```diff
@@ -122,8 +122,15 @@ def lambda_handler(event, context):
...
         file_obj = file_obj or imap_data_access.ScienceFilePath(filename)
-    except imap_data_access.ScienceFilePath.InvalidScienceFileError as e:
-        # No science file type matched, return an error with the
+    except imap_data_access.ScienceFilePath.InvalidScienceFileError:
+        # Not a SCIENCE file, continue on to ancillary files
+        pass
+
+    try:
+        # file_obj will be None if it's not a SPICE file
+        file_obj = imap_data_access.AncillaryFilePath(filename)
```
maxinelasp: AncillaryFilePath wasn't added in imap_data_access until v0.13.0! If you update the version in pyproject.toml and then run `poetry lock --no-update` and include it here, it should start working!

sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
```diff
@@ -122,8 +122,15 @@ def lambda_handler(event, context):
...
+    except imap_data_access.AncillaryFilePath.InvalidAncillaryFileError as e:
+        # No ancillary, science, or spice file type matched, return an error with the
         # exception message indicating how to fix it to the user
         logger.error(str(e))
         return {"statusCode": 400, "body": str(e)}
```
maxinelasp: Can you update this string "e" to clarify that NO file path was matched? So maybe instead of passing along the ancillary file error, return "error: file name does not match ancillary, science, or SPICE file paths." 
Could you also add "logger.info" messages to the other `except` statements for ScienceFilePath and SPICEFilePath so we get extra visibility into failures there? 

Long term I would also be interested in a better logging/return error system here, but that's out of scope for this ticket... could you open a new issue to come back and improve logging here? I think some of my changes from the dependency abstraction class would help with that.

maxinelasp (commented):
Just reviewed upload as Daralynn said in slack that was the piece  that was finished. Overall looks great, just a few comments and once query_api.py is removed from this PR (or I get the thumbs up to review that too) then I will approve! Nice work!

daralynnrhode:
I added a test for uploading ancillary files but the portion in which it tries to upload the same file again is failing, any ideas why? 
edit: Do I need to add this file to the S3 manually? I believe that is what the comments in the code is suggesting 

>  
    # Try to upload over a pre-existing file and we should get a 409
    # Note that we are using pre-signed urls so we haven't actually
    # uploaded anything in the previous call, only gotten back a url
    # So we need to upload a file to s3 to simulate a pre-existing file
    
How should I do this?

sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
daralynnrhode: @maxinelasp I went in a created an issue In the student table, but wasn't too sure about what improving it would look like so I largely left it blank.

maxinelasp: That's perfect, just writing it down somewhere so it doesn't get lost in the aether is sufficient. Thanks!

maxinelasp:
I think that does imply that there should be a pre-existing test file to try and upload over - does your ancillary file exist in S3? Does the test file used in the other upload test exist in S3? First step would be to check those. I also saw in git blame that @greglucas wrote those tests - so maybe he could clarify for us?

greglucas:
This shouldn't be touching any external files / s3, it should all be mocked. I think what might be happening is you are putting an object at the ancillary file path, then you are calling upload_lambda() which is trying to put an object in s3 but my guess is that might be a different place in s3 than the path you did put_object on (because there is some extra directory mapping going on for the various types of files).

daralynnrhode:
@maxinelasp I did not end up using the `construct_path()` method to make the `ancillary_file`. It was leading to some weird errors when it came to running the actual test. 
The two paths being used to upload would both match, for both the initial upload, and the second try for a duplicate error, but the error would not be triggered. Instead it would pass. The paths would be `/home/runner/work/sds-data-manager/sds-data-manager/data/imap/ancillary/swe/imap_swe_test-ancillary-description_20100101_v000.cdf`.
 I was pretty sure these were too long and needed to be trimmed down, but between the path and what the `put_object()` `key` should be I couldn't figure out what needed to be trimmed where. If that makes sense.

maxinelasp:
@daralynnrhode Ok, that sounds good, thanks for investigating! Are you intending query_api.py to be included in this review? I think the download pieces look good so I'm ready to approve if you remove the query_api file from this review!

maxinelasp (approved):
Nice work!

PR #467, opened 2025-02-27
Reconfigure lambda to use non-proxy integration and queryStringParameters

torimarbois:
# Change Summary
I recently configured this endpoint, but upon testing today realized that something was missing. The issue was that it was neither fully set up to use a proxy integration OR a non-proxy integration, which are the two options offered by AWS APIGateway. 

I went with the non-proxy, 'custom' integration because it made more sense to me than the proxy version.

## Updated Files

   - remove the `use_path_params` variable, which is meant to be used with a proxy integration
   - add a memory parameter, because it's used in other lambda function definitions
- ialirt_catalog_api.py
   - add queryStringParameters wrapping to the request parameters
- test_ialirt_catalog_api.py
   - add queryStringParameters wrapping to the request parameters

## Testing

Updated the unit tests, they all pass. I also deployed the IalirtStack with these changes, and was able to get a successful response back from the endpoint using Postman.

### DIFF
--- a/sds_data_manager/constructs/ialirt_api_manager_construct.py
+++ b/sds_data_manager/constructs/ialirt_api_manager_construct.py
@@ -124,6 +124,7 @@ class IalirtApiManager(Construct):
             handler="IAlirtCode.ialirt_catalog_api.lambda_handler",
             runtime=lambda_.Runtime.PYTHON_3_12,
             timeout=cdk.Duration.minutes(1),
+            memory_size=1000,
             environment={
                 "S3_BUCKET": data_bucket.bucket_name,
                 "REGION": env.region,
@@ -138,7 +139,6 @@ class IalirtApiManager(Construct):
             route="ialirt-catalog",
             http_method="GET",
             lambda_function=catalog_api,
-            use_path_params=True,
         )
 
         ialirt_db_query_handler = lambda_.Function(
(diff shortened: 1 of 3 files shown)


greglucas (approved):

sds_data_manager/lambda_code/IAlirtCode/ialirt_catalog_api.py
```diff
@@ -60,8 +60,10 @@ def lambda_handler(event, context) -> dict:
     """
```
laspsandoval, line 65: Update your docstring. Right now it is:
        Ex:
            event = {
                "start_date": "2025-02-06",  # inclusive
                "end_date": "2025-02-07",  # exclusive
                "station": "station1",
            }
            
And also please add an example of what you would put in Postman to perform a query (in this above docstring).

sds_data_manager/constructs/ialirt_api_manager_construct.py
```diff
@@ -138,7 +139,6 @@ def __init__(
             route="ialirt-catalog",
             http_method="GET",
             lambda_function=catalog_api,
-            use_path_params=True,
```
laspsandoval, line 141: I think this is ok. Could you write what the api call will be though?

laspsandoval (commented):
Just a couple of questions before I approve.

PR #472, opened 2025-03-10
Fixing requirements.txt file for lambda

maxinelasp:
# Change Summary
Lambda was using an outdated requirements.txt file. This updates.


greglucas (approved):
Thanks for tracking this down! ❤️

A few questions to try and get your thoughts on how we can improve this workflow to avoid it happening so frequently.

1. The poetry.lock file was previously made with 1.8.3, you've updated with 1.8.0. I feel like need a way to check that we only let the pre-commit hook make the poetry lock file (poetry running in its own virtual environment, not a user's local poetry environment which may be more recent/upgraded).
2. Can/should we force requirements.txt to be created every time? That might mean a lot more layer deployments whenever a new version of anything comes in though, or does that not happen because it is taken from poetry.lock which freezes versions?
3. I don't quite understand why this version bumping would lead to the missing psycopg binary previously. The package was present in `requirements.txt` before. I wonder if the "fix" has been deploying locally rather than from a linux VM in GitHub Actions. i.e. GitHub Actions might not have the underlying postrgres library installed and therefore that doesn't get linked in properly, whereas we do have that locally and building a wheel brings that along. (This is just a thought of what could be going wrong)

maxinelasp:
@greglucas response:

> 1. The poetry.lock file was previously made with 1.8.3, you've updated with 1.8.0. I feel like need a way to check that we only let the pre-commit hook make the poetry lock file (poetry running in its own virtual environment, not a user's local poetry environment which may be more recent/upgraded).

I don't think everyone has pre-commit installed locally. I'm fine with saying "you have to use pre-commit in this way" but just wanted to call that out. I'm not sure how we could enforce pre-commit creating the lock file... Poetry doesn't care about which version of it that was used to create the lock file, so even though `poetry lock` runs on every commit, most of the time it doesn't update. Maybe the best way is to remove the poetry-check step so it doesn't tell the user to run poetry lock and then depend on the poetry lock pre-commit check only? That might just cause users with pre-commit installed to always use the pre-commit generated version.

(for the record, both my version of Poetry and the version of pre-commit's Poetry are 1.8.0)

> 2. Can/should we force requirements.txt to be created every time? That might mean a lot more layer deployments whenever a new version of anything comes in though, or does that not happen because it is taken from poetry.lock which freezes versions?

I think this would be fine because it wouldn't update unless poetry.lock updates. However, I thought it _was_ running every pre-commit, but for some reason the step in pre-commit is skipped a lot. I was just testing it and even when the step runs, it doesn't update requirements.txt. 

Unfortunately I suspect this is some combination of Poetry being broken and pre-commit being broken... 

Maybe we could try deleting the file and then re-generating it every time? It shouldn't trigger redeploys due to the lock file, but it would make the pre-commit step slower. 

> 3.I don't quite understand why this version bumping would lead to the missing psycopg binary previously. The package was present in requirements.txt before. I wonder if the "fix" has been deploying locally rather than from a linux VM in GitHub Actions. i.e. GitHub Actions might not have the underlying postrgres library installed and therefore that doesn't get linked in properly, whereas we do have that locally and building a wheel brings that along. (This is just a thought of what could be going wrong)

This is an interesting theory. I did see the error when deploying locally with mismatched poetry and requirements.txt versions, which is why I think the issue is related to that. I am wondering if the layer is created with one version, and the code uses a different version, or something weird like that. I'm not able to find much information on why the error is occurring, and in the past it was fixed by redeploying and bumping the version, so that says to me it is related to something in our dependency system.

maxinelasp:
@greglucas and as another strike against the local vs remote deploy theory, I just deployed this branch via Github action (and confirmed the lambda did get updated) and it seems like everything is working.

greglucas:
The lambda was updated, but I don't think the underlying Layer was updated?

https://github.com/IMAP-Science-Operations-Center/sds-data-manager/actions/runs/13792398602/job/38575419226#step:9:161

PR #481, opened 2025-03-14
Add Ancillary file handling in batch starter lambda 

lacoak21:
# Change Summary

## Overview
This adds handling for when an ancillary file triggers the batch starter lambda. It also incorporates Maxine's shared dependency code. The batch starter now kicks off jobs by using ProcessingInputCollection.serialize() as the dependencies in the imap-processing cli command.

## Updated Files
   - Handling for ancillary files
   - Creates a ProcessingInputCollection and uses the serialized version of that in the processing command.
   - Removed code to get the end_date from file name. As of now we dont use end date for science files and if for whatever reason the file did have an end date, the science file table doesnt have a column "end_date". 

## Testing
Add more batch starter tests relevant to the updates above


tech3371:
I have one major requests for this PR which is to move dependencies lookup from batch starter to dependency lambda(API). I am following up on this on Slack since it's big requests and may require some discussion.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -60,74 +65,34 @@ def _get_url_response(url: str):
         raise IMAPDependencyFinderError(message) from e
 
 
-def _get_dependencies(dependency_events):
+def _get_dependencies(dependency_events: dict):
+    """Return dependencies for the input dependency requirements."""
```
maxinelasp: Need to document dependency_events and add a return type here

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -60,74 +65,34 @@ def _get_url_response(url: str):
...
     with _get_url_response(url) as response:
         # Retrieve the response as a list of dictionaries containing the dependency
         # information
-        dependency_repsonse = response.read().decode("utf-8")
-        # Parse the JSON responses
-        dependencies = json.loads(dependency_repsonse)
-        logger.debug(f"Received dependencies: {dependencies}")
+        dependency_response = response.read().decode("utf-8")
+        logger.debug(f"Received dependencies: {dependency_response}")
+    # If start_date is supplied, the dependency api will return a serialized
```
maxinelasp: Why only if start_date is provided? What is the other case here?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -198,125 +169,53 @@ def try_to_submit_job(session, job_info, start_date, version):
...
-        start_date=datetime.strptime(start_date, "%Y%m%d"),
-        version=version,
-    )
+            return
+
+        # All of our upstream requirements have been met.
+        # Try to insert a record into the Processing Jobs table
+        # If this job already exists, then we will get an integrity error
+        # and know that some other process has already taken care of it
+        processing_job = models.ProcessingJob(
```
maxinelasp: Just as a note, this is used as a lock because there is a unique constraint on the database table which ensures we can only have one process in progress at a time for a given file across the entire system. This lock lasts until after processing finishes so we don't get duplicate processes. I thought I'd point it out since you asked about it.

tests/conftest.py
```diff
@@ -83,3 +90,208 @@ def setup_dynamodb():
...
+        "version": version,
+        "trigger_source": trigger_source,
+    }
+    for param, value in optional_params.items():
+        if value:
+            if isinstance(value, list):
+                single_val = value[0]
+            else:
+                single_val = value
+            event["queryStringParameters"][param] = single_val
```
greglucas: What would a list of items mean, and is it OK to just take the first item and drop the rest?
```suggestion
        if value is not None:
            event["queryStringParameters"][param] = value
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
```diff
@@ -164,6 +164,7 @@ def s3_event_handler(event):
         #     "data_level": self.data_level,
         #     "descriptor": self.descriptor,
         #     "start_date": datetime.strptime(self.startdate, "%Y%m%d"),
+        #     "end_date": datetime.strptime(self.enddate, "%Y%m%d"),
```
greglucas: You added it to the comments here, but deleted it below? I'm not sure which you want.

tests/conftest.py
greglucas, line 1: Should these changes be in the `tests/lambda_endpoints/conftest.py` file? (down a directory)

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -263,8 +179,69 @@ def test_lambda_handler_multiple_events(session, mock_urlopen):
...
+        ]
+    }
+
+    context = {"context": "sample_context"}
+    with patch.object(batch_starter, "BATCH_CLIENT", Mock()) as mock_batch_client:
+        lambda_handler(events, context)
+        # There should be two different jobs submitted for one swe l1b ancillary file
+        assert mock_batch_client.submit_job.call_count == 2
+        # Assert_called_with only works on the last call
+        # Check that the last call is what we expect with the corrected
```
greglucas, line 208: Is it deterministic in how it decides which job to submit first/second, or is there chances of a race condition?

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -263,8 +179,69 @@ def test_lambda_handler_multiple_events(session, mock_urlopen):
...
+                    dependencies.serialize(),
+                    "--upload-to-sdc",
+                ]
+            },
+        )
+        # Submit a second job with the same file as input which will try to kick
+        # off a duplicate job. We expect the submit_job method to not be called
+        # so make sure it is still only called three times from our previous iteration.
+        lambda_handler(events, context)
+        assert mock_batch_client.submit_job.call_count == 2
```
greglucas: Might want to reset this to a value you control and/or add assert_not_called() after this?
```suggestion
        # so make sure it is still only called two times from our previous iteration.
        mock_batch_client.submit_job.call_count == 0
        lambda_handler(events, context)
        assert mock_batch_client.submit_job.call_count == 0
```

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -283,6 +260,30 @@ def test_lambda_handler_no_dependencies(session, mock_urlopen):
...
+    }
+    context = {"context": "sample_context"}
+    with caplog.at_level(logging.DEBUG):
+        lambda_handler(events, context)
+        # Verify the info statement was logged.
+        assert any(
+            "Upstream dependency not found for: swe, l2, sci, 20000101, None, l1b, v001"
+            in message
+            for message in caplog.text.splitlines()
+        )
```
```suggestion
        assert "Upstream dependency not found for: swe, l2, sci, 20000101, None, l1b, v001" in caplog.text
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -60,74 +65,34 @@ def _get_url_response(url: str):
...
         # Retrieve the response as a list of dictionaries containing the dependency
         # information
-        dependency_repsonse = response.read().decode("utf-8")
-        # Parse the JSON responses
-        dependencies = json.loads(dependency_repsonse)
-        logger.debug(f"Received dependencies: {dependencies}")
+        dependency_response = response.read().decode("utf-8")
+        logger.debug(f"Received dependencies: {dependency_response}")
+    # If start_date is supplied, the dependency api will return a serialized
+    # ProcessingInputCollection instance.
```
greglucas: Do we need to handle both cases? Is this to support the previous way of requesting dependencies too?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -60,74 +65,34 @@ def _get_url_response(url: str):
...
-    return record
-
-
 def is_job_in_processing_table(
     session: db.Session,
     instrument: str,
-    data_level: str,
     descriptor: str,
-    start_date: str,
+    data_level: str,
```
greglucas, line 121: Why are we changing the order of these? Our filenames are instrument_data-level_descriptor I thought, so I think it would make sense to be consistent with that.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -198,125 +169,53 @@ def try_to_submit_job(session, job_info, start_date, version):
...
-        record = get_file(
-            session,
-            upstream_source,
-            upstream_data_type,
-            upstream_descriptor,
-            upstream_start_date,
-            upstream_version,
-        )
-        if record is None:
+    with db.Session() as session:
```
greglucas: Why is this moved inside here now rather than passed in as previously? I think there was a reason to avoid creating new sessions every time we make calls through our stack. But I don't remember all of the details.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -354,6 +253,94 @@ def try_to_submit_job(session, job_info, start_date, version):
...
+            f"{version}"
+        )
+        return  # Exit the loop early
+
+    logger.info(f"All dependencies found for the job: {job}")
+    # Find science processingInputs that have the same source as the potential job
+    for dep in upstream_dependencies.get_science_files():
+        if job["data_source"] == dep.source and isinstance(
+            dep, processing_input.ScienceInput
+        ):
```
greglucas: @tech3371 and @lacoak21, I think this is something to consider between your two PRs.

Tenzin, I believe you are returning only "paths", not the objects, so it might be worth considering whether we need another function or a different return type on the other PR.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -354,6 +253,94 @@ def try_to_submit_job(session, job_info, start_date, version):
...
+                # TODO when do we bump version?
+                try_to_submit_job(
+                    job, job_start_date, job_version, upstream_dependencies
+                )
+        # TODO should we break?
+        # For example if the potential job is mag_l1c (dependent on
+        # mag_l1b_norm_20250101 and mag_l1b_burst__20250101), we only want to
+        # kick off one mag l1c_20250101 job once and run it with both
+        # mag_l1b_norm and mag_l1b_burst as dependencies.
+        break
```
greglucas: Isn't this OK? We don't allow multiple jobs to be running at the same time, so one would fail and the other would succeed.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -52,6 +60,17 @@ def valid_source(self) -> list[str]:
...
 
+def valid_science(data_type) -> bool:
+    """Check if data_type is in valid data levels.
+
+    Returns
+    -------
+    bool
+        True if the data_type is in VALID_DATALEVELS.
+    """
+    return data_type in [*imap_data_access.VALID_DATALEVELS]
```
greglucas: What does data_type mean? Should this be a data_level check instead, or are you really after a try/except turning it into a science file in the calling routines?

pyproject.toml
tech3371, line 1: some of these other file changes may go away after rebase. Can you do a rebase?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -371,6 +705,33 @@ def lambda_handler(event, context):
                     "descriptor": "historical",
                 },
             ]
+        If "start_date" is supplied "version" and "ancillary_trigger" are required and
```
tech3371:
```suggestion
        If "start_date" is supplied, "version" and "ancillary_trigger" are required and
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
greglucas, line 1: There are a few files here that I think should be separate PRs or were brought in from other PRs (requirements.txt for example). Can you rebase and fix these up so that this only shows your changes? I'm hoping we can avoid the psycopg errors by not updating requirements.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -391,9 +752,46 @@ def lambda_handler(event, context):
...
+                "'trigger_source' is required.",
+            }
+
+        # Get and convert end_date in one line if it exists
+        end_date = (
+            datetime.strptime(query_params.get("end_date"), "%Y%m%d")
+            if query_params.get("end_date")
+            else None
+        )
+        # TODO this only works for upstream deps right now. Do we need to ever get files
```
tech3371, line 812: good question. We can visit that in future PR if we need it. I can't think of any situation at the moment.

greglucas (commented):
This is a lot of new code, so I can't say I was able to review it all or understand everything that is going on, so take my comments with a grain of salt.

This is a PR where we should think about MVPs to get a smaller review in first, and extension PRs after that to add additional capabilities.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+    """
+    return (
+        query["data_source"] == dependency["data_source"]
+        and valid_science(dependency["data_type"])
+        and valid_science(query["data_type"])
+    )
+
+
+def get_dependency_processing_input(
+    query: dict,
```
tech3371: can you change this to `query_params` like you input names?
```suggestion
    query_params: dict,
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+    version: str,
+    trigger_source: str,
+    end_date: Optional[datetime] = None,
+):
+    """Construct a ProcessingInputCollection of dependency files.
+
+    Parameters
+    ----------
+    query : dict
+        Api query.
```
tech3371:
```suggestion
        Query parameters received from API calls.
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+       Upstream or downstream dependency from the query.
+
+    Returns
+    -------
+    bool
+        True if dependency is a primary science dependency, False otherwise
+
+    Examples
+    --------
+    swe_l1b_sci is a primary science dependency of swe_l1a_sci.
```
tech3371, line 417: really good function naming and doc string.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+
+def primary_science_dep(query: dict, dependency: dict) -> bool:
+    """Check if the dependency is a primary science dependency.
+
+    A primary science dependency exists when both the upstream and downstream
+    dependencies have the same data source and both are valid science data types.
+
+    Parameters
+    ----------
+    query : dict
```
tech3371: same here if you can change it to `query_params`

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+    start_date : datetime
+        Start date of the event data.
+    version : str
+        Version of the event data.
+    end_date: datetime, optional
+        End date of the event data.
+    exact_science_date: bool, optional
+        When True, query for science files with a match to the start time. It is assumed
+        that the dependency is a primary science dependency and the trigger source is of
+        the same data_source. Default is False.
```
tech3371:
```suggestion
        When True, query for science files with a match to the start time because it assumed
        that the dependency is a primary science dependency and the trigger source is of
        the same data_source. Default is False.
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+            # Trigger source: swe_l1b-flight-cal_20250102-20250104
+            # Downstream: swe_l1b_sci
+            # Upstream: Look for swe_l1a_sci with start dates in range
+            # 20250102-20250104
+            type_specific_conditions.append(
+                and_(
+                    models.ScienceFiles.start_date >= start_date,
+                    models.ScienceFiles.start_date <= end_date,
+                )
+            )
```
tech3371, line 633: Everything in this function looks great. It was easy to follow and those examples were so helpful. Even after looking at this few times before, I still got lost. Nice job!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+            # the start_date (start_date comes from an ancillary file, so we
+            # cannot use the exact date.)
+            # Example:
+            # Trigger source: swe_l1b-flight-cal_20250102_v001.pkts
+            # Downstream: swe_l1b_sci
+            # Upstream: Look for swe_l1a_sci with start dates greater than or
+            # equal to 20250102
+            type_specific_conditions.append(
+                models.ScienceFiles.start_date >= start_date
+            )
```
tech3371: I am little lost on this one. Is this different from `exact_science_date`? I tried to look at the example and doc string but I am still little confused.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+
+    filter_conditions = [
+        table.instrument == dependency["data_source"],
+        table.descriptor == dependency["descriptor"],
+        # table.version == version,
+        *type_specific_conditions,
+    ]
+
+    # Create a subquery that makes a column with the max version grouped by start_date
+    subquery = (
```
tech3371, line 667: If I understand correctly here, you are getting latest version. Can you add TODO to check if version is supplied, otherwise default to latest.

tech3371 (approved):
Huge Kudos to the amazing work! This was not easy but you wrote code that was easy to follow even when logic is hard. Really nice work!! I had few minor suggestions and one main questions. Otherwise, it looks great to me!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371, line 1: oops. I forget to review this file. Reviewing it now

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: Yeah, I am going to undo that function so that Luisa's code here doesn't get messed up and then create new function for the ones I need since they are expecting different things.

tech3371 (commented):
batch starter seems like it's in good shape as it is. I know we are planning to make more improvement in future PR for soft dependency. This is a great foundation work to start with.

lacoak21:
> This is a lot of new code, so I can't say I was able to review it all or understand everything that is going on, so take my comments with a grain of salt.
> 
> This is a PR where we should think about MVPs to get a smaller review in first, and extension PRs after that to add additional capabilities.

Yes i definitely agree! This was a learning lesson for next time.

tests/conftest.py
lacoak21: I added the list functionality because I wanted to create a dependency api event directly from a url when mocking the call to the dependency lambda handler. I do that by calling urllib.parse.parse_qs which returns a dictionary of lists for each param. The url should only ever contain one item per param.

    parsed_url = parse.urlparse(url)
    params = parse.parse_qs(parsed_url.query) 

This is used in test_batch_starter.urlopen_side_effect() function

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
lacoak21: Thanks for the catch! I added the comment because i saw it below then realized we dont have end date in the science files table and removed. I then forgot to go back and remove this comment.

tests/lambda_endpoints/test_batch_starter.py
lacoak21: In this test case it would be the order in which the upstream files appear in the science files table. It should always be this order for testing unless the order they get added changes in conftest.py

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Yes - batch starter still uses the "old way" when looking for downstream dependencies.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: I moved it back out! I definitely dont need to create new sessions.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Good point. I am really after seeing if the dependency is a science file by checking the data_type. Data_type is either "science" "ancillary" or a science data_level.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Thanks!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Yes this one is different than exact_science_date because the only time when we can use the exact start date is when the file that triggered the batch starter is also a primary science dep. 

For example if swe_l1a_sci_20240501_v001.cdf triggered the batch starter we first look downstream to swe_l0_raw then look back upstream and get the existing files. The start_date above actually comes from the trigger file so in this case it is "20240501". Since the trigger file is the same source as the downstream and upstream file we are looking for then we can use this as the date since they should be the same. 

BUT if the trigger file is mag_l1b_sci_20240510_v001.cdf and we find swe_l1a_sci as a downstream and are looking upstream to find all the swe l10 files, using "20240510" as the exact start date, we might not get the right swe l0 file because it might not have the same date as the mag file. 


Does that make more sense? I also might be confused here and we might actually be able to use the mag start date if they should be on the same day.

tech3371: I see. I have been thinking `primary` as meaning it is its own instrument that's kicking off the job. So in this case, it's MAG l1b that's causing SWE L0 to process. Can you put that example here? I think many can relate to this since MAG will be a HARD dependency type for many instrument at higher level. And can you rename that variable from `primary_sci_dep` to something like `sci_anc_dep` or `is_another_sci_dep`. I am not great with names but something without the `primary_` in it.

lacoak21: Ah I see that makes sense - I was thinking primary when looking for direct upstream/ downstream regardless of what kicked off the job. I can definitely rename it. What about ancillary_trigger_sci_dep?

tech3371: how about this `sci_anc_trigger_dep`?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+    """Filter primary science dependencies for unprocessed downstream dependencies.
+
+    Parameters
+    ----------
+    session : orm session
+        Database session.
+    records : list
+        Science file records.
+    query_data_type : str
+        The query data type.
```
maxinelasp: What is the query data type? I see it's a string, what does it describe?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+    session: db.Session, records: list, query_data_type: str
+):
+    """Filter primary science dependencies for unprocessed downstream dependencies.
+
+    Parameters
+    ----------
+    session : orm session
+        Database session.
+    records : list
+        Science file records.
```
maxinelasp, line 363: What is the type of the items in the list?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+
+    Returns
+    -------
+    bool
+        True if dependency is a primary science dependency, False otherwise
+
+    Examples
+    --------
+    swe_l1b_sci is a primary science dependency of swe_l1a_sci.
+    mag_l1c_norm-mago is a primary science dependency of mag_l1b_norm-burst.
```
maxinelasp:
```suggestion
    mag_l1c_norm-mago is a primary science dependency of mag_l1b_norm-mago.
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+    dependencies: list,
+    start_date: datetime,
+    version: str,
+    trigger_source: str,
+    end_date: Optional[datetime] = None,
+):
+    """Construct a ProcessingInputCollection of dependency files.
+
+    Parameters
+    ----------
```
maxinelasp, line 442: Can I get more details on all the parameters described here? What is the structure you are looking for?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+        Version to find dependent files with.
+    trigger_source : str
+        Data source of the file that triggered the batch starter.
+    end_date : datetime, optional
+        End date to find dependent files with.
+
+    Returns
+    -------
+    ProcessingInputCollection
+        Dependency files that can include Ancillary, SPICE, or Science inputs
```
maxinelasp: Could you add some logging in this function? I suspect we will end up debugging it a bit...

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+            # table. If the file already exists, the l1a file is ignored.
+            if primary_sci_dep:
+                filenames = filter_primary_science_dependencies(
+                    session, records, query_params["data_type"]
+                )
+                if not filenames:
+                    return dependency_inputs
+            # Create a processingInput instance and add it to the collection
+            if dep["data_type"] == DataType.ANCILLARY:
+                inputs.append(processing_input.AncillaryInput(*filenames))
```
maxinelasp, line 534: You can just add these files directly to `dependency_inputs` using `dependency_inputs.add()` rather than putting them in a list first

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -331,6 +350,313 @@ def get_dependencies(node, dependency_type, relationship):
...
+            # Query for science files matching the start date and version
+            # Example:
+            # Trigger source: swe_l0_raw_20250102_v001.pkts
+            # Downstream: swe_l1a_sci
+            # Upstream: Look for swe_l0_raw with start date == 20250102 and
+            # version == v001
+            type_specific_conditions.extend(
+                [
+                    models.ScienceFiles.start_date == start_date,
+                    table.version == version,
```
maxinelasp, line 616: I don't think we want to pin the version numbers here... will this work without that constraint? We can leave it for a future revision to figure out the versioning, I think. 

For example, what if there is a soft dependency on another instrument and then we have conflicting versions? I.E. we're on SWE_L2_v004 and this was triggered by an incoming MAG_L2_v004. We would want to run SWE_L2_v005. 

Unfortunately this might actually need the CRID work from reprocessing to actually figure out. (In that design, we can calculate the inputs of every file from a shared Composite Release ID so we know what versions of each input file were used to generate it, and we can then check that to see if the file we have would change the output CRID of the potentially triggered processing. [More info](https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/221740849/Reprocessing+Spike)) 

All this sounds like a future problem, just wanted to call it out.

tests/lambda_endpoints/test_dependency_api.py
```diff
@@ -115,3 +215,177 @@ def test_dependency_class(mock_load_dependencies):
...
+    assert len(records) == 2
+
+    for record in records:
+        assert record.instrument == "lo"
+        assert record.descriptor == "sci"
+
+    assert records[0].start_date == datetime(2010, 1, 1)
+    assert records[1].start_date == datetime(2010, 1, 2)
+    assert records[0].version == "v002"
+    assert records[1].version == "v003"
```
maxinelasp, line 391: Nice tests. In the future we should add some to handle versioning conflicts or some of the more weird/complicated ancillary file requirements. Just noting that here for posterity.

maxinelasp (commented):
Looks great overall! Nice job, this was loads of effort but it's great that we have a mostly-complete outcome. I had some comments but hopefully nothing too major...

maxinelasp (approved):

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: If start_date is not provided, this indicates that we want to just have a list of dependency dictionaries returned. This functionality is still used when we are looking at downstream dependencies and just want to know the source, data_type, and descriptor. 

For example: 

```
If "start_date" is not supplied return list of dictionaries:
        statusCode and body containing list of dictionary containing
        the dependencies information like this:

   [
                {
                    "data_source": "hit",
                    "data_type": "l1a",
                    "descriptor": "all",
                },
                {
                    "data_source": "hit",
                    "data_type": "l1b",
                    "descriptor": "hk",
                },
                {
                    "data_source": "sc_attitude",
                    "data_type": "spice",
                    "descriptor": "historical",
                },
            ]
```
        If "start_date" is supplied, "version" and "ancillary_trigger" are required and
        "end_date" is optional. Return a ProcessingInputCollection of files that exist
        on s3.
       ```
     [
                {
                    "type": "ancillary",
                    "files": [
                        "imap_mag_l1b-cal_20250101_v001.cdf",
                        "imap_mag_l1b-cal_20250103-20250104_v002.cdf"
                    ]
                }
            ]
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Its the  data_type of the dependency used to query the api. E.g, "ancillary" or "l1b". I will update that docstring.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Oops good catch. I think I will use mag_l1b_burst-mago to explain that the descriptor can be different.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
maxinelasp: Can you quickly document that in this line? :)

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: The version is only set when exact_science_date is true. I will rename this variable to be exact_science_date_and_version. It is for the case when it was triggered by the same source as the current downstream and upstream dependency. The example you showed above would be "exact_science_date" == False but if it was:

"we're on SWE_L2_v004 and this was triggered by an incoming SWE_L1a_v004. We would want to run SWE_L2_v004."

Does that make sense?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Yes!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
maxinelasp: I see... There are some situations where I could see this causing issues. (Weird ones!)

For example, MAG L1C takes in normal mode L1B OR burst mode L1B OR both of them. If we get a late arriving file, then:

L1b_burst_v003 and L1c_v003 could both exist already.
Then L1b_norm_v001 comes in... we would want to trigger L1c_v004, with the inputs L1b_norm_v001 and L1b_burst_v003. 

But that is the only case and maybe the only instrument where that might be a problem, and it's very odd. So that is better than I thought!

lacoak21: Ohhh right. Ok then I agree that it might be worth removing that set versioning for now.

maxinelasp: I am ok with leaving it in since that is a truly esoteric case! Up to you, though

lacoak21: OK I left it in but left a todo to revisit it.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: I actually fixed this in the branch where I am adding SOFT dependency handling so I will merge in that change in the next PR

PR #484, opened 2025-03-19
Adding Query by Ingestion Start and End Date

### LINKED ISSUE #483: Add querying by ingestion date to the API
### Summary

MAG has requested the ability to query files based on ingestion date to the API. This would allow them to determine when files are updated. 

This is already in the database and accessed by the query API lambda, we just need to add a new option to the curl call. This could be with some kind of ingestion_start and ingestion_end dates, or with one ingestion_range and greater than or less than operators. 

This isn't required in imap-data-access for now as they're using the API directly.

### Requirements

The more specific the better.
  - [ ] Create a query for files that arrived later than a certain date
  - [ ] Create a query for files that arrived sooner than a certain date
  - [ ] Should work for science, ancillary, or SPICE files

### Related tickets

_No response_

daralynnrhode:
# ingestion_start_date and ingestion_end_date added to query parameters

## Overview

Per request, adding the ability to query by ingestion date. I followed the same format that start and end date used since it seemed to be the same thing, just with the `ingestion_date` parameter in the database.

I'll go ahead and add this functionality to `mmap-data-access` just so the two querying options stay synced. 

Closes #483 
## Updated Files

- sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py


daralynnrhode:
Tests are failing because of the way the `ingestion_date` is formatted as a dateTime object. 
Mainly need to decide in the query lambda if the ingestion dates should be filter by just the `%Y-%m-%d` or by the date and time. 
I would assume just by the date because it would be tedious to expect users to also supply time.

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
```diff
@@ -84,6 +87,18 @@ def lambda_handler(event, context):
                 models.ScienceFiles.start_date
                 <= datetime.datetime.strptime(value, "%Y%m%d")
             )
+        elif param == "ingestion_start_date":
+            # filtering by ingestion date
+
+            query = query.where(
+                func.date(models.ScienceFiles.ingestion_date)
+                >= datetime.datetime.strptime(value, "%Y-%m-%d")
```
maxinelasp: Is the format different for ingestion date vs start_date? looks like start_date uses `%Y%m%d`

maxinelasp (commented):
Looks good overall, but we should match the format of start_date and end_data (YYYYMMDD instead of YYYY-MM-DD)

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
daralynnrhode: I originally had it matching, but that caused a lot of tests to fail. Ill have to go in and rework the tests for it

daralynnrhode:
There is a _single_ test failing now and I don't have the brain capacity to figure it out so Ill come back to this tomorrow.

tests/lambda_endpoints/test_query_api.py
```diff
@@ -119,22 +119,98 @@ def test_empty_end_date_query(session):
...
     assert returned_query["body"] == expected_response
 
 
-def test_non_date_query(session, expected_response):
-    """Test that a non-date parameters can be queried."""
+def test_empty_ingestion_end_date_query(session):
+    """Test that an ingestion_end_date query with no matches returns an empty list."""
     _populate_test_data(session)
-    event = {"queryStringParameters": {"instrument": "hit"}}
+    event = {"queryStringParameters": {"ingestion_end_date": "20261101"}}
```
maxinelasp: You have a typo - should be `20251101`

tests/lambda_endpoints/test_query_api.py
```diff
@@ -22,7 +22,7 @@ def _populate_test_data(session):
         "version": "v001",
         "extension": "pkts",
         "ingestion_date": datetime.datetime.strptime(
-            "2025-11-07 10:13:12+00:00", "%Y-%m-%d %H:%M:%S%z"
+            "20251107 10:13:12+00:00", "%Y%m%d %H:%M:%S%z"
```
tech3371: If you do this, then you will need to update date format in this [function](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/bf9a64e8fbe96e4bbb7060d71d1f7de6e809b791/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py#L22) too. I believe that function `get_file_ingestion_date()` is used by science, ancillary, and may be SPICE table too (it's not implemented yet but plans to use the same function to write it). It may require update to tests of indexer like you did here.

tech3371 (commented):
The changes looks good. I have one change requests which is to update the ingestion date function to match date format changes in this PR.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
```diff
@@ -182,7 +183,20 @@ def s3_event_handler(event):
...
-        ingestion_date_object = get_file_ingestion_date(s3_filepath)
+        # ingestion_date_object = get_file_ingestion_date(s3_filepath)
+
+        # Get ingestion date str
+        ingestion_date_str = get_file_ingestion_date(s3_filepath)
+
+        # convert str to datetime with no hyphens
+        ingestion_date_no_hyphens: datetime = datetime.strptime(
+            ingestion_date_str, "%Y%m%d %H:%M:%S%z"
+        )
```
tech3371: can you move this line of converting str to datetime object in `get_file_ingestion_date()` to remove these lines? I thought we were returning datetime object. My mistake.

tests/lambda_endpoints/test_query_api.py
daralynnrhode: @tech3371 Is this function ever used by anyone to find what the ingestion date is? I'm running into the problem of when I change the output of `get_file_ingestion_date` it must be changed to a `string` in order to return the ingestion date without hyphens. This however affects the calls made in `s3_event_handler()`. In order for the sql queries to run it must be in a date time object, and simply converting won't solve the problem because a mismatch of formats will still occur. 
I am under the impression that the way I implemented the ingestion start/end dates allows the query to run with the hyphen format and then converts it before and after the query is actually made. That way the format stored in the database tables is `%Y-%m-%d %H:%M:%S%z`, I assume this is the same format the `start_time` is stored as. 
To my understanding, `get_file_ingestion_date()` is what formats _how_ the ingestion date gets stored in the database and my implementation of `ingestion_start_date` and `ingestion_end_date` takes that into account. 
Am I correct? Does this actually need to be changed?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
```diff
@@ -43,9 +43,10 @@ def get_file_ingestion_date(file_path):
     response = s3_client.head_object(Bucket=bucket_name, Key=file_path)
     file_ingestion_date = response["LastModified"]
 
+    file_ingestion_date_formatted = file_ingestion_date.strftime("%Y%m%d %H:%M:%S%z")
```
tech3371:
```suggestion
    file_ingestion_date_formatted = file_ingestion_date.strftime("%Y%m%d %H:%M:%S%z")
    ingestion_date_object = datetime.strptime(
            file_ingestion_date_formatted, "%Y%m%d %H:%M:%S%z"
        )
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
```diff
@@ -43,9 +43,10 @@ def get_file_ingestion_date(file_path):
...
     # LastModified looks like this:
-    # 2024-01-25 23:35:26+00:00
-    return file_ingestion_date
+    # 20240125 23:35:26+00:00
+    return file_ingestion_date_formatted
```
tech3371:
```suggestion
    return ingestion_date_object
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
```diff
@@ -182,7 +183,20 @@ def s3_event_handler(event):
...
+
+        # convert str to datetime with no hyphens
+        ingestion_date_no_hyphens: datetime = datetime.strptime(
+            ingestion_date_str, "%Y%m%d %H:%M:%S%z"
+        )
+
+        # add hyphens
+        ingestion_date_object = ingestion_date_no_hyphens.strftime(
+            "%Y-%m-%d %H:%M:%S%z"
+        )
```
tech3371:
```suggestion
        ingestion_date_object = get_file_ingestion_date(s3_filepath)
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
```diff
@@ -218,7 +232,20 @@ def s3_event_handler(event):
...
+            # convert str to datetime with no hyphens
+            ingestion_date_no_hyphens: datetime = datetime.strptime(
+                ingestion_date_str, "%Y%m%d %H:%M:%S%z"
+            )
+
+            # add hyphens
+            ingestion_date_object = ingestion_date_no_hyphens.strftime(
+                "%Y-%m-%d %H:%M:%S%z"
+            )
+
```
tech3371:
```suggestion
           ingestion_date_object = get_file_ingestion_date(s3_filepath)
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
tech3371: And you don't need to change it back to `"%Y-%m-%d %H:%M:%S%z"` if we want `"%Y%m%d %H:%M:%S%z"` going forward. Looks like that's what Maxine and you decided. That's why I am suggesting that we change the format to match when we write to the database. I made code suggestions to show what I mean.

tests/lambda_endpoints/test_query_api.py
tech3371: I hope my PR response helped with this. If not, let me know and I am happy to provide more clarifications through here or in call.

maxinelasp:
@tech3371 Could you summarize your changes in a comment? I'm having trouble keeping track. What is the format of the API call vs the format of the database? I think it's fine to leave the database as `%Y-%m-%d` if that's what we use for the file timestamp, even if it doesn't match the API call. The API call for file timestamps is `%Y%m%d` which was my main concern with the changes, making sure that if you're filtering by file timestamp and by ingestion time those things match. I think it's ok if that doesn't match the database format, since we won't be mixing those timestamps together.

tech3371:
> @tech3371 Could you summarize your changes in a comment? I'm having trouble keeping track. What is the format of the API call vs the format of the database? I think it's fine to leave the database as `%Y-%m-%d` if that's what we use for the file timestamp, even if it doesn't match the API call. The API call for file timestamps is `%Y%m%d` which was my main concern with the changes, making sure that if you're filtering by file timestamp and by ingestion time those things match. I think it's ok if that doesn't match the database format, since we won't be mixing those timestamps together.

Sorry @daralynnrhode . Please ignore my comments. If @maxinelasp is ok with not needing to change write to database with new date format, then please ignore my comments and don't hesitate to go back to what you had before.

daralynnrhode:
@tech3371  or @maxinelasp It's ready for review whenever one of you has the chance.

tests/lambda_endpoints/test_query_api.py
tech3371: Since this test setup function is writing to the database, you will want that in the same date format as this [function](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/dev/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py#L22) since this is the date format we write  to the database on file ingestion. 
```suggestion
            "2025-11-07 10:13:12+00:00", "%Y-%m-%d %H:%M:%S%z"
```

tech3371 (commented):
I just had one comment. Besides that, I will ask Maxine to review this and approve this since you both are more familiar with this code change. Thank you for working on this!

maxinelasp (approved):
Looks great! Thanks so much!!

PR #485, opened 2025-03-20
Update with example soft dependencies (MAG)

maxinelasp:
# Change Summary
Update with example soft dependencies for MAG so Luisa can use them in development.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
@@ -104,11 +104,27 @@ mag, l1a, norm-mago, mag, l1b, norm-mago, HARD, DOWNSTREAM
 mag, l1a, norm-magi, mag, l1b, norm-magi, HARD, DOWNSTREAM
 mag, l1a, burst-mago, mag, l1b, burst-mago, HARD, DOWNSTREAM
 mag, l1a, burst-magi, mag, l1b, burst-magi, HARD, DOWNSTREAM
+mag, ancillary, l1b-calibration, mag, l1b, all, HARD, DOWNSTREAM
 
+# For L1C, we want to trigger on either burst or norm or both.
 mag, l1b, norm-mago, mag, l1c, norm-mago, SOFT, DOWNSTREAM
 mag, l1b, burst-mago, mag, l1c, norm-mago, SOFT, DOWNSTREAM
 mag, l1b, norm-magi, mag, l1c, norm-magi, SOFT, DOWNSTREAM
 mag, l1b, burst-magi, mag, l1c, norm-magi, SOFT, DOWNSTREAM
+
+mag, l1c, norm-mago, mag, l1d, norm, HARD, DOWNSTREAM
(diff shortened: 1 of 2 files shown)


lacoak21 (approved):
Thanks Maxine!

maxinelasp:
@lacoak21 would appreciate a second look here as I added a few other examples. Still won't merge this until after you're done though!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
```diff
@@ -104,11 +104,27 @@ mag, l1a, norm-mago, mag, l1b, norm-mago, HARD, DOWNSTREAM
...
+
+mag, l1c, norm-mago, mag, l1d, norm, HARD, DOWNSTREAM
+mag, l1c, norm-magi, mag, l1d, norm, HARD, DOWNSTREAM
+mag, l1b, burst-mago, mag, l1d, burst, HARD, DOWNSTREAM
+mag, l1b, burst-magi, mag, l1d, burst, HARD, DOWNSTREAM
+mag, ancillary, l1d-calibration, mag, l1d, all, HARD, DOWNSTREAM
+
+# For level 2, we actually only want to trigger on the arrival of the offsets file.
+# This offsets file is tightly tied to the L1C and L1B files, so those files will be
+# retrieved out of the ancillary file and not through the dependency API.
```
lacoak21, line 123: Interesting! I am trying to figure out if this will cause a issue with the start date because I use the "primary science" upstream dependencies to get the start dates for the future jobs to run. In this case there isnt a "primary science" dependency in the way that others have because the ancillary offset is the acting like the "primary science".

lacoak21: I will give this more thought! It will require an update I think.

maxinelasp: Aha, interesting. If that breaks everything I can cheese it by putting a dependency on a science file in and handling it in processing (which I kind of expected to have to do anyway.) This is an actually crazy case so I would say don't put extra work in when I can just discard extra data. As long as we can trigger on the arrival of the ancillary file, it's ok if we trigger a few extra times after that and I can check if the output file has already been produced with the expected parent files.

maxinelasp: We can talk more about this case but I can basically guarantee no other instrument will have this problem lol

lacoak21: Ok sounds good! Haha classic mag. I think you're work idea around to add the science file as a dep would definitely work but maybe we can figure something better out.

tech3371:
Is this PR still relevant?

tech3371:
@maxinelasp can we close this?

lacoak21:
oops that was my bad. I think I was supposed to merge this in a while ago. Not sure if anything has changed.

maxinelasp:
No longer relevant - I confirmed all the inputs in my latest PR.

PR #489, opened 2025-03-24
MNT: Ran poetry export command to fix psycopg issue

tech3371:
# Change Summary

## Overview
Updated `imap-data-access` again and ran poetry export command. Seems to fix the error.

## Updated Files
- pyproject.toml
   - Added note about the poetry export command
- lambda_layer/python/requirements.txt
   - Running above command doesn't change anything in the lock or requirements file but somehow fixes the issue in infrastructure.


daralynnrhode (approved):
I took notes when Maxine and I did our deployment and had the same problem, but I probably thought to my self "thats a simple fix to a known problem. I don't need to write that down". And look at where we are now, lol

daralynnrhode:
I believe the simple `poetry export ` command is all that is needed to synch the files up.

tech3371:
> I took notes when Maxine and I did our deployment and had the same problem, but I probably thought to my self "thats a simple fix to a known problem. I don't need to write that down". And look at where we are now, lol

I couldn't agree more.

PR #490, opened 2025-03-25
I-ALiRT - Add IP address to security group for Kiel

laspsandoval:
# Change Summary

## Overview
This allows traffic into the security group from our Kiel partner.

## Updated Files
   - Allows traffic into the security group from our Kiel partner.

## Testing
We will test with Kiel once this is implemented.

### DIFF
--- a/sds_data_manager/constructs/ialirt_processing_construct.py
+++ b/sds_data_manager/constructs/ialirt_processing_construct.py
@@ -81,7 +81,7 @@ class IalirtProcessing(Construct):
 
         # Allow inbound and outbound traffic from a specific port and IP.
         # IPs: LASP IP, BlueNet (tlm relay)
-        ip_ranges = ["128.138.131.0/24", "198.118.1.14/32"]
+        ip_ranges = ["128.138.131.0/24", "198.118.1.14/32", "193.174.22.3/32"]
         for port in self.ports:
             for ip_range in ip_ranges:
                 self.ecs_security_group.add_ingress_rule(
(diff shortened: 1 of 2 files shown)


sds_data_manager/constructs/ialirt_processing_construct.py
```diff
@@ -81,7 +81,7 @@ def create_ecs_security_group(self):
 
         # Allow inbound and outbound traffic from a specific port and IP.
         # IPs: LASP IP, BlueNet (tlm relay)
-        ip_ranges = ["128.138.131.0/24", "198.118.1.14/32"]
+        ip_ranges = ["128.138.131.0/24", "198.118.1.14/32", "193.174.22.3/32"]
```
greglucas: I'd suggest adding comments inline to the IPs you're using. You have them above for the previous two, but it might be easier to bring these onto separate lines and add comments next to them to make it clear.
```suggestion
        ip_ranges = ["128.138.131.0/24",  # LASP IP
                              "198.118.1.14/32",  # BlueNet
                              "193.174.22.3/32"]  # Kiel
```

greglucas: I thought we were going to open up a single port per antenna partner? With this, Kiel could talk over a port that someone else was assigned to use, or if their computer has malware / gets compromised they could gain alternative access through another port or start spamming things on other channels.

laspsandoval: Yes. You have a point. Take a look at the change and make certain it looks ok.

greglucas (approved):
Nice, looks good to me! I think this makes sense, so as long as it works I think it will help with security. Thanks for looking into that and the quick fix!

PR #495, opened 2025-03-30
Spice indexing

bryan-harter:
# Change Summary

## Overview
Adding SPICE indexing to the SDS, in addition to moving the file to the proper location on the EFS. 

I'll be on vacation for 3 days after I submit this, so I'd appreciate a thorough review in the next few days so I can just hop into it when I return Thursday! If people are in a hurry for MSIM, someone is definitely welcome to take over the commit and make the changes. 

## New Dependencies
None

## New Files
   - NOT technically this is not a new file, but simply moving the efs_lambda.py file for a more accurate name. 
   - The new code occurs starting at "index_spice_file(file_path)" in the lambda handler. You can follow the logic from there to see what the new functionality is. 
- tests/lambda_endpoints/test_spice_files/imap_2025_118_2025_120_001.ah.bc
   - Test file
- tests/lambda_endpoints/test_spice_files/imapsclk_0012.tsc
   - Test file
- tests/lambda_endpoints/test_spice_files/imapsclk_0012.tsc
   - Test file

## Updated Files
   - changing "file_path" to "file_name" as the primary key. The path might be different on each machine. 

## Testing
   - Tests the new indexer lambda. The test mimics a leapsecond kernel being placed on the SDS, followed by a spacecraft clock kernel, and then an attitude file. Then we ensure the database contains the proper information about those files.


bryan-harter:
FYI, the unit tests are failing at the moment *on github* because they rely on the changes I made to imap-data-access here: 

https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/147

I can confirm that they all work locally, with these new changes, however.

pyproject.toml
```diff
@@ -68,8 +68,10 @@ requests = "^2.28.2"
 psycopg = "^3.1"
 psycopg2-binary = "^2.9.9"
 SQLAlchemy = "<=3.0.0"
-spiceypy = "^6.0.0"
 
+
+# Used for installing with pip
+spiceypy = "^6.0.0"
```
greglucas: I think this can be as it was before?

pyproject.toml
```diff
@@ -105,12 +107,12 @@ lint.select = ["B", "D", "E", "F", "I", "N", "S", "W", "PL", "PT", "UP", "RUF"]
 # PLR0913: Too many arguments in function definition
 # PLR2004: Magic value in comparison
 # RUF200: pyproject missing field (poetry doesn't follow the spec)
-lint.ignore = ["D104", "D203", "D213", "D300", "D413", "S101", "S104", "PLR0913", "PLR2004", "RUF200"]
+lint.ignore = ["D104", "D203", "D213", "D300", "D413", "S101", "S104", "S108", "PLR0913", "PLR2004", "RUF200"]
```
greglucas:
```suggestion
lint.ignore = ["D104", "D203", "D213", "D300", "D413", "S101", "S104", "PLR0913", "PLR2004", "RUF200"]
```

pyproject.toml
```diff
@@ -105,12 +107,12 @@ lint.select = ["B", "D", "E", "F", "I", "N", "S", "W", "PL", "PT", "UP", "RUF"]
...
 
 [tool.ruff.lint.per-file-ignores]
 # TODO: Too many statements, this could be refactored to separate
 #       it into a few smaller pieces
 "sds_data_manager/utils/stackbuilder.py" = ["PLR0915"]
 # subprocess calls within these modules are expected
 "tests/lambda_endpoints/conftest.py" = ["S"]
-"tests/infrastructure/test_website_cloudfront_function.py" = ["S"]
+"tests/infrastructure/test_website_cloudfront_function.py" = ["S"]
```
greglucas:
```suggestion
"tests/infrastructure/test_website_cloudfront_function.py" = ["S"]

```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,369 @@
...
+from ..database import database as db
+from ..database import models
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+# Define the paths
+SPACECRAFT_ID = -43
+minimum_mission_time = datetime(2023, 1, 1)
```
greglucas: Does this need to be 2010 for when SHCOARSE == 0?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,369 @@
...
+    spice_params["min_date_sclk"] = file_coverage_sclk[0][0]
+    spice_params["max_date_sclk"] = file_coverage_sclk[-1][-1]
+    spice_params["file_intervals_sclk"] = file_coverage_sclk
+    spice_params["lsk_kernel"] = str(latest_lsk)
+    spice_params["sclk_kernel"] = str(latest_sclk)
+
+    # Open db session and determine whether to update the database
+    # or insert a new record
+    with db.Session() as session, session.begin():
+        # Check if the record already exists
```
greglucas: Out of curiosity, can we use an `update_if_exists` somehow? That way this could be a one-liner and you'd just `session.add(models.SPICEFiles(**spice_params))` and let it work itself out.

https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#sqlalchemy.dialects.postgresql.Insert.on_conflict_do_update

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,369 @@
...
+    results_j2000 = []
+    results_sclk = []
+    results_datetime = []
+
+    if spice_file.suffix == ".bc":
+        coverage_function = spiceypy.ckcov
+    elif spice_file.suffix == ".bsp":
+        coverage_function = spiceypy.spkcov
+    else:
+        raise Exception(
```
```suggestion
        raise ValueError(
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,369 @@
...
+        )
+        if existing_entry:
+            for key, value in spice_params.items():
+                if key == "file_name":
+                    continue
+                setattr(existing_entry, key, value)  # Update existing record
+        else:
+            # Record does not exist, add it
+            session.add(models.SPICEFiles(**spice_params))
+    logger.info("Wrote data to the SPICEFiles table")
```
greglucas: log what you added for helpful future debugging?
```suggestion
    logger.info(f"Wrote {spice_params} to the SPICEFiles table")
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,369 @@
...
+        The full name and path the SPICE file to index
+    """
+    latest_lsk = None
+    latest_sclk = None
+    spice_object = SPICEFilePath(spice_file)
+    spice_metadata = SPICEFilePath(spice_file).spice_metadata
+    try:
+        latest_lsk = furnish_best_spice_file(spice_file.parent.parent / "lsk")
+        latest_sclk = furnish_best_spice_file(spice_file.parent.parent / "sclk")
+    except Exception as e:
```
greglucas: This is a bare Exception catch, that I suggest changing to a ValueError in the other location.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,369 @@
...
+        if spice_metadata["type"] in ("leapseconds", "spacecraft_clock"):
+            # This block will likely only be reached if this is the very first
+            # leapsecond or spacecraft_clock kernel placed on the SDS. In this case,
+            # we'll insert default data and continue.
```
greglucas, line 211: Shouldn't this be minimum datetime intervals then so we don't infer anything? Otherwise we'll think these kernels are valid for the entire time period if we were to request data for say 2100 on accident I think.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,369 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+# Define the paths
+SPACECRAFT_ID = -43
+minimum_mission_time = datetime(2023, 1, 1)
+maximum_mission_time = datetime(2145, 1, 1)
+MAXIMUM_DATETIME_INTERVAL = [[minimum_mission_time, maximum_mission_time]]
```
greglucas, line 24: Why are all of these doubly nested? Can you have these be a tuple of (start, end) and then later add these tuples to a list if you get a list of files.

tests/lambda_endpoints/conftest.py
```diff
@@ -27,6 +27,7 @@ def _set_env(monkeypatch):
     # Mock the api gateway url
     # This is used in batch_starter.py
     monkeypatch.setenv("IMAP_DATA_ACCESS_URL", "https://test.url")
+    monkeypatch.setenv("EFS_SPICE_MOUNT_PATH", "/tmp")
```
greglucas: This should be a temporary directory in tests (bring in the fixture above).
```suggestion
    monkeypatch.setenv("EFS_SPICE_MOUNT_PATH", str(tmpdir))
```

tests/lambda_endpoints/test_spice_files/imapsclk_0012.tsc
greglucas, line 1: We are bringing these around in a few different repositories. Is it already available somewhere that we could reference so we are all referencing the same file in imap_processing, this etc...

tests/lambda_endpoints/test_spice_indexer_lambda.py
```diff
@@ -0,0 +1,141 @@
+"""Tests for the SPICE indexer lambda."""
+
+import os
+from datetime import datetime
+
+from sds_data_manager.lambda_code.SDSCode.database import models
+from sds_data_manager.lambda_code.SDSCode.pipeline_lambdas import spice_indexer
+
+
+def put_local_file_in_bucket(s3_client, path_in_s3, path_local):
```
greglucas, line 10: Are we able to use the imap-data-access upload/download routines instead of this explicit put?

tests/lambda_endpoints/test_spice_indexer_lambda.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    current_path = os.path.dirname(os.path.abspath(__file__))
+
+    # Insert leapsecond spice kernel
+    leapsecond_event = put_local_file_in_bucket(
+        s3_client,
+        "spice/lsk/naif0012.tls",
+        current_path + "/test_spice_files/naif0012.tls",
+    )
+    spice_indexer.lambda_handler(leapsecond_event, None)
```
greglucas, line 70: Do we need to add a trigger for this case still? I figured this would automatically be handled and not need an explicit kickoff. Can we add a ticket for this if that is still work to be done?

tests/lambda_endpoints/test_spice_indexer_lambda.py
```diff
@@ -0,0 +1,141 @@
...
+
+    # Verify the database has 3 files:
+    assert len(result) == 3
+
+    # Loop through the 3 database entries and ensure their accuracy
+    leapsecond_kernel_reached = False
+    clock_kernel_reached = False
+    attitude_kernel_reached = False
+    for r in result:
+        print(r.file_name)
```
greglucas:
```suggestion
```

tests/lambda_endpoints/test_spice_indexer_lambda.py
```diff
@@ -0,0 +1,141 @@
...
+    result = session.query(models.SPICEFiles).all()
+
+    # Verify the database has 3 files:
+    assert len(result) == 3
+
+    # Loop through the 3 database entries and ensure their accuracy
+    leapsecond_kernel_reached = False
+    clock_kernel_reached = False
+    attitude_kernel_reached = False
+    for r in result:
```
greglucas: Rather than looping through, we should try to query for an individual file and assert on that. Do 3 separate queries here on your expectations.

greglucas (commented):
One question/comment that I think would help consistency across the project: Can we use `IMAP_DATA_DIR = /mnt` on the EFS, and then let imap-data-access and your recent SPICEFiles directory mapping take care of the EFS file placement from there. Then we can also use imap_data_access.download() routines to automatically populate that structure. Then we don't have a separate EFS_MOUNT_POINT environment variable and other directory remappings.

pyproject.toml
greglucas: I pushed changes for these updates in pyproject.toml

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
bryan-harter: Oh sure, yeah I just picked a random one. It's only used within this file, and only to ensure something is filled in in the database. I'm not necessarily sure if it matters as long as its within a time range we expect to query (or if we want to be consistent with other time=0 elsewhere in the code)

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
bryan-harter: Sure we can try that out, I was just copying the approach from the file indexer

tmplummer (commented):
Everything looks good to me from a SPICE, not infrastructure or DB, perspective.

tests/lambda_endpoints/test_spice_indexer_lambda.py
bryan-harter: Would those work for mocked buckets? This is just to upload the local testing files into the mocked s3 bucket for unit testing

tests/lambda_endpoints/test_spice_indexer_lambda.py
bryan-harter: Do you have it set up so that the mocked infrastructure triggers the appropriate lambdas upon an upload? If so, I completely missed that!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
bryan-harter: The intervals are stored in the database as a list of lists like this, for the attitude and ephem files they are usually more complex [[date1, date2], [date3, date4], [date5, date6]...etc]. But just to keep the same structure, default intervals would be kept in this format of [[min_date, max_date]]

For the kinds of SPICE files that are essentially "timeless" because they don't have a date associated with them (like frames kernels, leapseconds kernels, planetary constants, etc), in the database their value for the columns "file_intervals_j2000/datetime/sclk" will always show up as this [[min_date, max_date]]

That way, when we make a query to the future metakernel API, something like "select spice_files where time>{date1} and time< {date2}", all these "timeless" kernels show up in the results as well. 

We could change it so that the times for these kinds of files have "None" in those columns, and just have the metakernel API add in these timeless kernels as needed. But I figure this way makes sense as well, because technically files like the instrument kernels are valid for those time ranges, they're valid for any time range.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
bryan-harter: Ok I implemented it in the latest version

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
bryan-harter: That was actually the behavior I was going for, technically the leapseconds and spacecraft clock kernels are valid out until very far into the future. 

They don't really have a set end date, its just that a new version of them will probably appear at some point (except I heard we are actually done with leapseconds, so maybe not that one?)

tests/lambda_endpoints/test_spice_indexer_lambda.py
bryan-harter: True, I was also thinking this would help show that no more than those 3 were inserted, but I guess we don't need to test that

bryan-harter:
Looks like the unit tests are working now, and I believe I addressed all of greg's comments

bryan-harter:
Ping!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
greglucas: Some of the test data we've gotten has SHCOARSE starting from 0 and other 2019 data and such, so I'd suggest setting this to (2009, 12, 31) or similar to bracket the entire time we would get data. Unless we are actually trying to weed those times out as well and then I'd switch this to 9/23/2025 as the actual start of potential data.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
greglucas: Ahh, that makes sense, thanks for clarifying! I didn't realize this was also to account for various gaps in files where there may be multiple versions. This makes sense.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,386 @@
...
+
+
+# Define constants needed in the file
+SPACECRAFT_ID = -43
+minimum_mission_time = datetime(2023, 1, 1)
+maximum_mission_time = datetime(2145, 1, 1)
+MAXIMUM_DATETIME_INTERVAL = [[minimum_mission_time, maximum_mission_time]]
+MAXIMUM_SCLK_INTERVAL = [
+    ["1/0410227203:00000", "1/4288750963:38093"]
+]  # Calculated from the above datetimes seperately
```
greglucas, line 27: Ahh, I actually wonder if we shouldn't say the SCLK_INTERVAL should be 0 to a very large number, so make this your decision point and calculate the others from this. I'm a bit confused what the `/` and `:` are here, I thought SCLK was just an incrementing uint counter?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,386 @@
...
+MAXIMUM_J2000_INTERVAL = [
+    [725803269.1839136, 4575787269.183866]
+]  # Calculated from the above datetimes seperately
+
+# Set constants for the time interval calculations
+COVERAGE_ANGULAR_VELOCITY_ONLY = False  # Only include segments with angular velocity?
+COVERAGE_SPICE_ARRAY_LENGTH = 10000  # Use an array size of 10000 for coverage calc
+COVERAGE_LEVEL = "INTERVAL"  # the granularity at which the coverage is examined
+COVERAGE_TOLERANCE = 1000000.0  # tolerance value expressed in ticks of the spacecraft
```
greglucas, line 36: Is this roughly 20 seconds? Assuming 20 microseconds / tick which I think I've seen elsewhere. I have no idea what a reasonable number is, but as an FYI that is longer than a spin which is roughly 15-seconds if that matters.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,386 @@
...
+        "lsk_kernel": str(latest_lsk),
+        "sclk_kernel": str(latest_sclk),
+    }
+
+    with db.Session() as session:
+        # Execute the statement as a single "insert-or-update" operation
+        stmt = (
+            insert(models.SPICEFiles)
+            .values(**spice_params)
+            .on_conflict_do_update(
```
greglucas, line 178: Very nice! I was wondering if this was possible and this seems like a pretty straightforward way to handle it.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,386 @@
...
+                index_elements=["file_name"],  # or name of a unique constraint
+                set_={  # Remove the "file_name" from the update dict
```
greglucas, line 180: Since it is unique and you've "found" this entry, does it matter if we update it or not here? i.e. could you just `set_=spice_params`

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
greglucas: OK, if this first file is the file we want to extrapolate with then 👍 , I just wasn't sure if we wanted to error out if we actually were beyond the valid data interval or something like that as a warning.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,386 @@
...
+        spice_object,
+        file_coverage_j2000,
+        file_coverage_datetime,
+        file_coverage_sclk,
+        latest_lsk,
+        latest_sclk,
+    )
+
+
+def create_symlink(source_path: Path, destination_path: Path) -> None:
```
greglucas: I think this function is now unused now, can you remove it?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -0,0 +1,386 @@
...
+        The local location of the SPICE file
+
+    """
+    # Create an S3 client
+    s3_client = boto3.client("s3")
+
+    # Remove 'spice/' prefix from the s3 key. See key example below.
+    #   Eg. spice/spin/imap_2025_122_2025_122_02.spin.csv
+    # Keep remaining folder path after `spice/` to match the folder structure
+    # defined in imap-data-access library.
```
greglucas: I'm not sure I follow this portion. I thought our file structure was supposed to be the same in s3 as on disk?

Can you set `imap_data_access.config["DATA_DIR"] = spice_mount_path.parent` and then create the `imap_data_access.SPICEFilePath(os.path.basename(s3_key))` which I think would fill out the proper directory structures for you to be consistent everywhere?

tests/lambda_endpoints/conftest.py
```diff
@@ -33,6 +33,7 @@ def _set_env(monkeypatch):
     # Mock the api gateway url
     # This is used in batch_starter.py
     monkeypatch.setenv("IMAP_DATA_ACCESS_URL", "https://test.url")
+    monkeypatch.setenv("EFS_SPICE_MOUNT_PATH", str(tmpdir))
```
greglucas, line 36: IMAP_DATA_DIR of which the SPICEFilePath is relative to? I think this can be a follow-up PR now that I'm reading it closer. But it would be nice to not have these mutliple different ways to point to different paths.

tests/lambda_endpoints/test_spice_indexer_lambda.py
greglucas: Ahh, I didn't look closely enough that this was for the testing. Good call!

tests/lambda_endpoints/test_spice_indexer_lambda.py
greglucas: I am again not paying enough attention to this being a test file :(
It would be great if we did have this all set up like that! I don't think we do, but am not positive either.

greglucas (approved):
Thanks for the ping again. Looks great Bryan! Sorry for the slow review here. I have a few more minor comments, but approving now so you can address those without waiting on me. It is in a good state in my opinion and my comments are mostly nits.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
tmplummer: @greglucas 
These are what SPICE calls SCLK strings and relate directly to how the SCLK kernel defines the clock. Basically, they are `"<partition_number>/<SHCOARSE>/<SHFINE>"`.

greglucas: Makes sense, thanks for clarifying!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
bryan-harter: Yeah we can definitely fiddle with this number later, the problem was I was taking the test data we received and set COVERAGE_TOLERANCE to "100000" (an order of magnitude less than the number above), and the number of gaps was HUGE! Like 100+ per file. I doubt there really was that much missing data. Only when I set it this high did it only spit out the single 40 minute gap in the test data, which I assume was an actual data gap. 

We might need to experiment when we get to processing with the files later what kind of gap size is tolerable, but for now this should cover the major gaps.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
bryan-harter: Yup! I was wondering about that

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
bryan-harter: Yeah I tried that initially and it still gave an error unfortunately, it still doesn't like you trying to modify the file_name field (even though it is the same)

PR #496, opened 2025-03-31
MNT/FIX: Only build Lambda Layers and use Lambdas with x86 architecture

greglucas:
This was causing issues where people with Mac / ARM processors would build a different layer than on GitHub Actions CI on a Linux x86 system. We can specify the platform for the Layer explicitly and indicate what it is compatible with. Then we can remove all ARM64 architectures for the Lambda's we were specifying which aren't needed anymore if we build the layer as an x86.

### DIFF
--- a/sds_data_manager/constructs/lambda_layer_construct.py
+++ b/sds_data_manager/constructs/lambda_layer_construct.py
@@ -39,6 +39,14 @@ class IMAPLambdaLayer(lambda_.LayerVersion):
             layer_dependencies_dir,
             bundling=cdk.BundlingOptions(
                 image=runtime.bundling_image,
+                # NOTE: We need to use the x86_64 architecture for the lambda layer
+                #      otherwise people with mac's will produce different shared object
+                #      files from those produced on our CI runners.
+                #      To debug you can look at the psycopg2 linked binaries in the
+                #      assets by adding the following to the commands below:
+                #      ls -al /asset-output/python/psycopg2
+                platform="linux/amd64",
+                environment={"DOCKER_DEFAULT_PLATFORM": "linux/amd64"},
                 command=[
                     "bash",
                     "-c",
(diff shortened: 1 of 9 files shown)


sds_data_manager/constructs/lambda_layer_construct.py
```diff
@@ -39,6 +39,14 @@ def __init__(
             layer_dependencies_dir,
             bundling=cdk.BundlingOptions(
                 image=runtime.bundling_image,
+                # NOTE: We need to use the x86_64 architecture for the lambda layer
+                #      otherwise people with mac's will produce different shared object
+                #      files from those produced on our CI runners.
+                #      To debug you can look at the psycopg2 linked binaries in the
+                #      assets by adding the following to the commands below:
+                #      ls -al /asset-output/python/psycopg2
+                platform="linux/amd64",
```
lacoak21, line 48: Out of curiosity why do we need to set the platform and the env variable? I would think setting the platform would be enough.

lacoak21 (approved):

sds_data_manager/constructs/lambda_layer_construct.py
```diff
@@ -55,5 +63,6 @@ def __init__(
             id=f"{id}-Layer",
             code=code_bundle,
             compatible_runtimes=[runtime],
+            compatible_architectures=[lambda_.Architecture.X86_64],
```
tech3371, line 66: I am trying to make sure I understood this correctly. In this PR, we are changing to use `x86_64` architecture now for all lambdas by not passing `architecture` parameter and let it use default which is `Architecture.X86_64`. Then you are building layer with `linux/amd64`. Now, here you are saying compatible architecture is `x86_64`. Will lambda layer result in mismatch? I have not tried to deploy this yet locally, but I think it may fail on newer mac. I don't know yet.

tech3371: I learned that docker `linux/amd64` is same as AWS `Architecture.X86_64` in the call.

tech3371 (approved):

PR #500, opened 2025-04-01
I-ALiRT - root certificate

### LINKED ISSUE #498: BUG - I-ALiRT <-> SDC Stack linkage
### Description of the issue

There is one remaining export from the SDCStack that the I-ALiRT Stack references. It would be good to break this link if we can.

Currently, our root_certificate is made in our sdc_stack
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/10abeb661d94b23617034f643efec8d887ed3cf2/sds_data_manager/utils/stackbuilder.py#L94-L95

this is then referenced in the I-ALiRT APIGateway Stack
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/10abeb661d94b23617034f643efec8d887ed3cf2/sds_data_manager/utils/stackbuilder.py#L309-L315

I think I-ALiRT could create a new certificate with the same content but that it owns, or change it to a `i-alirt.{domain}` specific version instead.

_No response_

_No response_

_No response_

### Code Snippet:

```python
Code
```

_No response_

laspsandoval:
# Change Summary

## Overview
Removed dependency on sds stack by creating ialirt's own root certificate

- stackbuilder.py
   - Added ialirt root certificate 

## Testing
Deployed to AWS.

### DIFF
--- a/sds_data_manager/utils/stackbuilder.py
+++ b/sds_data_manager/utils/stackbuilder.py
@@ -283,6 +283,18 @@ def build_sds(
 
     ialirt_stack = Stack(scope, "IalirtStack", cross_region_references=True, env=env)
 
+    ialirt_root_certificate = None
+    if domain is not None:
+        ialirt_root_certificate = acm.Certificate(
+            ialirt_stack,
+            "IAlirtDomainRegionCertificate",
+            domain_name=f"*.{domain_name}",  # *.imap-mission.com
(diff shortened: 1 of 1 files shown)


greglucas (approved):
Nice if this works! That would be great to break that link and only have I-ALiRT depend on the networking stack.

PR #501, opened 2025-04-01
Soft Dependency Handling

lacoak21:
# Change Summary
Add soft dependency handling to batch starter and dependency lambda. 

## Overview
Batch starter now checks for downstream soft dependencies and potentially kicks off soft dependency jobs. 

Maxine brought up the case of Mag l1c: 

" MAG L1C would preferably take in two L1B input files, norm and burst. However, it is possible that for a given day only norm mode or only burst mode data is available, in which case they will only use one file"

In this case, batch starter will check the processing table for every soft dependency and see if it is currently "INPROGRESS" If there is a soft dependency in progress, then batch starter will return and let that file kick off batch starter again when it completes processing and arrives in s3. This is default soft dependency behavior and only works for upstream primary science dependencies e.g. "mag" and "mag", since an upstream non primary dependency might have a different version and start_date. 

## Updated Files
   - Checks upstream and downstream soft dependencies
   - Can look for upstream or downstream soft dependencies

## Testing
Add Mag l1c example tests to make sure batch starter is kicking off jobs we expect.


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -139,23 +144,30 @@ def is_job_in_processing_table(
...
     bool
         True if a duplicate job is found, False otherwise.
     """
+    # If in_progress_only is True, only check for jobs that are currently processing
+    if in_progress_only:
+        status_clause = models.ProcessingJob.status == models.Status.INPROGRESS.value
+    else:
+        status_clause = models.ProcessingJob.status.in_(
+            [models.Status.INPROGRESS.value, models.Status.SUCCEEDED.value]
+        )
```
tech3371: I am wondering if we need this if statement because wouldn't the second query result in true if a job was in process. Eg. if something is processing, then that `in_()` would return the job that's in progress. It wouldn't return the succeeded one. Now if something has finished processing, then it would return the succeeded one this time. In any case, we should expect one return value for this, either in progress or succeeded one. That's because, we update the same row with either in progress or succeeded status. I think that's why whether it's a soft or hard dependency, I think what you had before should work, I think. What do you think?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -327,16 +341,44 @@ def submit_all_jobs(
...
+
+    # If soft nor hard dependencies are found, return.
+    if not upstream_dependencies:
         logger.info(f"Upstream dependency not found for: {dependency_event_msg}")
-        return  # Exit the loop early
+        return
+
+    logger.info(f"All required dependencies found for the job: {job}")
+    # Get soft dependencies if they are available
+    base_event_msg["relationship"] = "SOFT"
```
tech3371: I see. Can you change above `upstream_dependencies` to be `all_hard_upstream_dependenices`. For a second, I assumed that we should have both hard and soft dependencies in that.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -327,16 +341,44 @@ def submit_all_jobs(
...
+    all_soft_upstream_dependencies = _get_dependencies(base_event_msg)
+    # Check if any soft dependencies are currently processing.
+    # If a job is in progress, then return and let that file trigger batch starter
+    # When it is done processing.
```
tech3371: I personally feel little concern about this approach. What if job failed, then it won't kick of the waiting job. I personally think it defeat the purpose of soft dependencies. If soft dependencies doesn't exist, it should kick off processing with what's available. If that job finished and soft dependencies is available, should we trigger processing and then version bump? We should ask instrument if that's what they want or not.

tech3371 (commented):
I made some initial comments with some questions.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: This boolean was added for SOFT dependencies. Based of the MAG l1c situation described in the overview, I think here we want to check ONLY if the job is currently processing to see if we should wait for the SOFT dep. If one of the two upstream l1b files triggers processing it is expected that the other l1b job is in progress.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Your point about the job failing is a really good one and a concern here for sure. Maxine said she is planning on reviewing this PR when she has time, and maybe has thought about this situation. It might be worth it like you said to let the second SOFT dep trigger anther job with a version bump so we dont have the case where nothing ends up processing and it falls through the cracks. Right now im currently not handling any version bumping and the code would see that the mag l1c job already ran and would not kick off any processing.

Ive been reading through this [reprocessing spike](https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/221740849/Reprocessing+Spike) after menlo brought up many reprocessing questions this morning. Maybe we should have a follow up meeting (SDC)? I have a lot of questions 😅

tech3371: Totally! me too. Let's tag-up when Maxine in back.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
```diff
@@ -90,10 +90,10 @@ mag, l1a, norm-magi, mag, l1b, norm-magi, HARD, DOWNSTREAM
 mag, l1a, burst-mago, mag, l1b, burst-mago, HARD, DOWNSTREAM
 mag, l1a, burst-magi, mag, l1b, burst-magi, HARD, DOWNSTREAM
 
-mag, l1b, norm-mago, mag, l1c, norm-mago, HARD, DOWNSTREAM
-mag, l1b, burst-mago, mag, l1c, norm-mago, HARD, DOWNSTREAM
-mag, l1b, norm-magi, mag, l1c, norm-magi, HARD, DOWNSTREAM
-mag, l1b, burst-magi, mag, l1c, norm-magi, HARD, DOWNSTREAM
+mag, l1b, norm-mago, mag, l1c, norm-mago, SOFT, DOWNSTREAM
```
maxinelasp: If it would be useful to merge my dependency_config PR, go ahead and do that whenever. I just didn't want to break things before we were ready.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
maxinelasp: +1

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
maxinelasp: Hmm this is a good point. I agree on the tag-up. Let me think about this one. The problem I am trying to address here, is that most of the time, we would expect both files to be available. So if we trigger on both, we always result in an immediate reprocess (or actually, I think due to our checks, it wouldn't even trigger the second time, because the first time would be in-progress and currently we assume that means we shouldn't run again.) However, I suspect we can fix this with queues....

maxinelasp: Maybe in if the other dependency is in the processing table, we can pass both into processing and then the processing can try to retrieve both and fail gracefully. That way, we can indicate that we expect both to be available, but due to the processing queues on each instrument we are guaranteed to finish all L1B processing before L1C processing. That would be nice and safe, but it does assume that we have those per-instrument queues to enforce that requirement. However, I think regardless, for a given day we always want to enforce that processing error within instruments.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -327,16 +341,44 @@ def submit_all_jobs(
...
+            datetime.strptime(start_date, "%Y%m%d"),
+            version,
+            in_progress_only=True,
+        ):
+            logger.info(
+                f"Soft dependency {dep} is currently processing. Job will be"
+                " triggered when dependency processing completes and file is "
+                "available in S3."
+            )
+            return
```
maxinelasp: So basically, to handle the issue brought up by Tenzin, we would include the in-progress soft dependency within the dependencies returned here, and any processing code expecting that file should have some failure path if that file is actually not available. (I think this should be the case regardless, since the soft dependencies are not required, so if they aren't where you expect then we can continue onwards anyway. However, this is a division of responsibilities and adds some risks around future changes.)

maxinelasp (commented):
Overall looks good for a first pass, I agree with Tenzin's concerns about failed jobs, which I did not consider... I added some comments to the discussion, we can meet next week and talk out loud about this.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -433,10 +478,14 @@ def lambda_handler(events: dict, context):
...
+        dependency_event_msg["relationship"] = "SOFT"
+        potential_soft_jobs = _get_dependencies(dependency_event_msg)
+
+        if not potential_jobs and not potential_soft_jobs:
             logger.info(f"Found no dependencies for {dependency_event_msg}.")
             return
 
         with db.Session() as session:
-            for job in potential_jobs:
+            for job in potential_jobs + potential_soft_jobs:
```
lacoak21, line 473: Tenzin brought up a good point that SOFT dependencies should not trigger processing.

maxinelasp: I need soft dependencies to trigger processing. Or we need some other kind of class of things that is not strictly required, but does trigger processing. I actually think in most cases, we do want to trigger processing for soft dependencies.

maxinelasp: Unfortunately, that means we are definitely running into the versioning and reprocessing issues now.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Coming back to this after our Tuesday discussion. 

To clarify (or just reword for my own sanity check) when we get the first trigger for either norm or burst we SHOULD have both files due to the processing queue so we pass in both. We try to retrieve both files and if it cant find the file that was inprogress, we know that this is because the job that was processing that file failed so we continue processing with just the original file that triggered processing in the first place. If it succeeds with both files and then the second file triggers, it would not get reprocessed in this case because the CRID would be the same.

lacoak21: Another potential option would be to pass in the job-id of the job in progress into the batch starter command as a "depends on" instead of setting up queues unless there are other usages for that. 

The new command would look like: 

  BATCH_CLIENT.submit_job(
        jobName=job_name,
        jobQueue=job_queue,
        jobDefinition=job_definition,
        containerOverrides={
            "command": batch_command,
        },
        depends_on=[example_soft_trigger_id]
    )

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
lacoak21: I was planning on merging that right before I merge this!

lacoak21:
Ok @tech3371 and @maxinelasp this is ready for another review whenever you have a chance! As this stage, there is nothing stopping something from reprocessing.

In the mag l1c case, It would in fact process twice most of the time due to the 2 l1b files triggering one after another with the second getting a version bump. This will be fixed in another PR that will (as of now) implement proposed fix 2 as documented in [this page](https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/256944836/Handling+Soft+Triggering+Dependencies)

There will also be reprocessing checks when the CRID is implemented.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -147,11 +147,15 @@ class Relationship:
 
     Valid data relationships are:
         1. HARD - the data is required for the pipeline to run
-        2. SOFT - the data is optional for the pipeline to run
+        2. SOFT_TRIGGER - the data is optional for the pipeline to run. A new file will
+            trigger processing.
+        2. SOFT_NO_TRIGGER - the data is optional for the pipeline to run. A new file
```
tech3371:
```suggestion
        3. SOFT_NO_TRIGGER - the data is optional for the pipeline to run. A new file
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -100,6 +99,10 @@ def _get_dependencies(dependency_events: dict):
         # Retrieve the response as a list of dictionaries containing the dependency
         # information
         dependency_response = response.read().decode("utf-8")
+        # Check for a 206 status code
+        if response.status == 206:
+            logger.info(f"Partial content received: {dependency_response}")
```
tech3371: Does this log mean that some dependencies are missing? If so, can you rephrase this log?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -137,39 +155,39 @@ def is_job_in_processing_table(
...
-    bool
-        True if a duplicate job is found, False otherwise.
+     str
+        The highest version number bumped by 1 if the file has been already processed,
+        otherwise "v000".
     """
-    # check in the processing table if the job is already in progress
-    # for this instrument, data level, version, and descriptor
-    query = select(models.ProcessingJob.__table__).where(
+    # TODO should I be making a separate call to the scienceFilesTable for the
```
tech3371, line 165: Good question. I think we should leave previous `is_job_in_processing_table()` function as it. Then create another function for this `determine_job_version()` and call `ScienceFiles` table as you are thinking. Because I am little concerns with combining these two together. For `is_job_in_processing_table()` function, it makes sense to check for job in progress or succeeded but for version bump, I feel like we should be definite that job was completed successfully before we kick off another job. Therefore, it makes sense for `determine_job_version` function to make calls to science files table.

tech3371 (commented):
I am leaving some comments but I will look more closely tomorrow. I would highly recommend getting feedback from @maxinelasp since you both worked through this new scenario.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Yes the dependency response will be "At least one dependency is missing." in this case. so the full log would look like:

"Partial content received: At least one dependency is missing." 

Would it be better to reword it to 

"Dependency api response: At least one dependency is missing."

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: > "For version bump, I feel like we should be definite that job was completed successfully before we kick off another job. Therefore, it makes sense for `determine_job_version` function to make calls to science files table."

I am a little confused on this part because I thought we wouldn't want to hold off processing if a job is "in progress" in case that job fails. 

The only issue I had with this though, is what if V002 is in progress and when we go look at the science files for the version to bump, it would return v001 and we kick of another job with v002. That job would fail if the first job succeeded because it would try to upload a duplicate file. There might be a case where this is NOT actually a duplicate run but it would look like a duplicate from the outside. (MAG l1c case! Hopefully we can solve this in the future with the deduplication queue or something similar.)

On the other hand, if we bump the version that is in the processing table, we might have a gap in versions but at least we have that newer file.

lacoak21: I might be making this more complicated than it needs to be because the CRID check should stop a lot of these duplicate scenarios.

tech3371: same. I may be making it more complicated too. This case would happen when there is a race, is that right? If so, I think SQS with FIFO was added to reduce that part of complication. So that we know everything is coming in some kind of sequential order.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -202,22 +218,25 @@ def try_to_submit_job(
...
+    version = determine_job_version(
         session=session,
         instrument=instrument,
         descriptor=descriptor,
         start_date=start_date,
-        version=version,
         data_level=data_level,
-    ):
+    )
+    if version == "v000":
```
maxinelasp: Are we counting up from 000 or from 001?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: ** CRID check would not fix this.

maxinelasp (approved):
Looks good! Thanks for all the iteration!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Oh good point. I think we have been doing "v001" ill make that update.

PR #503, opened 2025-04-02
1564 hi dependencies updates

tmplummer:
# Change Summary

## Overview
Updating Hi dependencies

## Updated Files
   - Change l1b HK dependency to L0
   - Remove downstream dependency from l1a histogram. There will be some quicklooks added eventually, but no L1B.
   - Add 90sensor dependency entries

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
@@ -43,8 +43,6 @@ hi, l1b, 45sensor-de, hi, l1c, 45sensor-pset, HARD, DOWNSTREAM
 hi, l1a, 90sensor-de, hi, l1b, 90sensor-de, HARD, DOWNSTREAM
 hi, l1b, 90sensor-de, hi, l1c, 90sensor-pset, HARD, DOWNSTREAM
 
-# TODO: add IMAP-Hi 90 sensor data products
-
 # <---- HIT Dependencies ---->
 
 hit, l0, raw, hit, l1a, all, HARD, DOWNSTREAM
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
@@ -35,11 +35,13 @@ glows, l2, hist, glows, l3a, hist, HARD, DOWNSTREAM
 # <---- HI Dependencies ---->
 
 hi, l0, raw, hi, l1a, all, HARD, DOWNSTREAM
+hi, l0, raw, hi, l1b, hk, HARD, DOWNSTREAM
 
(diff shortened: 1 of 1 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
```diff
@@ -35,13 +35,13 @@ glows, l2, hist, glows, l3a, hist, HARD, DOWNSTREAM
 # <---- HI Dependencies ---->
 
 hi, l0, raw, hi, l1a, all, HARD, DOWNSTREAM
+hi, l0, raw, hi, l1b, hk, HARD, DOWNSTREAM
```
tech3371, line 38: @greglucas had suggestions for handling this differently. Greg did something like this [here](https://github.com/IMAP-Science-Operations-Center/imap_processing/blob/dev/imap_processing/spacecraft/quaternions.py#L75). I want to mention it in case you like to do that approach. I am ok with either approach.

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
greglucas: Thanks for the reminder that I needed to add a ticket to update this too!

@subagonsouth I am doing both l1a and l1b within the same processing step, so only going to add one job here that is dependent on l0 packets coming in and that job will produce two output datasets. (your case you could use the same packet file with `use_derived_value=True/False` separately or something like that if you'd prefer. It also doesn't matter to me which way this goes for everyone.

tmplummer: I don't understand how that code is related to this dependency change.

greglucas: It isn't directly. I think @tech3371 was insinuating that you could leave this line out if you would prefer to do both l1a and l1b at the same time on the processing side of the fence.

tmplummer: Oh, I see. That does make the changes in the `Hi.do_processing()` function a bit cleaner.

The thing that is really bothering me about this is that the packet decom has to be done twice either way I do it. I would really love to be able to get the raw and derived values out of `packet_file_to_datasets`.

greglucas: > The thing that is really bothering me about this is that the packet decom has to be done twice either way I do it. I would really love to be able to get the raw and derived values out of packet_file_to_datasets.

I completely agree. I would suggest tackling that later though as I think there are some other design decisions around that function like would it return a tuple of two datasets `(raw, derived)` or some other way of combining things into a single dataset.

The other issue I've seen is that instrument teams seem to produce their validation data as only the analog conversion values, but not the Enum/State variables. So we produce files with text "ON" / "OFF" for instance in our derived datasets, but they actually generally keep those as integers "0" / "1" instead, so there is a mix of how much derivation people want 😂

PR #507, opened 2025-04-02
Database Bump and Protections Added

laspsandoval:
# Change Summary

## Overview
Added protection for RDS and S3. Version bump for RDS.

## Updated Files
- stackbuilder.py
   - Version pump for RDS
- database_construct.py
   - Prevents deletion of database unless delete protection is changed
- data_bucket_construct.py
   - Prevents deletion if files in s3

### DIFF
--- a/sds_data_manager/utils/stackbuilder.py
+++ b/sds_data_manager/utils/stackbuilder.py
@@ -161,7 +161,7 @@ def build_sds(
         scope=sdc_stack,
         construct_id="RDS",
         vpc=networking.vpc,
-        engine_version=rds.PostgresEngineVersion.VER_15_6,
+        engine_version=rds.PostgresEngineVersion.VER_15_7,
         instance_size=ec2.InstanceSize[rds_size],
         instance_class=ec2.InstanceClass[rds_class],
         max_allocated_storage=rds_storage,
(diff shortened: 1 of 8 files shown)


tech3371 (approved):
Looks like tests was failing otherwise, it looks good. Thank you for fixing that!

sds_data_manager/utils/stackbuilder.py
```diff
@@ -161,7 +161,7 @@ def build_sds(
         scope=sdc_stack,
         construct_id="RDS",
         vpc=networking.vpc,
-        engine_version=rds.PostgresEngineVersion.VER_15_6,
+        engine_version=rds.PostgresEngineVersion.VER_15_7,
```
greglucas, line 164: Is there a reason to only bump this one minor version rather than go to the latest (do we fail with higher versions)? Otherwise, I think we are going to be in the same situation in another couple of months.

greglucas:
It looks like this didn't deploy from this merge point.
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/actions/runs/14232549374/job/39886934910

We should note down all of the manual steps involved or what things need human intervention to make sure we can fully deploy this automatically in the future. Could @laspsandoval, @tech3371, and @lacoak21 work together today to try and take notes of these things.

PR #508, opened 2025-04-02
I-ALiRT - Lambda layers

laspsandoval:
# Change Summary

## Overview
Separated the lambda layers as dependent on the type.

## Updated Files
- stackbuilder.py
   - updated filepath
- .pre-commit-config.yaml
   - Updated layer names
- /database/requirements.txt
   - Requirements for this layer
- /python/requirements.txt
   - Requirements for this layer
- /sdc/requirements.txt
   - Requirements for this layer
- pyproject.toml
   - Added groups and assigned tools
- poetry.lock


lacoak21 (commented):
Other than the issues brought up by Greg this looks good!

pyproject.toml
```diff
@@ -58,16 +58,28 @@ moto = "^4.1.3"
 boto3 = "^1.26.78"
 SQLAlchemy = "<=3.0.0"
 
+# Example to create requirements.txt for lambda layers:
+# poetry export -f requirements.txt --with lambda-database -o
+# lambda_layer/database/requirements.txt --without-urls
+[tool.poetry.group.lambda-sdc]
```
tech3371: can you rename this to something like `lambda-imap-processing`? because my mind thinks that `lambda-sdc` should contain everything that SDC needs which are what's in `lambda-database`.

while we are it, should we rename all `lambda` to `layer` instead? Something like this. This way it kind of matches how we are calling it in the code or how it's used.
```suggestion
[tool.poetry.group.imap-processing-layer]
```

.github/workflows/deploy.yml
```diff
@@ -24,7 +24,7 @@ jobs:
 
       - name: Install dependencies and app
         run: |
-          poetry install --with lambda-dev
+          poetry install --with lambda-database --with lambda-python
```
tech3371: Are you not adding what is currently called `lambda-sdc` for a reason? Same question for the change of tests.yml and pre-commit-config.yaml file change.

laspsandoval: @greglucas  said he would prefer imap_processing to be in the python lambda layer. At least I think I interpreted his message correctly.

pyproject.toml
laspsandoval: Ok. I renamed everything from lambda to layer

lacoak21 (approved):

tech3371 (approved):
Thank you Laura for doing this!

laspsandoval:
@tech3371 @lacoak21 Why do you think the SPICE error is happening? I don't think that has anything to do with this PR.

lacoak21:
> @tech3371 @lacoak21 Why do you think the SPICE error is happening? I don't think that has anything to do with this PR.

I think imap-data-access 0.17.0 updated SPICEFilePath and now the test test_spice_file() in tests/lambda_endpoints/test_batch_starter.py is not working properly. I am about to meet with the IDEX team but can look into it further after.

.github/workflows/deploy.yml
greglucas: I'm not sure what I said 😂 My preference would be for the database layer to be only psycopg / DB things. The other layers to contain imap_processing/SPICE. So that was my only real preference was to keep those separate. But if they can all be in one layer that is fine too.

bryan-harter:
oh yeah, imap-data-access requires a more strict filename convention now, so having "yyyy" in the file name won't work any more, it needs numbers

laspsandoval:
Good grief. I've got to get this PR merged before any other random things happen. That's what I get for being careless. :)

PR #511, opened 2025-04-03
Return most recent upstream ancillary file with the max version number from dependency API

### LINKED ISSUE #510: Return the latest upstream ancillary file with the max version number
Right now the dependency api is returning all the upstream ancillary files with the max version. For example, If there are two ancillary files with version "v003" it will return both of them. We need to ONLY return the one with the most recent start_date.

lacoak21:
# Change Summary

closes #510
## Overview
Right now the dependency api is returning all of the files that have the max start date. This PR updates this logic to return ONLY the ancillary file with the latest start date and max version number. 

## Updated Files
-sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
   - Update get_files() to return the latest ancillary file

## Testing
Updates test


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -652,36 +652,48 @@ def get_files(
             # Downstream: swe_l1a_sci
             # Upstream: Look for mag_l1d_sci with start dates less than or equal
             # to 20250102
-            # TODO returns a lot of ancillary files with start dates before 20250102
             type_specific_conditions.append(
                 models.ScienceFiles.start_date <= start_date
             )
+            return_latest_ancillary = True
```
tech3371, line 664: Are these two place holder for future handling? I think we will have case where some may want to specific version to use. But we can extend it in future PR to keep this PR scope small.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -652,36 +652,48 @@ def get_files(
...
+    # Only group by start date if return_latest_ancillary is false.
+    # If true, we only want to return one ancillary file (including science files of
+    # another instrument) with the most recent start date and greatest version number,
+    # otherwise we want to return the max version for each start_date.
+    if return_latest_ancillary:
+        subquery = max_version_query.subquery()
+        joiner = table.version == subquery.c.latest_version
+    else:
+        # Group by start_date
+        max_version_query = max_version_query.group_by(table.start_date)
```
tech3371, line 705: Can you add tests for this scenario or example in the comment? It would helpful for me

tech3371 (approved):
This looks like a good foundation. Version is something I still struggle with and going to need some more work on it  during reprocessing design.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: I was not planning on having them be placeholders. I added the boolean because I cant just rely on whether the type is ancillary because sometimes it will be science of another instrument. Do we know when people would want a specific version of an ancillary (including science of another instrument) and not just the latest one? How will batch starter/ dependency api know which version to use?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Yes! test_get_files_max_version() addresses this and I will add an example in the comment.

lacoak21: OK I added another test and comments.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: good to know. This may come up during reprocessing discussion. Let's revisit this then.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: cool. Thank you!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: But don't let this comment stop you from moving forward with this PR.

tech3371:
My previous approval still stand given our discussion!

PR #512, opened 2025-04-04
Fix bug in dependency api

lacoak21:
# Change Summary

## Overview
Fixes bug in the dependency api. The function filter_primary_source_files() was incorrectly using the descriptor of the upstream dep instead of the downstream dependency "potential job". 

This caused swapi l1 to not be processed because instead of checking if swapi_l1_sci had already been processed it checked for swapi_l1_hk.

## Updated Files
   - Fixed logs. 
   - Check existing records for potential job descriptor.


maxinelasp (approved):
Could you add a test for the dependency lambda? Approving to get this unblocked though!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -376,7 +378,7 @@ def filter_primary_science_dependencies(
         # already exists.
         query = select(models.ScienceFiles.__table__).where(
             models.ScienceFiles.instrument == record.instrument,
```
greglucas, line 381: Does this also need to be `query_instrument`? i.e. should all of these be query related rather than record related.

If SWE is dependent on MAG, won't this look up the SWE-descriptor rather than MAG-descriptor?

Alternatively should the "records" really be the "query" here rather than passing in a separate query related thing + record.

lacoak21: This function is only called on dependencies that are "primary science dependencies" meaning swe_l1a_sci is a downstream primary science dependency of swe_l0_raw. So in this case the query instrument and the record.instrument are already guaranteed to be the same but the descriptor and the data_level can be different. For this iteration the start_date and version should also be the same. 

This function is trying to filter primary science downstream dependencies that already exist ("records"). 

swe_l1a_sci = query 

So If the upstream dependencies are (records): 
     - swe_l0_raw_24_v001
     - swe_l0_raw_25_v001

That means we want to process: 
     - swe_l1a_sci_24_v001
     - swe_l1a_sci_25_v001

but if swe_l1a_sci_25_v001 already exists we can check that by querying the the science table for the record instrument ( or query instrument), the query descriptor, query data_level, and the record start date and version. 

Does that logic make sense?

greglucas: I'm not sure I fully follow and this is likely good enough for now, so feel free to consider this for later.

To me (coming in without prior knowledge of how this works), this bug looks like it crept in because of a minor bookkeeping mistake. My suggestion is to try and create exactly what it is you're querying for rather than combining instruments/descriptors across records. Can you construct "upstream_records" and "downstream_records" to be explicit in what you're looking for.

lacoak21: The idea of this function was to keep the dependency api returning a lot of primary science files dependent on an ancillary file if a new one with a large date range was uploaded. In the future some of these should be reprocessed so I will have to revisit this.

lacoak21: OK that is a good idea I will keep that in mind for the next iteration.

PR #516, opened 2025-04-04
Bugfix: batch starter skipped files 

maxinelasp:
# Change Summary
Batch starter can receive multiple files/events in one go. The code currently returns when it finds the first file that has no downstream dependencies, which is incorrect behavior. Now, instead of returning, we just skip the rest of that loop iteration.


lacoak21 (approved):
Good catch!!

PR #517, opened 2025-04-04
Automated packet downloads

greglucas:
# Change Summary

## Overview

This is the first cut at automated packet downloads. It is triggered based on repointing files arriving in our bucket.

We then look for the past two files uploaded to determine the time between contacts that we should be querying for. Let me know if there are better ideas for the logic here. This was just my attempt at quickly figuring it out. I think we should look into making this more robust in the future and potentially keeping track in a database or something similar. There is another internal API that we could access for this information, but I didn't have time to set that up yet and I am not positive it is what we want either.

There is a manual step where we need to make sure we have the `webpoda-token` secret added in AWS. This is from an external source, so I am not sure if there is a better way to do this. Let me know if so and I'm happy to add that in.


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
```diff
@@ -0,0 +1,130 @@
...
+    start_time, end_time = times
+
+    response = SECRETS_MANAGER.get_secret_value(SecretId="webpoda-token")
+    if "SecretString" not in response:
+        return {
+            "statusCode": 500,
+            "body": "Secret not found in Secrets Manager.",
+        }
+    # Update the token to be used for use in subsequent requests
+    imap_data_access.config["WEBPODA_TOKEN"] = response["SecretString"]
```
lacoak21, line 97: Do we know how long these tokens will last/ Is there any chance of it expiring in this duration?

greglucas: This one is an infinite one and manually uploaded. It is based on a service account we have with LASP that the web team set up for us. If we need to yank it, we will have to regenerate and upload to AWS.

I'll add some notes about that internally.

lacoak21: Got it 👍

sds_data_manager/constructs/packet_downloader_lambda_construct.py
```diff
@@ -0,0 +1,85 @@
+"""Configure monitoring formatter lambda."""
```
laspsandoval: Update the docstring.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
```diff
@@ -0,0 +1,130 @@
...
+
+    # Download the repointing file from S3 for use in the repointing downloads
+    S3_CLIENT.download_file(
+        Bucket=bucket_name,
+        Key=repointing_key,
+        Filename=repointing_file,
+    )
+
+    # ENA imagers group by repointing
```
laspsandoval, line 108: Does the download need to be deleted from the temp directory?

laspsandoval (approved):
Great job Greg. This will help so much.

sds_data_manager/constructs/packet_downloader_lambda_construct.py
greglucas: Thanks!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
greglucas: I don't think so, especially with it running so infrequently I'd assume the temp space gets cleared out. But time will tell how much space this all takes up and whether we even need more space than is available on a lambda.

I'll write a ticket for this as follow-up.

greglucas:
As an FYI to other reviewers. I merged so that we have this ready to go for tomorrow. Don't hesitate to leave comments here or open new issues and I'm happy to address them.

I have a separate branch for adding tests to this, but it was taking too long to figure out the mocks per usual :(

PR #520, opened 2025-04-08
Updating data access version

maxinelasp:
# Change Summary
Version bump, plus a new convenience script for the different layers requirements files.


generate_requirements_files.sh
greglucas, line 1: This should be run within `pre-commit` and not by users manually IMO. I think that has led to some of the issues we've been seeing because a user may have a different poetry version locally than what pre-commit is on.

Can you move this script into the `lambda_layer/` folder and figure out how to call this script within the pre-commit configuration? Otherwise I'd suggest just keeping the current 3 items within the exports of pre-commit.

tech3371 (approved):

greglucas:
This broke my workflow as I don't have poetry installed locally. So this still uses the local user's poetry version in the script apparently and not pre-commit's installed version in the separate venv. I am in favor of reverting that change and will push a PR for that now that I'll tag everyone on.

PR #522, opened 2025-04-09
SPICE Query API

bryan-harter:
# Change Summary

## Overview
Adds an API to our API gateway to enable querying for SPICE files

## New Files
- spice_query_api.py
   - The logic for querying the spice database
- test_spice_query_api.py
   - Unit tests for the SPICE queries 

## Updated Files
- tests/infrastructure/test_sds_api_manager_construct.py
   - Fixing the unit test now that we have a new endpoint
   - The spice indexer tests will use the lambda API for queries 

Minor Update:
- .devcontainer/Dockerfile
   - Updated the devcontainer to run on python 3.12

## Testing
New unit tests are in test_spice_query_api.py, and I update the unit tests in test_spice_indexer_lambda.py to use the new logic

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,10 +1,7 @@
 # OS
-ARG VARIANT=bullseye
-FROM --platform=linux/amd64 mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
-RUN apt-get update && export DEBIAN_FRONTEND=noninteractive && apt-get install -y firefox-esr
+FROM --platform=linux/amd64 mcr.microsoft.com/vscode/devcontainers/python:3.12
 RUN sudo apt-get update
-RUN sudo apt-get install -y libgtk-3-dev
-RUN curl -fsSL https://deb.nodesource.com/setup_18.x | sudo -E bash - && sudo apt-get install -y nodejs
+RUN curl -fsSL https://deb.nodesource.com/setup_20.x | sudo -E bash - && sudo apt-get install -y nodejs
 RUN npm install -g aws-cdk
 RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
 RUN unzip awscliv2.zip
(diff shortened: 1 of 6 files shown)


bryan-harter:
Some general questions:

1) Do we want a separate SPICE query API or should this somehow be rolled into the science query API? 

2) Should we be able to query by datetime? If so, should we have the option for someone to query in either seconds or datetime string? On EMM we allow querying by seconds, datetime, and sclk, with a query_parameter of "time_units={sclk | datetime | j2000}" but maybe we should just choose one and stick with it? 

3) Anything else we should be able to query on? Really the only other options are version number and ingestion_date, but I don't know if there is a huge demand for querying either of those.  

4) Should the ability to create a metakernel go in *THIS* API, or a separate one? For example, we could add a query_parameter called "as_metakernel" and it could return the contents of a metakernel file.

tech3371 (approved):

tech3371:
> Some general questions:
> 
> 1. Do we want a separate SPICE query API or should this somehow be rolled into the science query API?
> 2. Should we be able to query by datetime? If so, should we have the option for someone to query in either seconds or datetime string? On EMM we allow querying by seconds, datetime, and sclk, with a query_parameter of "time_units={sclk | datetime | j2000}" but maybe we should just choose one and stick with it?
> 3. Anything else we should be able to query on? Really the only other options are version number and ingestion_date, but I don't know if there is a huge demand for querying either of those.
> 4. Should the ability to create a metakernel go in _THIS_ API, or a separate one? For example, we could add a query_parameter called "as_metakernel" and it could return the contents of a metakernel file.

These are really good questions. I don't know answer to those yet.

bryan-harter:
If it looks alright to you, we could just deploy as is for the moment, and come back to add features to it later when we understand how SPICE will be used in the future. 

Or we could wait until we understand the use case further. But having this query code in place sooner rather than later will be useful for the metakernel generation code, which needs to query our database

tech3371:
> If it looks alright to you, we could just deploy as is for the moment, and come back to add features to it later when we understand how SPICE will be used in the future.
> 
> Or we could wait until we understand the use case further. But having this query code in place sooner rather than later will be useful for the metakernel generation code, which needs to query our database

Yeah, That sounds good!

PR #524, opened 2025-04-11
Add HI L2 triggering HI L3 to dependency config

pleasant-menlo:
# Change Summary

## Overview
Add dependency from HI L2 to HI L3.

## New Dependencies
N/A

## New Files
N/A

## Deleted Files
N/A

## Updated Files
- dependency_config.csv
  - Add line to trigger HI L3 pipeline based on HI L2 dependency 

## Testing
N/A


tech3371 (approved):
The syntax looks good. I would check with Tim if he will produce that L2 product. Besides that, LGTM

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
```diff
@@ -49,6 +49,9 @@ hi, l1b, 45sensor-de, hi, l1c, 45sensor-pset, HARD, DOWNSTREAM
 hi, l1a, 90sensor-de, hi, l1b, 90sensor-de, HARD, DOWNSTREAM
 hi, l1b, 90sensor-de, hi, l1c, 90sensor-pset, HARD, DOWNSTREAM
 
+hi, l2, 90sensor-spacecraft-full-4deg-map, hi, l3, 90sensor-spacecraft-survival-full-4deg-map, HARD, DOWNSTREAM
```
tmplummer: Map descriptors are a WIP. Dan sent out an email with proposed descriptors but I don't think any discussion is planned until next week. Proposal is as follows:

> We need to come up with short map CDF descriptor elements to concatenate together for the “Map_descriptor” field (e.g. h45-hf-sp-ram-hae-3mo).  I will throw out the following suggestions:
3-letter instrument descriptor: ilo, h45, h90, u45, u90, glx, idx
species: helium (he), oxygen (oxy) – only use if it’s not hydrogen.
frame: spacecraft (sf), heliospheric (hf), intermediate heliospheric (cg)
survival (probability) corrected: sp, or nothing if not corrected  (this could be ‘sc’, but that looks too much like it is short for ‘spacecraft’)
spin phase: ram (ram), anti-ram (anti), nothing if full spin
coordinate system: hae, hgi, rc, etc.
duration: 3mo, 6mo, 1yr

I don't see any proposal for the map resolution but think that `4deg` seems reasonable. 
I think that we can probably remove the `map` field as all Hi L2 products are maps.

pleasant-menlo: Thanks for the suggestion. We've updated our dependency lines to use the abbreviations. The descriptor format will also have to take into account combined 90 and 45 maps for ultra and hi as well as spectral index maps.

For a starting point on spectral index maps, we used your proposed map descriptor with '-spectral' on the end.

For testing purposes, is it possible to deploy this before the map descriptors have been finalized? If so, are there any additional steps we should take: opening an issue? Adding a comment that this is WIP or a TODO?

tmplummer: Sure, this looks good for now.

tmplummer (approved):
Thanks for the change. I will bring up the additional missing pieces from the descriptor that you pointed out. Thanks!

tech3371:
@pleasant-menlo Can you rebase and then I can merge it?

tmplummer (approved):

PR #525, opened 2025-04-14
Adding processing time to ProcessingJob table

tech3371:
# Change Summary

## Overview
This is pre-work of API that returns information about Batch job. 

## Updated Files
   - update table use datetime type
   - update code to write datetime object to DB

## Testing

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -88,7 +88,7 @@ class ProcessingJob(Base):
     job_log_stream_id = Column(String)
     container_image = Column(String)
     container_command = Column(String)
-    start_at = Column(DateTime(timezone=True))
+    started_at = Column(DateTime(timezone=True))
     stopped_at = Column(DateTime(timezone=True))
 
     __table_args__ = (
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -88,7 +88,8 @@ class ProcessingJob(Base):
     job_log_stream_id = Column(String)
     container_image = Column(String)
     container_command = Column(String)
-    processing_time = Column(DateTime(timezone=True))
+    start_at = Column(DateTime(timezone=True))
+    stopped_at = Column(DateTime(timezone=True))
 
     __table_args__ = (
         # Partial unique index to ensure only one INPROGRESS or COMPLETED for a record
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -88,7 +88,7 @@ class ProcessingJob(Base):
     job_log_stream_id = Column(String)
     container_image = Column(String)
     container_command = Column(String)
-    processing_time = Column(Integer)
+    processing_time = Column(DateTime(timezone=True))
 
     __table_args__ = (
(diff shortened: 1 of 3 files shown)


sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -88,7 +88,7 @@ class ProcessingJob(Base):
     job_log_stream_id = Column(String)
     container_image = Column(String)
     container_command = Column(String)
-    processing_time = Column(Integer)
+    processing_time = Column(DateTime(timezone=True))
```
greglucas: We use both `timezone=True` and `timezone=False` in this file, should we use `timezone=False` here and just "know" that we are using UTC as a standard?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
```diff
@@ -259,6 +260,9 @@ def batch_event_handler(event):
             ),
             "status": "FAILED",
             "statusReason": "some error message",
+            "createdAt": 1744396985534,
+            "startedAt": 1744397031734,
+            "stoppedAt": 1744397296519,
```
greglucas, line 265: Do we want to track all of these in the DB? I could see them all being useful. Particularly the startedAt and stoppedAt would get us how long the job took. Why do we use **stoppedAt** for 'processing_time` instead of **startedAt**?

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: Do you mean `timezone=False` in pytests? That's because of limitation of sqlite engine we are using in our pytest. I didn't see `timezone=False` in this file. Can you point me to that?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
tech3371: I can do that. call them `start_at` and `stopped_at` to be consistent with other naming convention?

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: 🤔 I swear I just looked at this locally and saw `timezone=False` somewhere 🤯 but now I don't... Ignore me, sorry about that!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
greglucas: Sounds good to me! (changed from processing_time)

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: I have Monday haze too!

bourque (approved):
LGTM!

greglucas (approved):

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -88,7 +88,8 @@ class ProcessingJob(Base):
     job_log_stream_id = Column(String)
     container_image = Column(String)
     container_command = Column(String)
-    processing_time = Column(Integer)
+    start_at = Column(DateTime(timezone=True))
```
greglucas:
```suggestion
    started_at = Column(DateTime(timezone=True))
```

tech3371: good catch!

PR #532, opened 2025-04-15
I-ALiRT - CDF Cron Job

laspsandoval:
# Change Summary

## Overview
Created lambda that will be kicked off daily by a cron job. The lambda will query the database and then (eventually) create a cdf and put the cdf in s3.

## New Files
- ialirt_archive_construct.py
   - Added cron event and lambda that will query the database
- Dockerfile.archive
   - Name change
- ialirt_archive.py
   - Lambda that will query the database

## Updated Files
- stackbuilder.py
   - Deploys ialirt_archive_lambda_construct in ialirt_stack
- .pre-commit-config.yaml
   - Added creation of requirements.txt for imap_processing
   - Small change to use renamed Dockerfile
- Dockerfile.ingest
   - Dockerfile used for ingest lambda

   
## Testing
- test_ialirt_archive_construct.py
   - Test ialirt_archive_construct.py
- test_ialirt_archive.py
   - Test ialirt_archive.py
- Successfully deployed to personal AWS account.


sds_data_manager/constructs/ialirt_archive_construct.py
```diff
@@ -0,0 +1,107 @@
...
+
+        # The resource is deleted when the stack is deleted.
+        ialirt_archive_lambda.apply_removal_policy(RemovalPolicy.DESTROY)
+
+        return ialirt_archive_lambda
+
+    def create_event_rule(
+        self, ialirt_archive_lambda: lambda_.DockerImageFunction
+    ) -> None:
+        """Create the event rule to trigger Lambda on S3 object creation."""
```
greglucas:
```suggestion
        """Create the event rule to trigger Lambda once per day."""
```

sds_data_manager/constructs/ialirt_archive_construct.py
```diff
@@ -0,0 +1,107 @@
...
+            timeout=Duration.minutes(1),
+            memory_size=1000,
+            role=lambda_role,
+            environment={
+                "ALGORITHM_TABLE": algorithm_data_table.table_name,
+                "S3_BUCKET": ialirt_bucket.bucket_name,
+            },
+        )
+
+        algorithm_data_table.grant_read_write_data(ialirt_archive_lambda)
```
greglucas: Just read permissions on the table, or does this one need to write as well?

sds_data_manager/lambda_code/IAlirtCode/Dockerfile.archive
greglucas, line 1: Can you have one Dockerfile / ImageFunction and switch the entrypoint/cmd when calling it? These contain basically the same thing it looks like, just a different file/command. So put both files in and then let the code figure out who called / what subfunction to call from there even.

sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
```diff
@@ -0,0 +1,62 @@
...
+    response : dict
+        The response from the DynamoDB query.
+    """
+    logger.info("Received event: %s", json.dumps(event))
+
+    algorithm_table_name = os.environ.get("ALGORITHM_TABLE")
+    dynamodb = boto3.resource("dynamodb")
+    algorithm_table = dynamodb.Table(algorithm_table_name)
+
+    now = datetime.utcnow()
```
greglucas:
```suggestion
    now = datetime.now(datetime.UTC)
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
```diff
@@ -0,0 +1,62 @@
...
+    response = algorithm_table.query(
+        IndexName="insert_time",
+        KeyConditionExpression="apid = :apid_val AND "
+        "insert_time BETWEEN :start AND :end",
+        ExpressionAttributeValues={
+            ":apid_val": 478,
+            ":start": start_iso,
+            ":end": end_iso,
+        },
+    )
```
greglucas: I think this should use the boto3 methods rather than creating the commands manually.
```
boto3.dynamodb.conditions.Key('apid').eq(478) &
boto3.dynamodb.conditions.Key('insert_time').between(start_iso, end_iso)
```

tests/lambda_endpoints/test_ialirt_archive.py
```diff
@@ -0,0 +1,46 @@
...
+import pytest
+
+from sds_data_manager.lambda_code.IAlirtCode.ialirt_archive import lambda_handler
+
+
+@pytest.fixture
+def populate_algorithm_table(setup_dynamodb):
+    """Populate the algorithm table with test entries."""
+    algorithm_table = setup_dynamodb["algorithm_table"]
+    now = datetime.utcnow()
```
greglucas: It is deprecated, so we should be moving away from this call.
https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow

sds_data_manager/constructs/ialirt_archive_construct.py
```diff
@@ -0,0 +1,107 @@
...
+    def __init__(
+        self,
+        scope: Construct,
+        construct_id: str,
+        algorithm_data_table: ddb.Table,
+        ialirt_bucket: aws_s3.Bucket,
+        docker_path: str = "sds_data_manager/lambda_code",
+        **kwargs,
+    ) -> None:
+        """Create ialirt cdf.
```
lacoak21:
```suggestion
        """Create ialirt cdf archive resources.
```

sds_data_manager/constructs/ialirt_archive_construct.py
```diff
@@ -0,0 +1,107 @@
...
+                ),
+            ],
+        )
+
+        # Lambda function
+        ialirt_archive_lambda = lambda_.DockerImageFunction(
+            self,
+            id="IalirtArchiveLambda",
+            code=lambda_.DockerImageCode.from_image_asset(
+                docker_path,
```
lacoak21, line 76: should docker path be "sds_data_manager/lambda_code/IAlirtCode/" and the file be just "Dockerfile.archive"?

lacoak21 (approved):
Looks good Laura! Just had a couple of comments.

sds_data_manager/constructs/ialirt_archive_construct.py
laspsandoval: You're right that pointing docker_path directly to "sds_data_manager/lambda_code/IAlirtCode" would simplify the Dockerfile. However, we will have other Lambda constructs at the same level as IAlirtCode that reuse the same processing/requirements.txt. By keeping the build context at lambda_code/, all of those Dockerfiles can share the same relative path for requirements and avoid duplication.

sds_data_manager/lambda_code/IAlirtCode/Dockerfile.archive
laspsandoval: AWS Lambda container images don’t support overriding the CMD at deployment. So I guess I could have another lambda that decides which Dockerfile to use, but I'm not certain that I like that either.

PR #534, opened 2025-04-16
Ultra - add ultra dependency info

laspsandoval:
# Change Summary

## Overview
Added (almost) all of the Ultra dependencies. There are still a couple of lookup tables that they will give to me. Based on this Galaxy page:
https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/256944034/Ultra+lookup+table+names

Note that some 45 or 90 luts are missing. In those cases I use the other instrument. 

## Updated Files
- dependency_config.csv
   - Added Ultra dependency info

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
@@ -135,13 +135,52 @@ ultra, l0, raw, ultra, l1a, all, HARD, DOWNSTREAM
 
 # Ultra-45 products
 ultra, l1a, 45sensor-de, ultra, l1b, 45sensor-de, HARD, DOWNSTREAM
+ultra, l1a, 45sensor-params, ultra, l1b, 45sensor-extendedspin, HARD, DOWNSTREAM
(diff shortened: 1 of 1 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
tmplummer, line 1: Out of curiosity, is there a reason that it seems like everyone specifies dependencies by DOWNSTREAM only?

tmplummer, line 1: As a follow on to my question, IMO, it might make dependency identification easier to group by the downstream dependency. For example, if I want to see what all of the upstream dependencies for Ultra L1B extended spin are, can I find them all in one place?

Do whatever makes the most sense to you. You are leading the way in integrating complex dependencies.

tmplummer (approved):
I'll let you decide what makes the most sense to you for my comment. This makes me feel lucky that so far, Hi dependencies are pretty simple.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
laspsandoval: That is a good question. I don't know why we even have the UPSTREAM/DOWNSTREAM option if we are only using downstream.

PR #536, opened 2025-04-18
First Pass at DPS architecture

laspsandoval:
# Change Summary

## Overview
Triggers lambda after kernels are dropped in spice/ck or spice/repoint.

Big picture: either ck or repoint files trigger the lambda. We generate a new dps kernel for the attitude timerange each time (regardless if we already have the same contents in another dps kernel). We don't check. We just create. It is what Tim and I call the "brute force" method. This is the just the architecture for it to get setup.

Note: we haven't figured out filenaming. That's probably next.

## New Files
- Dockerfile.dps
   - Dockerfile used for lambda environment
- dps_maker.py
   - Lambda that will be invoked for creating dps

## Updated Files
- .pre-commit-config.yaml
   - Precommit hook for generating requirements.txt for imap_processing
- .Dockerfile
   - Modified path
- indexer_lambda_construct.py
   - Added construct for dps lambda
- spice_indexer.py
   - Added customized event used to kick-off dps_maker.py
- requirements.txt
   - Path change
- stackbuilder.py
   - Calls on dps construct

## Testing
- test_dps_maker.py
- test_spice_indexer_lambda.py

Deployed to personal account and dropped a .csv into spice/repoint. Checked that dps_maker.py was kicked off:
[INFO]	2025-04-18T20:00:10.711Z	03e05d42-a1c5-4261-a45c-d2655c50933c	Retrieved filename: imap_2026_267_01.repoint.csv


greglucas (commented):
Per my comment on the other PR if we are going towards many container images.
https://github.com/aws-samples/aws-cdk-examples/blob/main/python/lambda-from-container/app.py
I am wondering if you can create the ecrImage and share that across your lambdas so we aren't building new containers all the time.
```
            ecr_image = aws_lambda.EcrImageCode.from_asset_image(
                directory = os.path.join(os.getcwd(), "lambda-image")
            )
```

Do we need this to be triggered on new ck files arriving, or can we rely on only new repointing files arriving? I don't think we should be watching EFS/eventbridge things, but rather just watching the s3 bucket for that repointing file like we are in the packet_downloader.

laspsandoval:
> Per my comment on the other PR if we are going towards many container images. https://github.com/aws-samples/aws-cdk-examples/blob/main/python/lambda-from-container/app.py I am wondering if you can create the ecrImage and share that across your lambdas so we aren't building new containers all the time.
> 
> ```
>             ecr_image = aws_lambda.EcrImageCode.from_asset_image(
>                 directory = os.path.join(os.getcwd(), "lambda-image")
>             )
> ```
> 
> Do we need this to be triggered on new ck files arriving, or can we rely on only new repointing files arriving? I don't think we should be watching EFS/eventbridge things, but rather just watching the s3 bucket for that repointing file like we are in the packet_downloader.

I will look into sharing the image.

If we only rely on new repointing files to arrive then there is a possibility that the attitude kernel won't have the timerange that we need for each pointing. And things get more complicated. Hence the brute-force method.

The lambda (dps_maker.py) will need to mount the EFS to create the dps kernel, which will need to have the new attitude kernel in it. So if we watch s3 we get into race conditions. Will the new attitude kernel be in EFS in time, etc.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -304,6 +304,32 @@ def write_data_to_efs(s3_key: str, s3_bucket: str, spice_mount_path: Path) -> Pa
...
+
+    Parameters
+    ----------
+    file_path : Path
+        The path to the file that was written to EFS
+    s3_key : str
+        S3 object key
+    """
+    prefix = s3_key.split("/")[1]
+    if prefix not in ["ck", "repoint"]:
```
tech3371, line 318: Another way you can do this, you can pass filename into `imap_data_access.SPICEFilePath`. Then lookup `.spice_metadata["type"]` and see if it's of `repoint` or `attitude_history` or `attitude_predict`. Does DPS cares if it's history or predict?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dps_maker.py
```diff
@@ -0,0 +1,33 @@
...
+import logging
+import os
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+def lambda_handler(event, context):
+    """Create dps kernel.
+
```
tech3371, line 13: Can you add in the doc string that you are expecting event input like this for future reader?
```
{
            "Source": "imap.spice.efs",
            "DetailType": "SPICE EFS Write Complete",
            "Detail": f'{{"path": "{file_path}", "prefix": "{prefix}"}}',
        }
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -304,6 +304,32 @@ def write_data_to_efs(s3_key: str, s3_bucket: str, spice_mount_path: Path) -> Pa
     return efs_spice_filename_and_path
 
 
+def send_event_to_trigger_dps_lambda(file_path: Path, s3_key: str):
```
tech3371, line 307: Can you change first parameter name to `efs_path` or something since it's EFS filepath now since `write_data_to_efs()` function returns `efs_spice_filename_and_path`?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -304,6 +304,32 @@ def write_data_to_efs(s3_key: str, s3_bucket: str, spice_mount_path: Path) -> Pa
...
+    if prefix not in ["ck", "repoint"]:
+        return
+
+    client = boto3.client("events")
+    client.put_events(
+        Entries=[
+            {
+                "Source": "imap.spice.efs",
+                "DetailType": "SPICE EFS Write Complete",
+                "Detail": f'{{"path": "{file_path}", "prefix": "{prefix}"}}',
```
tech3371, line 327: After looking at your DPS construct, you are expecting this `prefix` to `ck` or `repoint` for your rule. If you do above suggestion, you can change event input to be

```
{
"efs_path": "<full EFS path and filename>",
"kernel_type": imap_data_access.SPICEFilePath(filename).spice_metadata["type"]
}
```
Then in your Event rule in `DpsLambda`, you can update your rule to filter event on these
```
efs_write_event_rule = events.Rule(
            self,
            "TriggerDpsAfterSpiceEfsWrite",
            event_pattern=events.EventPattern(
                source=["imap.spice.efs"],
                detail_type=["SPICE EFS Write Complete"],
                detail={
                    "kernel_type": ["attitude_history", "attitude_predict", "repoint"],
                },
            ),
        )
```

sds_data_manager/constructs/indexer_lambda_construct.py
```diff
@@ -262,3 +265,106 @@ def __init__(
...
+            ],
+            assumed_by=iam.ServicePrincipal("lambda.amazonaws.com"),
+        )
+
+        dps_lambda = lambda_.DockerImageFunction(
+            self,
+            id="DpsLambda",
+            code=lambda_.DockerImageCode.from_image_asset(
+                docker_path,
+                file="SDSCode/Dockerfile.dps",
```
tech3371, line 330: I don't what the best folder this Dockerfile fits in but would be nice if this `Dockerfile.dps` can go into a folder.

sds_data_manager/lambda_code/processing/requirements.txt
tech3371, line 1: Curious why this was not in `lambda_layer/` folder? All of requirements were in that folder. It would be nice to group them in the same location.

tech3371 (commented):
I have some initial comments

laspsandoval:
This will be moved to a batch job.

laspsandoval:
This is WIP.

laspsandoval:
This is WIP.

tech3371:
@laspsandoval I think you can close this PR since DPS will be trigger from batch starter and now processed in batch like any other jobs.

PR #538, opened 2025-04-18
CRID calculation and checks for upstream files. 

### LINKED ISSUE #716: CRID calculation to reduce duplicate processing
Calculate the expected CRID for each input file dependency. If the CRID doesn't match what we expect, we can assume that an upstream file is being processed. In this case, hold off processing and allow that upstream file to retrigger processing when complete.

lacoak21:
# Change Summary

closes #716

## Overview
Add CRID calculation for reprocessing checks. 

Does not include tests yet because this is a very rough draft and might be subject to a lot of change. 

## Updated Files

   - Add column to ScienceFiles to keep track of CRID
   - Add CRID calculation, insertion and checks

## Testing
tests/lambda_endpoints/test_dependency_api.py
edit to work with dep node refactor


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -432,6 +434,154 @@ def primary_science_dep(query_params: dict, dependency: dict) -> bool:
...
+            upstream_records = get_files(
+                session,
+                upstream_dep["data_source"],
+                upstream_dep["data_type"],
+                upstream_dep["descriptor"],
+                record.start_date,
+                primary_sci_dep=primary_sci_dep,
+            )
+            if not upstream_records:
+                # TODO: How to handle this case
```
maxinelasp, line 486: I think we should just construct the CRID without the dependency.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -432,6 +434,154 @@ def primary_science_dep(query_params: dict, dependency: dict) -> bool:
...
+    ----------
+    session : db.Session
+        Database session.
+    record : models.ScienceFiles, models.AncillaryFiles, or models.SPICEFiles
+        The record for which the CRID is being calculated.
+
+    Returns
+    -------
+    str
+        The calculated CRID as a SHA-256 hash.
```
maxinelasp, line 518: I think rather than using a hash, we should actually use an encoding scheme so we can go back and get the input versions when we have the CRID (rather than SHA-256 where you can't go back.) That's my bad for using the wrong terminology, I didn't realize "hash" is specifically one-way. 

ChatGPT suggested packing the version numbers into 2 bytes and then using [base85](https://docs.python.org/3/library/base64.html#base64.a85encode), which seems like a very reasonable suggestion.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -432,6 +434,154 @@ def primary_science_dep(query_params: dict, dependency: dict) -> bool:
...
+        existing_crid = upstream_record.crid
+        if existing_crid:
+            # Check if the CRID already exists in the database
+            if existing_crid == crid:
+                logger.info(
+                    f"Found expected CRID for {upstream_record.file_path}. Continuing.."
+                )
+            else:
+                logger.info(
+                    f"Found unexpected CRID for {upstream_record.file_path}. "
```
maxinelasp, line 571: "unexpected" seems confusing. I would say "Found mismatched CRID..."

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -432,6 +434,154 @@ def primary_science_dep(query_params: dict, dependency: dict) -> bool:
...
+                    f"This indicates that we are expecting a reprocessing for"
+                    f" this file."
+                )
+                return False
```
maxinelasp, line 575: nitpick but I don't like returning in the middle of a function, especially in a for-loop like this. Could you use a variable and `break` instead?

maxinelasp (commented):
Overall this looks good, I would say you're close to being able to merge now (with tests) but I think this is a good checkpoint. Note that we'll need to remove and re-deploy the database to add the CRID so we should also add some of the loosened database rules suggested by [Greg](https://lasp.slack.com/archives/C04L7EUMJ8G/p1745272076758469) before running the destroy/deploy cycle.

tech3371:
@lacoak21 can we close this and revisit after SIT-4?

PR #540, opened 2025-04-18
Code changes to put invalid data into staging folder

### LINKED ISSUE #539: Creating stage folder in data bucket temporarily
MOC keeps changing filename convention on us and it's been hard to receive data from MOC through POC without causing upload to fail. This is temporarily workaround to receive files of MOC. Then when they narrow down on filename convention, then we can update our code to work and remove this workaround.

tech3371:
# Change Summary
closes #539 

## Overview
MOC keeps changing filename convention on us and it's been hard to receive data from MOC through POC without causing upload to fail. This is temporarily workaround to receive files of MOC. Then when they narrow down on filename convention, then we can update our code to work and remove this workaround.

## Updated Files
- sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
   - way to send data to staging folder

## Testing
tests/lambda_endpoints/test_upload_api.py


sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
```diff
@@ -130,24 +131,21 @@ def lambda_handler(event, context):
...
+                )
+                invalid_file_error = True
+                s3_key_path_str = f"staging/{filename}"
+
+    if not invalid_file_error:
+        s3_key_path = file_obj.construct_path()
+        # Strip off the data directory to get the upload path + name
+        # Must be posix style for the URL
+        s3_key_path_str = str(
+            s3_key_path.relative_to(imap_data_access.config["DATA_DIR"]).as_posix()
```
maxinelasp, line 148: Will this still return an error if the filename already exists? Or will we overwrite files? I think this design is fine as a temporary thing, but I'm wondering what happens in that case.

maxinelasp (approved):

sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
tech3371: good point. I will check that that happens.

tech3371: line 89 in `test_upload_api.py` tests that we throw `409` error if someone tries to upload a file again. And that tests passes with this change.

PR #395, opened 2024-11-12
BUG - Undid Code that breaks tests due to psycopg import error

tech3371:
# Change Summary
closes #394 

## Overview
We are unable to run pytest locally to test anything because of issue listed in the ticket. The logic in the file was added to tests `test_batch_starter.py`'s `test_duplicate_job()`. I undid that change since it's more important to test other tests. If we want to continue testing batch starter's duplicate job, we will need to revisit. I will create issue to capture that.

## Testing
- tests/lambda_endpoints/conftest.py
   - undid lines that create engine using postgresql for pytest

PR #549, opened 2025-04-23
MNT: Fixing some SPICE path to be able to work with upcoming change of DATA_DIR

### LINKED ISSUE #550: MNT: Fixing some SPICE path to be able to work with upcoming change of DATA_DIR
Read summary in PR for more details.

tech3371:
# Change Summary
closes #550 #565

## Overview
This PR updates SPICE query API to return filename without prefix of path. Instead, it moves adding S3 path prefix in metakernel API code. This code changes help line up with future work. Future work is mounting S3 to batch job. For that, it needs to know about s3 folder path of SPICE. Then when user or batch starter passes `spice_path` parameter to metakernel API, it will line nicely with DATA_DIR thinking and download API. If user pass in `DATA_DIR` as `/mnt/`, then the path in SPICE metakernel would result in `/mnt/spice/ck/<filename>`.

## Updated Files
- sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
   - Format fixes and adding S3 folder path
   - Removed adding folder prefix

## Testing
- tests/lambda_endpoints/test_spice_query_api.py

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
@@ -3,7 +3,7 @@
 import json
 import logging
 import textwrap
-from datetime import datetime
+from datetime import datetime, timezone
 from pathlib import Path
 
 logger = logging.getLogger(__name__)
@@ -46,16 +46,15 @@ class MetaKernel:
             self.spice_files[type] = []
             self.spice_gaps[type] = [[start_time, end_time]]
 
-        self.template_header = rf"""
+        self.template_header = f"""
+\\begintext
 
-       \begintext
+This is the most up to date Metakernel as of
+{datetime.now(timezone.utc)}.
 
-       This is the most up to date Metakernel as of
-       {datetime.now()}.
-
-       This attempts to cover data from
-       {self.start_time_j2000} to {self.end_time_j2000}
-       seconds since J2000.
+This attempts to cover data from
+{self.start_time_j2000} to {self.end_time_j2000}
+seconds since J2000.
 
         """
 
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
@@ -353,7 +353,7 @@ def lambda_handler(event, context):
 
     """
     # Define the paths
-    spice_mount_path = Path(os.getenv("EFS_SPICE_MOUNT_PATH"))  # Eg. /mnt/
+    spice_mount_path = Path(os.getenv("DATA_DIR"))  # Eg. /mnt/data
 
     # Retrieve the S3 bucket and key from the event
     s3_bucket = event["detail"]["bucket"]["name"]
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
@@ -382,5 +382,5 @@ def lambda_handler(event, context):
 
     return {
         "statusCode": 200,
-        "body": f"{s3_key} moved and index to table successfully",
+        "body": f"{s3_key} moved to EFS and indexed to table successfully",
     }
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
@@ -274,7 +274,7 @@ def write_data_to_efs(s3_key: str, s3_bucket: str, spice_mount_path: Path) -> Pa
     s3_bucket : str
         The S3 bucket
     spice_mount_path: Path
-        The path to the local SPICE directory
+        The path to the local SPICE directory. Eg. /mnt/
 
     Returns
     -------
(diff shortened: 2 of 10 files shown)


tests/lambda_endpoints/test_spice_query_api.py
```diff
@@ -123,7 +123,7 @@ def expected_ck_response():
                 "sclk_kernel": "naif0012.tls",
                 "lsk_kernel": "imap_sclk_0012.tsc",
                 "ingestion_date": "2025-04-09, 21:12:53",
-                "timestamp": 1744233173.0,
+                "timestamp": 1744254773.0,
```
tech3371: @bryan-harter Do you know why my local returns `1744254773.0 ` and GitHub testing returns `1744233173.0`?

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
```diff
@@ -46,16 +48,15 @@ def __init__(
...
-        self.template_header = rf"""
-
-       \begintext
+        self.template_header = f"""
+\\begintext
 
-       This is the most up to date Metakernel as of
-       {datetime.now()}.
+This is the most up to date Metakernel as of
+{datetime.now()}.
```
greglucas:
```suggestion
{datetime.now(datetime.UTC)}.
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
```diff
@@ -162,7 +163,13 @@ def return_tm_file(self, base_path: Path) -> str:
...
         kernelfiles = []
         for f in metakernel_files:
-            fn = base_path / f["file_name"]
+            # Set path prefix to be path of s3 folder
+            spice_obj = SPICEFilePath(f["file_name"])
+            s3_folder_path = (
+                f"{spice_obj.construct_path().parent.parent.name}/"
+                f"{spice_obj.construct_path().parent.name}"
+            )
+            fn = base_path / f"{s3_folder_path}/{f['file_name']}"
```
greglucas: This is confusing what is going on here with all the parents / names. Can you give an example of what is going on?

Perhaps using `"/".join(path.parts[-3:-1])` or something like that?

tests/lambda_endpoints/test_spice_query_api.py
tech3371: @greglucas Do you think it's because of timezone too? I commit what it expects but my local tests fails because of this.

greglucas: I don't know 🤷, but it looks like you removed our PostgreSQL tests a while ago (which were causing the previous time issues) in this PR: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/395
This is a bit concerning because we are running a postgres server in operations, but only testing on a sqlite3 db. These handle times very differently which is why we wanted/needed the previous setup. We should look into adding that capability back in.

tech3371: yeah, we had to do that because tests were failing and you were away during that. I looked and many of us didn't know how to fix that.

bryan-harter: Yeah somewhere along the line it does something weird with the time zones, since that number is exactly 6 hours ahead of UTC (and we're 6 hours behind). The unit tests on github are accurate, so the number should be 1744233173.0.  I think virtual environments (like my dev environment and the github testing environment) default to UTC, but your laptop is set to mountain time. 

We should probably go through the test and the indexing code itself, and make sure all the inserts/reads have the appropriate time zone set in the datetime objects.

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
```diff
@@ -6,6 +6,8 @@
 from datetime import datetime
 from pathlib import Path
 
+from imap_data_access import SPICEFilePath
```
bryan-harter: This isn't a big deal and you're welcome to change it if you want to, but I was thinking we would try to keep this metakernel.py file general purpose and free from any IMAP dependencies, so we could easily take this file out and put it in curryer (or something) at some point.

tech3371: Since we decided to use top approach in the purple boxes diagram I sent few days ago, we won't be generating metakernel anymore. I can undo this.

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
tech3371: I will be undoing this change given recent decisions.

tests/lambda_endpoints/test_spice_query_api.py
tech3371: Let's do that in future

tech3371:
I undid changes in SPICE query API. I see now why Bryan was doing that. I instead made some changes to EFS mount and EFS write function. Will test and fix more as needed on Monday.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -285,23 +286,22 @@ def write_data_to_efs(s3_key: str, s3_bucket: str, spice_mount_path: Path) -> Pa
...
-    # Keep the base folder name and filename from the s3 key
-    # i.e. "ck/file.bc"
+    # Get directory structure from the S3 key
     dirname, filename = os.path.split(s3_key)
-    s3_folder_path = os.path.basename(dirname)
-    # Download path to EFS
-    efs_spice_path = spice_mount_path / s3_folder_path
-    efs_spice_filename_and_path = efs_spice_path / filename
+    # Prepend EFS path to the s3 directory structure
+    efs_spice_path = spice_mount_path / dirname
```
greglucas: Do you want to call this "data" mount path now since it is the generic data folder instead of spice specific folder?
```suggestion
    efs_spice_path = data_mount_path / dirname
```

greglucas (approved):
Nice! I like where this heading with the DATA_DIR. I think this is going to help us moving forward in keeping everything co-aligned (data and spice) without needing multiple environment variables.

laspsandoval (approved):
Good for I-ALiRT

PR #559, opened 2025-04-28
Simplify Dependency API

### LINKED ISSUE #558: Simplify dependency api
Simplify dependency api to get upstream files in date range supplied to the query.
Image

lacoak21:
# Change Summary
Simplify dependency api to get upstream files in set range.

closes #558

## Overview
Tenzin and I realized that the dependency handler was getting overcomplicated and overloaded. It would make debugging a challenge. We decided to require batch starter to supply a start_date and end_date range to the dependency lambda and it would return all of the science files and ancillary files that are valid for that range. Before it was using convoluted logic to try and determine which files to return. Part of the reason why it was using this complicated logic was in preparation for returning multiple files as an input for the maps but these will now be triggered by cron jobs where the start_date and end_date will be explicitly set in the event data. 

## Updated Files
   - Add handling for determining date range for the upstream file dependency request. 
   - Simplify logic to return only the files valid for the given date range instead of trying to determine that within this code. 

## Testing
Remove some of the dependency tests that are no longer needed after the simplification.


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -436,7 +427,14 @@ def lambda_handler(events: dict, context):
...
         end_date = file_obj.end_date if hasattr(file_obj, "end_date") else None
-        version = file_obj.version
+
+        if not end_date:
+            if isinstance(file_obj, ScienceFilePath):
+                # Set end_date to start_date for science files
+                end_date = file_obj.start_date
+            else:
+                # Set end_date to today's date for ancillary or SPICE files
+                end_date = datetime.today().strftime("%Y%m%d")
```
tech3371, line 306: Can we look for end_date before we set now to be end date?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -782,34 +655,19 @@ def lambda_handler(event, context):
         else None
     )
     if start_date:
```
tech3371: Can we flip logic here? Something like this.
```
    if start_date is None:
        return {
            "statusCode": 200,  # Success
            "body": json.dumps(dependencies),
        }

    end_date = query_params.get("end_date")
    if not end_date:
        return {
            "statusCode": 400,  # Client error
            "body": "end_date not found. If 'start_date' is supplied, "
            "'end_date' is required.",
        }
    end_date = datetime.strptime(end_date, "%Y%m%d")

    upstream_dependencies_output = get_dependency_processing_input(
        query_params=query_params,
        dependencies=dependencies,
        start_date=start_date,
        end_date=end_date,
    )
    if isinstance(upstream_dependencies_output, str):
        return {
            "statusCode": 206,  # Partial content
            "body": dependencies_output,
        }
    upstream_dependencies_output = upstream_dependencies_output.serialize()
    return {
        "statusCode": 200,  # Success
        "body": upstream_dependencies_output,
    }
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -403,8 +403,6 @@ def get_dependency_processing_input(
     query_params: dict,
     dependencies: list,
     start_date: datetime,
```
tech3371, line 404: can we rename `get_dependency_processing_input()` to be `get_upstream_dependency_inputs()`?

tech3371: Looks like `query_params` is not used anymore. can we remove that?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -439,39 +433,12 @@ def get_dependency_processing_input(
...
-                session,
-                dep,
-                start_date,
-                None,
-                end_date,
-                primary_sci_trigger,
-                primary_sci_dep,
-            )
-
+            records = get_files(session, dep, start_date, end_date)
```
tech3371, line 437: I was looking at `get_files()` and noticed that we are query for ancillary files and science files little bit differently.

For ancillary,
```
input_start_date >= table_start_date or input_end_date <= table_end_date
```
For science,
```
input_start_date <= table_start_date or input_end_date >= table_end_date
```
I am trying to think about why.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -460,4 +458,4 @@ def lambda_handler(events: dict, context):
 
         with db.Session() as session:
             for job in potential_jobs + potential_soft_jobs:
-                submit_all_jobs(session, job, start_date, version, data_type, end_date)
+                submit_all_jobs(session, job, start_date, end_date)
```
tech3371: I am wondering if we can refactor some more here. Let me put down ideas on the Galaxy page.

tech3371 (commented):
It already looks simpler! I had some comments and some ideas. Let me comment in a little bit once I put it down and thought it through.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: If you look a few lines up it is only sets end_date to now if end_date is not set.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Yes good idea - I forgot query params are not used anymore.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Yes, because for ancillary we want to get ancillary files that are covering a start_date. 
 
For example if the start date is 20250101 and end_date is also  20250101 we want to get ancillary files with start_dates either before that date or equal to that date and end dates after or equal to the end date. 

                                    start_date (20250101) |--- science file --| end_date (20250101)

              start_date (2024000) |-------------------- ancillary file -------------------| end_date  (20250110)


For science files on the other hand, we want files that within the given range. 

If the range is 2024000 - 20250110 we want science files within that range so start_dates after or equal and end_dates before or equal. 

start_date (2024000) |-------------------- ancillary file -------------------| end_date  (20250110)

                                    start_date (20250101) |--- science file --| end_date (20250101)

lacoak21: basically we just want ancillary files that are covering the range given and science files within the range given.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371, line 1: I added refactor logic for this file on this galaxy page, https://lasp.colorado.edu/galaxy/spaces/IMAP/pages/262373504/Normal+Processing+Refactoring+-+Draft. Let me know if you like to discuss. It's in white text. Red text was old code diagram.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Ok I updated this so the s3_processing_event function will handle the "submit_all_jobs" logic and get the job version.

maxinelasp (commented):
Looks good to me, I think you probably want Tenzin's approval since you two worked on this together, but I didn't see any issues. Nice work going back and cleaning things up, I think this made things a lot simpler.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: I think there is more change to be made here. In above suggestions, I was hoping to return things as soon as possible and avoid other checks. It also helps with following logic easier.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -458,6 +341,101 @@ def lambda_handler(events: dict, context):
...
+        Event input
+    context : LambdaContext
+        Lambda context object
+    """
+    logger.info(f"Events: {events}")
+    logger.info(f"Context: {context}")
+
+    with db.Session() as session:
+        # handle s3 event from the SQS queue
+        s3_processing_event(session, events)
```
tech3371, line 425: logic here looks much better!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: cool. Thank you!

tech3371 (approved):
It's looking great. Thank you for doing this!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Oh that was my bad! I forgot to pop this stashed change. Thanks for catching that.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -458,6 +341,101 @@ def lambda_handler(events: dict, context):
...
+
+            logger.info(f"All required dependencies found for the job: {job}")
+            # Find the first science processingInput that has the same source as the
+            # potential job. Use this to determine the start date.
+            primary_science = upstream_dependencies.get_science_inputs(
+                job["data_source"]
+            )[0]
+            for filepath in primary_science.imap_file_paths:
+                job_start_date = datetime.strptime(filepath.start_date, "%Y%m%d")
+                job_version = determine_job_version(
```
tech3371, line 353: small request. Can you remove `bump_version_number()` function and do that in the this `determine_job_version()`? Since it's one liner, it would be nice to keep that in this function instead of standalone function.

lacoak21: yes definitely.

tech3371: Thank you!

lacoak21: Ofc. That was a good idea! For some reason I had thought I needed to call it somewhere else.

PR #570, opened 2025-05-01
ENH: Adding spin table and test data from MSIM3

### LINKED ISSUE #569: ENH: Create Spin table
### Summary

We went back and forth on this a lot. We decided to create table and log spin files to database to make batch starter logic simpler. This will allow batch start to query table like science, ancillary and SPICE and get latest files for given date range.

### Requirements

The more specific the better.
  - [ ] Create Spin table
  - [ ] Update SPICE indexer lambda to write to spin table when spin data is ingested

### Related tickets

_No response_

tech3371:
# Change Summary
closes #569 

## Overview
Adding spin table and update SPICE indexer to index to spin table also. This is done to simplify and support batch starter logics.

## Updated Files
   - create table
   - Update SPICE indexer to write to spin table too

## Testing
- tests/test-data/test_spice_files/imap_2026_267_2026_267_01.spin.csv
- tests/test-data/test_spice_files/imap_2026_267_2026_267_02.spin.csv


tests/lambda_endpoints/test_spice_indexer_lambda.py
```diff
@@ -147,3 +149,45 @@ def test_s3_spice_files(session, s3_client, events_client):
...
+        os.path.join(test_spice_data_dir, "imap_2026_267_2026_267_02.spin.csv"),
+    )
+    spice_indexer.lambda_handler(spin_file2_event, None)
+    query = select(models.SpinTable.__table__)
+    spin_table_rows = session.execute(query).all()
+    assert len(spin_table_rows) == 2
+    assert spin_table_rows[1].file_path == (
+        temp_path + "/imap/spice/spin/imap_2026_267_2026_267_02.spin.csv"
+    )
+    assert spin_table_rows[1].version == "02"
```
lacoak21, line 193: Maybe add an assert for first and last spin_utc?

lacoak21 (approved):
Im not super familiar with the spin table but the logic looks good!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -265,6 +266,38 @@ def index_spice_file(spice_file: Path):
...
+            "version": spin_metadata["version"],
+            "first_spin_utc": datetime.strptime(
+                spin_df["spin_start_utc"].values.min(),
+                "%Y-%m-%d %H:%M:%S.%f",
+            ),
+            "last_spin_utc": datetime.strptime(
+                spin_df["spin_start_utc"].values.max(),
+                "%Y-%m-%d %H:%M:%S.%f",
+            ),
+            "ingestion_date": datetime.now(),
```
```suggestion
            "ingestion_date": datetime.now(datetime.UTC),
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -265,6 +266,38 @@ def index_spice_file(spice_file: Path):
...
+            "end_date": spin_metadata["end_date"],
+            "version": spin_metadata["version"],
+            "first_spin_utc": datetime.strptime(
+                spin_df["spin_start_utc"].values.min(),
+                "%Y-%m-%d %H:%M:%S.%f",
+            ),
+            "last_spin_utc": datetime.strptime(
+                spin_df["spin_start_utc"].values.max(),
+                "%Y-%m-%d %H:%M:%S.%f",
+            ),
```
greglucas: I'm not entirely following what we will end up querying on, `start_date` or `first_spin_utc`. Will you use both, just one, ...? I'm not sure it really matters if we are adding more information to the table here, but I was thinking that this would be similar to our science files database table where we only reference whats in the filename to be able to search s3 easier.

tech3371: We will need to query just one. I like that suggestions. I will remove `first_spin_utc` and just keep date of filename to line up with file itself.

tests/lambda_endpoints/test_spice_indexer_lambda.py
tech3371: I will be removing first and last spin time from table. But good suggestions.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -265,6 +265,29 @@ def index_spice_file(spice_file: Path):
...
+
+    Parameters
+    ----------
+    spin_file: Path
+        The full name and path the spin file to index
+    """
+    with db.Session() as session:
+        spin_obj = SPICEFilePath(spin_file)
+        spin_metadata = spin_obj.spice_metadata
+        params = {
```
tmplummer, line 279: So are the `start_date` and `end_date` just coming from the filename? If so, will they just be the start of that day, e.g. `2025-05-01 00:00:00`? And then, how will the query be done to ensure that, for example, the returned spin table files for the full day 2025-05-02 cover that full day from midnight to midnight?

tech3371: Let's take an example. Let's say we have spin files such as this where half data is in first file and half is in second files
```
imap_2025_001_2025_002_01.spin.csv
imap_2025_002_2025_005_01.spin.csv
```

We can do query like this, `start_date>= input_date <= end_date` for a given `input_date==2025_002`. That will return both files. Then in `pre_processing`, we can read both of those files.

Let's try a different day with same query, `start_date>= input_date <= end_date`. If `input_data==2025_003`, that would return the second files.

This is my initial thoughts. I can think about it some more.

tech3371: We can end up with little bit more data than we want since we are passing spin files but I feel like that's ok as long as we pass data user needs.

tech3371:
I am going to merge this PR to keep work going. Please don't hesitate to leave comment if I missed anything.

PR #575, opened 2025-05-06
Bug fix for batch starter version calculation 

lacoak21:
# Change Summary

## Overview
The version calculation in the batch starter code contained two bugs: 
1. It only checked for jobs in progress in the processing table. It should get for both in progress and succeeded. 
2. If the descriptor is "all" batch starter only was checking for the max version for any "all" jobs but it should be looking for the max version of all of the jobs - meaning "all" plus any descriptor that falls under that all umbrella. 

## Updated Files
   - Bug fixes

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -152,11 +152,19 @@ def determine_job_version(
         conditions = [
             table.instrument == instrument,
             table.data_level == data_level,
-            table.descriptor == descriptor,
             table.start_date == start_date,
         ]
         if table == models.ProcessingJob:
-            conditions.append(table.status == models.Status.INPROGRESS)
+            conditions.append(
+                table.status.in_(
+                    [models.Status.INPROGRESS.value, models.Status.SUCCEEDED.value]
+                )
+            )
+        elif descriptor != "all":
+            # If the descriptor is not "all", filter by descriptor otherwise, look at
+            # all descriptors
+            conditions.append(table.descriptor == descriptor)
+
         return conditions
 
     # First check to see if there are any jobs in progress and get the max version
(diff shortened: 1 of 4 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -152,11 +152,18 @@ def filter_conditions(table):
...
         ]
         if table == models.ProcessingJob:
-            conditions.append(table.status == models.Status.INPROGRESS)
+            conditions.append(
+                table.status.in_(
+                    [models.Status.INPROGRESS.value, models.Status.SUCCEEDED.value]
+                )
+            )
+        if descriptor != "all":
+            # If the descriptor is all we want to check for all descriptors
```
maxinelasp, line 164: I think we still want to check for "all" in the processing jobs table, just not in the file tracker.... but I think this functionally is the same so if the matching behavior is preferred for you then what you have is ok too.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -152,11 +152,18 @@ def filter_conditions(table):
...
+            conditions.append(table.descriptor == descriptor)
```
maxinelasp, line 165: Could you add a test for this bug?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: This still would check the "all" jobs too I think. So if there was a "mag" "l1a" "all" "v001" job and a "mag" "l1a" "burst-mago" v002"  we would want to use the v002.

maxinelasp: Ok, your logic tracks, I'm happy with this the way it is!

maxinelasp (approved):
Looks good, thank you!

Issue #423, opened 2024-12-19
Add security to APIs


tech3371:
Add security to APIs
- Restrict permission on upload API. Only to POC, and SDC(or web team) and others as needed.
- Restrict permission to data until data is release to public. See #422 for more details

tech3371:
@greglucas completed this. We can create new issues if new requests. closing this issue.

PR #576, opened 2025-05-07
ENH: Add authorization to the API

greglucas:
# Change Summary

## Overview

This adds a JWT authorizer that is managed internally. It only adds the authorization to specific endpoints that are located under `auth` to maintain the public endpoints.

This required a change to HttpApi from RestApi to more easily add an authorizer and CORS handling.

For testing information, reach out to me and I can give you auth flows.

### DIFF
--- a/sds_data_manager/constructs/api_gateway_construct.py
+++ b/sds_data_manager/constructs/api_gateway_construct.py
@@ -10,13 +10,16 @@ https://ialirt.prod.imap-mission.com/ialirt-log-query
 from typing import Optional
 
 from aws_cdk import Duration, aws_sns
-from aws_cdk import aws_apigateway as apigw
+from aws_cdk import aws_apigatewayv2 as apigwv2
+from aws_cdk import aws_apigatewayv2_authorizers as apigwv2_authorizers
+from aws_cdk import aws_apigatewayv2_integrations as apigwv2_integrations
 from aws_cdk import aws_certificatemanager as acm
 from aws_cdk import aws_cloudwatch as cloudwatch
 from aws_cdk import aws_cloudwatch_actions as cloudwatch_actions
 from aws_cdk import aws_lambda as lambda_
 from aws_cdk import aws_route53 as route53
 from aws_cdk import aws_route53_targets as targets
+from aws_cdk import aws_ssm as ssm
 from constructs import Construct
 
 from sds_data_manager.constructs.route53_hosted_zone import DomainConstruct
@@ -50,7 +53,6 @@ class ApiGateway(Construct):
             Prefix for ialirt domain, Optional
         kwargs : dict
             Keyword arguments
-
         """
         super().__init__(scope, construct_id, **kwargs)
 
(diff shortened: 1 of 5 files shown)


sds_data_manager/constructs/sds_api_manager_construct.py
```diff
@@ -95,10 +95,14 @@ def __init__(
         upload_api_lambda.apply_removal_policy(cdk.RemovalPolicy.DESTROY)
 
         api.add_route(
-            route="upload",
+            route="/upload/{proxy+}",
+            http_method="GET",
+            lambda_function=upload_api_lambda,
+        )
+        api.add_route(
+            route="/authorized/upload/{proxy+}",
```
tech3371, line 104: Is this another way of passing ``/download/{filename}`` style routes? If so, can you add comment above upload and donwload route line about that?

sds_data_manager/constructs/sds_api_manager_construct.py
```diff
@@ -95,10 +95,14 @@ def __init__(
         upload_api_lambda.apply_removal_policy(cdk.RemovalPolicy.DESTROY)
 
         api.add_route(
-            route="upload",
+            route="/upload/{proxy+}",
```
tech3371, line 99: Thank you for keeping those around for now. It helps everyone to continue working with current setup.

tech3371 (approved):
Looks great to me. Thank you!

tech3371:
I believe this completes most of #423 work. We just need to remove and move some of API behind auth. But it would helpful to do those after SIT-4.

sds_data_manager/constructs/sds_api_manager_construct.py
greglucas: Added a comment above each of these :+1:

sds_data_manager/constructs/api_gateway_construct.py
```diff
@@ -61,13 +63,34 @@ def __init__(
...
-            description="API Gateway for lambda function endpoints.",
-            endpoint_types=[apigw.EndpointType.REGIONAL],
+        # Start with an empty domain name mapping and fill it in within
+        # the domain construct if necessary within the lower if-block
+        domain_mapping = None
+
+        # NOTE: We look these up from the account parameter store. To update
+        # these values, run the following command:
+        #
+        # aws ssm put-parameter --name lasp-auth-issuer --value <issuer> --type String
```
lacoak21, line 73: 👍 I think using parameter store works great here.

sds_data_manager/constructs/api_gateway_construct.py
```diff
@@ -76,33 +99,47 @@ def __init__(
...
             )
 
+        # Create a single HTTP API Gateway
+        self.api = apigwv2.HttpApi(
+            self,
+            f"{self.lowercase_prefix}HttpApi",
+            api_name=f"{self.prefix}HttpApi",
+            default_domain_mapping=domain_mapping,
+            description="HTTP API Gateway for lambda function endpoints.",
+            cors_preflight={
```
lacoak21, line 133: So v1 must have had some default cors enabled?

lacoak21 (approved):
LGTM!

tech3371 (approved):

sds_data_manager/constructs/api_gateway_construct.py
greglucas: No I don't think so, this was actually requested from the web team because their websites were getting rejected with CORS errors. I think the website just hasn't been used a lot yet and we do have some CORS headers added manually throughout the code, so good point here. I'll make a follow-up ticket to remove those CORS headers as they are now taken care of here.

PR #577, opened 2025-05-07
SPICE Metakernel improvements

### LINKED ISSUE #555: ENH: Add feature to query for metakernel with some filter options
### Summary

This feature is needed in processing pipeline and processing kickoff. SPICE is added to dependencies tree like this:
```
# primary_source, primary_data_type, primary_descriptor, dependent_source, dependent_data_type, dependent_descriptor, relationship, dependency_type

attitude_history, spice, historical, swapi, l3, sci, HARD, DOWNSTREAM
```
It means SWAPI L3 needs kernel type `attitude_history` and it's of `spice` data type and it only wants `historical` kernels. That `primary_source` line ups with what Bryan defined on `imap-data-access` [here](https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/main/imap_data_access/file_validation.py#L370) or we can change that. At one point we thought of using these as `primary_source`.
1. sc_attitude
2. sc_ephemeris
3. planet_ephemeris
4. time_kernel
5. thruster
6. sc_spin
7. sc_repoint
8. sdc_pointing_frame

Some filter options that people listed when we did SPIKE were these:

1. predict
2. historical
3. reconstruct
4. nominal
5. best(best means give me historical if that exist. otherwise give me predict)

I am bringing notes Tim put on past PR t help with above.
```
The notes I took have the following types of ephem predicts:

4-week
- The predicted position and velocity of the spacecraft as a function of time ==without any future maneuvers== with a duration of 4 weeks in the future.
- Burn
- The predicted position and velocity of the spacecraft as a function of time ==with any future maneuvers== with a duration of 4 weeks in the future.
- 90-Day
- The predicted position and velocity of the spacecraft as a function of time ==without any future maneuvers== with a duration of 90 days in the future.
- DSN Long Term
- The predicted position and velocity of the spacecraft as a function of time ==without any future maneuvers== with a duration of 6 months in the future.

Additionally, the "nominal" ephemeris is a predict of the entire mission:

Generated by mission design, it contains the latest design of the spacecraft trajectory for the entire mission, optimized to meet all science objectives.
```

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

bryan-harter:
# Change Summary

## Overview
Adding a few things requested from the previous telecon:

- "list_files=True" now returns only the list of files, rather than the big JSON with everything in it
- "require_coverage=True" will now return a "422" error and return no files if any gaps are still detected
- "file_types=x,y,z" will now limit the file types it uses to build the metakernel. You could use this method to get only the predicts or finals if you wanted

## Updated Files
- sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
   - Modified return_spice_files_in_order so that you can choose if it returns all data or just the list
   - Added helper function "contains_gaps" to determine if any gaps remain
- sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
   - I modified things to use Enums rather than hard-coding in lists of strings which was starting to feel clunky
   - If "file_types" is given as a query parameter, only spice files matching a string in "file_types" will be loaded
   - If "require_coverage" is given as a query parameter, a 422 status code is returned and a list of SPICE file types and gaps if any gaps in the data are detected.
   - Update the gap size based on what I was told by the MOC
   - Updated our test file to reflect that there are now 395 gaps in it because of change in previous file 🫢 
   - Added tests for both new query parameters, and modified the tests because of the change of behavior in "list_files" 

## Testing
Added a unit test for the new "require_coverage" and "file_types" parameters to the API.


bryan-harter:
One thing I wasn't sure about is if this should even be an API any more, or if it should just be wrapped up into Batch starter for simplicity. Is there a use case for allowing people to use it anywhere else?

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
```diff
@@ -12,6 +14,100 @@
 logger.setLevel(logging.INFO)
 
 
+class LeapsecondKernels(Enum):
+    """Container for Leapsecond Kernel Types."""
+
+    LEAPSECONDS = auto()
+
+    @classmethod
```
tmplummer: Is there a reason that all of these `spice_category_name` methods are `@classmethod`s instead of `@staticmethod`s?

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
```diff
@@ -57,85 +169,35 @@ def lambda_handler(event, context):
...
-        allowed_spice_types=[
-            "leapseconds",
-            "planetary_constants",
-            "frames",
-            "spacecraft_clock",
-            "planetary_ephemeris",
-            "spacecraft_ephemeris",
-            "spacecraft_attitude",
-        ],
+        allowed_spice_types=[c.spice_category_name() for c in IMAP_SPICE_LOAD_ORDER],
```
tmplummer, line 193: Cool, defining `IMAP_SPICE_LOAD_ORDER` really made this section simpler!

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
```diff
@@ -57,85 +169,35 @@ def lambda_handler(event, context):
...
-                "spacecraft_ephemeris",
-                "file_intervals_j2000",
-                priority_field="timestamp",
-            )
-
-    for attitude_type in ["attitude_history", "attitude_predict"]:
-        if len(metakernel.spice_gaps["spacecraft_attitude"]) > 0:
-            attitude_files = spice_query_api.lambda_handler(
+    for spice_category in IMAP_SPICE_LOAD_ORDER:
+        for spice_subtype in spice_category:
```
tmplummer, line 197: This means that the order in which each spice category Enum is defined affects the load order within that category. I see:
```
class SpacecraftAttitudeKernels(Enum):
    """Container for Spacecraft Attitude Kernel Types."""

    ATTITUDE_HISTORY = auto()
    ATTITUDE_PREDICT = auto()
```
Which I think means any attitude predict will get loaded after the attitude history. That should be reversed. It also looks like the `SpacecraftEphemerisKernels` are backwards.

If I am correct in this logic, can you fix and also add comment in the Enum definitions that the defined order is reverse priority order?

tmplummer (commented):
Mostly looks good. I have just one major concern to look in to.

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
```diff
@@ -142,7 +148,10 @@ def return_spice_files_in_order_detailed(self) -> list[dict]:
         for type in self.allowed_spice_types:
             if self.spice_files[type]:
                 metakernel_files.extend(reversed(self.spice_files[type]))
-        return metakernel_files
+        if detailed:
+            return metakernel_files
+        else:
+            return [f["file_name"] for f in metakernel_files]
```
tech3371, line 154: if not detailed, can we return just the filename with no path. Eg. `"imap_1000_065_1000_090_003.ap.bc"` instead of `"ck/imap_1000_065_1000_090_003.ap.bc"`?

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
```diff
@@ -12,6 +14,100 @@
 logger.setLevel(logging.INFO)
 
 
+class LeapsecondKernels(Enum):
```
tech3371, line 17: I don't what is the best place, but can you add somewhere that these names must match what's defined [here](https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/main/imap_data_access/file_validation.py#L420). This way, if someone where to make change, they will know to check here as well. Then may be in the future once we get to a good place, we should move these class to `imap-data-access`?

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
```diff
@@ -12,6 +14,100 @@
...
+    """Container for Spacecraft Attitude Kernel Types."""
+
+    ATTITUDE_HISTORY = auto()
+    ATTITUDE_PREDICT = auto()
+
+    @classmethod
+    def spice_category_name(cls):
+        """Category of SPICE file."""
+        return "spacecraft_attitude"
+
```
tech3371, line 109: Can you add class for the `pointing_attitude` that @laspsandoval and @subagonsouth added on `imap-data-access` recently?

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
bryan-harter: Yeah I had to think about it for a sec too so I know its confusing. So first it will look through all the attitude_history files to fill the time ranges, and then the attitude_predict files to fill in any remaining gaps. So we always want the highest priority files to be called with "load_spice" first, so we can determine what gaps are left over after them. 

But when it comes time to actually make the metakernel, the whole list of files for each file type is reversed (line 150 in metakernel.py), so that any attitude predict files are at the top of the metakernel, and the attitude_history are at the bottom. So attitude history still takes precedence in the end. 

Maybe "load" isn't the right word for the function, maybe something involving "check" or "analyze" or something. Because it doesn't necessarily add these files to the metakernel here, it just looks to see if any of the files provided could potentially fill in the gaps.

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
bryan-harter: Ok glad you think so, I was worried I was trying to get too fancy

tests/lambda_endpoints/test_spice_metakernel_api.py
```diff
@@ -216,3 +216,43 @@ def test_metakernel(session):
...
+    _insert_test_data(session)
+
+    result = spice_metakernel_api.lambda_handler(
+        {
+            "queryStringParameters": {
+                "start_time": 1,
+                "end_time": 100,
+                "spice_path": "",
+                "list_files": "True",
+                "file_types": "leapseconds,spacecraft_clock",
```
tech3371, line 252: can you show an example of how we can query for only historical or predict attitude or ephemeris here?

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
bryan-harter: Not really, my first pass at these functions actually did use the "cls" to retrieve the name of the class and strip off the word "Kernels" to get the category of file, but I was like that's way too complicated. I'm just going to hard-code it. So yeah they should be staticmethods now.

tests/lambda_endpoints/test_spice_metakernel_api.py
tech3371: Also out of curiosity, when we pass in `file_types` to the API, do we use the `spice_category_name()` defined in each class?

tech3371 (commented):
It's looking great. Thank you for doing that so quickly! I had few clarification questions.

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
bryan-harter: Sure, I can do that

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
bryan-harter: Yeah I think it would be great to have them in a centralized place so we ensure the strings match between the two. We can put these in imap-data-access at some point so there is only ever 1 spot where we need to actually write out the strings.

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
bryan-harter: Yeah I can add that in this commit

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
tech3371: Thank you!

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py
tech3371: Thank you!

tests/lambda_endpoints/test_spice_metakernel_api.py
bryan-harter: With the enums, I was having the class act as a container for all the sub-types of the kernels. So for things like "leapseconds" and "spacecraft_clock", we only have one "sub-type" of that kind of file, which are **also** called "leapseconds" and "spacecraft_clock". 

But with spacecraft attitude kernels, we have two sub-types, "attitude_history" and "attitude_predict". And with spacecraft ephemeris, we have much more. 

So what you are actually specifying in "file_types" here in the sub-types of the kernels. If you want, for instance, a full metakernel with *no predicts*, you'd specify:

`file_types=leapseconds,spacecraft_clock,planetary_constants,frames,planetary_ephemeris,ephemeris_reconstructed,attitude_history`. 

or if you wanted only the predicts:

`file_types=leapseconds,spacecraft_clock,planetary_constants,frames,planetary_ephemeris,ephemeris_nominal,ephemeris_predicted,ephemeris_90days,ephemeris_long,ephemeris_launch,attitude_predict`.

tech3371: ah ok. So we can pass in multiple types of historical and/or predict. Do we need to careful with order or do you handle that in the back? Eg. if I want predict/best ephemeris, do I need to be aware of order of types I pass in, `ephemeris_reconstructed, ephemeris_predict, ephemeris_nominal` or can those be passed in any order and metakernel figures out the order?

bryan-harter: You can pass them in any order!

tech3371: Perfect!

tech3371 (approved):
I created issue on `imap-data-access` to transition class from this PR to `imap-data-access` in the future. Besides that, I think I understand where we are going with this and looks great to me!

bryan-harter:
Ok I believe I've got everyone's comments answered or merged in. I'll leave it open for a little bit longer just in case, but I'll probably merge it by the end of the day

PR #580, opened 2025-05-08
Refactor ancillary fetching

### LINKED ISSUE #579: Ancillary filtering is incorrect for bulk reprocessing and cadence
Currently, when we are querying for upstream files in the dependency lambda we already require a data range. 

For example if it is a normal processing science file the start_date might be 20250101 and end_date might be 20250101. 

The dependency lambda will look for ancillary files with start_dates <= to the query start date and end dates >= query end_date. This works in the case above. 

If our range is greater than that - for bulk reprocessing or cadence, 

we need the ancillary query to look for <= the end date. This is because we could get back a science file with a start_date equal to the end date and we need to check for ancillary files that are valid for that end date. If we keep the original query we could be missing out on more recent/valid ancillary files with start_dates earlier than the query start date. 

For example if the range is 20240101 - 20250101 

Upstream science files:
  idex_l1a_sci_20250101

Ancillary_files: 
 idex_l1a-calibration_20241201-20251201

The original query would NOT return the ancillary file listed. 

To fix this, we need to update the ancillary query to look for files with start_dates <= query end date and end_dates >= query start_date. 

This will ensure that we have all valid ancillary files for any science file found in that range. Batch starter will then filter the ProcessingInputCollection for the most recent valid ancillary file.

lacoak21:
# Change Summary

closes #579
## Overview
Currently, when we are querying for upstream files in the dependency lambda we already require a data range.

For example if it is a normal processing science file the start_date might be 20250101 and end_date might be 20250101.

The dependency lambda will look for ancillary files with start_dates <= to the query start date and end dates >= query end_date. This works in the case above.

If our range is greater than that - for bulk reprocessing or cadence,

we need the ancillary query to look for <= the end date. This is because we could get back a science file with a start_date equal to the end date and we need to check for ancillary files that are valid for that end date. If we keep the original query we could be missing out on more recent/valid ancillary files with start_dates earlier than the query start date.

For example if the range is 20240101 - 20250101

Upstream science files:
idex_l1a_sci_20250101

Ancillary_files:
idex_l1a-calibration_20241201-20251201

The original query would NOT return the ancillary file listed.

To fix this, we need to update the ancillary query to look for files with start_dates <= query end date and end_dates >= query start_date.

This will ensure that we have all valid ancillary files for any science file found in that range. Batch starter will then filter the ProcessingInputCollection for the most recent valid ancillary file.

Updated FIles
   - use return_latest_ancillary boolean 
   - update query 

## Testing
Fix tests to work with updated ancillary file dates


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -485,19 +485,18 @@ def get_files(
...
         table = models.AncillaryFiles
-        # Query for ancillary files whose ranges cover the
-        # start date and end date.
-        # E.g., if the start date is '20250102', the query could return an ancillary
-        # file with the date range ('20250101', '20250103')
+        # Query for ancillary files where the start_date is less than or equal to
+        # the end_date, and the end_date is either greater than or equal to the
+        # start_date or is None. For example, if the start_date is '20250102' and
+        # the end_date is '20250103', the query could return an ancillary file with
+        # the date range ('20250101', '20250104').
```
tech3371: With this example, I couldn't differentiate which is input date and which is ancillary files' date. Can you state example like we drew on whiteboard. for given input, which files are returned with this query?

tests/lambda_endpoints/conftest.py
```diff
@@ -311,11 +311,22 @@ def _populate_file_catalog(session):
...
-            end_date=datetime(2024, 1, 4),
+            start_date=datetime(2023, 1, 2),
+            version="v001",
+            extension="cdf",
+            ingestion_date=datetime.strptime(
+                "2024-01-25 23:35:26+00:00", "%Y-%m-%d %H:%M:%S%z"
+            ),
+        ),
+        AncillaryFiles(
+            file_path="/path/to/imap_swe_l1b-in-flight-cal_20240105_20250104_v002.cdf",
```
tech3371: Can you change to something like this? For a second, I thought you flipped the start and end date but realized later that you had different year.
```suggestion
            file_path="/path/to/imap_swe_l1b-in-flight-cal_20240104_20240106_v002.cdf",
```

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -208,7 +208,7 @@ def test_lambda_handler_ancillary_event(
         # as an upstream dep.
         ancillary_in = [
             AncillaryInput(
-                "imap_swe_l1b-in-flight-cal_20231231_20240104_v002.cdf",
+                "imap_swe_l1b-in-flight-cal_20230102_v001.cdf",
```
tech3371, line 211: Below dependency tests makes sense why it return all those valid ancillary files. Then we filter that in batch starter before you pass that to batch job, right? I understand right now why you can't test it because it's inside a function but would be nice if we can validate that here somehow.

tech3371: may be the test on `imap-data-access` could be enough?

tech3371 (approved):
couple minor comments. besides that, it looks good to me.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Oops yes that is very confusing 😄 I will specify which one is the input.

tests/lambda_endpoints/test_batch_starter.py
lacoak21: So technically this is a validation that the ancillary files got filtered because as the comment above says:

```
Even though there are two imap_swe_l1b-in-flight-cal ancillary files that
have valid dates, there should be only be the most recent one returned
as an upstream dep.
```

It is hard to show that there are two valid ancillaries for that date but it is working as expected.

lacoak21: I had to update the date because I had changed it in another PR. It is so hard to keep these tests all stable 🥲

tech3371: Ah ok. So `mock_batch_client.submit_job.assert_called_with()` checks that batch job was submitted with that input? cool. I didn't know that.

lacoak21: Yes exactly

PR #582, opened 2025-05-09
Bulk reprocessing Event 

lacoak21:
# Change Summary

## Overview
Add handling for a bulk reprocessing event. This does not include the api to call this. That will be in a separate PR. 

## Updated Files
   - add bulk event
   - Handle finding dependencies when no instrument or data_level or descriptor is specified. 

## Testing
Bulk reprocessing tests.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -200,7 +200,7 @@ def submit_all_jobs(session, job, start_date, end_date):
     start_date : str
         Start date to query the data.
     end_date : str
-        Start date to query the data.
+        End date to query the data.
     """
     # Submit downstream jobs for each upstream primary science dependency file.
     # Find the files that this job depends on
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -151,7 +151,7 @@ def try_to_submit_job(
 
     # Reformat the upstream dependencies from dependency call to match
     # what batch job expects.
-
+    # TODO: do we need a reprocessing column to indicate if this is a reprocessing job?
     batch_command = [
         "--instrument",
         instrument,
@@ -187,6 +187,60 @@ def try_to_submit_job(
     )
     logger.info(f"Submitted job {job_name} with this command: {batch_command}")
 
+
+def submit_all_jobs(session, potential_jobs, start_date, end_date):
+    """Submit all jobs for the given job and upstream dependencies.
+
+    Parameters
+    ----------
+    session : orm session
+        Database session.
+    potential_jobs : list[dict]
+        List of job node to run.
+    start_date : str
+        Start date to query the data.
+    end_date : str
+        Start date to query the data.
+    """
+    for job in potential_jobs:
+        # Submit downstream jobs for each upstream primary science dependency file.
+        # Find the files that this job depends on
+        upstream_dependencies = dependency.get_jobs(
+            data_source=job["data_source"],
+            data_type=job["data_type"],
+            descriptor=job["descriptor"],
+            dependency_type="UPSTREAM",
+            relationship="ALL",
+            start_date=start_date,
+            end_date=end_date,
+        )
+        if not upstream_dependencies:
+            return
+
+        logger.info(f"All required dependencies found for the job: {job}")
+        # Find the first science processingInput that has the same source as the
+        # potential job. Use this to determine the start date.
+        primary_science = upstream_dependencies.get_science_inputs(job["data_source"])[
+            0
+        ]
+        for filepath in primary_science.imap_file_paths:
+            job_start_date = datetime.strptime(filepath.start_date, "%Y%m%d")
+            job_version = determine_job_version(
+                session=session,
+                instrument=job["data_source"],
+                descriptor=job["descriptor"],
+                start_date=job_start_date,
+                data_level=job["data_type"],
+            )
+            # Filter dependencies to get only files needed for this job
+            try_to_submit_job(
+                session,
+                job,
+                job_start_date,
+                job_version,
+                upstream_dependencies.get_valid_inputs_for_start_date(job_start_date,
+                                                                      return_latest_ancillary=True),
+            )
 
 def s3_processing_event(session, events):
     """Process SQS events that were triggered by S3 file arrivals.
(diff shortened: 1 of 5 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -272,6 +272,61 @@ def try_to_submit_job(
     logger.info(f"Submitted job {job_name} with this command: {batch_command}")
 
 
+def submit_all_jobs(session, job, start_date, end_date):
```
lacoak21: @tech3371  I know we decided to remove this function but there is so much repeated logic that I am hoping to add it back in.

tech3371: cool. And I will review this PR early next week.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -333,47 +388,72 @@ def s3_processing_event(session, events):
...
+                "If no instrument is provided, descriptor must not be specified."
+            )
+        # If data_level is not provided, we need to reprocess all levels.
+        # To get potential jobs, first we query for jobs directly downstream from l0 raw
+        # files. This will create a domino effect in which all downstream files get
+        # reprocessed.
+        event = {
+            "data_type": "l0",
+            "dependency_type": "DOWNSTREAM",
+            "relationship": "ALL",
```
tech3371: we don't have any SOFT_NO_TRIGGER, but I think we will want to exclude those from reprocessing if they existed.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -350,7 +350,21 @@ def get_dependencies(node, dependency_type, relationship):
 
     dependencies = []
     for rel in relationships:
-        deps = dependency_config.dependencies[rel][dependency_type].get(node, [])
+        all_deps_for_type = dependency_config.dependencies[rel][dependency_type]
+        if None in query_node:
```
tech3371: What caused this case? Just curious. `node` should always have these three parts, (source, data_type, descriptor). If we don't have that, then that's a concern to me.

lacoak21: For bulk reprocessing, there can be a case where we dont have a full node. So for example if we are looking for downstream dependencies for all instruments, there would be no instrument and no descriptor provided and the node would look like (None,l0, None). In this case we would look for nodes that have l0 at the second index.

tech3371: I am concern with passing none to this query. I am wondering if we can search what you need through new utils function. A function that looks up those things you need for bulk reprocessing. I am concern with making this more generic and causing confusion down the road.

tech3371: I am looking at where that could fit and wondering if you can create a method in `DependencyConfig()` class. That's first one that comes to mind.

lacoak21: Ok I see what you mean. I want to push back a little bit because the two functions will have so much overlap.

lacoak21: After more thought, maybe it would make more sense to have these separated out. I will think about the best way to do this.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -299,6 +299,36 @@ def _validate_node(self, node: tuple) -> bool:
         # data descriptor.
         return True
 
+    def kickoff_pipeline_jobs(self) -> list:
```
lacoak21, line 305: Let me know what you think @tech3371 ! Hopefully this will make more sense to people.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Ok found a potential solution. Instead of doing this, I added the function kickoff_pipeline_jobs()

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -538,30 +569,30 @@ def get_files(
...
 
 def get_jobs(
-    data_source: str,
-    data_type: str,
-    descriptor: str,
     dependency_type: str,
     relationship: str,
+    data_source: Optional[str] = None,
+    data_type: Optional[str] = None,
+    descriptor: Optional[str] = None,
```
tech3371: I think these are now required based on above changes.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: This function looks good.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -399,6 +399,63 @@ def test_spice_file(session):
...
+    }
+    context = {"context": "sample_context"}
+    with pytest.raises(ValueError, match="instrument and descriptor are required"):
+        lambda_handler(events, context)
+    # Add instrument and try again
+    events["instrument"] = "swe"
+    with patch.object(batch_starter, "try_to_submit_job") as mock_submit:
+        lambda_handler(events, context)
+    # There should be three different jobs submitted for swe l1b
+    assert mock_submit.call_count == 3
```
tech3371: Is this 3 jobs because there are three l1a file that can be reprocessed? If so, can you add that in the comment. For a moment, I thought it was because there are three downstream dependents that were kicked off.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -399,6 +399,63 @@ def test_spice_file(session):
...
+    """Tests ``lambda_handler`` when there is bulk reprocessing for all instruments."""
+    _populate_file_catalog(session)
+    # leave instrument, data_level and descriptor blank
+    events = {"reprocessing": True, "start_date": "20230101", "end_date": "20260101"}
+    context = {"context": "sample_context"}
+    # Add instrument and try again
+    with patch.object(batch_starter, "try_to_submit_job") as mock_submit:
+        lambda_handler(events, context)
+    # There should be two jobs submitted, one for swe, one for lo based off the existing
+    # l0 files
```
tech3371, line 425: Ah ok. This comment is exactly what I am looking for above.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -399,6 +399,63 @@ def test_spice_file(session):
...
+        "reprocessing": True,
+        "start_date": "20230101",
+        "end_date": "20260101",
+        "instrument": "swe",
+    }
+    context = {"context": "sample_context"}
+    # Add instrument and try again
+    with patch.object(batch_starter, "try_to_submit_job") as mock_submit:
+        lambda_handler(events, context)
+    # There should be one job submitted for swe
```
tech3371, line 445: same here

tech3371 (approved):
It looks great! I had few minor comments that non-blocking.

tests/lambda_endpoints/test_batch_starter.py
lacoak21: I will add a comment to clarify.

PR #588, opened 2025-05-12
Remove Dependency api

lacoak21:
# Change Summary

## Overview
Remove the dependency API construct and call handler directly from batch starter. Call the dependency handler again after finding the start date of the file to process instead of filtering the original collection. 

## Updated Files
   - Call dependency handler directly 
   -remove api handling

## Testing
Refactor tests to not mock the api call. 
Remove tests that test the api response.


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -328,11 +239,11 @@ def s3_processing_event(session, events):
         }
         # Potential jobs are the instruments that depend on the current file,
         # which are the downstream dependencies.
-        potential_jobs = _get_dependencies(dependency_event_msg)
-
+        # potential_jobs = _get_dependencies(dependency_event_msg)
+        potential_jobs = dependency.lambda_handler(dependency_event_msg)
```
greglucas: Rather than calling the method `lambda_handler(event)`, could you instead change this to a signature like: `dependency.get_jobs(start_date, end_date, relationship, ...)`? I think that would make it more clear exactly what is being passed around and what the method does.

greglucas (approved):
I like this a lot! Very nice refactor.

Minor suggestion on naming and making it more pythonic-based than AWS-based if that makes sense.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -328,11 +239,11 @@ def s3_processing_event(session, events):
         }
         # Potential jobs are the instruments that depend on the current file,
         # which are the downstream dependencies.
-        potential_jobs = _get_dependencies(dependency_event_msg)
-
+        # potential_jobs = _get_dependencies(dependency_event_msg)
```
```suggestion
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -366,15 +277,18 @@ def s3_processing_event(session, events):
...
+                upstream_deps_for_start_date = dependency.lambda_handler(event_msg)
                 try_to_submit_job(
                     session,
                     job,
                     job_start_date,
                     job_version,
-                    upstream_dependencies.get_valid_inputs_for_start_date(
-                        job_start_date, return_latest_ancillary=True
-                    ),
+                    upstream_deps_for_start_date,
```
tech3371, line 297: Thank you for changing this. This will help with SPICE work I have coming up.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: I feel like we documented enough in the lambda handler to understand what that is doing. I personally like passing event but I don't feel strongly about it.

tech3371 (approved):
LGTM

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: That is a good point. I think I like changing it to be more python-based. Passing around the event now seems to be outdated since it is no longer an api.

PR #590, opened 2025-05-13
ENH: Adding SPICE logics to batch starter and dependency

### LINKED ISSUE #586: ENH: Implement SPICE dependency in batch starter
This includes work of implementing SPICE dependency tree logic in batch starter lambda.

tech3371:
# Change Summary
closes #586

## Overview
Main changes for adding SPICE dependency.

## Updated Files
   - updated how inputs to dependency are constructed based on what file type it is.
   - main changes. Added logics to query and build spice dependencies
   - added handling for spin, repoint, and kernels.

## Testing
updated tests to test above cases

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -212,7 +212,8 @@ def s3_processing_event(session, events):
         if isinstance(file_obj, SPICEFilePath):
             # Add source, data_type, and descriptor to the downstream event message
             spice_input = SPICEInput(filename)
-            data_source = spice_input.source
+            # TODO: fix this in future release of imap_data_access
+            data_source = file_obj.spice_metadata["type"]
             descriptor = spice_input.descriptor
             data_type = spice_input.data_type
             # Set the start and end dates for the upstream event message.
@@ -262,6 +263,10 @@ def s3_processing_event(session, events):
             dependency_type="DOWNSTREAM",
             relationship="SOFT_TRIGGER",
         )
+        logger.info(
+            f"Potential jobs: {potential_jobs} and potential soft jobs: "
+            f"{potential_soft_jobs}"
+        )
 
         for job in potential_jobs + potential_soft_jobs:
             # Submit downstream jobs for each upstream primary science dependency file.
@@ -278,9 +283,35 @@ def s3_processing_event(session, events):
             )
 
             if not upstream_dependencies:
-                return
+                continue
 
             logger.info(f"All required dependencies found for the job: {job}")
+            # data_source -> spacecraft and pointing kernel is a unqiue case
+            # where we need to kick off with what's in upstream dependencies
+            # without filtering.
+            if (
+                job["data_source"] == "spacecraft"
+                and job["descriptor"] == "pointing_attitude"
+            ):
+                # Convert to datetime object
+                job_start_date = datetime.strptime(start_date, "%Y%m%d")
+                job_version = determine_job_version(
+                    session=session,
+                    instrument=job["data_source"],
+                    descriptor=job["descriptor"],
+                    start_date=job_start_date,
+                    data_level=job["data_type"],
+                )
+                # submit the job
+                try_to_submit_job(
+                    session,
+                    job,
+                    job_start_date,
+                    job_version,
+                    upstream_dependencies,
+                )
+                continue
+
             # Find the first science processingInput that has the same source as the
             # potential job. Use this to determine the start date.
             primary_science = upstream_dependencies.get_science_inputs(
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -242,7 +242,7 @@ def s3_processing_event(session, events):
             # Set the start and end dates for the upstream event message
             start_date = file_obj.start_date
             # Ancillary files can have an end date.
-            end_date = file_obj.end_date if hasattr(file_obj, "end_date") else None
+            end_date = getattr(file_obj, "end_date", None)
 
         # Potential jobs are the instruments that depend on the current file,
         # which are the downstream dependencies.
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -217,11 +217,8 @@ def s3_processing_event(session, events):
             descriptor = spice_input.descriptor
             data_type = spice_input.data_type
             # Set the start and end dates for the upstream event message.
-            # TODO: fix date range if file was repoint file.
-            if file_obj.spice_metadata["type"] == "repoint":
-                # Repoint file doesn't have a start date.
-                # TODO: So set start date to be what?
-                start_date = file_obj.spice_metadata["start_date"]
+            # TODO: fix date range if/when repoint file ingestion event is
+            # passed to batch starter to kickoff HARD or SOFT_TRIGGER downstream jobs.
             # Convert datetime object to string of format YYYYMMDD
             start_date = file_obj.spice_metadata["start_date"].strftime("%Y%m%d")
             end_date = file_obj.spice_metadata["end_date"].strftime("%Y%m%d")
@@ -286,31 +283,6 @@ def s3_processing_event(session, events):
                 continue
 
             logger.info(f"All required dependencies found for the job: {job}")
-            # data_source -> spacecraft and pointing kernel is a unqiue case
-            # where we need to kick off with what's in upstream dependencies
-            # without filtering.
-            if (
-                job["data_source"] == "spacecraft"
-                and job["descriptor"] == "pointing_attitude"
-            ):
-                # Convert to datetime object
-                job_start_date = datetime.strptime(start_date, "%Y%m%d")
-                job_version = determine_job_version(
-                    session=session,
-                    instrument=job["data_source"],
-                    descriptor=job["descriptor"],
-                    start_date=job_start_date,
-                    data_level=job["data_type"],
-                )
-                # submit the job
-                try_to_submit_job(
-                    session,
-                    job,
-                    job_start_date,
-                    job_version,
-                    upstream_dependencies,
-                )
-                continue
 
             # Find the first science processingInput that has the same source as the
             # potential job. Use this to determine the start date.
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -7,10 +7,15 @@ from datetime import datetime
 import boto3
 import imap_data_access
 from imap_data_access import (
+    AncillaryFilePath,
     ScienceFilePath,
     SPICEFilePath,
 )
-from imap_data_access.processing_input import ProcessingInputCollection
+from imap_data_access.processing_input import (
+    ProcessingInputCollection,
+    ProcessingInputType,
+    SPICEInput,
+)
 from sqlalchemy import func
 from sqlalchemy.exc import IntegrityError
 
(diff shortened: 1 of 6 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -204,43 +209,55 @@ def s3_processing_event(session, events):
...
-            else:
-                # Set end_date to today's date for ancillary or SPICE files
-                end_date = datetime.today().strftime("%Y%m%d")
-
-        # TODO: handle spice once implemented
-        data_type = (
-            file_obj.data_level if hasattr(file_obj, "data_level") else "ancillary"
-        )
+            # Add source, data_type, and descriptor to the downstream event message
+            spice_input = SPICEInput(filename)
```
lacoak21, line 214: You could create an "input" for each type to be consistent, and then in the get_jobs call use input.data_source, input.descriptor and input.data_type.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -387,10 +390,132 @@ def primary_science_dep(query_params: dict, dependency: dict) -> bool:
...
+    Parameters
+    ----------
+    dependency : dict
+        Dependency dictionary containing the data source and data type.
+
+    Returns
+    -------
+    str
+        Combined kernel sources separated by commans. Eg.
+        "attitude_history, attitude_predict,..."
```
lacoak21:
```suggestion
        "attitude_history,attitude_predict,..."
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -387,10 +390,132 @@ def primary_science_dep(query_params: dict, dependency: dict) -> bool:
...
+    latest = max(repoint_files, key=lambda x: (x[0], x[1]))
+    latest_file_date = latest[0]
+
+    # Check that input end is within latest repoint file end date
+    if latest_file_date < end_date:
+        logger.info(
+            f"Latest repoint file end date {latest_file_date} "
+            f"is before input end date {end_date}"
+        )
+        return None
```
lacoak21, line 505: So we would never want to return the latest file that is valid for the input end date? In other words we only want the file if it is truly the latest one and it is valid for the input end date?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -387,10 +390,132 @@ def primary_science_dep(query_params: dict, dependency: dict) -> bool:
...
+
+    # Otherwise, return the latest repoint file without the path prefix
+    return latest[2].split("/")[-1]
```
lacoak21: nitpick, but to be consistent with get_spice_files above maybe use basename(latest[2])?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -415,7 +540,91 @@ def get_upstream_dependency_inputs(
...
+                # Define J2000 epoch: 2000-01-01T12:00:00 UTC
+                j2000 = datetime(2000, 1, 1, 12, 0, 0, tzinfo=timezone.utc)
+
+                # Compute seconds difference
+                delta = dt - j2000
+                return delta.total_seconds()
+
+            start_time = yyyymmdd_to_seconds_since_j2000(start_date.strftime("%Y%m%d"))
+            end_time = yyyymmdd_to_seconds_since_j2000(end_date.strftime("%Y%m%d"))
+            metakernel_response = spice_metakernel_api.lambda_handler(
```
lacoak21, line 591: Will this eventually call the api instead? If so we need to add the API URL env variable back in to the batch starter construct.

lacoak21 (commented):
Great job with the SPICE integration Tenzin! I had a few questions.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: I think there were some difference using one vs the other. Let me look into that.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: yeah, I thought about that. That is a good question to us.

Right now, it returns latest repoint file irrespective of what input date was as long as input date is equal or before repoint date. But I don't return latest repoint file if the input date is later than repoint file's date which means that repoint doesn't have repoint data needed by someone.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: no. I am calling lambda directly like you did with dependency. I am not going through API for this.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -204,43 +209,55 @@ def s3_processing_event(session, events):
...
+        elif isinstance(file_obj, AncillaryFilePath):
+            # Add source, data_type, and descriptor to the downstream event message
+            data_source = file_obj.instrument
+            descriptor = file_obj.descriptor
+            # data_type ==> "ancillary"
+            data_type = ProcessingInputType.ANCILLARY_FILE.value
+            # Set the start and end dates for the upstream event message
+            start_date = file_obj.start_date
+            # Ancillary files can have an end date.
+            end_date = file_obj.end_date if hasattr(file_obj, "end_date") else None
```
greglucas:
```suggestion
            end_date = getattr(file_obj, "end_date", None)
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -387,10 +390,132 @@ def primary_science_dep(query_params: dict, dependency: dict) -> bool:
...
+    str
+        Combined kernel sources separated by commans. Eg.
+        "attitude_history, attitude_predict,..."
+    """
+    file_types = []
+    for dep in dependency:
+        if (
+            dep["data_source"] != processing_input.SPICESource.REPOINT.value
+            and dep["data_source"] != processing_input.SPICESource.SPIN.value
+            and dep["data_source"] in spice_metakernel_api.KernelCollection().file_types
```
greglucas: Do you need the first two checks? I assume that a `dep["data_source"]` equal to a SPIN or REPOINT would not be a part of the `KernelCollection().file_types` and thus that would be sufficient here.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: yeah, we don't have that capabilities like we do with file validator. I will add that feature to imap-data-access and then update here. I will do that in new PR.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
greglucas: What about adding this logic to your loop above so that you would only add repoint files to your list if they satisfy your criteria while looping. Then you could do something like: `sorted(repoint_files, lambda x: x[0], x[1])[-1]` on your known satisfied list.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -415,7 +540,91 @@ def get_upstream_dependency_inputs(
     """
     dependency_inputs = processing_input.ProcessingInputCollection()
     with db.Session() as session:
-        for dep in dependencies:
+        # -----------------------------
+        # Check for SPICE dependencies
+        # -----------------------------
+        # If spin is a dependency, query spin table for given date range
+        has_spin_dep = [dep for dep in dependencies if dep["data_source"] == "spin"]
+        if has_spin_dep:
```
greglucas: If you don't need to keep track of the items in the list you can use any() here instead to short-circuit the loop and not store references to the other items. Same with all of these conditionals.
```suggestion
        has_spin_dep = any(dep["data_source"] == "spin" for dep in dependencies)
        if has_spin_dep:
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -415,7 +540,91 @@ def get_upstream_dependency_inputs(
...
+            for dep in dependencies
+            if dep["data_source"] != "spin"
+            and dep["data_source"] != "repoint"
+            and dep["data_type"] == "spice"
+        ]
+        if has_kernel_dep:
+            combined_kernel_sources = combine_kernel_sources(dependencies)
+
+            # convert start_date and end_date in seconds after j2000.
+            # TODO: remove this once Bryan changes takes in 'yyyymmdd' format
```
greglucas, line 577: @bryan-harter are you aware of this request / do we have a ticket for it?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
greglucas: This is another kind of odd mix of internal api calls. Are you able to call the functions directly instead? You are calling the APIs through Python method calls, getting a JSON response and then parsing that. I'm wondering if there is a way we can organize these to allow for nicer call forms internally for us and then have the `lambda_handler()` map to those internal calls. That would allow you to call Python methods instead of API methods here.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: good point, I didn't think about that way.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Oh i see. I dont feel strongly about it because it is clear what you are doing now.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: I asked Bryan already :). I don't think we have a ticket for that though.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: yeah, I was afraid of ripple effect from last PR suggestions :). Can we do that in future PR? I want to get SPICE logic in as soon as possible to give other people chance to test their code before SIT-4.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: cool. I will leave as it is for this PR and then update in future PR once I add that in imap-data-access.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -204,47 +209,64 @@ def s3_processing_event(session, events):
...
+            spice_input = SPICEInput(filename)
+            # TODO: fix this in future release of imap_data_access
+            data_source = file_obj.spice_metadata["type"]
+            descriptor = spice_input.descriptor
+            data_type = spice_input.data_type
+            # Set the start and end dates for the upstream event message.
+            # TODO: fix date range if file was repoint file.
+            if file_obj.spice_metadata["type"] == "repoint":
+                # Repoint file doesn't have a start date.
+                # TODO: So set start date to be what?
```
tmplummer: Could we set start to the mission start date? Any fixed value prior to launch should work.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -261,9 +283,35 @@ def s3_processing_event(session, events):
             )
 
             if not upstream_dependencies:
-                return
+                continue
 
             logger.info(f"All required dependencies found for the job: {job}")
+            # data_source -> spacecraft and pointing kernel is a unqiue case
```
tmplummer: I don't understand what this comment means. Can you help explain it to someone unfamiliar with filtering and kicking off upstream dependencies?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -261,9 +283,35 @@ def s3_processing_event(session, events):
...
+            # where we need to kick off with what's in upstream dependencies
+            # without filtering.
+            if (
+                job["data_source"] == "spacecraft"
+                and job["descriptor"] == "pointing_attitude"
+            ):
+                # Convert to datetime object
+                job_start_date = datetime.strptime(start_date, "%Y%m%d")
+                job_version = determine_job_version(
```
tmplummer: This may be unique for this job as well. We are going to produce one kernel file per pointing and there is no way to know what the file version number will be a priori. We should discuss these details.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -387,10 +390,128 @@ def primary_science_dep(query_params: dict, dependency: dict) -> bool:
...
+        Dependency list containing dictionary of SPICE dependencies.
+    start_date : datetime
+        Start date to find dependent files with.
+    end_date : datetime
+        End date to find dependent files with.
+
+    Returns
+    -------
+    list
+        List of spin files.
```
tmplummer, line 440: This is great! I will use this comment to generate a new ticket.

The spin data functionality will need to join data from multiple spin-tables.

tmplummer (commented):
A couple of comments, but nothing major. Looking good to me.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: We discussed it on a call.

tech3371:
I think I address all of comments or capture in issue if can't be address here. Can someone review again and approve it? Appreciate it. I will open new PR to incorporate new changes from `imap-data-access` to keep this going forward.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -415,7 +536,89 @@ def get_upstream_dependency_inputs(
...
+            end_time = yyyymmdd_to_seconds_since_j2000(end_date.strftime("%Y%m%d"))
+            metakernel_response = spice_metakernel_api.lambda_handler(
+                {
+                    "queryStringParameters": {
+                        "start_time": start_time,
+                        "end_time": end_time,
+                        "list_files": "True",
+                        "file_types": combined_kernel_sources,
+                        # TODO: revisit this after SIT-4
+                        # "require_coverage": "True",
```
tech3371, line 599: We will need to revisit this after SIT-4.

tmplummer (approved):
👍

PR #591, opened 2025-05-13
Remove api domain name from batch starter

### LINKED ISSUE #589: Add environment variables to the Batch Processing Jobs
### Summary

In our Batch Processing Jobs we need to have specific environment variables set for whether we are in dev vs prod or some other environment.

These are set here:
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/bc585067df567be0e65af357660705f9288a5755/sds_data_manager/constructs/processing_construct.py#L101

* The DATA_DIR should be IMAP_DATA_DIR to match what `imap-data-access` expects.
* The IMAP_DATA_ACCESS_URL can be set deterministically from the current context we are deploying from I think. Something like what is done in the account_config: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/bc585067df567be0e65af357660705f9288a5755/sds_data_manager/utils/stackbuilder.py#L65 and https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/bc585067df567be0e65af357660705f9288a5755/app.py#L23

```
                environment={"IMAP_DATA_DIR": "/mnt/data",
                             "IMAP_DATA_ACCESS_URL": f"https://api.{self.context.get(environment)}.imap-mission.com"},
```

We also need to remove references to `api_domain_name` where that makes sense too.

### Related tickets

_No response_

lacoak21:
# Change Summary

closes #589

## Overview
In our Batch Processing Jobs we need to specify an env var for the api endpoint url which depends on whether we are in dev vs prod or some other environment.

It also removes references to self.api_domain_name
## Updated Files
-sds_data_manager/constructs/api_gateway_construct.py
   - Make domain name local var
   - remove references to the api domain name. 
   - use current context to determine api endpoint url


tech3371 (approved):

greglucas (approved):

greglucas:
This will need a rebase now, but should be pretty straightforward I think.

PR #607, opened 2025-05-19
Add an API endpoint to trigger bulk reprocessing 

lacoak21:
# Change Summary

## Overview
Add an api endpoint to trigger bulk reprocessing. Add a simple python script to send a POST request to this API. 

TODO: Should the endpoint be /auth/reprocess? Is that ready? 

## Updated Files

   - Add api route for reprocessing
- sds_data_manager/utils/trigger_bulk_reprocessing.py
   - Add simple python script to trigger processing


sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -111,3 +115,10 @@ def __init__(
         # different instruments will be processed in parallel, with multiple instances
         # of the batch_starter lambda.
         self.instrument_lambda.add_event_source(SqsEventSource(sqs_queue))
+
+        # Add api route for triggering batch starter with a bulk reprocessing request
+        api.add_route(
+            route="/reprocess",  # TODO replace with /auth/reprocess
```
tech3371: I think it's ready but we not using `/auth/<endpoint>` yet to simplify SIT-4 preparation. I believe we will test that in SIT-5.

sds_data_manager/utils/trigger_bulk_reprocessing.py
```diff
@@ -0,0 +1,165 @@
...
+            query_params[param] = locals()[param]
+    base = "https://api.dev.imap-mission.com/auth/reprocess?"
+    url = f"{base}{urlencode(query_params)}"
+
+    print(f"Triggering bulk reprocessing for {query_params} with url {url}")
+    with _get_url_response(url) as response:
+        # Retrieve the response
+        response_text = response.read().decode("utf-8")
+        if response.status == 200:
+            print("Reprocessing triggered successfully.")
```
tech3371: should these print statement be logger instead? Just curious. I am ok either way since it's tests

sds_data_manager/utils/trigger_bulk_reprocessing.py
```diff
@@ -0,0 +1,165 @@
...
+        else:
+            print(f"Reprocessing triggered with api result: {result}")
+
+
+if __name__ == "__main__":
```
tech3371: Oh I didn't realized that it's an CLI script. Is that direction Greg wanted to go for this? If so, I am wondering what's the best place for this. Should this in `imap-data-access` or something else.

tech3371 (commented):
Might be good to get feedback from Greg on this.

sds_data_manager/utils/trigger_bulk_reprocessing.py
lacoak21: Im not exactly sure.. I dont think we would want imap-data-access because we only want members of the SDC running this script.

lacoak21: Maxine and I talked in the sdc-dev channel about having a simple script available to only SDC folks but I can definitely change this.

tech3371: I was thinking about this yesterday evening, `imap-data-access` does have upload API and webPODA support as well. But those are also will only be used by SDC and POC. In this case, it makes sense to move this over there and figure out limitation in SIT-5.

sds_data_manager/utils/trigger_bulk_reprocessing.py
```diff
@@ -0,0 +1,165 @@
...
+from urllib.error import HTTPError, URLError
+from urllib.parse import urlencode
+
+# Logger setup
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+@contextlib.contextmanager
+def _get_url_response(url: str):
```
greglucas: If we are already importing imap-data-access can you use the code there which has been refactored to use requests now?

sds_data_manager/utils/trigger_bulk_reprocessing.py
greglucas: Oh interesting. I think probably this does make sense to put over in imap-data-access as well similar to webpoda. So it'd be called something like `imap-data-access reprocess --instrument X --end-date Y ...`

sds_data_manager/utils/trigger_bulk_reprocessing.py
```diff
@@ -0,0 +1,165 @@
...
+            print(f"Reprocessing triggered with api result: {result}")
+
+
+if __name__ == "__main__":
+    # Parse the CLI command for the input data:
+    parser = argparse.ArgumentParser(
+        description="Trigger bulk reprocessing of data for a specific time range."
+    )
+    parser.add_argument(
+        "-s",
```
greglucas: I don't think we should add shorthands for these. We should be explicit because "s" and "e" are somewhat ambiguous.

sds_data_manager/constructs/instrument_lambdas.py
greglucas: Yes, probably add both for now like we are doing for the query/download/upload. `/auth/reprocess` and `/reprocess`

tech3371 (approved):
besides moving reprocessing scripts to `imap-data-access`. this looks good to me.

sds_data_manager/utils/trigger_bulk_reprocessing.py
lacoak21: OK sounds good! Will move that over to imap_data_access

sds_data_manager/utils/trigger_bulk_reprocessing.py
lacoak21: Ok good point.

PR #611, opened 2025-05-19
Add ancillary end date handling back to batch starter

lacoak21:
# Change Summary

## Overview
Somehow in all the shuffling of batch starter last week, the logic that sets the end_date of an ancillary file to datetime.today(), if it is None was removed. This PR adds it back in and should fix at least some of the problems (if not all) that menlo was having.

## Updated Files
   - Add ancillary end_date handling back in

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -292,7 +292,10 @@ def s3_processing_event(session, events):
             start_date = file_obj.start_date
             # Ancillary files can have an end date.
             end_date = getattr(file_obj, "end_date", None)
-
+            # If there is no end date for the ancillary file, then it is implicitly
+            # valid through today.
+            if not end_date:
+                end_date = datetime.today().strftime("%Y%m%d")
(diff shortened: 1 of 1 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -292,7 +292,10 @@ def s3_processing_event(session, events):
             start_date = file_obj.start_date
             # Ancillary files can have an end date.
             end_date = getattr(file_obj, "end_date", None)
-
+            # If there is no end date for the ancillary file, then it is implicitly
+            # valid through today.
+            if not end_date:
```
tech3371, line 297: Thank you for catching that. I think I made that change based on feedback and didn't check closely the impact of it.

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Ahh that was my fault to because I think I noticed it was gone but somehow thought we also didnt need it!

PR #613, opened 2025-05-20
Special case processing for IDEX and ultra, lo, and hi l3 maps 

lacoak21:
# Change Summary

## Overview
Add special case handling in batch starter. This covers IDEX l2b and lo, hi, and ultra l3 survival probability maps that need a range of glows l3e files. 

## Updated Files
   - Add handling for special cases
   - fix logging
   
## Testing
Special case test section

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -3,6 +3,7 @@
 import datetime
 import json
 import logging
+from dataclasses import dataclass
 from datetime import datetime as dt
 
 import boto3
@@ -21,7 +22,7 @@ from sqlalchemy.exc import IntegrityError
 from ..database import database as db
 from ..database import models
 from . import dependency
-from .dependency import DependencyConfig
+from .dependency import DependencyConfig, get_jobs
 
 # import dependency
 
@@ -38,6 +39,40 @@ SPECIAL_CASE_JOBS = [
     ("ultra", "l3", "u90-spx-hsf-sp-full-hae-nside8-3mo"),
     ("idex", "l2b", "sci-1week"),
 ]
+
+
+@dataclass
+class Cadence:
+    """Valid cadences for processing jobs triggered by cron jobs.
+
+    Valid cadences can be in either months or years
+    """
+
+    months3: str = "3mo"
+    months6: str = "6mo"
+    years1: str = "1yr"
+
+    @property
+    def valid_source(self) -> list[str]:
+        """Get all Cadences.
+
+        Returns
+        -------
+        list[str]
+            list of valid cadences.
+        """
+        return [self.years1, self.months3, self.months6]
+
+    @property
+    def days(self) -> dict:
+        """Cadence to days.
+
+        Returns
+        -------
+        dict
+            Cadence values in days.
+        """
+        return {self.years1: 365, self.months3: 90, self.months6: 180}
 
 
 def determine_job_version(
@@ -120,35 +155,49 @@ def handle_special_case_jobs(session, job_node, start_date, end_date):
     end_date : str
         End date to query the data.
     """
-    pass
-    # def find_most_recent_dep_start_date(dep, sd):
-    #     # Find the most recent start_date that matches the filters
-    #     return (
-    #         session.query(func.max(models.ScienceFiles.start_date))
-    #         .filter(
-    #             models.ScienceFiles.instrument == dep[0],
-    #             models.ScienceFiles.data_level == dep[1],
-    #             models.ScienceFiles.descriptor == dep[2],
-    #             models.ScienceFiles.start_date < dt.strptime(sd, "%Y%m%d"),
-    #         )
-    #         .scalar()
-    #     )
-    #
-    # if job_node["data_source"] == "idex":
-    #     # if job_node["data_source"] == "idex" and job_node["data_type"] == "l2b":
-    #     #     # IDEX l2b needs all the l1b evt datasets from the last 7 days
-    #     start_date_dt = dt.strptime(start_date, "%Y%m%d")
-    #
-    # # Check for ena l3 special case maps.
-    # else:
-    #     # Each of these maps has a l2 map and corresponding glows l3e files.
-    #     # To run this job, we need to find the most recent l2 map file and use that
-    #     # date range to query the glows l3e files.
-    #     deps = get_jobs("UPSTREAM", "HARD", *job_node)
-    #     # Get the l2 upstream dependency. There should only be one
-    #     l2_dep = [dep for dep in deps if dep[1] == "l2"][0]
-    #     # Find the most recent l2 map file
-    #     sd = find_most_recent_dep_start_date(l2_dep, start_date)
+
+    def find_most_recent_dep_start_date(dep, start_date):
+        # Find the most recent start_date that matches the filters
+        return (
+            session.query(func.max(models.ScienceFiles.start_date))
+            .filter(
+                models.ScienceFiles.instrument == dep[0],
+                models.ScienceFiles.data_level == dep[1],
+                models.ScienceFiles.descriptor == dep[2],
+                models.ScienceFiles.start_date <= dt.strptime(start_date, "%Y%m%d"),
+            )
+            .scalar()
+        )
+
+    start_date_dt = dt.strptime(start_date, "%Y%m%d")
+    if job_node["data_source"] == "idex":
+        # IDEX l2b needs all the l1b evt datasets since the last l2b job
+        # We need to find the most recent l2b job and use that start date in the query
+        # for upstream dependencies.
+        start_date_dt = start_date_dt - datetime.timedelta(days=1)
+        new_start_date = find_most_recent_dep_start_date(job_node, start_date_dt)
+        submit_all_jobs(
+            session, job_node, new_start_date, end_date, filter_dependencies=False
+        )
+    else:
+        # Each of these maps has a l2 map and corresponding glows l3e files.
+        # To run this job, we need to find the most recent l2 map file and use that
+        # date range to query the glows l3e files.
+        deps = get_jobs("UPSTREAM", "HARD", *job_node)
+        # Get the l2 upstream dependency. There should only be one
+        l2_dep = next(dep for dep in deps if dep[1] == "l2")
+        # Find the most recent l2 map file
+        new_start_date = find_most_recent_dep_start_date(l2_dep, start_date_dt)
+        # Get the number of days the map was created for
+        map_days = Cadence().days[job_node["descriptor"]]
+        new_end_date = start_date_dt + datetime.timedelta(days=map_days)
+        submit_all_jobs(
+            session,
+            job_node,
+            new_start_date,
+            new_end_date.strftime("%Y%m%d"),
+            filter_dependencies=False,
+        )
 
 
 def try_to_submit_job(
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -243,7 +243,9 @@ def handle_special_case_jobs(session, job_node, start_date, end_date):
             descriptor=job_node["descriptor"],
         )
         # Get the l2 upstream dependency (there should only be one).
-        l2_dep = next(dep for dep in deps if dep["data_type"] == "l2")
+        l2_dep = next((dep for dep in deps if dep["data_type"] == "l2"), None)
+        if l2_dep is None:
+            raise ValueError(f"Missing required l2 dependency for job: {job_node}.")
         # Find the most recent l2 map file start_date.
         new_start_date = find_most_recent_start_date(l2_dep, start_date)
         if not new_start_date:
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -1,8 +1,9 @@
 """Functions for supporting the batch starter component of the architecture."""
 
+import datetime
 import json
 import logging
-from datetime import datetime
+from datetime import datetime as dt
 
 import boto3
 import imap_data_access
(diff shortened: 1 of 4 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -31,6 +33,67 @@
...
+
+
+@dataclass
+class Cadence:
+    """Valid cadences for processing jobs triggered by cron jobs.
+
+    Valid cadences can be in either months or years
+    """
+
+    months3: str = "3mo"
```
tech3371, line 70: do you need `1week` for IDEX?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -31,6 +33,67 @@
...
+    months6: str = "6mo"
+    years1: str = "1yr"
+
+    @property
+    def valid_source(self) -> list[str]:
```
tech3371: can we rename this?
```suggestion
    def valid_cadence(self) -> list[str]:
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -200,6 +350,13 @@ def submit_all_jobs(session, job_node, start_date, end_date):
         Start date to query the data.
     end_date : str
         End date to query the data.
+    filter_dependencies : bool
+        If True, filter the upstream dependencies to only include the files valid for
+        upstream primary science start_date. For cadence jobs, we should never filter
+        the upstream dependencies because we want all the files found within the cadence
+        job range. For IDEX l2b, this is a special case where we need the last 7 days
+        of housekeeping event message data, so we should not filter the dependencies.
```
tech3371: add a comment `Default is set to True`?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -224,7 +381,7 @@ def submit_all_jobs(session, job_node, start_date, end_date):
     num_jobs = len(primary_science.imap_file_paths)
     logger.info(f"Found {num_jobs} jobs to process.")
     for filepath in primary_science.imap_file_paths:
-        job_start_date = datetime.strptime(filepath.start_date, "%Y%m%d")
+        job_start_date = dt.strptime(filepath.start_date, "%Y%m%d")
```
tech3371: can we revert this abbreviation? Previous is more readable.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -234,7 +391,7 @@ def submit_all_jobs(session, job_node, start_date, end_date):
         )
         # If there is only one file to process, then we can use upstream dependencies
         # that have already been queried.
-        if num_jobs > 1:
+        if num_jobs > 1 or filter_dependencies:
```
tech3371, line 427: nice!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -275,6 +450,17 @@ def s3_processing_event(session, events):
 
         file_obj = imap_data_access.file_validation.generate_imap_file_path(filename)
         input_obj = imap_data_access.processing_input.generate_imap_input(filename)
+
+        if input_obj.source == "glows" and input_obj.data_type == "l3e":
+            if triggered_from_glows_l3e:
+                logger.info(
+                    f"Already tried to submit job from a glows l3e file."
+                    f"Skipping trigger from filename {filename}"
+                )
```
tech3371: do we want to check in database if this is the case? Eg. something similar to `determine_job_version()` but in a new function where you call `determine_job_version()` which will check if job was in progress or succeeded and determine new version. Then check for sure if job with that version is in progress or not.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -322,7 +508,10 @@ def s3_processing_event(session, events):
             logger.info(f"No downstream dependencies found for the file: {filename}")
             continue
         for job in potential_jobs + potential_soft_jobs:
-            submit_all_jobs(session, job, start_date, end_date)
+            if job in SPECIAL_CASE_JOBS:
```
tech3371, line 531: I am concerned here. I am afraid of this variable becoming out of date when people update `dependency_config.csv` since descriptor is still being worked on and can be improved. I am wondering if we can check if `3mo or 6mo or 1yr` in descriptor and treat those a special case jobs. Also, can we rename `handle_special_case_job` to be `handle_cadence_job`?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -98,6 +161,106 @@ def filter_conditions(table):
     return f"v{int(max_version[1:]) + 1:03d}" if max_version else "v001"
 
 
+def handle_special_case_jobs(session, job_node, start_date, end_date):
+    """Handle special case jobs that require specific dependency querying.
```
tech3371: I was trying to think through this function, it took me a while to understand this. But the logics are pretty solid and really thought out. Nice job!!!

Can you expand the doc string explaining what is being done in this function? If I understand correctly, it's trying to figure out new start and end querying date for given situation. It's not building or looking for any upstream dependency or anything. It's just looking for upstream dependency to find start time and then set end date based on cadence of the job. Does that sound close?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -98,6 +161,106 @@ def filter_conditions(table):
...
+            .scalar()
+        )
+
+    start_date = dt.strptime(start_date, "%Y%m%d")
+    if job_node["data_source"] == "idex":
+        # Special case for IDEX l2b jobs:
+        # IDEX l2b requires all l1b event datasets since the last l2b job.
+        # Find the most recent l2b job before the current start date.
+        one_day = datetime.timedelta(days=1)
+        new_start_date = find_most_recent_start_date(job_node, start_date - one_day)
```
tech3371, line 230: why is this minus where as you are adding in ENA case?

tech3371: I remember you mentioning that you need 7 days worth of HK to process IDEX L1B. Is that why? I would add that here for future reference and understanding.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -98,6 +161,106 @@ def filter_conditions(table):
...
+        if not new_start_date:
+            # If there are no l2b jobs, subtract 7 days from the start date.
+            new_start_date = start_date - datetime.timedelta(days=7)
+        # Add one day from the most recent l2b job start date
+        # We need idex l1b evt files AFTER the last l2b job.
+        new_start_date = (new_start_date + one_day).strftime("%Y%m%d")
+        new_end_date = end_date
```
tech3371, line 240: I almost said that shouldn't `new_end_date=start_date` but then I remembered that both `start_date` and `end_date` are same for in-situ science files besides GLOWS. Can you add comment about why you can do this, in case someone gets confused?

tech3371 (commented):
I had clarification questions and few suggestions. Sorry for commenting in disorder. I commented by following logics. If you look at comments through `Files changed` tabs, I believe they will show in better order. Sorry in advance!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -98,6 +161,93 @@ def filter_conditions(table):
...
+        # files.
+        deps = get_jobs(
+            dependency_type="UPSTREAM",
+            relationship="HARD",
+            data_source=job_node["data_source"],
+            data_type=job_node["data_type"],
+            descriptor=job_node["descriptor"],
+        )
+        # Get the l2 upstream dependency (there should only be one).
+        l2_dep = next(dep for dep in deps if dep["data_type"] == "l2")
```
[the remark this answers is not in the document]
tech3371: I think ValueError is a good suggestions. I was wondering that too what happens if we don't find L2 or more than one L2.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Not anymore. I decided to not do a cadence job for IDEX l2b because the l2a files might not come exactly a week apart.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: I added this because we were importing datetime.datetime as datetime and I need to access datetime.today(). I can use the full datetime.datetime.strptime if that is better.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: The reason why I didnt check the database is because we will running the exact same code and submitting the exact same job. I added added flag that gets flipped on the first occurrence so we know that it already tried to submit it.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: Looks like we were importing like this `from datetime import datetime` and then you can use `datetime.strptime()` or `datatime.now()`.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: cool. I was just concerns with assumption we are making here and may be I am over complicating than it needs to.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: These are actually not cadence jobs. These are the map jobs that depend on cadence jobs and glows l3e jobs that menlo was talking about in the last meeting. These are the only jobs that deal with this particular issue so the only thing we might have to change is the descriptor but we shouldn't have to add anymore to this list. 

This is the diagram depicting the dependencies for these l3 sp-correlated maps: 
Screenshot 2025-05-07 at 2 54 59 PM


[This PR](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/616) is the cadence pr

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: yes that is the goal! I will expand on the docstring.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: I will expand on the comment! i see now how that comment might not be enough.

lacoak21: Its because idex l2b is triggered by an idex l2a file and we want the last l2a file start_date not the current one so we subtract a day for the query. ( I added this to the comments"

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: I think the issue is that i need datetime.datetime and datetime. So they are overriding eachother. That is why i changed datetime.datetime to dt.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Discussed in person

lacoak21:
OK @tech3371 I added the handling for special case reprocessing!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -683,10 +684,13 @@ def yyyymmdd_to_seconds_since_j2000(date_str: str) -> float:
...
 
-            logger.info(f"Searching for files matching dep={dep_string}")
+            logger.info(f"Searching for files matching {dep_string}")
 
             records = get_files(session, dep, start_date, end_date)
-            if not records and relationship == Relationship.HARD:
+            if not records and relationship in [
+                Relationship.HARD,
+                Relationship.HARD_NO_TRIGGER,
+            ]:
```
tech3371, line 695: good catch!

tech3371 (approved):
It looks good to me! Thank you for doing this!

maxinelasp (approved):
This looks great, I like the dataclass!!

lacoak21:
> This looks great, I like the dataclass!!

Awesome thanks for the review!

lacoak21:
@pleasant-menlo I am merging this to not block processing but please add any comments you may have and I will add them to a new branch if need be.

pleasant-menlo:
@lacoak21 Sorry for the delayed response. We've taken a look and everything seems to be good as far as we can tell.

 Our one question is: are all of the incoming GLOWS jobs are processed all at once? We upload them one at a time during the L3e processing; will this trigger the pipeline separately for each update, or is this batched in this case? We see that the case where they are batched is being handled appropriately.
 
 Thanks,
 Karn and George

maxinelasp:
@pleasant-menlo We have discussed this case in the team. We will tweak the batching on our side to try to get all the GLOWS L3e files at the same time, but our code should handle the case where some of them arrive in a different run. As long as they are uploaded one at a time in a tight loop within your code (so all files are uploaded at the end of processing rather than being generated and uploaded one at a time) we expect they will arrive in the same batch here, but even if they don't we shouldn't see any issues. 

Let me or Luisa know if you have further questions!

pleasant-menlo:
@lacoak21 @maxinelasp That sounds good. We don't expect the files to ever be generated over more than one run; they are generated in a single loop in the code that outputs them in close succession.

maxinelasp:
Perfect, in that case we can tweak the batching settings and be confident we will get all the files in one go. Thanks for clarifying!

PR #616, opened 2025-05-20
Cadence Processing Event

lacoak21:
# Change Summary

## Overview
Adds EventBridge jobs for cadence events. All l2 jobs are considered cadence jobs. Hopefully that is an assumption we can keep. 

## Updated Files
   - Add event bridge triggers for each type of cadence
   - Add handling for cadence jobs in batch starter

## Testing
Add a test to make sure all the cadence jobs for 3months are triggered.


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -384,6 +451,59 @@ def bulk_reprocessing_event(session, events):
...
+        # For the map.
+        job_start_date = primary_science.get_time_range()[0]
+        job_version = determine_job_version(
+            session=session,
+            instrument=job_node[0],
+            data_level=job_node[1],
+            descriptor=job_node[2],
+            start_date=job_start_date,
+        )
+        # TODO write out upstream_dependencies.serialize() to a json file
```
lacoak21: I will add this feature in another PR

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -111,3 +110,47 @@ def __init__(
         # different instruments will be processed in parallel, with multiple instances
         # of the batch_starter lambda.
         self.instrument_lambda.add_event_source(SqsEventSource(sqs_queue))
+
+        # Set up eventBridge rules to trigger batch starter lambda.
+        # Many l2 jobs create maps and need 3-12 months worth of data to run.
+        # Create an eventBridge rule for 3, 6, and 12, month cadences.
+        # TODO: IMWG is still deciding on the number of days for each cadence.
+        self.month3job = aws_events.Rule(
```
greglucas: I would invert the naming here so that 3 is before month similar to your strings below.
`self.event_3month`

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -111,3 +110,47 @@ def __init__(
...
+            scope=scope,
+            id="ProcessingCadenceJob3month",
+            rule_name="ProcessingCadenceJob3month",
+            description="Trigger processing jobs every 3 months (30 days)",
+            schedule=aws_events.Schedule.rate(Duration.days(91)),
```
greglucas: Why 91 days if you say 3 months worth of 30 days? Same with the ones below that are 182.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -32,6 +38,73 @@
...
+    as_str : bool
+        If True, return the start and end dates as strings. Default is False.
+
+    Returns
+    -------
+    tuple(datetime, datetime)
+        The start date and end date of the cadence. The end_date is set to today
+    """
+    cadence_obj = Cadence()
+    if cadence not in cadence_obj.valid_cadence:
```
greglucas: Can `valid_cadence` be a class property? Then you wouldn't have to create this unused object above just to reference a valid_cadence attribute.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -32,6 +38,73 @@
...
+
+    return start_date, end_date
+
+
+@dataclass
+class Cadence:
+    """Valid cadences for processing jobs triggered by cron jobs.
+
+    Valid cadences can be in either months or years
+    """
```
greglucas: Would an Enum suffice here which might be more descriptive? Then you could also have a separate mapping on it if you want too, or just keep a mapping of `CADENCE_STR_TO_DAYS = {"3mo": 90, ...}` and lookup from there. This class seems like it adds a lot of complexity for not a lot of gain.

```python
class CadenceDays(IntEnum):
    THREE_MONTHS = 90
    SIX_MONTHS = 180
    ONE_YEAR = 360

    def str_lookup(input_str):
        return {"3mo": self.THREE_MONTHS, ...}[input_str]
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -384,6 +465,108 @@ def bulk_reprocessing_event(session, events):
...
+            session=session,
+            instrument=job_node[0],
+            data_level=job_node[1],
+            descriptor=job_node[2],
+            start_date=job_start_date,
+        )
+        # Serialize the upstream dependencies to a JSON file. This is necessary for map
+        # jobs with many dependencies to avoid passing a long list of dependencies
+        # directly to the batch job. Imap processing code will read the JSON file and
+        # deserialize the dependencies.
```
greglucas, line 795: Is this something we should do for all jobs so it is easy to track what command was input and used for every job and to make things consistent across the board? This is similar to what I was trying to suggest a while ago where we would not use the CLI commands anymore and use a file instead, so I'm definitely in favor of this approach here. The CLI command is quite hefty to parse now IMO with all the dependencies even for simple commands.

sds_data_manager/constructs/instrument_lambdas.py
lacoak21: @nkerman  and I had a conversation about this yesterday. I think there are still some questions from the IMWG about what exactly this should be. Can you weigh in here Nat? Maybe the naming here should change and definitely the outdated comment.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Yes definitely. That simplifies things a lot.

sds_data_manager/constructs/instrument_lambdas.py
nkerman: Yes - Luisa is following my advice here based on attending the IMWG meetings. I've tried to get them to formalize their definition of a 3/6/9/12-month period but they haven't wanted to do so yet. What has become _pretty_ clear is that: 
1. What matters at its core here is not calendar year but fraction through a 365.25 day orbit. 
2. 3 month periods should be equivalent to one another (if you did it as Jan through Mar, etc. you would sometimes get 90 days in 3 months, sometimes 91, sometimes 92, etc.) It should be based on some day number which is a certain fraction of a 365.25 day year. 
3. The same 3 months should align in the orbit as closely as possible.

Given all of this, I suggested picking an integer day number for 3 months which would come closest to splitting up 365.25 days evenly into 4. 365.25/4 = 91.3125, so you could pick either 91 or 92; however I leaned towards avoiding overlap between days, towards likely needing to throw out an occasional day.

nkerman: This is all probably over-explaining it. The core reason is that `91 * 4` is closer to `365` than `90 * 4`.

greglucas: I would say that this is not what we actually want then because this is going to be dependent on deploy time. i.e. the cadences are based on when the last event was triggered / deployed.

I think we want to be explicit with the dates and not cadences for the event then. So set these to be on specific dates/times to align properly.
https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-scheduled-rule-pattern.html#eb-cron-expressions

I don't think we need to restrict ourselves to integer values here either. You can probably add a daily duration to an hourly duration or something along those lines?

nkerman: > I  would say that this is not what we actually want then because this is going to be dependent on deploy time. i.e. the cadences are based on when the last event was triggered / deployed.

Oh, yikes - yeah we don't want it to be indexed off of the deploy time. I figured we could set a sort of index date (e.g. Jan 1 or the start of post-commissioning mission operations), but that makes sense. 

> I think we want to be explicit with the dates and not cadences for the event then. So set these to be on specific dates/times to align properly.

Agreed, my only real opinion is to use 91 days here. We can just pick a `t0` (I suggest Jan 1 for now unless there's a definite idea of exactly when regular mission operations start? It shouldn't be far off from January anyways). Then we just pick t0+ 91 days; t0 + 182 days, etc. 

>I don't think we need to restrict ourselves to integer values here either. You can probably add a daily duration to an hourly duration or something along those lines?

I'm only really thinking of the ENAs, but we're really quantized by 1-day pointings, which is why I had been thinking in terms of integer days; however the kickoff might as well be every `365.25/4 = 91.3125` = **91 days, 7 hours, 30 minutes**

nkerman: It doesn't look like we could use the `rate` syntax, so I think a separate cron job rule would need to be scheduled for each individual process. Here's what I'm thinking

```
Starting at t0: 2026-01-01 00:00:00
To Kick off the 1th 3mo map at 2026-04-02 07:30:00:
	30 07 02 04 * 2026
To Kick off the 2th 3mo map at 2026-07-02 15:00:00:
	00 15 02 07 * 2026
To Kick off the 3th 3mo map at 2026-10-01 22:30:00:
	30 22 01 10 * 2026
To Kick off the 4th 3mo map at 2027-01-01 06:00:00:
	00 06 01 01 * 2027
To Kick off the 5th 3mo map at 2027-04-02 13:30:00:
	30 13 02 04 * 2027


To Kick off the 1th 6mo map at 2026-07-02 15:00:00:
	00 15 02 07 * 2026
To Kick off the 2th 6mo map at 2027-01-01 06:00:00:
	00 06 01 01 * 2027
To Kick off the 3th 6mo map at 2027-07-02 21:00:00:
	00 21 02 07 * 2027
To Kick off the 4th 6mo map at 2028-01-01 12:00:00:
	00 12 01 01 * 2028
To Kick off the 5th 6mo map at 2028-07-02 03:00:00:
	00 03 02 07 * 2028
```

lacoak21: Ah right that makes sense I should have realized that we wouldnt want this dependent on deploy time. I chose not to use a cron expression originally because then we cant be explicit with the amount of days a map represents. I guess i am still a little confused with how we would do this still if we want it to be 91 days.

lacoak21: Oh I see, the page didnt refresh and i missed your message above, Nat

lacoak21: Im not sure if I like the separate cron job event for each one. I also do not have any better suggestions that fulfill this though. We could add a config file with the list of cron patterns to add or dynamically create these. What would be our stop date?

nkerman: Would it make sense to have an automated way of generating these lambdas? (I made the above few automatically, based on a t0 of Jan 1 2026.) We could have a lambda that runs every day or month or year to set new cron jobs. 

Alternatively, we could just have a lambda that knows the starting time t0, and runs every day, kicks off jobs based on if the time difference from t0 is divisible by 3 months. I might be oversimplifying stuff I don't understand here though - I like a config file listing the kickoff times; then we can always specify the kickoff times for 100 years.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: OK sure. Definitely would be nice to have a simple way of tracking. Eventually it could be nice too if someone downloaded a file with the cli, they could specify if they wanted the input json file downloaded as well.

sds_data_manager/constructs/instrument_lambdas.py
lacoak21: The lambda idea is a really good one! If we have to use 92 days AND base it off a t0 it would make sense to skip the cron job altogether. The pros for the config file/ cron though, are it is nice to know when exactly we are expecting maps and we dont need to run a lambda everyday ( not that it would do a lot)

lacoak21: Oh wait this could be amazing. I didnt realize event bridge has a "scheduler" so we may be able to specify a rate event AND specify a start date (t0)... https://us-west-2.console.aws.amazon.com/scheduler/home?region=us-west-2#create-schedule

lacoak21: Nevermind. I was hoping if you specified a start date in the past it could calculate next expected trigger time but it requires a future or current start date.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: yeah, I have been thinking the same @greglucas. Can we switch to json file for everything? This can simplify local testings too. It's much easier to change json file than edit CLI string. This would mean that we would have lot more file in `cadence` folder and we would download more. Is that a concerns? If not, then we may need to rename the folder from `cadence` to `processing_input` on `imap-data-access`.

lacoak21: We also should probably rename CadenceFilePath. Should I do this in a new PR or here?

tech3371: I would say in a new PR to keep things moving and "small" scope

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -127,3 +131,83 @@ def __init__(
             http_method="POST",
             lambda_function=self.instrument_lambda,
         )
+
+        # Set up eventBridge rules to trigger batch starter lambda.
```
lacoak21, line 135: @nkerman or @greglucas. Can you take a look at this again? 

Please let me know if you think we should switch to making a lambda that checks if batch starter should be triggered.

poetry.lock
tech3371, line 1: you might need to rebase for these files

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -127,3 +131,92 @@ def __init__(
...
+        #    - 3 month map jobs (every 365.25 / 4 days)
+        #    - 6 month map jobs (every 365.25 / 2 days)
+        #    - 1 year map jobs (every 365.25 days)
+        # Note: each map job trigger will have its own eventBridge rule, because we need
+        # them to run every x days starting from the same date (t0).
+        # TODO what would be a good start date?
+        t0_date = datetime.datetime(2026, 1, 1)
+        today = datetime.datetime.now()
+        # Create rules for 15 years (far beyond what we expect as a precaution)
+        # AWS event bridge supports up to 500 rules, so this is well within the limit.
```
tech3371, line 155: oh interesting. Couldn't we do recurring schedule? Something like this from https://crontab.guru/#0_0_1_*/3_*
```
0 0 1 */3 *
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
```diff
@@ -123,21 +123,29 @@ def lambda_handler(event, context):
...
-                        "not match ancillary, "
-                        "science, or SPICE file "
-                        "naming convention."
-                    ),
-                }
+            except imap_data_access.ImapFilePath.InvalidImapFileError:
+                # Not a ANCILLARY file, continue on to cadence files
+                logger.info(f"Filename {filename} is not a valid ANCILLARY file.")
+                try:
+                    file_obj = imap_data_access.file_validation.CadenceFilePath(
```
tech3371: I think we can update upload_api.py to use `generate_imap_file_path` from `imap_data_access`. This will catch for cadence too. we can remove all the try except except last one.

tech3371 (approved):
I had couple comments but in general, logic flows really nice in batch starter and dependency.

sds_data_manager/constructs/instrument_lambdas.py
lacoak21: Resolved in person. Basically we need each map to consist of the same amount of days worth of data.

sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
lacoak21: Oh yeah good catch.

sds_data_manager/constructs/instrument_lambdas.py
tech3371: we can't do this based on discussion we had and it's unique to IMAP.

lacoak21:
pre-commit.ci autofix

PR #617, opened 2025-05-20
First Pass: Querying Between db Tables

daralynnrhode:
# Adding functionality to switch between tables when querying

## Overview

Added '--table' as a new query parameter to indicate what table should be selected. Science falls as the default if no table is provided. I toyed with the idea of adding subcommands but found that the nesting caused unnecessary confusion on my part and for when using the CLI.

For the sds side of things, I tried to keep it pretty versatile and applicable for both science and ancillary tables. Obviously if a parameter for the science that is not in the ancillary table is queried (against the ancillary table) it will fail, but the changes on the imap-data-access side will ensure that no parameters will be passed to the query_lambda that don't match the desired table. 
This may cause some problems for when queries are made directly to the api, now that I think about it. It could become common that queries will fail if invalid parameters are matched with the incorrect table, there is no error catch on this side for it. 

Things to consider, I guess. 

## Updated Files

- query_api.py

## Testing

I'll make sure to add some tests to ensure table are selected correctly.


sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
```diff
@@ -33,16 +33,28 @@ def lambda_handler(event, context):
 
     logger.info("Received event: " + json.dumps(event, indent=2))
 
+    table_models = {
```
maxinelasp: I'd suggest that this should be a NamedTuple, since you don't want later code modifying it and it would be nice to get some easy type management here

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
```diff
@@ -33,16 +33,28 @@ def lambda_handler(event, context):
...
+        "science": models.ScienceFiles,
+        "ancillary": models.AncillaryFiles,
+        "SPICE": models.SPICEFiles,
```
maxinelasp: Either make them all uppercase or all lowercase (I would say all lowercase)

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
```diff
@@ -33,16 +33,28 @@ def lambda_handler(event, context):
...
+    }
+
     # add session, pick model like in indexer and add query to filter_as
     query_params = event["queryStringParameters"]
+    # get desired table for query
+    if query_params.get("table") is not None:
```
maxinelasp: You can replace this if/else with `query_params.get("table", "science")`

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
```diff
@@ -33,16 +33,28 @@ def lambda_handler(event, context):
...
+        query_table = query_params.get("table")
+    else:
+        query_table = "science"
+
+    logger.info(f"Querying table: {query_table}")
+    model = table_models[query_table]
```
maxinelasp: Nice

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
```diff
@@ -54,6 +66,9 @@ def lambda_handler(event, context):
 
     # go through each query parameter to set up sqlalchemy query conditions
     for param, value in query_params.items():
+        # skip the table parameter
+        if param == "table":
```
maxinelasp, line 72: Could you add to the response below the model you're querying against? So update it to be like `f"{param} is not a valid query parameter for table {query_params.get("table")}. "`

maxinelasp (commented):
Overall looks good, just had a few comments!

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
```diff
@@ -14,7 +14,8 @@
 logger.setLevel(logging.INFO)
 
 
-def lambda_handler(event, context):
+# Too many conditional branches
```
bourque: I've been running into this problem in my CoDICE pipeline, I am going to steal this code 😅

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
bourque: Agreed, I really like the concept you came up with here.

bourque (approved):
Nice work!

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
```diff
@@ -33,16 +35,27 @@ def lambda_handler(event, context):
 
     logger.info("Received event: " + json.dumps(event, indent=2))
 
+    TableModels = namedtuple("TableModels", ["science", "ancillary", "spice"])
```
maxinelasp, line 37: You know what, I said namedtuple but I was actually thinking of an Enum (like[ this](https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/main/imap_data_access/processing_input.py#L49)) 😅 Sorry! This looks good if you want to leave it as-is (works the same way) but I apologize for the weird advice!

maxinelasp (approved):
Nice, thank you!

PR #618, opened 2025-05-21
BUG: Fixing permissions and trigger events of indexer lambdas. Fixin logs to be easier to follow.

tech3371:
# Change Summary

## Overview
Fixing few bugs in this PR
1. Fixing event trigger of sci/anc indexer lambda. It was getting SPICE too.
2. Adding permissions to SPICE indexer lambda.
3. Fixing metakernel API to return 404 if no files found.
4. Clean up logs to be able to follow easier on CloudWatch logs

## Updated Files
- sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_metakernel_api.py

### DIFF
--- a/sds_data_manager/constructs/indexer_lambda_construct.py
+++ b/sds_data_manager/constructs/indexer_lambda_construct.py
@@ -106,10 +106,10 @@ class IndexerLambda(Construct):
 
         # Write science data info to db with
         # status SUCCEEDED
-        event_rule_prefixes = [
+        science_event_prefixes = [
             {"prefix": f"imap/{instrument}/"} for instrument in VALID_INSTRUMENTS
         ]
-        event_rule_prefixes.append({"prefix": "imap/ancillary/"})
+        science_event_prefixes.append({"prefix": "imap/ancillary/"})
         imap_data_arrival_rule = events.Rule(
             self,
             "ImapDataArrival",
@@ -120,7 +120,7 @@ class IndexerLambda(Construct):
                 detail={
                     "bucket": {"name": [data_bucket.bucket_name]},
                     "object": {
-                        "key": event_rule_prefixes,
+                        "key": science_event_prefixes,
                     },
                 },
             ),
--- a/sds_data_manager/constructs/indexer_lambda_construct.py
+++ b/sds_data_manager/constructs/indexer_lambda_construct.py
@@ -10,6 +10,7 @@ from aws_cdk import aws_lambda as lambda_
 from aws_cdk import aws_s3 as s3
 from aws_cdk import aws_secretsmanager as secrets
 from constructs import Construct
+from imap_data_access import VALID_INSTRUMENTS
 
 from .efs_construct import EFSConstruct
 
@@ -105,6 +106,10 @@ class IndexerLambda(Construct):
 
         # Write science data info to db with
         # status SUCCEEDED
+        event_rule_prefixes = [
+            {"prefix": f"imap/{instrument}/"} for instrument in VALID_INSTRUMENTS
+        ]
+        event_rule_prefixes.append({"prefix": "imap/ancillary/"})
         imap_data_arrival_rule = events.Rule(
             self,
             "ImapDataArrival",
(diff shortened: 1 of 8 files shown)


sds_data_manager/constructs/indexer_lambda_construct.py
```diff
@@ -105,6 +106,10 @@ def __init__(
 
         # Write science data info to db with
         # status SUCCEEDED
+        event_rule_prefixes = [
+            {"prefix": f"imap/{instrument}/"} for instrument in VALID_INSTRUMENTS
+        ]
+        event_rule_prefixes.append({"prefix": "imap/ancillary/"})
```
greglucas: Maybe call this `science_event`? I'm a bit confused though because it contains ancillary too? So this is everything but spice then instead of just science?
```suggestion
        science_event_prefixes = [
            {"prefix": f"imap/{instrument}/"} for instrument in VALID_INSTRUMENTS
        ]
        science_event_prefixes.append({"prefix": "imap/ancillary/"})
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -683,7 +682,10 @@ def yyyymmdd_to_seconds_since_j2000(date_str: str) -> float:
 
             dep_string = f"{dep=}\n{start_date=}\n{end_date=}"
 
-            logger.info(f"Searching for files matching dep={dep_string}")
+            logger.info(
+                "Looking up current upstream dependencies with these "
+                f"info:={dep_string}"
```
greglucas:
```suggestion
                f"Searching for upstream dependencies with dependency string: {dep_string}"
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -914,6 +919,11 @@ def get_jobs(
         start_date=start_date,
         end_date=end_date,
     )
+    if upstream_dependencies_output is None:
+        logger.info(
+            f"No dependencies found for {dependencies}, {start_date=}, {end_date=}"
+        )
+        return None
```
greglucas: What about adding this to the info log below so you don't have a separate branch? `Dependencies for {start_date=} - {end_date=}: {dependencies}`. Which would then print None in this case, but also add the start/end info to the other portion.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -320,10 +320,12 @@ def write_data_to_efs(s3_key: str, s3_bucket: str, data_mount_path: Path) -> Pat
         # Download path to the EFS path
         efs_spice_filename_and_path = efs_spice_path / filename
         # Download file from S3 to the EFS path
+        logger.info(f"Downloading {s3_key} to {efs_spice_filename_and_path}")
         s3_client.download_file(s3_bucket, s3_key, efs_spice_filename_and_path)
-        logger.info(f"{s3_key} file downloaded successfully")
+        logger.info("Download Successfull")
     except Exception as e:
         logger.error(f"Error downloading file: {e!s}")
```
greglucas: Remove the logger.error if you are raising a ValueError here with the same info below?

greglucas (approved):
Some minor nits on log message formatting, but looks good otherwise.

sds_data_manager/constructs/indexer_lambda_construct.py
tech3371: yep. It's just science and ancillary event.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: I do like to explicit say that it didn't find it. Sometimes it's hard to see what happened when we are looking through the logs. This helps in spotting quickly what happened without having to derive what happened. But I do like the syntax of `{start_date=} - {end_date=}: {dependencies}`. I will add that below as well.

sds_data_manager/constructs/indexer_lambda_construct.py
tech3371: Doing this filtered indexer lambda event to what it needs because we are going to add more subfolder under `imap/` and doing this helps in ignoring other events that could come in the future, such as quicklook products or cadence inputs or SPICE(which is handled by SPICE indexer lambda).

PR #624, opened 2025-05-22
I-ALiRT - algorithm db ingest

laspsandoval:
# Change Summary

## Overview
This is the final (hopefully) version of the ialirt_ingest.py. It:
1. Queries s3 for last 5 minutes of packets
2. Parses packets and combines them into a dataset
3. Removes duplicates in the dataset that have the same epoch
4. Compares the "algorithm" DynamoDB with the dataset to access if the data needs to be inserted
5. Inserts data to algorithm DB

In this PR I also:
1. Remove the "ingest" database since it was no longer needed
2. Change the naming of "insert_time" to "utc"

## Updated Files
   - Remove ingest database, change insert_time to utc
- ialirt_archive.py
   - Change insert_time to utc
- ialirt_db_query_api.py
   - Change insert_time to utc
- ialirt_ingest.py
   - Changes stated in summary
- confest.py
   - Remove ingest table

## Testing
test_ialirt_ingest_lambda_construct.py
test_ialirt_archive.py
test_ialirt_db_query_api.py
test_ialirt_ingest.py

Verified deployment and functionality, as expected.


pyproject.toml
```diff
@@ -83,7 +83,7 @@ requests = "^2.28.2"
 spiceypy = "^6.0.0"
 
 [tool.poetry.group.layer-processing.dependencies]
-imap-processing = "0.12.0"
+imap-processing = "0.14.0"
```
tech3371: I think it's at version `0.15.0` now. Do you want to update it?

sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
```diff
@@ -141,12 +97,12 @@ def create_algorithm_dynamodb_table(self) -> aws_dynamodb.Table:
 
         # Add a GSI for ingest time.
         self.algorithm_data_table.add_global_secondary_index(
-            index_name="insert_time",
+            index_name="utc",
```
tech3371, line 100: I liked the previous variable name. It says something about what that time means. I am wondering if you want to capture the fact that it's in UTC time in some docstring instead and keep your previous variable name, `insert_time` or something that describe what that time is. `utc` may be too generic and can confuse later on.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -100,6 +74,131 @@ def query_filenames(bucket: str, region: str, now: datetime):
...
+    xtce_ialirt_path = (
+        imap_module_directory / "ialirt" / "packet_definitions" / "ialirt.xml"
+    )
+    datasets = []
+
+    for filename in filenames:
+        local_path = download_dir / Path(filename).name
+        s3.download_file(bucket, filename, str(local_path))
+        xarray_data = packet_file_to_datasets(
+            local_path, xtce_ialirt_path, use_derived_value=False
```
tech3371:
```suggestion
            local_path, xtce_ialirt_path
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -100,6 +74,131 @@ def query_filenames(bucket: str, region: str, now: datetime):
...
+    """
+    s3 = boto3.client("s3")
+    xtce_ialirt_path = (
+        imap_module_directory / "ialirt" / "packet_definitions" / "ialirt.xml"
+    )
+    datasets = []
+
+    for filename in filenames:
+        local_path = download_dir / Path(filename).name
+        s3.download_file(bucket, filename, str(local_path))
```
tech3371, line 107: Looks like this code may get run on a lambda. If so, you may want to be careful with downloading files to some directory. You may want to download to a `tm/` directory to not use up all the memory of lambda because lambda doesn't cleanup storage unless you do.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -100,6 +74,131 @@ def query_filenames(bucket: str, region: str, now: datetime):
...
+    """Insert or update database row, depending on content of item.
+
+    Parameters
+    ----------
+    data : list[dict]
+        Data product produced from processing respectively instrument.
+    algorithm_table : dynamodb.Table
+        The DynamoDB table to insert or update the data.
+    product_prefix : str
+        The prefix for the product name.
```
tech3371: can you use `instrument` instead of `prefix`?
```suggestion
    for instrument, process_func in processors:
        insert_data(process_func(combined), algorithm_table, instrument)


def insert_data(data: list[dict], algorithm_table, instrument: str):
    """Insert or update database row, depending on content of item.

    Parameters
    ----------
    data : list[dict]
        Data product produced from processing respectively instrument.
    algorithm_table : dynamodb.Table
        The DynamoDB table to insert or update the data.
    instrument : str
        The instrument of the product.
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -100,6 +74,131 @@ def query_filenames(bucket: str, region: str, now: datetime):
...
+    """
+    apid = data[0]["apid"]
```
tech3371, line 150: Will there be other apid at `data[n]` other than zero index? Just curious

tests/lambda_endpoints/test_ialirt_ingest.py
```diff
@@ -75,32 +56,53 @@ def test_lambda_handler(setup_dynamodb):
...
+    """Test parse_packet function that duplicate packets are removed."""
+    # Simulate two datasets with the same epoch.
+    ds1 = xr.Dataset({"data": (["epoch"], [1.0])}, coords={"epoch": (["epoch"], [100])})
+    ds2 = xr.Dataset({"data": (["epoch"], [2.0])}, coords={"epoch": (["epoch"], [100])})
+
+    # Each time the function packet_file_to_datasets() is called
+    # return the next item from this list.
+    mock_packet_file_to_datasets.side_effect = [
+        {478: ds1},
+        {478: ds2},
```
tech3371, line 91: This is what I am wondering in my last comment. Here we are returning two key of same apid. I thought packet_file_to_datasets returns dictionary with unique key of apid. Is that right @greglucas?

tech3371 (approved):
I had few minor questions and comments. I understand lot of these on high level. Sorry if I missed anything :). I know you are on time constraint. So I am approving this to keep you moving.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: There will only be a single apid. So it will always be zero index.

tests/lambda_endpoints/test_ialirt_ingest.py
laspsandoval: You are right. But this is just mocking two calls of packet_file_to_datasets.

sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
laspsandoval: I changed insert_time because the instruments will each have different insert times. So it wouldn't be a uniform global secondary index in the database, which was causing problems.

PR #633, opened 2025-05-27
MNT/FIX: Fixing indexing of pointing attitude kernels with new filename convention

tech3371:
# Change Summary

## Overview
Fixing SPICE indexing error since we updated pointing_attitude kernel type. The reason were
1. That new filename conventions has both start and end date
2. Because of first statement, we go in else statement of the code change. We tried setting `idcode` to be DPS_FRAME_ID=-43901 since that kernel needs that and I didn't change anything besides that but `get_coverage_dictionary()` doesn't work for this kernel. I don't know enough about SPICE to know this. That's why I end up doing a separate elif statement to set date range need for SPICE indexing for this kernel type.

@bryan-harter Do you know why second point's first try I did was not working?

## Updated Files
   - update for above changes


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -240,12 +240,31 @@ def index_spice_file(spice_file: Path):
...
+                [
+                    spiceypy.datetime2et(spice_metadata["start_date"]),
+                    spiceypy.datetime2et(spice_metadata["end_date"]),
+                ]
+            ]
+            file_coverage_sclk = [
+                [
+                    spiceypy.sce2s(SPACECRAFT_ID, file_coverage_j2000[0][0]),
+                    spiceypy.sce2s(SPACECRAFT_ID, file_coverage_j2000[0][1]),
+                ]
```
tmplummer, line 258: This should give valid coverage with day precision. Do we want precision down to the second? Perhaps it won't matter based on how the repoint table interacts with the generation of this kernel, but there is certainly higher precision start/end date/times available using SPICE to the kernel.

tmplummer (approved):
Just a question about the precision start and end date required.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
tech3371: I don't know enough here. Sorry. May be Bryan knows!

tmplummer: Sorry, I wasn't very clear. Say we have a Pointing Attitude kernel that provides coverage for `[2025-01-01 12:00:00.0, 2025-01-15 10:00:00.0]`. It will have a filename: `imap_dps_2025_001_2025_015_01.ah.bc`. As implemented, I think that the file coverage inserted in the SPICE table will be: `[2025-01-01 00:00:00, 2025-01-15 00:00:00.0]`. So, the inserted times don't reflect that the first 12-hours of 2025-01-01 are not covered and the first 10-hours of 2025-01-15 are covered.
I would say for now, the implementation should be the same as for the attitude_history kernel which queries SPICE to get the exact file coverage rather than relying on the year_doy string in the filename.

tech3371: Ah ok. Then I may have to pass this request to Bryan. I tried getting coverage similar to attitude kernel but it failed and I didn't know enough.

greglucas:
@tech3371 is this ready to be merged?

tech3371:
> @tech3371 is this ready to be merged?

@greglucas I think so. I don't have answer for question Tim had. Should we revisit that question after SIT-4?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
greglucas: @bryan-harter for your awareness

greglucas:
> @greglucas I think so. I don't have answer for question Tim had. Should we revisit that question after SIT-4?

Is this PR needed for SIT-4? I don't know how all of this stuff is being indexed either, but I would tend to agree that we want the most accurate times possible if they are there and not floor timestamps.

tech3371:
> > @greglucas I think so. I don't have answer for question Tim had. Should we revisit that question after SIT-4?
> 
> Is this PR needed for SIT-4? I don't know how all of this stuff is being indexed either, but I would tend to agree that we want the most accurate times possible if they are there and not floor timestamps.

Yes. it's needed for SIT-4.

greglucas:
Then I'd suggest taking the approval and opening a new issue saying you'll improve upon the timestamps in a follow-up PR.

tech3371:
I am going to be out tomorrow and our SIT-4 dry run is on Monday. Would be nice to get this in. Let me try again and see if I can solve it.

tech3371:
per discussion with Tim and Greg's suggestion. I am going to merge this PR with less accurate date and revisit this after SIT-4.

bryan-harter:
If you tried to add it to the unit test suite, it seems to have failed because the imap_sclk kernel in the test_data folder isn't the one we have in our S3 bucket. It looks like the sclk file changed recently. 

But anyway, once I modified the sclk file to look like the one we currently have in our S3 bucket, then it worked with just the simple change:

```
if spice_metadata['type'] == "pointing_attitude":
   function_arguments["idcode"] = -43901
```

at least, it did for the file "imap_dps_2025_105_2026_105_01.ah.bc". 

Do you recall if the error you received was something about the end_date being before the start_date somewhere?

tech3371:
> If you tried to add it to the unit test suite, it seems to have failed because the imap_sclk kernel in the test_data folder isn't the one we have in our S3 bucket. It looks like the sclk file changed recently.
> 
> But anyway, once I modified the sclk file to look like the one we currently have in our S3 bucket, then it worked with just the simple change:
> 
> ```
> if spice_metadata['type'] == "pointing_attitude":
>    function_arguments["idcode"] = -43901
> ```
> 
> at least, it did for the file "imap_dps_2025_105_2026_105_01.ah.bc".
> 
> Do you recall if the error you received was something about the end_date being before the start_date somewhere?

@bryan-harter I tried to do what you suggested because Tim suggested that too. But I came across some issue when testing in infrastructure.

I was seeing something like this:
> [ERROR] IndexError: list index out of range
  File "/var/task/SDSCode/pipeline_lambdas/spice_indexer.py", line 463, in lambda_handler
    index_spice_file(file_path)
  File "/var/task/SDSCode/pipeline_lambdas/spice_indexer.py", line 259, in index_spice_file
    _upsert_into_spice_table(
  File "/var/task/SDSCode/pipeline_lambdas/spice_indexer.py", line 160, in _upsert_into_spice_table
    "min_date_j2000": file_coverage_j2000[0][0],

from event:

>{
    "version": "0",
    "id": "5b246219-cd03-4682-1d66-c24973dd895e",
    "detail-type": "Object Created",
    "source": "aws.s3",
    "account": "449431850278",
    "time": "2025-05-27T20:18:15Z",
    "region": "us-west-2",
    "resources": [
        "arn:aws:s3:::sds-data-449431850278"
    ],
    "detail": {
        "version": "0",
        "bucket": {
            "name": "sds-data-449431850278"
        },
        "object": {
            "key": "imap/spice/ck/imap_dps_2025_105_2026_105_01.ah.bc",
            "size": 62464,
            "etag": "6df3b0a2b93ba0981eb7fc93973a4ed3",
            "version-id": "6xufKALH1t4jF3meL8Hk94WiI1TbLMZU",
            "sequencer": "0068361E0789149314"
        },
        "request-id": "VAQ89PMSMRGX02ZP",
        "requester": "449431850278",
        "source-ip-address": "67.168.251.19",
        "reason": "PutObject"
    }
}

Log stream link [here](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252Fspice-file-indexer/log-events/2025$252F05$252F27$252F$255B$2524LATEST$255D784381b8115b4220ad565928d112b151)

tmplummer:
@tech3371, I can write a new ticket to update how to get the DPS coverage and will take that on.

tech3371:
> @tech3371, I can write a new ticket to update how to get the DPS coverage and will take that on.

Thank you @subagonsouth. This is what I wrote in the main summary of this PR if it helps little bit.
>That new filename conventions has both start and end date. Because of that statement, we go in else statement of the code change. We tried setting idcode to be DPS_FRAME_ID=-43901 since that kernel needs that and I didn't change anything besides that but get_coverage_dictionary() doesn't work for this kernel. I don't know enough about SPICE to know this. That's why I end up doing a separate elif statement to set date range need for SPICE indexing for this kernel type.

PR #636, opened 2025-05-28
Swapi l1b dep

tech3371:
# Change Summary

## Overview
correcting SWAPI L1 science dependency

## Updated Files
   - update dependency

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
@@ -324,7 +324,8 @@ leapseconds, spice, historical, swapi, l1, hk, HARD_NO_TRIGGER, DOWNSTREAM
 spacecraft_clock, spice, historical, swapi, l1, hk, HARD_NO_TRIGGER, DOWNSTREAM
 
 swapi, l0, raw, swapi, l1, sci, HARD, DOWNSTREAM
-swapi, l1b, hk, swapi, l1, sci, HARD, DOWNSTREAM
+# Need to use l1a HK data to be able to set bit-wise flags in L1 Science file.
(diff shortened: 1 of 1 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
```diff
@@ -324,7 +324,8 @@ leapseconds, spice, historical, swapi, l1, hk, HARD_NO_TRIGGER, DOWNSTREAM
 spacecraft_clock, spice, historical, swapi, l1, hk, HARD_NO_TRIGGER, DOWNSTREAM
 
 swapi, l0, raw, swapi, l1, sci, HARD, DOWNSTREAM
-swapi, l1b, hk, swapi, l1, sci, HARD, DOWNSTREAM
+# Need to use l1a HK data to be able to set bit-wise flags in L1 Science file.
```
greglucas, line 327: All of the data should be there in both l1a and l1b, just named differently I think?

greglucas (approved):

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
tech3371: yeah, same data should be in both l1a and l1b. But in l1b, the housekeeping flags data is now in derived form which is usually a string, such as 'OK' or 'ERR'. Eg.
```
SWP_HK | PCEM_RATE_ST | 0 | OK | GOOD
SWP_HK | PCEM_RATE_ST | 1 | ERR | BAD
SWP_HK | SCEM_RATE_ST | 0 | OK | GOOD
SWP_HK | SCEM_RATE_ST | 1 | ERR | BAD
```
That's why I need to use l1a hk instead of l1b hk in science since I just need the raw bit value to set those bit-wise flag.

PR #637, opened 2025-05-28
Add L3 dependencies for CoDICE-Lo 3d Distribution data products

pleasant-menlo:
# Change Summary
Defined dependencies for 4 new L3 CoDICE data products.
Updated the L3-direct-events product dependency definition to include a new ancillary.

## Updated Files
dependency_config.csv
 - added dependency lines for 4 new 3d CoDICE data products


tech3371 (approved):
The syntax looks good to me. Just be cautious of using 'HARD' on ancillary. It will cause reprocessing if new one is uploaded.

PR #642, opened 2025-05-29
Harrison L3 2569 - Update CoDICE L3a dependency config

pleasant-menlo:
# Change Summary

## Overview
Some of our dependency config lines had 2 spaces in front of the output descriptor which we believe was causing them to be
read incorrectly. 

## Updated Files
dependency_config.csv
 - removed whitespaces on CoDICE 3d distribution lines

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
@@ -45,16 +45,16 @@ mag, l1d, norm-mago, codice, l3b, hi-pitch-angle, HARD, DOWNSTREAM
 codice, l2, hi-sectored, codice, l3b, hi-pitch-angle, HARD, DOWNSTREAM
 
 codice, l3a, lo-direct-events, codice, l3a, lo-hplus-3d-distribution, HARD, DOWNSTREAM
-codice, l1a, lo-sw-priority, codice, l3a, lo-hplus-3d-distribution, HARD, DOWNSTREAM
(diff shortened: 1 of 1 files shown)


pleasant-menlo:
We have also updated the L1a dependencies of the 3d distributions products to be HARD_NO_TRIGGER.
When they were HARD dependencies, if a new version of an L1a file were to be produced it would have triggered both L3a lo-direct-events, and L3a lo-3d-distributions. Once the new version of L3a lo-direct-events was produced L3a-3d-distributions would be triggered again. It would be triggered twice for one new L1a file.

Our understanding of HARD_NO_TRIGGER is it will behave like a HARD dependency, but avoid the above case. Now when a new L1a file comes out it will first trigger a new L3a direct-events, and this will trigger 3d-distributions which will pull in the new L1a file. Is our understanding correct? Does a different dependency type work better for this case?

tech3371 (approved):
That's interesting that the space was causing issue. We can look into that after SIT-4. Can you create an issue for this?

tech3371:
> We have also updated the L1a dependencies of the 3d distributions products to be HARD_NO_TRIGGER. When they were HARD dependencies, if a new version of an L1a file were to be produced it would have triggered both L3a lo-direct-events, and L3a lo-3d-distributions. Once the new version of L3a lo-direct-events was produced L3a-3d-distributions would be triggered again. It would be triggered twice for one new L1a file.
> 
> Our understanding of HARD_NO_TRIGGER is it will behave like a HARD dependency, but avoid the above case. Now when a new L1a file comes out it will first trigger a new L3a direct-events, and this will trigger 3d-distributions which will pull in the new L1a file. Is our understanding correct? Does a different dependency type work better for this case?

Sorry, I went straight to PR files change and missed the question. Let me think about it and get back to you. Looping @lacoak21 as well.

PR #643, opened 2025-05-29
Add cr db column

maxinelasp:
# Change Summary
This change will ultimately require more updates in batch-starter, but I want to get this piece in separately. 

When looking through, the only piece that could break from this change is in indexer.py, but I think with the new version of imap-data-access and a careful deployment it will be ok. In order to deploy this change without destroying our DB, I was thinking I could manually add a new column for testing, and then during the destroy/redeploy step for SIT-4 where we lose all the data this CDK code will be used to create the database.

## Overview

- Add new CR column to the file tracker database
- Update imap-data-access version
- Not a change, but I did search for all usages of extract_filename_components to make sure this wasn't going to break anything and it won't. The column name does need to match the component name of "cr"


tech3371 (approved):
LGTM

PR #654, opened 2025-06-02
Fix end_time when making a call to the metakernel api 

lacoak21:
# Change Summary

## Overview
@subagonsouth and I noticed that the start/ end_date query parameters to the metakernel api, were being set to that date at 00:00. When start date == end, ( as is the case for many jobs during normal processing/ reprocessing) It was only querying for files at that exact day/hour/minute. See log below from the metakernel API:

logger:INFO Attempting to find file to cover 813931200 to 813931200

This resulted in some files not being returned. To fix this Tim proposed adding 24 hours to the end date. This will still need to be revisited after SIT-4 because there are still issues for instruments that have data for a Pointing rather than a Day.

## Updated Files
   - After parsing end_date into a string, add 24 hours to that value so the metakernel API returns files through that date.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
@@ -616,6 +616,7 @@ def get_latest_repoint_file(end_date: datetime) -> Optional[str]:
     return basename(latest[2])
 
 
+# ruff: noqa: PLR0915
 def get_upstream_dependency_inputs(
     dependencies: list,
     start_date: datetime,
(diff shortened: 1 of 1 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -681,10 +681,15 @@ def get_upstream_dependency_inputs(
 
             # convert start_date and end_date in seconds after j2000.
             # TODO: remove this once Bryan changes takes in 'yyyymmdd' format
-            def yyyymmdd_to_seconds_since_j2000(date_str: str) -> float:
+            def yyyymmdd_to_seconds_since_j2000(
```
tech3371, line 685: Actually, can you remove this function now? This was a place holder and Bryan did make changes to take in string date now and I didn't get to remove this.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -693,7 +698,9 @@ def yyyymmdd_to_seconds_since_j2000(date_str: str) -> float:
                 return delta.total_seconds()
 
             start_time = yyyymmdd_to_seconds_since_j2000(start_date.strftime("%Y%m%d"))
-            end_time = yyyymmdd_to_seconds_since_j2000(end_date.strftime("%Y%m%d"))
+            end_time = yyyymmdd_to_seconds_since_j2000(
+                end_date.strftime("%Y%m%d"), True
+            )
```
tech3371: With remove above funciton, now you can add the 24 hour time delta here and then convert to string as input here.

and may be we can check if start and end date is same(such as science files ingestion). if it is, then may be add 24 delta time? Just a thought.

tech3371 (commented):
couple comments. Otherwise looks good to me.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tmplummer: I like the idea of only adding 24 hours if `start_date == end_date`.

tmplummer (approved):
I'm going to pre approve this in case there is a possibility of getting it merged in time for the SIT4 dry run.

tech3371 (approved):

PR #660, opened 2025-06-02
SQS decrease retry count 

lacoak21:
# Change Summary

## Overview
Reduce sqs retry count to 1. We want failed events to move directly into the dead letter queue. 

## Updated Files
- sds_data_manager/constructs/sqs_construct.py
   - Reduce retry count.


tech3371 (approved):
Thank you for doing that!

PR #664, opened 2025-06-02
BUG FIX for niche case with multiple dependent ancillary files

lacoak21:
# Change Summary

## Overview
This PR fixes a bug that was caused by a rare scenario: 

When a new ancillary file is uploaded, the dependency handler might find jobs to run where the uploaded file is valid for the job's start_date, but another required ancillary file is not. 

We recently updated the batch starter to check for upstream dependencies twice, once to get the start dates for the jobs to run and then second to get the dependencies for that specific start date. The first request checks if the value returned was None (meaning missing dependency) but the second request did not check this and caused an error when submitting the job with None type as the processingInputCollection. In the specific scenario above, the second request would determine that there was no valid file for the second required ancillary dependency, and return None, causing an error. 

## Updated Files
   - Check if the second upstream dependency request returned None
   - If it did return none, log a message and continue with the next job.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -437,7 +437,8 @@ def submit_all_jobs(session, job_node, start_date, end_date, filter_dependencies
     )
     if not upstream_dependencies:
         logger.info(
-            f"Skiping job submission for {job_node} because no upstream dependencies."
+            f"Skipping job submission for {job_node} because of a missing upstream "
+            f"dependency."
         )
         return
 
@@ -473,6 +474,12 @@ def submit_all_jobs(session, job_node, start_date, end_date, filter_dependencies
                 start_date=filepath.start_date,
                 end_date=filepath.start_date,
             )
+            if not upstream_deps_for_job:
+                logger.info(
+                    f"Skipping job submission for {job_node} with start_date: "
+                    f"{start_date} because of a missing upstream dependency."
+                )
+                continue
         else:
             upstream_deps_for_job = upstream_dependencies
         try_to_submit_job(
--- a/tests/lambda_endpoints/test_batch_starter.py
+++ b/tests/lambda_endpoints/test_batch_starter.py
@@ -20,6 +20,7 @@ from sqlalchemy.exc import IntegrityError
 
 from sds_data_manager.lambda_code.SDSCode.database import models
 from sds_data_manager.lambda_code.SDSCode.database.models import (
+    AncillaryFiles,
     ProcessingJob,
     ScienceFiles,
     SPICEFiles,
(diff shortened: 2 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -473,6 +474,12 @@ def submit_all_jobs(session, job_node, start_date, end_date, filter_dependencies
                 start_date=filepath.start_date,
                 end_date=filepath.start_date,
             )
+            if not upstream_deps_for_job:
```
maxinelasp, line 477: Will this work if one upstream dependency is found but the other one is missing?

lacoak21: Yes - the dependency code will return None if _any_ of the HARD or HARD_NO_TRIGGER dependencies are missing. So in this case if one ancillary is missing, but the other was found that function will still return None.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -269,6 +270,75 @@ def test_lambda_handler_missing_upstream_dependency(session, caplog):
...
+                    "2024-01-25 23:35:26+00:00", "%Y-%m-%d %H:%M:%S%z"
+                ),
+            ),
+        ]
+    )
+    multiple_events = {
+        "Records": [
+            {
+                "body": '{"detail": '
+                '{"object": {"key": "imap_mag_l2-calibration_20250117_v001.cdf"}}'
```
tech3371, line 322: isn't this a duplicate file and should fail on upload?

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -269,6 +270,75 @@ def test_lambda_handler_missing_upstream_dependency(session, caplog):
...
+                descriptor="l2-norm-offsets",
+                start_date=datetime(2025, 4, 21),
+                version="v001",
+                extension="cdf",
+                ingestion_date=datetime.strptime(
+                    "2024-01-25 23:35:26+00:00", "%Y-%m-%d %H:%M:%S%z"
+                ),
+            ),
+        ]
+    )
```
tech3371, line 317: also you might want to add `session.commit()`. It caused me issue in the past.

maxinelasp (approved):
Thanks for the quick fix, this looks good to me!

tests/lambda_endpoints/test_batch_starter.py
lacoak21: I dont think so because I basically am just mocking the indexer by first adding it to the database and then mocking the event.

tests/lambda_endpoints/test_batch_starter.py
lacoak21: good catch

tests/lambda_endpoints/test_batch_starter.py
tech3371: Ah I see.

tech3371 (approved):
I don't fully follow the reason behind this but the fix makes sense to me. Thank you for looking into this and fixing it!

PR #670, opened 2025-06-03
MNT/FIX: removed EFS mount volume from batch. HIT L1B dep fix

tech3371:
# Change Summary

## Overview
This removes EFS volume from batch job. Because IMAP_DATA_DIR is EFS path, it was causing to download sci/anc into EFS volume. And in addition, some processing code was removing SPICE files from EFS as clean up, we think. Therefore, we removed EFS mount and let batch job download from API and load kernels that way. And we tested that it works in AWS and it worked even after removing EFS mount volume.

## Updated Files
   - remove volume from batch job construct
   - update HIT depedency

### DIFF
--- a/sds_data_manager/constructs/processing_construct.py
+++ b/sds_data_manager/constructs/processing_construct.py
@@ -66,8 +66,6 @@ class ProcessingConstruct(Construct):
             ],
         )
 
-        self.volumes = volumes
-
     def add_job(self, job_name: str):
         """Create an ECR repo and a job definition for the given job.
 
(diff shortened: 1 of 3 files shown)


sdhoyt (approved):
Thanks for doing this!

laspsandoval:
I haven't gotten too far with I-ALiRT SPICE work yet. So it's hard to say how it would be affected. Just keep it in mind. I'm guessing I'll need to create an apid to grab the latest kernels that I need.

tech3371:
> I haven't gotten too far with I-ALiRT SPICE work yet. So it's hard to say how it would be affected. Just keep it in mind. I'm guessing I'll need to create an apid to grab the latest kernels that I need.

yeah, @laspsandoval I am also unsure how much this will impact I-ALiRT work. I looked quickly and didn't see anything obvious. That's why I shared for your awareness :). btw, we are not deleting EFS code or SPICE indexer and EFS write features. So EFS will be available still for I-ALiRT. That's why Greg and I didn't remove EFS in case you need it.

tech3371:
merging this in for SIT-4 dry-run

PR #672, opened 2025-06-03
MNT/ENH: Add bucket replication configuration to the CDK

greglucas:
Right now we have this specified as manual steps to go in and update it in the console. We can create the replication rule with L1 constructs, so not as elegant as the L2 methods, but still doable in code rather than manual intervention.

The backup bucket arn needs to exist for this deployment to work.

### DIFF
--- a/docs/source/cdk/s3-replication.rst
+++ b/docs/source/cdk/s3-replication.rst
@@ -1,31 +1,21 @@
 S3 Replication
 ==============
 
-Once you have finished :doc:`backup-deploy`, you can set up replication in the source bucket.
-
-Create a replication rule
-^^^^^^^^^^^^^^^^^^^^^^^^^
-
-#. Open your source S3 bucket in the AWS console
-#. Under the "Management" tab, find the "Replication Rules" section
-#. Click "Create replication rule" to create a new rule
-#. Enter an ID. If this is a developer bucket, include your initials. Preferably, it should be clear this is a backup rule.
-#. Under "Source Bucket", select "Apply to all objects in this bucket"
-#. Under "Destination" select "Specify a bucket in another account" and enter the account number and bucket name
-#. Under "IAM role", select the "SdsDataManager-{account_name}-BackupRole"
-#. Then, save the rule, and select "Do not replicate existing objects" in the popup
-
-Now, items placed in the source bucket will be automatically replicated to the backup bucket!
+Once you have finished :doc:`backup-deploy`, the new items are automatically replicated
+from the source bucket to the backup bucket.
 
 Copying items back into source account bucket
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-To restore files from the backup bucket into the main account bucket, we are going to [assume the role](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-role-prereqs) created by SdsDataManager stack in the source account. Here, "source account" refers to the account with SdsDataManager deployed into it (i.e. dev or prod).
+To restore files from the backup bucket into the main account bucket, we are going to
+[assume the role](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-role-prereqs)
+created by SdsDataManager stack in the source account. Here, "source account" refers to
+the account with SdsDataManager deployed into it (i.e. dev or prod).
 
 #. Update ``~/.aws/config`` to include a new profile for this role. Here, ``source_profile`` should be the source account profile
 ```
 [profile backup-role]
-role_arn = arn:aws:iam::<source account number>:role/<role name>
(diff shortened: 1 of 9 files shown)


maxinelasp (approved):
Looks good, nice update! What if you deploy this and the backup bucket arn doesn't exist?

maxinelasp:
Does this create a circular dependency with the backup role, actually? The backup role needs to exist before the table does, but I think it's for the main bucket, so the main bucket needs to be created first. Then the backup bucket uses the backup role to point to the main bucket, but then this piece points to the backup bucket so it can't actually work. I think instead we need to have this try to create, if the backup bucket doesn't exist then continue on anyway and skip that replication step, and then the basic steps would be: 
1. deploy this bucket, have the replication bit fail gracefully
2. Deploy backup bucket with the correct permissions
3. re-deploy this bucket, now with replication succeeding

Or, maybe put replication into a different stack entirely

maxinelasp:
Alternatively, Sean mentioned just having the backup bucket in the same account but in a different region - that would also solve all these problems and a bunch of other problems

greglucas:
If the bucket doesn't exist, this will not deploy. There is an error that the backup bucket arn must exist and it rolls back gracefully to the previous state.

Good question about the circular reference. I tested this in the same account with a different test bucket arn that I created manually and that did work. Can we deploy the backup stack first? I think we could actually update the backup stack to infer the BackupRole ARN similar to how we are doing it for the BackupBucket ARN, which would remove circular references?

https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/2ac3ca4f1f82986eb75d2a83895efae93cde3304/sds_data_manager/constructs/backup_bucket_construct.py#L52-L55
Change that to:
```python
role_arn = f"arn:aws:iam::{source_account}:role/BackupRole"
```

Will the backup stack deploy with that change, or does that role need to exist first so we are in a chicken and egg situation?

maxinelasp:
Actually, looking at the current BackupRole, you're right that something changed (presumably in the stack collapsing refactor) that means we can assume the BackupRole is just what you said. Before, it always added some random CDK numbers to it, but now it looks like the role just is named BackupRole. So that solves the problem, I think, as long as the backup is always deployed. 

I do think we should have the option to successfully deploy (maybe with lots of warnings? idk) without the backup bucket because we don't want that in dev.

greglucas:
OK, I added in a check for whether we are in the "sit-4" or "prod" account_names and only creating the backup roles in those situations, so avoiding creating any of the backup stuff at all in the dev account.

I updated the documentation to reflect the current state of things.

I'm still a bit unclear whether the backup stack can deploy without the production stack/bucket or not. Can you verify whether that role needs to be there or not? If not, then I think we can say that the backup stack must be deployed first and then the production stack can be deployed and all will be good.

maxinelasp:
Just tested it, I ended up in kind of a weird state but I _think_ everything deployed properly...? I can test more after the SIT.

maxinelasp:
So I think we can deploy backup first and then do the main one

docs/source/cdk/s3-replication.rst
```diff
@@ -1,31 +1,21 @@
...
+from the source bucket to the backup bucket.
 
 Copying items back into source account bucket
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-To restore files from the backup bucket into the main account bucket, we are going to [assume the role](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-role-prereqs) created by SdsDataManager stack in the source account. Here, "source account" refers to the account with SdsDataManager deployed into it (i.e. dev or prod).
+To restore files from the backup bucket into the main account bucket, we are going to
+[assume the role](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-role-prereqs)
+created by SdsDataManager stack in the source account. Here, "source account" refers to
+the account with SdsDataManager deployed into it (i.e. dev or prod).
```
tech3371:
```suggestion
created by SDCStack stack in the source account. Here, "source account" refers to
the account with SDCStack deployed into it (i.e. dev or prod).
```

sds_data_manager/constructs/backup_bucket_construct.py
```diff
@@ -14,15 +14,7 @@ class BackupBucket(Construct):
     should be the AWS account for the source bucket.
 
     For replication to work, you also need to deploy SdsDataManager and create
```
tech3371: Can you replace `SdsDataManager` with `SDCStack`? It must be remains from a while ago.
```suggestion
    For replication to work, you also need to deploy SDCStack and create
```

tech3371 (approved):
couple minor comments. Otherwise, looks good to me on high-level.

greglucas:
This is out of date, I'm going to close it and we can add it back in later if we want. I won't delete the branch yet.

PR #676, opened 2025-06-04
Add retry strategy for batch jobs 

lacoak21:
# Change Summary

## Overview
Retry batch jobs that failed due to status reason = "Your Spot Task was interrupted." 

## Updated Files
   - Add batch job retry strategy.

## Testing
Fix expected calls to include retry strategies.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -34,6 +34,19 @@ logger.setLevel(logging.INFO)
 
 # Create a batch client
 BATCH_CLIENT = boto3.client("batch", region_name="us-west-2")
+# Define the retry strategy for batch jobs
+BATCH_JOB_RETRY_STRATEGY = (
+    {
+        "attempts": 2,
+        "evaluateOnExit": [
+            {
+                "onStatusReason": "Your Spot Task was interrupted.",
+                "action": "RETRY",
+            },
+            {"onReason": "*", "action": "EXIT"},
+        ],
+    },
+)
 # Create an sqs client
 SQS_CLIENT = boto3.client("sqs", region_name="us-west-2")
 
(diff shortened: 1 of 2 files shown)


maxinelasp (approved):
That was fast! Nice :)

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -34,6 +34,19 @@
...
+# Define the retry strategy for batch jobs
+BATCH_JOB_RETRY_STRATEGY = (
+    {
+        "attempts": 2,
+        "evaluateOnExit": [
+            {
+                "onStatusReason": "Your Spot Task was interrupted.",
+                "action": "RETRY",
+            },
+            {"onReason": "*", "action": "EXIT"},
```
tech3371, line 46: Should we be checking for this action?
```
Your Spot Task was interrupted.
```

tech3371: This status reason mean it failed with valid reasons, I believe.
```
Essential container in task exited
```

lacoak21: I think so! Today during the sit idex l2b processing didnt even start and this was the status reason. The AWS docs say this status reason is due to: 

Amazon EC2 can interrupt your Spot Instance when it needs it back. EC2 reclaims your instance mainly to repurpose capacity, but it can also occur for other reasons such as host maintenance or hardware decommission.


https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-interruptions.html

So it wasnt that anything was wrong with idex l2b processing. In this case it was just that the instance was needed.

lacoak21: So it was a valid reason that it failed but we should retry it because it had nothing to do with the actual processing.

tech3371: oh ok. My mistake. You had this `"onStatusReason": "Your Spot Task was interrupted.",`. I totally missed it. Should not do two things at once.

tech3371 (approved):
Thank you for fixing that so quickly!

PR #687, opened 2025-06-09
Remove cadence start and end date parameters (used for SIT4)

### LINKED ISSUE #683: Remove cadence event start and end date parameters
The parameters were used temporarily for SIT-4.

lacoak21:
# Change Summary

closes #683
## Overview
The "start_date" and "end_date" parameters were used temporarily for SIT-4. In production, the start and end dates will be calculated from the time the batch starter was triggered by the EventBridge event. 

## Updated Files
   - Remove handling for start_date and end_date parameters


lacoak21:
Do we want to hold off on merging this for testing purposes?

greglucas (approved):
No, we don't need to keep it around for testing. We have a tagged commit that we are using for deploying to our SIT-4 account if necessary, so we should keep progressing forward here.

PR #690, opened 2025-06-10
Write all dependencies for processing to a JSON file

### LINKED ISSUE #639: Write all serialized dependencies from batch starter to a json file.
Currently only cadence jobs are doing this. This should happen for all jobs.

lacoak21:
# Change Summary

closes #639

## Overview
This PR updates the 'dependency' keyword in the batch job command to indicate a JSON file containing the serialized dependencies for the job instead of the actual serialized dependencies. For each processing job submitted there will be a corresponding dependency JSON file. 

## Updated Files
<!--List each updated file and add bullets to describe the purpose/function of the updates.
   - Upload json file of dependencies to s3 and use the filename as the --dependency argument in the batch job command. 

## Testing
Fixed test to check that the correct json filename is passed into the batch job command. The 'try_to_submit_job' function is mocked in many tests still check that the upstream dependencies are the ones we would expect.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -749,12 +749,12 @@ def bulk_reprocessing_event(session, events):
 
 
 def upload_dependency_file(dependency_file_path: Path, serialized_dependencies: str):
-    """Upload a JSON file containing a cadence job's dependencies to S3.
+    """Upload a JSON file containing a job's dependencies to S3.
 
     Parameters
     ----------
     dependency_file_path : Path
-        The cadence JSON file to upload.
+        The dependency JSON file to upload.
     serialized_dependencies : str
         The serialized upstream dependencies to upload.
     """
@@ -769,9 +769,10 @@ def upload_dependency_file(dependency_file_path: Path, serialized_dependencies: 
     )
     if signed_url["statusCode"] != 200:
         logger.error(
-            f"Failed to get signed URL for cadence file upload: {signed_url['body']}, "
-            f"with status code: {signed_url['statusCode']}. Cadence file upload failed "
-            f"the job did not get kicked off."
+            f"Failed to get S3 pre-signed URL for file: {dependency_file_path}. "
+            f"As a result, failed to kick off job. "
+            f"Error message: {signed_url['body']}, "
+            f"with status code: {signed_url['statusCode']}."
         )
         return None
     try:
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -16,9 +16,6 @@ from imap_data_access import (
     SPICEFilePath,
 )
 from imap_data_access.file_validation import CadenceFilePath
-from imap_data_access.processing_input import (
-    ProcessingInputCollection,
-)
 from sqlalchemy import func
 from sqlalchemy.exc import IntegrityError
 
@@ -320,7 +317,7 @@ def try_to_submit_job(
     job_info: dict,
     start_date: datetime,
     version: str,
-    upstream_dependencies: ProcessingInputCollection | str,
+    serialized_dependencies: str,
 ):
     """Try to submit a batch job with the given job information.
 
@@ -334,9 +331,8 @@ def try_to_submit_job(
         Start date of the data.
     version : str
         Version of the job.
-    upstream_dependencies : ProcessingInputCollection or str
-        Either a filename string of a JSON file containing the serialized upstream
-        dependencies. Otherwise, a ProcessingInputCollection of the upstream
+    serialized_dependencies : str
+        The serialized ProcessingInputCollection of the upstream
         dependencies.
     """
     instrument = job_info["data_source"]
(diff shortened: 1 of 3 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -366,16 +362,26 @@ def try_to_submit_job(
...
+    # release
+    dependency_file = CadenceFilePath.generate_from_inputs(
+        instrument=instrument,
+        data_level=data_level,
+        descriptor=descriptor,
+        start_time=start_date.strftime("%Y%m%d"),
+        version=version,
+        extension="json",
+    )
+    dependency_file_path = Path(dependency_file.construct_path())
```
greglucas: Isn't this already a Path object?
```suggestion
    dependency_file_path = dependency_file.construct_path()
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -366,16 +362,26 @@ def try_to_submit_job(
...
+    response = upload_cadence_file(dependency_file_path, serialized_dependencies)
+    # If response is None, then the upload failed and we should skip submitting the job.
```
greglucas, line 382: Is this an error and we should log something?

Edit: I see this is logged in the function call directly. Feel free to ignore.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -389,7 +395,7 @@ def try_to_submit_job(
         "--version",
         version,
         "--dependency",
-        upstream_dependencies,
+        os.path.basename(dependency_file_path),
```
greglucas: If we already have a Path object, we should use the pathlib methods.
```suggestion
        dependency_file_path.name,
```

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -103,16 +102,32 @@ def test_lambda_handler(session):
...
+        lambda_handler(events, context)
+        mock_submit.assert_called_with(
+            session,
+            {"data_source": "swe", "data_type": "l1a", "descriptor": "sci"},
+            dt.datetime(2024, 1, 10, 0, 0),
+            "v002",
+            processing_input.serialize(),
+        )
+
+    # Clean up: Delete object from the test bucket to avoid conflicts in future tests
```
greglucas: Note that if there is an error in this test above this line the object will still be there and subsequent tests would fail. Cleaning up like this is generally frowned upon because it can leave things hanging around like extra files and things when errors crop up.

I thought we were resetting/repopulating the bucket fresh in each test? If not, would it be worth doing that?

greglucas (commented):
Does the json file only contain the dependencies, or does it contain all job information like the instrument, descriptor, start_date, etc...?

lacoak21:
> Does the json file only contain the dependencies, or does it contain all job information like the instrument, descriptor, start_date, etc...?

It only contains the dependencies.

tests/lambda_endpoints/test_batch_starter.py
lacoak21: That is a great point. I dont think we are resetting the bucket. I did a few tests to check that. I think it might be best to do that.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -738,27 +748,36 @@ def bulk_reprocessing_event(session, events):
...
 
-def upload_cadence_file(cadence_file_path: Path, upstream_dependencies):
+def upload_dependency_file(dependency_file_path: Path, serialized_dependencies: str):
     """Upload a JSON file containing a cadence job's dependencies to S3.
 
     Parameters
     ----------
-    cadence_file_path : Path
+    dependency_file_path : Path
         The cadence JSON file to upload.
```
```suggestion
    dependency_file_path : Path
        The dependency JSON file to upload.
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -738,27 +748,36 @@ def bulk_reprocessing_event(session, events):
             submit_all_jobs(session, job, start_date, end_date)
 
 
-def upload_cadence_file(cadence_file_path: Path, upstream_dependencies):
+def upload_dependency_file(dependency_file_path: Path, serialized_dependencies: str):
     """Upload a JSON file containing a cadence job's dependencies to S3.
```
```suggestion
    """Upload a JSON file containing a job's dependencies to S3.
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -738,27 +748,36 @@ def bulk_reprocessing_event(session, events):
...
     # call the upload API handler directly
     signed_url = upload_api.lambda_handler(
-        {"pathParameters": {"proxy": cadence_file_path.as_posix()}}, None
+        {"pathParameters": {"proxy": dependency_file_path.as_posix()}}, None
     )
+    if signed_url["statusCode"] != 200:
+        logger.error(
+            f"Failed to get signed URL for cadence file upload: {signed_url['body']}, "
+            f"with status code: {signed_url['statusCode']}. Cadence file upload failed "
+            f"the job did not get kicked off."
```
```suggestion
            f"Failed to get S3 pre-signed URL for file: {dependency_file_path}. "
            f"As a result, failed to kick off job. "
            f"Error message: {signed_url['body']}, "
            f"with status code: {signed_url['statusCode']}."
```

tech3371 (approved):
minor suggestions. otherwise, LGTM

PR #691, opened 2025-06-10
Adjust cadence map filename start dates

### LINKED ISSUE #682: Fix Cadence map start_dates
I (luisa) thought that the cadence map start date should be the start of the date range for the map. It should actually be the start date of the earliest pset. This defect was discovered during a SIT-4 dry run.

lacoak21:
# Change Summary

closes #682

## Overview
I had mistakenly assumed that the start date to assign to a cadence mapping job should be the end date (the date when the cadence job was triggered) minus the number of days associated with the cadence job. After some conversations with the team during a SIT-4 dry run last week, it was determined that the expected start date should actually be the date of the first pset. 

For example if an ultra l2, 3 month map job got kicked off and the date range to query for files was :

start_date= 20250101
end_date= 20250401

The earliest pset start_date: 
start_date = 20250110

The batch starter code was originally assigning the date to be 20250101 and this PR fixes it to be 20250110.

## Updated Files
   - Fix map start dates

## Testing
Slightly edit the cadence job test so that the earliest pset is slightly after the start date of the cadence date_range. This ensures that the correct date is being used.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -804,8 +804,8 @@ def cadence_processing_event(session, events):
     for job_node in potential_jobs:
         node = {
             "data_source": job_node[0],
-            "descriptor": job_node[2],
             "data_type": job_node[1],
+            "descriptor": job_node[2],
         }
         upstream_dependencies = dependency.get_jobs(
             data_source=node["data_source"],
@@ -832,7 +832,7 @@ def cadence_processing_event(session, events):
             science_input.get_time_range()[0]
             for science_input in primary_science_inputs
         ]
-        cadence_start_date = sorted(start_dates)[0]
+        cadence_start_date = min(start_dates)
 
         job_version = determine_job_version(
             session=session,
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -802,10 +802,15 @@ def cadence_processing_event(session, events):
     logger.info(f"Using {start_date=} and {end_date=} for cadence jobs.")
 
     for job_node in potential_jobs:
+        node = {
+            "data_source": job_node[0],
+            "descriptor": job_node[2],
+            "data_type": job_node[1],
+        }
         upstream_dependencies = dependency.get_jobs(
-            data_source=job_node[0],
-            data_type=job_node[1],
-            descriptor=job_node[2],
+            data_source=node["data_source"],
+            data_type=node["data_type"],
+            descriptor=node["descriptor"],
             dependency_type="UPSTREAM",
             relationship="ALL",
             start_date=start_date,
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -802,10 +802,15 @@ def cadence_processing_event(session, events):
     logger.info(f"Using {start_date=} and {end_date=} for cadence jobs.")
 
     for job_node in potential_jobs:
+        node = {
+            "data_source": job_node[0],
+            "descriptor": job_node[2],
+            "data_type": job_node[1],
```
greglucas: Switch the order here so it is 0, 1, 2?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -815,12 +820,26 @@ def cadence_processing_event(session, events):
...
+        # For cadence jobs, it is assumed there is at least one primary ScienceInput,
+        # typically only one.
+        primary_science_inputs = upstream_dependencies.get_science_inputs(
+            node["data_source"]
+        )
+        start_dates = [
+            science_input.get_time_range()[0]
+            for science_input in primary_science_inputs
+        ]
+        cadence_start_date = sorted(start_dates)[0]
```
greglucas: Can this be written as `min(...)` rather than needing to sort and then index into the 0th place.

greglucas (commented):
Is this actually what we want for a map? Lets say I wanted a 3-month map from Jan 1 -> march 30, but the Hi instrument happens to be off for 3 days at the start of the year to do anomaly resolution, but the Lo and Ultra teams are still collecting data. I don't think I'd want to see a Hi data product with a map called `20260103` while the Lo/Ultra were `20260101` for the same map cadence start.

@subagonsouth If for some reason we get 0 events for a day, do we still make an empty pset? I think this gets to your discussion yesterday where I'm curious if we would make the pset so that you at least have the exposure time showing you were observing and just nothing was collected.

sdhoyt:
> Is this actually what we want for a map? Lets say I wanted a 3-month map from Jan 1 -> march 30, but the Hi instrument happens to be off for 3 days at the start of the year to do anomaly resolution, but the Lo and Ultra teams are still collecting data. I don't think I'd want to see a Hi data product with a map called `20260103` while the Lo/Ultra were `20260101` for the same map cadence start.
> 
> @subagonsouth If for some reason we get 0 events for a day, do we still make an empty pset? I think this gets to your discussion yesterday where I'm curious if we would make the pset so that you at least have the exposure time showing you were observing and just nothing was collected.

The plan was to use the start of the data like Luisa did so we considered that a defect for the SIT, but I see what you're saying about inconsistent dates. 

With your question for Tim about the empty psets, are you talking about creating an empty pset from a packet that has no events or are you talking about creating an empty pset for any days with missing packets?

greglucas:
> With your question for Tim about the empty psets, are you talking about creating an empty pset from a packet that has no events or are you talking about creating an empty pset for any days with missing packets?

This is a great question too. If the instrument is "on", then that is an accumulated exposure. So I think it would depend on whether the instrument produces packets when there are no events but it is on... (This is likely a non-issue because we will be getting enough events on ENAs, but IDEX for example only gets handfuls of events so could potentially happen there)

tech3371 (commented):
I am going to ask one of ENA members to review and approve this. The code changes looks good but I don't know enough about the logic behind these though.

tmplummer:
> After some conversations with the team during a SIT-4 dry run last week, it was determined that the expected start date should actually be the date of the first pset.

Did I miss this conversation, or was I just confused... I think that as implemented it is correct. As @greglucas points out, we want the maps across the ENAs to be consistently named so I think that the initial implementation is correct.

> Is this actually what we want for a map? Lets say I wanted a 3-month map from Jan 1 -> march 30, but the Hi instrument happens to be off for 3 days at the start of the year to do anomaly resolution, but the Lo and Ultra teams are still collecting data. I don't think I'd want to see a Hi data product with a map called `20260103` while the Lo/Ultra were `20260101` for the same map cadence start.

I agree that we want whatever start date was used when searching for PSET data to use, not the first PSET that is found.

> @subagonsouth If for some reason we get 0 events for a day, do we still make an empty pset? I think this gets to your discussion yesterday where I'm curious if we would make the pset so that you at least have the exposure time showing you were observing and just nothing was collected.

If the instrument is taking data but there are no direct events, we will make an empty PSET. If the instrument is not taking science, no L1A de, L1B de, nor L1C pset product will be produced.

lacoak21:
Ok that makes sense. I actually cant remember if you were a part of this conversation @subagonsouth. I am happy to keep the code the way it is. Does that seem ok to you @sdhoyt?

lacoak21:
After the SDC tagup on 06/23/25, we decided as a team to keep the current implementation and close this PR

PR #694, opened 2025-06-16
I-ALiRT - Added way to furnish the SPICE kernels

laspsandoval:
# Change Summary

## Overview
Add ialirt efs, create way to query, download, and furnish spice kernels.

## Updated Files
- ialirt_efs_construct.py
   - Creates EFS that is exactly the same as SDC, but with a different name.

   - Removes EFS security group (not needed anymore since we are using ialirt EFS now)
   - Puts lambda in private with egress subnet. This was necessary to connect w/ EFS. I tested DynamoDB and S3 to make certain the lambda still connects with them

- ialirt_ingest.py
   - Lambda that now queries for the most recent SPICE kernels and downloads them to the EFS

- stackbuilder.py
   - Adds EFS construct

## Testing
- test_ialirt_ingest.py
   - Tests ialirt_ingest.py
- Deployed and tested that I was able to query and download SPICE kernels to EFS. Also tested that s3 and dynamoDB still connect.

### DIFF
--- a/sds_data_manager/constructs/ialirt_efs_construct.py
+++ b/sds_data_manager/constructs/ialirt_efs_construct.py
@@ -1,0 +1,101 @@
+"""Configure the EFS."""
+
+import aws_cdk as cdk
+from aws_cdk import aws_ec2 as ec2
+from aws_cdk import aws_efs as efs
+from constructs import Construct
+
+
+class IAlirtEFSConstruct(Construct):
+    """Elastic File System for storing various software and data.
+
+    This file system can be mounted by multiple resources. It
+    will be attached to AWS resources that run the
+    models (ECS, EC2, Lambda) and produce output data.  It will
+    also be mounted by the visualizer app to view that model data.
+    This allows model data to be seamlessly transferred between
+    resources without having to push/pull from s3. For the
+    visualizer display, it should have decent I/O capabilities
+    to read and write the model output data to the frontend quickly.
+
+    NOTE: The EFS file systems will not be removed during a CDK
+    destroy. This is a protection mechanism to
+    prevent data from being deleted. After destroying this
+    stack you will need to manually delete these
+    resources via the AWS console or CLI.
+    """
+
+    def __init__(
+        self,
+        scope: Construct,
+        construct_id: str,
+        vpc: ec2.Vpc,
+        **kwargs,
+    ) -> None:
+        """Construct the EFS.
+
+        Parameters
+        ----------
+        scope : Construct
+            Parent construct.
+        construct_id : str
+            A unique string identifier for this construct.
+        vpc : ec2.Vpc
+            VPC into which to put the resources that require networking.
+        kwargs : dict
+            Keyword arguments
+
+        """
+        super().__init__(scope, construct_id, **kwargs)
+
+        # Initialize EFS related information that other resources
+        # will need to access EFS or mount EFS.
+        self.volume_name = "IALIRT-SPICE-EFS"
+        self.efs_path = "/data"
+
+        # Define EFS security group, ports are added in EC2 stack
+        self.efs_security_group = ec2.SecurityGroup(
+            self,
+            "IAlirtEFSSecurityGroup",
+            vpc=vpc,
+            description="No outbound rule for EFS",
+            allow_all_outbound=False,
+            security_group_name="IAlirtEFSSecurityGroup",
+        )
+
+        # Add inbound rule for TCP port 2049
+        self.efs_security_group.connections.allow_from_any_ipv4(
+            ec2.Port.tcp(2049),
+            "Allow services to mount the EFS",
+        )
+
+        self.efs = efs.FileSystem(
+            self,
+            "IAlirtEFS",
+            vpc=vpc,
+            file_system_name=self.volume_name,
+            # This will automatically downscale infrequently accessed
+            # data to a cheaper storage tier after 14 days of inactivity.
+            # TODO: Investigate what lifecycle policy we need or if
+            # we should remove if it effects read latency.
+            lifecycle_policy=efs.LifecyclePolicy.AFTER_14_DAYS,
+            # TODO: Investigate what performance characteristics we need
+            # For now, use general purpose, but we can experiment later.
+            performance_mode=efs.PerformanceMode.GENERAL_PURPOSE,
+            security_group=self.efs_security_group,
+            # Need to add EFS mount points to all Private subnets
+            # This will allow EC2 in either AZ to connect without data
+            # transfer charges
+            vpc_subnets=ec2.SubnetSelection(
+                subnet_type=ec2.SubnetType.PRIVATE_ISOLATED
+            ),
+            removal_policy=cdk.RemovalPolicy.DESTROY,
+        )
+
+        # Make an access point for others to be able to access the drive
+        self.spice_access_point = self.efs.add_access_point(
+            "IAlirtSpiceAccessPoint",
+            create_acl=efs.Acl(owner_gid="1000", owner_uid="1000", permissions="750"),
+            path=self.efs_path,
+            posix_user=efs.PosixUser(gid="1000", uid="1000"),
+        )
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
@@ -3,13 +3,12 @@
 import json
 import logging
 import os
-from datetime import datetime, timezone, timedelta
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
-import requests
-import shutil
-import spiceypy
-import imap_data_access
 
+import imap_data_access
+import requests
+import spiceypy
 from imap_data_access.processing_input import (
     ProcessingInputCollection,
     SPICEInput,
@@ -24,8 +23,8 @@ from boto3.dynamodb.conditions import Key
 from imap_processing import imap_module_directory
 from imap_processing.ialirt.l0.process_hit import process_hit
 from imap_processing.ialirt.l0.process_swe import process_swe
-from imap_processing.utils import packet_file_to_datasets
 from imap_processing.spice.time import str_to_et
+from imap_processing.utils import packet_file_to_datasets
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
@@ -43,13 +42,11 @@ EFS_BASE_PATH = Path("/mnt/data")
 
 
 def get_latest_spice_kernels() -> ProcessingInputCollection:
-    """
-    Query the SPICE metakernel API to retrieve the latest SPICE kernel filenames
-    for the past week.
+    """Query the SPICE metakernel API for latest SPICE kernel filenames.
 
     Returns
     -------
-    ProcessingInputCollection
+    dependency_inputs: ProcessingInputCollection
         A collection containing a SPICEInput object with the list of kernel filenames
         returned from the metakernel API.
     """
@@ -72,7 +69,7 @@ def get_latest_spice_kernels() -> ProcessingInputCollection:
     }
 
     logger.info(f"Sending request to {url} with params: {params}")
-    response = requests.get(url, params=params)
+    response = requests.get(url, params=params, timeout=10)
     metakernel_files = response.json()
 
     logger.info(f"Found metakernel files: {metakernel_files}. Adding to collection.")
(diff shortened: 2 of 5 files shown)


laspsandoval:
pre-commit.ci autofix

laspsandoval:
pre-commit.ci autofix

laspsandoval:
pre-commit.ci autofix

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -19,6 +27,93 @@
...
 logger.setLevel(logging.INFO)
 
+KERNELS = {
+    "ephemeris_predicted",
+    "ephemeris_90days",
+    "planetary_ephemeris",
+    "spacecraft_clock",
+    "leapseconds",
+    "imap_frames",
+    "science_frames",
```
tmplummer, line 37: Is `sds-data-manager` a dependency in the iALiRT package? If so, it might be worth getting these names from there.
```
SpacecraftEphemerisKernels.EPHEMERIS_PREDICTED.name.lower(),
etc
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -19,6 +27,93 @@
...
+def get_ephemeris_time():
+    """Get the current ephemeris time in seconds since J2000.
+
+    Returns
+    -------
+    et: float
+        The current ephemeris time in seconds since J2000.
+    """
+    now = datetime.now(timezone.utc)
+    j2000 = datetime(2000, 1, 1, 12, 0, 0, tzinfo=timezone.utc)
```
tmplummer: I'm not sure what the needed accuracy of this function is, but it is not quite correct.
This is not the correct epoch for J2000.
From the imap_processing.spice.time.py file:
```
# The UTC string was generated by:
# >>> spiceypy.et2utc(spiceypy.unitim(0, "TT", "ET"), "ISOC", 9)
TTJ2000_EPOCH = np.datetime64("2000-01-01T11:58:55.816", "ns")
```
Furthermore, ET is TDB which has a non-fixed offset from TT. The safest approach would be to use spiceypy for this calculation but perhaps that precision is not needed?

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -19,6 +27,93 @@
...
+    dependency_inputs: ProcessingInputCollection
+        A collection containing a SPICEInput object with the list of kernel filenames
+        returned from the metakernel API.
+    """
+    dependency_inputs = ProcessingInputCollection()
+
+    now = datetime.now(timezone.utc)
+    one_week_ago = now - timedelta(weeks=1)
+    # Define J2000 epoch: 2000-01-01T12:00:00 UTC
+    j2000 = datetime(2000, 1, 1, 12, 0, 0, tzinfo=timezone.utc)
```
tmplummer: This is a duplicated definition of the J2000 epoch. It should be moved to a common global variable if it is going to be used in multiple places. See above comment... this is not the correct definition.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -19,6 +27,93 @@
...
+    end_time = (now - j2000).total_seconds()
+    start_time = (one_week_ago - j2000).total_seconds()
```
tmplummer: This is converting to ET. Might be worth naming these variables to reflect that.
```suggestion
    et_end_time = (now - j2000).total_seconds()
    et_start_time = (one_week_ago - j2000).total_seconds()
```

IMO, we should ask Bryan to add UTC time strings as a query option. For now, this estimated time approach is probably OK.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -19,6 +27,93 @@
...
+    return dependency_inputs
+
+
+def download_spice_file(dependencies) -> list[Path]:
+    """Download SPICE kernel files from the IMAP data archive and store them in EFS.
+
+    Parameters
+    ----------
+    dependencies: ProcessingInputCollection
+        A collection containing a SPICEInput object with the list of kernel filenames
```
tmplummer, line 88: It is probably worth noting that this list is priority ordered such that furnishing in order results in the correct SPICE priority.

tmplummer (commented):
I didn't review the infrastructure code. I do have a few comments about the time conversions.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: The ialirt stack is separate from the sdc stack. The Friday before I went on vacation I tried to use the spice_query_api (where the names are defined), but the problem is that spice_query_api imports spice_query_api, which relies on the database. And that opened a whole can of worms. It would mean that I-ALIRT becomes dependent on the SDC stack (not ideal) and also opens a lot of issues with networking (security groups, subnets), additional dependencies on libraries required for the database, etc. So I think I might have to keep this a standalone list.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: I ran into the chicken/egg situation here. I needed approximate timestamps to furnish the most recent kernels. But to furnish the most recent kernels I needed the leapseconds kernel to use your library: str_to_et. So I ended up copying what dependency.py does:

```
#TODO: remove this once Bryan changes takes in 'yyyymmdd' format
            def yyyymmdd_to_seconds_since_j2000(
                date_str: str, add_24_hrs=False
            ) -> float:
                # Parse input date string
                dt = datetime.strptime(date_str, "%Y%m%d").replace(tzinfo=timezone.utc)
                if add_24_hrs:
                    dt += timedelta(hours=24)
                # Define J2000 epoch: 2000-01-01T12:00:00 UTC
                j2000 = datetime(2000, 1, 1, 12, 0, 0, tzinfo=timezone.utc)

                # Compute seconds difference
                delta = dt - j2000
                return delta.total_seconds()
```

I can add a TODO to change when Bryan changes takes in 'yyyymmdd' format. Otherwise I think this might be the best that I can do for now.

tmplummer: Ok, I would still suggest fixing the J2000 epoch to reduce the error.
```suggestion
    j2000 = datetime(2000, 1, 1, 11, 58, 55.816, tzinfo=timezone.utc)
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: Yes. I forgot to delete the function above. Fixed the typo.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: Ok. I put a note. Everything in the code is ok, though, right?

tmplummer (approved):
The pieces for querying metakernel API, downloading, and furnishing kernels looks good to me.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
bryan-harter: This might be opening a whole can of worms, but if the stacks are so different and they don't interact much with each other, should I-ALIRT just be its own repository?

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -19,6 +27,84 @@
...
+    """
+    dependency_inputs = ProcessingInputCollection()
+
+    now = datetime.now(timezone.utc)
+    one_week_ago = now - timedelta(weeks=1)
+    # Define J2000 epoch: 2000-01-01T12:00:00 UTC
+    # TODO: remove this once Bryan changes takes in 'yyyymmdd' format
+    j2000 = datetime(2000, 1, 1, 11, 58, 56, tzinfo=timezone.utc)
+    et_end_time = (now - j2000).total_seconds()
+    et_start_time = (one_week_ago - j2000).total_seconds()
```
bryan-harter, line 59: Technically this will be off from the "real" ET that SPICE uses by 5 seconds (because there has been 5 leap seconds since the year 2000). But if you don't care about that level of precision, this is a good approximation.

bryan-harter:
Oops sorry, reviewed after it was merged lol. But it looked good to me!

sds_data_manager/constructs/ialirt_efs_construct.py
```diff
@@ -0,0 +1,101 @@
+"""Configure the EFS."""
+
+import aws_cdk as cdk
+from aws_cdk import aws_ec2 as ec2
+from aws_cdk import aws_efs as efs
+from constructs import Construct
+
+
+class IAlirtEFSConstruct(Construct):
```
tech3371, line 9: I don't know if creating different EFS for I-ALiRT will help here. If you create new EFS, then SPICE data won't go in it unless you update SPICE indexer lambda to write to it. Otherwise, it will be empty EFS mounted to a lambda. This means that dependency on SDC stack still remains. But SDC is not using EFS mount anymore in batch job. We download all SPICE through download API instead. It has been working. I am wondering if you want to do the same here for I-ALiRT. If you do that, then we don't need to have EFS for both SDC and I-ALiRT and will remove dependency between those components.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
tech3371: Also, Bryan updated SPICE query to take in human readable time. We didn't get to remove function that converts to j2000 because of SIT-4 rush. We mean to do that soon. But for you, you can pass in date like this 'yyyymmdd' right now.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -19,6 +27,84 @@
...
+    Returns
+    -------
+    spice_files: list[Path]
+        A list of Path objects representing the SPICE files stored in EFS.
+
+    Notes
+    -----
+    List is priority ordered so furnishing in order results in correct SPICE priority.
+    """
+    imap_data_access.config["DATA_DIR"] = EFS_BASE_PATH
```
tech3371, line 100: I see. You are download SPICE to EFS and that's how you are writing to EFS. If that's what you want, that looks good to me. I would be careful of keeping things around. It can become out of sync. Also, I would be careful with setting `DATA_DIR` as EFS path because this will result in download other files such as your l0 files or ancillary files into EFS too. That's what we found before SIT-4 and that's why we decided to move away from EFS mounting approach. Just FYI.

tech3371 (commented):
I left few comments.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
tmplummer: @tech3371 and @bryan-harter , I'm not sure that the human readable date format is working for the spice-query or the metakernel APIs.
https://api.dev.imap-mission.com/metakernel?start_time=20250415&end_time=20250416&file_types=leapseconds,spacecraft_clock&list_files=True
https://api.dev.imap-mission.com/spice-query?start_time=20250416&end_time=20250417&type=leapseconds

bryan-harter: Yeah i don't think that is working yet, I think I did make changes to it, but in retrospect it requires that the leapsecond kernel already be furnished before it can convert datetimes over to ET. So it needs to be fixed to handle that first.

tech3371: Oh I see. We were seeing with kernel not loaded error before and we didn't know what caused it. Now, this explains.

laspsandoval: Got it! It really is just an estimate to get the most recent kernel. Thanks for your responses. I'll update this part when that feature becomes available.

sds_data_manager/constructs/ialirt_efs_construct.py
laspsandoval: Thanks Tenzin! We only need EFS for I-ALiRT to store the spice kernels (especially the planetary kernel)

Issue #554, opened 2025-04-25
BUG - Inconsistent ingestion time handling


greglucas:
### Description of the issue

We currently store "ingestion time" in different formats across the DB Tables which makes it hard to follow and inconsistent. We should settle upon a single string format for all of these.

SPICE (a comma between date and time): "%Y-%m-%d, %H:%M:%S"
Acnillary: "%Y%m%d %H:%M:%S"
Science: "%Y-%m-%d %H:%M:%S%z"

Instead of worrying about calculating all of these ourselves, should we instead let the DB take care of this for us by setting the `created_at` and `last_updated` automatic fields?
https://stackoverflow.com/questions/13370317/sqlalchemy-default-datetime

### Code Snippet:

```python
Code
```

tech3371:
@greglucas can you confirm if mentioned PR fixed this issue? If so, can you close this?

greglucas:
No, I think we should still update this. Adding links to what I mean with inconsistencies.

I think we should remove the comma from here:
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/8d0ac6f0a531ebe7418624c597499b860195a80f/sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_query_api.py#L146-L147

https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/8d0ac6f0a531ebe7418624c597499b860195a80f/sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py#L137

tech3371:
cool. Then I will look into that.

tech3371:
Looks like it is completed with update in `get_file_ingestion_date()`.

PR #696, opened 2025-06-17
MNT: SPICE ingest time

### LINKED ISSUE #556: MNT: Update SPICE indexer to use UTC time
Yeah somewhere along the line it does something weird with the time zones, since that number is exactly 6 hours ahead of UTC (and we're 6 hours behind). The unit tests on github are accurate, so the number should be 1744233173.0.  I think virtual environments (like my dev environment and the github testing environment) default to UTC, but your laptop is set to mountain time. 

We should probably go through the test and the indexing code itself, and make sure all the inserts/reads have the appropriate time zone set in the datetime objects.

_Originally posted by @bryan-harter in https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/549#discussion_r2060518467_

tech3371:
# Change Summary
close #556 and may fix this as well, #554.

## Overview
Fixing SPICE ingestion date to S3 modified date. This will help in sync database correctly from S3 in future PR.
Fixed timezone diff bug.

## Updated Files
   - added s3 env
   - Fix timezone issue
   - Fix ingestion date

### DIFF
--- a/sds_data_manager/constructs/indexer_lambda_construct.py
+++ b/sds_data_manager/constructs/indexer_lambda_construct.py
@@ -246,6 +246,7 @@ class SPICEIndexerLambda(Construct):
             environment={
                 "DATA_DIR": spice_mount_path,
                 "SECRET_NAME": db_secret_name,
+                "S3_BUCKET": data_bucket.bucket_name,
             },
         )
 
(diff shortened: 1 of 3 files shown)


bryan-harter:
Looks good to me!

bryan-harter (approved):

PR #703, opened 2025-06-24
Ialirt sit testing

laspsandoval:
# Change Summary

## Overview
Updates to include codice and swapi.

## Updated Files
- ialirt_ingest.py
   - Updates to include codice and swapi.

## Testing
Deployment and test.

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
@@ -22,7 +22,9 @@ from imap_data_access.processing_input import (
 from imap_processing import imap_module_directory
 from imap_processing.cdf.utils import load_cdf
 from imap_processing.ialirt.l0.parse_mag import process_packet
+from imap_processing.ialirt.l0.process_codice import process_codice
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -255,11 +260,21 @@ def process_algorithms(combined: xr.Dataset, algorithm_table):
             download_path = get_ancillary(instrument, "l1b-calibration")
             calibration_data = load_cdf(download_path)
             result, _ = process_func(combined, calibration_data)
+        elif instrument == "codicelo":
+            result, _ = process_func(combined)
+        elif instrument == "codicehi":
+            _, result = process_func(combined)
```
bourque, line 266: Just wanted to note here that this is something I would like to revisit after the SIT. I think the way my code is set up right now isn't very efficient (since both `codicelo` and `codicehi` is processed regardless of which instrument is needed). But this is good enough for now I think.

bourque (approved):
Looks good!

torimarbois (approved):

Issue #518, opened 2025-04-04
Clear out temp space on packet download Lambda


greglucas:
Right now, we are downloading files to the local disk and then uploading them to s3. After this (or upon entering the lambda), we could ensure that there is enough tmp space by removing all files currently on the Lambda.

Something like:
```
import shutil
shutil.rmtree('/tmp')
```
or glob all the files and iterate through and remove them individually.

tech3371:
@greglucas Can you specify which lambda we need this for? Do we still need this?

PR #707, opened 2025-06-25
MNT: Remove EFS and refactor SPICE indexer

### LINKED ISSUE #698: MNT: Delete SDC EFS stack
Now that SDC is not using EFS and I-ALiRT has it's own EFS, we are deleting EFS stack from SDC stack.

tech3371:
# Change Summary
closes #698 

## Overview
Deleting EFS. EFS was connect to SPICE indexer lambda logic. SPICE indexer was reading latest leapseconds and clock kernels from EFS. Fixed those logic to query SPICE query API and download from S3.

## Deleted Files
- sds_data_manager/constructs/efs_construct.py
   - remove EFS constructs

## Updated Files
- CDK construct clean up
   - added `file_path` to setup for future work to update DB sync lambda to sync SPICE db.
   - Main change of this PR.
   - removed EFS write functions
   - updated fursh latest spice kernel to query and download from s3.
   - updated SPICE indexer to write to DB with new added key

## Testing
Fixing tests to new changes
- tests/lambda_endpoints/test_spice_query_api.py


sds_data_manager/constructs/indexer_lambda_construct.py
```diff
@@ -234,17 +223,13 @@ def __init__(
...
-                efs_construct.spice_access_point, spice_mount_path
-            ),
             timeout=cdk.Duration.minutes(1),
             architecture=lambda_.Architecture.X86_64,
             layers=layers,
             memory_size=1000,
             security_groups=[rds_security_group],
             environment={
-                "DATA_DIR": spice_mount_path,
+                "DATA_DIR": "/tmp/",  # noqa: S108
```
greglucas, line 232: Do we need DATA_DIR specified, or can we just let it go to the current directory on the lambda?

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -154,6 +154,7 @@ class SPICEFiles(Base):
     __tablename__ = "spice_files"
 
     file_name = Column(String, nullable=False, primary_key=True, unique=True)
+    file_path = Column(String, nullable=False, unique=True)
```
greglucas: Why do we need both of these? Can we just have the path be the path+name?

greglucas (commented):
A minor nit that I don't think we should put both file_name and file_path in the DB since I think they can be determined from one another through `imap_data_access.FilePath(filename).construct_path()` or `path.name` the other way. So I think we should just store one or the other and construct the other from that when we need it.

sds_data_manager/constructs/indexer_lambda_construct.py
tech3371: I added that because I didn't want download of ingested SPICE file to take all the space of Lambda. We download those files to parse and index it to DB. Setting this to tmp cleans up after itself.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: I thought about this. I didn't feel comfortable removing `file_name` because it is used often and I didn't know enough. But if it's important, I can look into it. It may take little bit more time fixing and updating tests(this takes more time usually).

sds_data_manager/constructs/indexer_lambda_construct.py
greglucas: Actually, I am unclear if this would even work without it, so it may be required to have this even? i.e. more important than cleaning up after itself.

This comment pertains to the cleaning up of `/tmp` space in https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/518, it is actually not as temporary as one might think on each lambda invocation. One invocation after another may or may not start fresh, so even this tmp space could start filling up. I guess we can just keep this in the back of our minds for now.

tests/lambda_endpoints/test_spice_metakernel_api.py
```diff
@@ -29,6 +29,7 @@ def _insert_test_file(session, filename, intervals, upload_time=0):
     version = spice_object.spice_metadata["version"]
     metadata_params = {
         "file_name": filename,
+        "file_path": f"spice/{filename}",
```
greglucas:
```suggestion
        "file_path": f"imap/spice/{filename}",
```

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: OK, now I see that this is the only place we use `file_name` and in Science/Ancillary etc we use file_path. I think we should try to keep things consistent and change over to file_path instead of file_name. These are details for us in the database and I think consistency is important still. So my preference is to see how hard it would be to replace file_name altogether.

sds_data_manager/constructs/indexer_lambda_construct.py
tech3371: Cool. Let's revisit that in the ticket.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: yeah, it's just an extra field in SPICE table, but otherwise, it matches with what Science/Ancillary is doing.

greglucas: Yes, and I think I'd still prefer to remove "file_name" altogether so there isn't any confusion. I think it is redundant.

tech3371: Can I leave `file_name` for this PR? It's touching lots of component in Metakernel() that I don't understand.

tech3371: As we discussed in Slack, going to make this change only in this PR.
>Maybe that is the trick here... You make file_path the primary key now and leave file_name as the extra portion

greglucas (approved):
I'll write a follow-up ticket to remove the `file_name` column.

tech3371:
> I'll write a follow-up ticket to remove the `file_name` column.

Thank you for creating that issue!

PR #708, opened 2025-06-26
MNT: Change routes for batch jobs API

greglucas:
* batch -> processing
* job id query param -> path param

Batch is an implementation detail IMO. We really want to refer to these more generally as "processing jobs" I think.

APIs should use path params for getting an ID from a route. This is a bit odd with the id containing slashes, but we can get it to work with a proxy+ url.

### DIFF
--- a/sds_data_manager/constructs/sds_api_manager_construct.py
+++ b/sds_data_manager/constructs/sds_api_manager_construct.py
@@ -256,7 +256,12 @@ class SdsApiManager(Construct):
             layers=layers,
         )
         api.add_route(
-            route="/batch-job",
+            route="/processing-jobs",
+            http_method="GET",
+            lambda_function=batch_job_query_api_lambda,
+        )
+        api.add_route(
+            route="/authorized/processing-jobs",
             http_method="GET",
             lambda_function=batch_job_query_api_lambda,
         )
@@ -275,7 +280,16 @@ class SdsApiManager(Construct):
             layers=layers,
         )
         api.add_route(
-            route="/batch-logs",
+            # {id+} is used to allow for any pathParams after /batch-logs/
+            # This is needed because the log stream ID can contain slashes
+            route="/processing-logs/{id+}",
+            http_method="GET",
+            lambda_function=batch_logs_api_lambda,
+        )
+        api.add_route(
+            # {id+} is used to allow for any pathParams after /batch-logs/
+            # This is needed because the log stream ID can contain slashes
+            route="/authorized/processing-logs/{id+}",
             http_method="GET",
             lambda_function=batch_logs_api_lambda,
         )
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/batch_logs_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/batch_logs_api.py
@@ -2,6 +2,7 @@
 
 import json
 import logging
+from urllib.parse import unquote
 
 import boto3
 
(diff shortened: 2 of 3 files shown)


greglucas:
Example query: https://api.dev.imap-mission.com/processing-logs/ProcessingJob-idex/default/2c4234d191bb4a13bd303b2ac6a72b81

sds_data_manager/constructs/sds_api_manager_construct.py
```diff
@@ -260,6 +260,11 @@ def __init__(
             http_method="GET",
             lambda_function=batch_job_query_api_lambda,
         )
+        api.add_route(
```
tech3371, line 263: should we change above API to be this one instead?

sds_data_manager/constructs/sds_api_manager_construct.py
```diff
@@ -275,7 +280,9 @@ def __init__(
             layers=layers,
         )
         api.add_route(
-            route="/batch-logs",
+            # {id+} is used to allow for any pathParams after /batch-logs/
+            # This is needed because the log stream ID can contain slashes
+            route="/processing-logs/{id+}",
```
tech3371, line 285: Thank you for the comments!

sds_data_manager/lambda_code/SDSCode/api_lambdas/batch_logs_api.py
```diff
@@ -30,25 +31,8 @@ def lambda_handler(event, context):
...
-
-    if not job_log_stream_id:
-        logger.info("No job_log_stream_id provided in query parameters.")
-        return {
-            "statusCode": 400,
-            "body": json.dumps({"error": "job_log_stream_id is required."}),
-        }
-    logger.info(f"Fetching logs for job_log_stream_id: {job_log_stream_id}")
+    # e.g. ProcessingJob-codice-l3/default/bcc41f2cc3f146eb818a12eec1d7177c
+    job_log_stream_id = unquote(event["pathParameters"]["id"])
```
tech3371, line 38: can we check and return missing input status?

tech3371 (commented):
Thank you for doing this!

sds_data_manager/lambda_code/SDSCode/api_lambdas/batch_logs_api.py
greglucas: If you test with a bad id you get back an error saying that log stream doesn't exist: https://api.dev.imap-mission.com/processing-logs/abc

```
{"error": "Could not fetch logs for id [abc]: An error occurred (ResourceNotFoundException) when calling the GetLogEvents operation: The specified log stream does not exist."}
```

The way the API is defined, it must have an "id" field and if it is missing that is a non-existent route and would never make it to this lambda. This will be helped by the documentation of the API endpoint.

sds_data_manager/constructs/sds_api_manager_construct.py
greglucas: 😂 I had a stale review comment saying I was leaving the other one around for now and asking whether this was OK to replace our existing one but I never clicked submit review. I guess you answered my question here that we should do that :) Good eye, thanks!

greglucas:
I have now renamed the `batch-jobs` endpoint to `processing-jobs`. I have also added the `/authorized` versions of these as I think we should do that and remove the unauthorized portions later. For now, lets keep both though.

sds_data_manager/lambda_code/SDSCode/api_lambdas/batch_logs_api.py
tech3371: Thank you! That's good to know. Can you add a comment mentioning that the API handles missing input error?

tech3371 (approved):
Besides minor one requests, it looks great! Thank you for adding the authorized one!

sds_data_manager/lambda_code/SDSCode/api_lambdas/batch_logs_api.py
greglucas: Sure, I added it above the "id" request as a NOTE, is that where you wanted it?

tech3371: What you had was great. Sorry I missed that!

greglucas: I just added it, so I don't think you missed anything :)

PR #710, opened 2025-06-30
Updating all queues and some tests

maxinelasp:
# Change Summary
Add a delay queue and set it up to receive the MAG L1B events. 

This required some updates to batch starter to handle events from any queues. I went ahead and made it all generic enough that we won't need to make updates in the future if we want to change the queues again.

## Overview

To address the race condition at MAG L1C, we added a delay queue to give the L1B processing jobs extra time to complete. To do this, I added an additional EventBridge rule for MAG L1B, a new queue with a delay on it, and updated batch starter to delete the queue items from multiple queues instead of just one. 

## Testing

I needed to fix a few tests to include the queue information for deleting. This is included by default on all queue items. I also added a test to make sure the queue the deletion.

Note: this cannot be merged until after the I-ALiRT SIT


sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -114,11 +113,15 @@ def __init__(
...
+        # This sets up the lambda to be triggered by the SQS queues. Since they are FIFO
+        # queues, each instrument will have messages processed in order. However,
         # different instruments will be processed in parallel, with multiple instances
         # of the batch_starter lambda.
-        self.instrument_lambda.add_event_source(SqsEventSource(sqs_queue))
+        # The nominal case is for there to be a file arrived queue and a delayed
+        # file arrived queue. On the batch starter side, all events will look the
+        # same from the two queues.
+        for q in sqs_queues:
+            self.instrument_lambda.add_event_source(SqsEventSource(q))
```
tech3371, line 124: Thank you for making this generic!

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -60,7 +60,7 @@ def __init__(
             RDS security group.
         vpc : ec2.Vpc
             VPC into which to put the resources that require networking.
-        sqs_queue: sqs.Queue
+        sqs_queues: sqs.Queue
```
tech3371: Should this be this now?
```suggestion
        sqs_queues: list[sqs.Queue]
```

sds_data_manager/constructs/sqs_construct.py
```diff
@@ -60,7 +60,69 @@ def __init__(
...
+            instrument_event_match = {
+                "object": {
+                    "key": [{"exists": True}],
+                    "instrument": [instrument_name],
+                },
+            }
+
+            if instrument_name in delayed_cases:
+                data_level = delayed_cases[instrument_name]
+                # Don't send L1B events to the main queue, they go to the delay queue
```
tech3371: Can we rephrase this?
```suggestion
                # Update main queue filters to not send these exception cases. They will go to the delay queue
```

sds_data_manager/constructs/sqs_construct.py
```diff
@@ -60,7 +60,69 @@ def __init__(
...
+                instrument_event_match["object"]["data_level"] = [
+                    {"anything-but": data_level}
+                ]
```
tech3371, line 100: should we add instrument name here also? Otherwise, it will not send all L1B to main queue which is not what we want here.

sds_data_manager/constructs/sqs_construct.py
```diff
@@ -60,7 +60,69 @@ def __init__(
...
+                    self,
+                    "MagFileArrivedL1B",
+                    rule_name="mag_file_arrived_l1b",
+                    event_pattern=events.EventPattern(
+                        source=["imap.lambda"],
+                        detail_type=["Processed File"],
+                        detail={
+                            "object": {
+                                "key": [{"exists": True}],
+                                "instrument": ["mag"],
```
tech3371:
```suggestion
                                "instrument": [instrument_name],
```

sds_data_manager/constructs/sqs_construct.py
```diff
@@ -60,7 +60,69 @@ def __init__(
...
+                                "data_level": [data_level],
+                            },
+                        },
+                    ),
+                )
```
tech3371, line 118: I am noticing that we are carrying MAG in this event rule. Can we make this event rule to be generic too?

sds_data_manager/constructs/sqs_construct.py
tech3371, line 1: Can you test this components in the AWS? Changes looks good to me but sometimes, event rule can be very particular about syntax.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -498,6 +498,33 @@ def submit_all_jobs(session, job_node, start_date, end_date, filter_dependencies
...
+        the normal file arrived queue or the delay queue.
+    """
+    source_arn = event[
+        "eventSourceARN"
+    ]  # e.g., arn:aws:sqs:us-east-1:123456789012:my-queue-name.fifo
+    queue_name = source_arn.split(":")[-1]
+    region = source_arn.split(":")[3]
+    account_id = source_arn.split(":")[4]
+    queue_url = f"https://sqs.{region}.amazonaws.com/{account_id}/{queue_name}"
+    return queue_url
```
tech3371, line 525: good refactor!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
```diff
@@ -110,13 +110,21 @@ def send_event_from_indexer(file_obj):
...
     detail = {
-        "object": {"key": str(file_obj.filename), "instrument": file_obj.instrument}
+        "object": {
+            "key": str(file_obj.filename),
+            "instrument": file_obj.instrument,
+            "data_level": "ancillary",
+        }
     }
 
+    # used to filter MAG L1B events later
```
tech3371:
```suggestion
    # used to filter other event types for SQS here
```

tech3371 (commented):
Thank you for refactoring and making good generic! This will help in the future for sure.

I had minor requests.

sds_data_manager/constructs/sqs_construct.py
maxinelasp: Since this is inside the `if instrument_name in delayed_cases:`, it is only applied to instruments which have that delayed case, which in this case is just mag. Good catch though, it was good to double check that.

tech3371: That's what I thought too but if we look closely, this `instrument_event_match` is used by main queue. Am I missing something, lol?

maxinelasp:
Proof of queue working to delay 4 MAG L1B file arrivals: 
image

sds_data_manager/constructs/sqs_construct.py
maxinelasp: So in the case where instrument_name is in the delayed cases, we do want to filter the specific main queue for that instrument only. So here, it is saying "for the instrument instrument_name, if we're delaying one level, filter that instrument's level events out." 

I just tested this with GLOWS and the GLOWS L1B events skipped the delay queue and triggered L2. So we're good!

tech3371 (approved):
Thank you for confirming it. Event rule is tricky. The rules looks good.

PR #726, opened 2025-07-15
Mag config updates

maxinelasp:
# Change Summary
Add spice dependencies for MAG L2.

## Overview

MAG L2 will generate 4 output files from the same run, but the config only says one so the processing only runs once. I think most people will depend on the SRF file so that should work.


lacoak21 (approved):
LGTM!

bourque (approved):

PR #727, opened 2025-07-16
ENH: Add API Key authorization

greglucas:
This adds api key authorization through a lambda function authorizer. The keys are stored in AWS SSM and retrieved from there. There is a new script added to manage the keys, listing, removing, and adding keys that I tested out and currently is available in dev for you to try as well.

Then you can make a request like:
```
IMAP_API_KEY=key-you-made-from-script IMAP_DATA_ACCESS_URL=https://api.dev.imap-mission.com/api-key imap-data-access --debug query --instrument swapi
```

and you should get data back.

### DIFF
--- a/sds_data_manager/constructs/api_gateway_construct.py
+++ b/sds_data_manager/constructs/api_gateway_construct.py
@@ -16,6 +16,7 @@ from aws_cdk import aws_apigatewayv2_integrations as apigwv2_integrations
 from aws_cdk import aws_certificatemanager as acm
 from aws_cdk import aws_cloudwatch as cloudwatch
 from aws_cdk import aws_cloudwatch_actions as cloudwatch_actions
+from aws_cdk import aws_iam as iam
 from aws_cdk import aws_lambda as lambda_
 from aws_cdk import aws_route53 as route53
 from aws_cdk import aws_route53_targets as targets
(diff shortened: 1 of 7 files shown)


greglucas:
Note that this adds new routes for query, download, upload that use the `/api-key/query` path. So we have 3 routes now:
* `/query` -> public files
* `/authorized/query` -> oauth verification / used by the public website and human logins
* `/api-key/query` -> token verification / used by scripts and automated users

sds_data_manager/constructs/api_gateway_construct.py
```diff
@@ -198,8 +225,14 @@ def add_route(
...
         # Add the authorizer to the route if it is a route that requires authentication
-        authorizer = self.authorizer if route.startswith("/auth") else None
-        authorization_scopes = [self.auth_scope] if route.startswith("/auth") else None
+        authorizer = None
+        authorization_scopes = None
+        if route.startswith("/api-key"):
+            authorizer = self.api_key_authorizer
+        elif route.startswith("/authorized"):
+            authorizer = self.authorizer
+            authorization_scopes = [self.auth_scope]
```
lacoak21, line 234: Is the auth scope like the group of users?

sds_data_manager/lambda_code/authorization/lambda_api_key_authorizer.py
```diff
@@ -0,0 +1,27 @@
+"""Authorization for API Keys within the SDS."""
+
+import json
+
+import boto3
+
+# Retrieve the API keys from AWS Systems Manager Parameter Store
```
lacoak21, line 7: Oh that is clever caching them outside of the handler.

sds_data_manager/lambda_code/authorization/manage_api_keys.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+"""Simple script to manage API keys in AWS SSM Parameter Store (SecureString).
+
+AWS_PROFILE and AWS_REGION environment variables should be set to the appropriate values
+for the account and region where the SSM parameter is stored.
+
+Usage:
```
lacoak21, line 7: I tested list, add, and remove 👍

lacoak21 (approved):
LGTM

sds_data_manager/constructs/api_gateway_construct.py
greglucas: Yes, that is defined by our internal auth group and then they can specify which users get what scopes attached to them.

PR #728, opened 2025-07-16
ENH: Add Repoint date range for ENA and GLOWS and kickoff pointing attitude job

### LINKED ISSUE #629: Create Repoint DB table and refactor code that find latest repoint file using the table
I think we have discussed this in the past and decided on not using a DB table, but this seems like there will be a lot of downloading and reading the repointing file just to get the dates for every ENA instrument's files. Now that we are using this in more places should we add a database table for it to be able to query by repoint_id and get the start/end time periods?

_Originally posted by @greglucas in https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/628#discussion_r2109750042_

tech3371:
# Change Summary
close #629
## Updated Files

- usds_data_manager/lambda_code/SDSCode/database/models.py
   - add repoint table
   - add logics to query repoint table for date range
   - update indexer to write to table

## Testing

- tests/test-data/test_spice_files/imap_2000_056_03.repoint.csv


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +526,142 @@ def generate_queue_url(event):
...
+    str
+        The file name for the latest repoint file.
+    """
+    # Look for latest and greatest file_path in the repointing table.
+    query = (
+        session.query(models.RepointTable)
+        .order_by(desc(models.RepointTable.file_path))
+        .limit(1)
+    )
+    latest_repoint_file = query.first()
```
greglucas: The limit(1) and first() are redundant and can be combined.
```suggestion
    latest_repoint_file = (
        session.query(models.RepointTable)
        .order_by(desc(models.RepointTable.file_path))
        .first()
    )
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +526,142 @@ def generate_queue_url(event):
...
+    if not latest_repoint_file:
+        raise ValueError("No repoint file found in the database.")
+
+    latest_repoint_file = latest_repoint_file.file_path
+    logger.info(f"Latest repoint file found: {latest_repoint_file}")
+    return latest_repoint_file
+
+
+def calculate_repoint_date_range(session, file_obj):
```
greglucas: This almost seems like we should just make another DB Table to store the items you need. Use this function to update the table with these start/end dates for all pointings in the file via upserts whenever we receive a repointing file. Then query that table directly for the information you need.

I'm still torn on which direction to go, but this seems like a decently complex version to do every time an ENA instrument gets kicked off in batch starter where you wouldn't want to download, open up files etc. unnecessarily.

tech3371: I agree. I am also torn. Let me look into storing the whole repoint data in db in indexing lambda. I started to look at other case that Tim mentioned in linked ticket and we do need to store the filename to be able to narrow down for pointing_attitude job.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -201,6 +201,17 @@ class SpinTable(Base):
     ingestion_date = Column(DateTime(timezone=True))
 
 
+class RepointTable(Base):
+    """Repoint table."""
+
+    __tablename__ = "repoint_table"
```
maxinelasp, line 219: Should we connect this table to the science file table, where we have repoint already?

Could we rename to "repointing" also, to match that existing column?

tech3371: It's been while I did that. Can you expand on what use case this will benefit? I want to keep it separate unless it's needed.

tech3371:
@greglucas I added changes to determine date range using repoint files. Based on this, let me know if you think it makes sense to store repoint data to its table still. Either way, we need to download and read repoint file in some place using pandas. Don't know which one is better. After we decide on what path forward, I will fix tests. It's effecting many areas.

I prefer current implementation to keep things separate.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: After looking through, I am going to suggest keeping table separate. If we combine it, tests setup are going to get even more complicated.

greglucas: Is this table actually duplicated in a sense and included in the SPICEFilesTable already? i.e. are they already indexed into that via Bryan's work. We do have the directory mapping:
https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/bf4756bce99553dab2e462a09fd62eda2df276d3/imap_data_access/file_validation.py#L462

greglucas:
> @greglucas I added changes to determine date range using repoint files

I don't think I follow the latest updates. The date range of a file is from the start of the mission to the end of the file, so the start_date should be the first repointing in the file and the end date the last. I don't understand what the two files thing is calculating.

My suggestion for another table was to make a table that has basically these same exact fields from the file itself just so you could query against a database rather than needing a file. The complication is that you have to do logic on the i and i+1 rows and not just each row individually. So the computation might be the benefit/nice thing to do in a table form.

```
pointing_id,pointing_start_utc,pointing_end_utc
```

Then every time a new file comes in you just update everything in the table with the latest and greatest data. Now all batch starter jobs just make a quick query against this simple table that you've already prepared upon ingest.

I'm leaning this direction now that I type it out...

tech3371:
> > @greglucas I added changes to determine date range using repoint files
> 
> I don't think I follow the latest updates. The date range of a file is from the start of the mission to the end of the file, so the start_date should be the first repointing in the file and the end date the last. I don't understand what the two files thing is calculating.
> 
> My suggestion for another table was to make a table that has basically these same exact fields from the file itself just so you could query against a database rather than needing a file. The complication is that you have to do logic on the i and i+1 rows and not just each row individually. So the computation might be the benefit/nice thing to do in a table form.
> 
> ```
> pointing_id,pointing_start_utc,pointing_end_utc
> ```
> 
> Then every time a new file comes in you just update everything in the table with the latest and greatest data. Now all batch starter jobs just make a quick query against this simple table that you've already prepared upon ingest.
> 
> I'm leaning this direction now that I type it out...

That change is related to first question Tim had, using repoint as trigger for pointing_attitude job. Eg. when a repoint file is ingested, it needs to trigger pointing_attitude kernel job in batch job since that's the only file that is not HARD_NO_TRIGGER. So when repoint file ingestion happens, we need to determine date range for querying upstream dependency. Since ingested file in repoint file, we need to use last two ingested file's date to determine date range. I am happy to jump on call if this is confusing.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: No. Repoint file is not indexed in SPICEFiles. SPICEFiles only stores kernels.

tech3371: In this line, https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/dev/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py#L455, we don't index repoint file based on past discussion. But in this PR, I am adding that capability based on your last suggestion on my last PR.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +544,187 @@ def generate_queue_url(event):
...
+    """
+    if isinstance(file_obj, SPICEFilePath):
+        # TODO: fix date range if/when repoint file ingestion event is
+        # passed to batch starter to kickoff HARD or SOFT_TRIGGER downstream jobs.
+        # Convert datetime object to string of format YYYYMMDD
+        file_type = file_obj.spice_metadata["type"]
+        if file_type == "repoint":
+            # Get last two repoint files from the database.
+            # Then calculate the date range based on those two files.
+            start_date, end_date = calculate_pointing_date_range(session)
```
tech3371: Based on discussion with Greg on zoom, we are changing what database to query for this. We are using the new table discussed above. Query spice_table and query for pointing_attitude type and get date range from that? If we do that, DPS file format is this:
>    dps_file_pattern = (
        r"(?P<type>imap_dps)_"
        r"(?P<start_year_doy>\d{4}_\d{3})_"
        r"(?P<end_year_doy>[\d]{4}_[\d]{3})_"
        r"(?P<version>\d+)\."
        r"(?P<extension>ah\.bc)"
    )

we query for the latest and greatest dps file, then parse and get end_date of that latest file. that end date becomes the start date to query repoint file. all pointing records that have pointing_start_utc >=end_date of the last DPS file.

```suggestion
            # Get last two repoint files from the database.
            # Then calculate the date range based on those two files.
            start_date, end_date = calculate_pointing_date_range(session)
```

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -201,6 +201,17 @@ class SpinTable(Base):
...
+    file_path = Column(String, nullable=False, primary_key=True, unique=True)
+    # end date from file name
+    end_date = Column(DateTime, nullable=False)
+    version = Column(String(2), nullable=False)
+    ingestion_date = Column(DateTime(timezone=True))
+
```
tech3371, line 224: Pulling general comments here:
>My suggestion for another table was to make a table that has basically these same exact fields from the file itself just so you could query against a database rather than needing a file. The complication is that you have to do logic on the i and i+1 rows and not just each row individually. So the computation might be the benefit/nice thing to do in a table form.

pointing_id,pointing_start_utc,pointing_end_utc
Then every time a new file comes in you just update everything in the table with the latest and greatest data. Now all batch starter jobs just make a quick query against this simple table that you've already prepared upon ingest.

Tenzin - we create table with `repoint_id,pointing_start_utc,pointing_end_utc` where pointing_start_utc is i_pointing -1, pointing_end_utc is i_pointing.

tech3371: `repoint_id_i, pointing_start_utc(repoint_end_utc_i-1), pointing_end_utc (repoint_start_utc_i)` where `repoint_end_utc_i-1` is the right side of first two red lines. Then `repoint_start_utc_i` is the start repoint time of next repoint.
IMG_9086

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +544,187 @@ def generate_queue_url(event):
...
+    ].iloc[0]
+    start_date = datetime.datetime.strptime(
+        start_date, "%Y-%m-%dT%H:%M:%S.%f"
+    ).strftime("%Y%m%d")
+    end_date = repoint_df.loc[
+        repoint_df["repoint_id"] == file_obj.repointing + 1, "repoint_end_utc"
+    ].iloc[0]
+    end_date = datetime.datetime.strptime(end_date, "%Y-%m-%dT%H:%M:%S.%f").strftime(
+        "%Y%m%d"
+    )
```
tech3371: start_date = pointing_start_utc (repoint_end_utc_i-1) and end_date = pointing_end_utc(repoint_start_utc_i) from the new table.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +544,187 @@ def generate_queue_url(event):
...
+    end_date = first_obj.spice_metadata["end_date"].strftime("%Y%m%d")
+    # Calculate end date using second file's repoint's end date from file_obj.
+    second_repoint_file = latest_repoint_files[1].file_path
+    second_obj = SPICEFilePath(os.path.basename(second_repoint_file))
+    start_date = second_obj.spice_metadata["end_date"].strftime("%Y%m%d")
+
+    logger.info(
+        f"repointing date range, start_date: {start_date}, end_date: {end_date}"
+    )
+    return start_date, end_date
```
tech3371: now query from pointing table and dps files and determine that way.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: we may not even need the pointing now because pointing_attitude job code doesn't use the input date but uses the date of attitude kernel. So we can set date range of this unique job to use the start and end date of attitude kernel and then get latest and great repoint and pass that. This is done to have one to one between attitude kernel and dps kernel. This makes sense to us.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: resolved this in the meeting.

tech3371 (commented):
Lots of comments and discussion from PR tag-up. Please ignore some of these comments. We found few different ways and solution after discussing and looking at the processing code itself. I will make more updates to this and request review after that.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +544,187 @@ def generate_queue_url(event):
...
+
+
+def calculate_ena_date_range(session, file_obj):
```
tmplummer: If I am understanding this function, it is calculating the Pointing date range for a science file that has a pointing number associated with it. The Pointing is defined to span from the end of one repoint maneuver to the end of the next repoint maneuver. Can the function be renamed and comments added that specify these things?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tmplummer: Just double checking... Do we always want the date range of the last repoint table csv? Or might there be a case where we want the date range of a specific repoint table? I think that `calculate_pointing_dare_range()` always provides the former.

sds_data_manager/lambda_code/SDSCode/database/models.py
tmplummer: I read this comment thread to mean that the initial plan was to ingest the data from each repoint table csv into a repoint database table. Later, it was decided that this repoint table was not needed. Am I understanding that correctly?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +544,187 @@ def generate_queue_url(event):
...
+
+    if not latest_repoint_files:
+        raise ValueError("No repoint file found in the database.")
+
+    # Calculate start date using first file's repoint's end date from file_obj.
+    first_repoint_file = latest_repoint_files[0].file_path
+    first_obj = SPICEFilePath(os.path.basename(first_repoint_file))
+    end_date = first_obj.spice_metadata["end_date"].strftime("%Y%m%d")
+    # Calculate end date using second file's repoint's end date from file_obj.
+    second_repoint_file = latest_repoint_files[1].file_path
```
tmplummer: What happens if there is only one repoint file in the DB? Can we default the `start_date` to something reasonable like the launch date?

tmplummer (commented):
I made a few comments. It might be worth sitting down to go over this PR in detail because I am having a little trouble figuring out all of the pieces and how they fit together.

tech3371:
> I made a few comments. It might be worth sitting down to go over this PR in detail because I am having a little trouble figuring out all of the pieces and how they fit together.

@subagonsouth A tat-up might be helpful since this touches few areas. I am making update still since talk with Greg. Greg and we didn't know few things that you might know more. Let me know when you are available to tag-up through Slack.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +545,111 @@ def generate_queue_url(event):
...
+    repoint_record = (
+        session.query(models.PointingTable).filter(
+            models.PointingTable.pointing_id == repoint_id
+        )
+    ).first()
+
+    if not repoint_record:
+        raise ValueError(f"No PointingTable record found for ID: {repoint_id}")
+
+    # TODO: double check which time to use for the start and end dates.
```
tech3371: What is there is good.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -201,6 +201,29 @@ class SpinTable(Base):
...
+    repoint_start_utc = Column(DateTime(timezone=True))
+    repoint_end_utc = Column(DateTime(timezone=True))
+
+
+class RepointTable(Base):
+    """Repoint table."""
+
+    __tablename__ = "repoint_table"
+    file_path = Column(String, nullable=False, primary_key=True, unique=True)
+    # TODO: determine start date in upcoming PR
```
tech3371: derived_start_date = <end date of previous file>. Right now, I only store 'yyyymmdd' precision. 'hhmmss' precision?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +545,111 @@ def generate_queue_url(event):
...
+    """
+    if isinstance(file_obj, SPICEFilePath):
+        file_type = file_obj.spice_metadata["type"]
+        if file_type == "repoint":
+            # TODO: fix calculations for repoint date range in upcoming PR
+            # when we work on pointing attitude job.
+            start_date = (
+                file_obj.spice_metadata["end_date"] - datetime.timedelta(days=1)
+            ).strftime("%Y%m%d")
+            end_date = file_obj.spice_metadata["end_date"].strftime("%Y%m%d")
```
tech3371: If a repoint file is ingested, then what the date range to be. End date should be end date of the repoint file. Start date is -1 of end_date?

tech3371 (commented):
From discussion with Tim

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: We decided not to for now.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: Tim confirm that this implementation is ok.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +545,111 @@ def generate_queue_url(event):
...
 
 
+def calculate_pointing_date_range(session, repoint_id):
+    """Calculate date range for the pointing data range using repoint files.
+
+    Parameters
+    ----------
+    session : sqlalchemy.orm.Session
+        Database session.
+    repoint_id : int
```
tmplummer: I think that this should be renamed to `pointing_id` to be consistent about terminology.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +545,111 @@ def generate_queue_url(event):
...
+        The ID of the repointing record.
+
+    Returns
+    -------
+    tuple
+        A tuple containing the start date and end date in the format YYYYMMDD.
+    """
+    # Query the pointing table to find the repointing information.
+    repoint_record = (
```
tmplummer: Same here... 
```suggestion
    pointing_record = (
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -525,17 +545,111 @@ def generate_queue_url(event):
...
+    Returns
+    -------
+    tuple
+        A tuple containing the start date and end date in the format YYYYMMDD.
+    """
+    if isinstance(file_obj, SPICEFilePath):
+        file_type = file_obj.spice_metadata["type"]
+        if file_type == "repoint":
+            # TODO: fix calculations for repoint date range in upcoming PR
+            # when we work on pointing attitude job.
```
tmplummer: Does this TODO go away now?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tmplummer: Can just add a comment that says the only job expected to trigger off of the repoint table is the Pointing Attitude kernel job. For that job, we will query for a spacecraft attitude history SPK that covers this 1 day date range which will get the correct SPK that pairs with the repoint table.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -335,13 +336,89 @@ def index_spin_file(s3_key: Path):
...
+                    repoint_start_utc = pd.to_datetime(
+                        next_row.iloc[0]["repoint_start_utc"]
+                    )
+                    repoint_end_utc = pd.to_datetime(
+                        next_row.iloc[0]["repoint_end_utc"]
+                    )
+                else:
+                    repoint_start_utc = None
+                    repoint_end_utc = None
+                    pointing_end_utc = None
```
tmplummer, line 437: I have a concern that these Null values will not get updated when the next repoint table is delivered. Is that covered in a test?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -335,13 +336,89 @@ def index_spin_file(s3_key: Path):
...
+                continue
+
+            params = {
+                "pointing_id": repoint_id,
+                "pointing_start_utc": pointing_start_utc,
+                "pointing_end_utc": pointing_end_utc,
+                "repoint_start_utc": repoint_start_utc,
+                "repoint_end_utc": repoint_end_utc,
+            }
+            # print(f"Processing repoint_id: {params}")
```
tmplummer:
```suggestion
```

tmplummer (commented):
Looking really good. I have a few minor suggestions. One concern about Null (None) values persisting in the PointingTable when a subsequent repoint table is delivered that should fill those values in.

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_query_api.py
```diff
@@ -105,7 +105,7 @@ def lambda_handler(event, context):
     search_results = [
         _convert_spice_metadata_model_to_dict(result) for result in search_results
     ]
-    logger.info(
+    logger.debug(
```
tech3371, line 119: These logs are only needed if SPICE dependency query API failed. Otherwise, it crowds the logs

tech3371:
IMG_9150
IMG_9147
IMG_9146
IMG_9145
Some whiteboard drawing that can help understand why we landed on this decision for future reference

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: yep

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
tech3371: Good catch. I didn't. Writing test for that.

tests/lambda_endpoints/test_spice_metakernel_api.py
```diff
@@ -274,6 +275,8 @@ def test_metakernel_filtered_file_types(session):
     assert result["body"] == "No files found."
 
 
+# skip this test
+@pytest.mark.skip(reason="Need to fix the spiceypy.datetime2et() call")
```
tech3371: address this in mentioned issue.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
tech3371, line 1: address 'best' ephemeris query

tech3371:
@greglucas and @subagonsouth, sorry for this large PR. It caused lots of things to rethink and they all kind of relate to one another. I addressed all of the comments and address few other cases after discussing with @bryan-harter, can you review it again? I can make small updates as needed based on feedback. Otherwise, it would be great if we can move this forward and I can address larger feedback in separate PR.

Thank you!

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_query_api.py
```diff
@@ -12,7 +12,7 @@
 
 # Logger setup
 logger = logging.getLogger(__name__)
-logger.setLevel(logging.INFO)
+logger.setLevel(logging.DEBUG)
```
greglucas: Do you want to move this back to INFO and only move it to debug when we are actually debugging an event? Otherwise the end result here is the same as previously.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -202,6 +202,27 @@ class SpinTable(Base):
...
+
+    __tablename__ = "pointing_table"
+    pointing_id = Column(Integer, nullable=False, primary_key=True, unique=True)
+    pointing_start_utc = Column(DateTime(timezone=True), nullable=False)
+    pointing_end_utc = Column(DateTime(timezone=True))
+    repoint_start_utc = Column(DateTime(timezone=True))
+    repoint_end_utc = Column(DateTime(timezone=True))
+
+
+class RepointTable(Base):
```
greglucas: Would this be easier to distinguish if we added in a "Files" qualifier because it is really just a file tracker.
```suggestion
class RepointFiles(Base):
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -623,17 +643,110 @@ def generate_queue_url(event):
...
+        file_type = file_obj.spice_metadata["type"]
+        if file_type == "repoint":
+            # NOTE:
+            # Repoint file is used to kicks off pointing_attitude job only.
+            # This date range is used to query attitude kernel file(s). If
+            # Other is dependent on the repoint file, please revisit this logic.
+            start_date = (
+                file_obj.spice_metadata["end_date"] - datetime.timedelta(days=1)
+            ).strftime("%Y%m%d")
+            end_date = file_obj.spice_metadata["end_date"].strftime("%Y%m%d")
```
greglucas, line 703: I'm not sure I'm following this portion still. What happens if we miss a daily repoint maneuver, don't we want 2 days worth of data and not just the previous one day? Does the start date not matter and this is an arbitrary choice? It might be nice to add a quick comment about the choice of 1-day.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -304,7 +302,7 @@ def _validate_node(self, node: tuple) -> bool:
                 f"Invalid data source: {node[0]}. "
                 f"Valid data sources: {self.data_source.valid_source}"
             )
-        if node[1] not in self.data_type.valid_type:
+        if node[1] != "best" and node[1] not in self.data_type.valid_type:
```
greglucas: Should best be a `valid_type`?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -578,44 +576,116 @@ def get_latest_repoint_file(end_date: datetime) -> Optional[str]:
...
-    if not repoint_files:
-        return None
+    with db.Session() as session:
+        latest_repoint_file = (
+            session.query(models.RepointTable)
+            .order_by(desc(models.RepointTable.file_path))
+            .first()
+        )
+        if not latest_repoint_file:
+            raise ValueError("No repoint file found in the database.")
```
greglucas: Duplicate condition. Keep the one outside the session?
```suggestion
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -578,44 +576,116 @@ def get_latest_repoint_file(end_date: datetime) -> Optional[str]:
...
             f"is before input end date {end_date}"
         )
         return None
 
-    # Otherwise, return the latest repoint file without the path prefix
-    return basename(latest[2])
+    return basename(latest_repoint_file.file_path)
+
+
+def check_requested_kernels(combined_kernel_sources, metakernel_files):
```
tmplummer, line 599: This function looks reasonable, but I am not sure I understand the purpose. I would suggest that in processing, we should not be using metakernels as dependencies at all.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -578,44 +576,116 @@ def get_latest_repoint_file(end_date: datetime) -> Optional[str]:
...
+        kernel_type = file_obj.spice_metadata["type"]
+        if "ephemeris_" in kernel_type:
+            ephemeris_found.add(kernel_type)
+        else:
+            other_kernels_found.add(kernel_type)
+
+    # Check if all other requested kernels are found
+    if expected_other_kernels == other_kernels_found:
+        pass
+    else:
```
greglucas: Invert the condition
```suggestion
    if expected_other_kernels != other_kernels_found:
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -578,44 +576,116 @@ def get_latest_repoint_file(end_date: datetime) -> Optional[str]:
...
+        logger.error(
+            f"Non-ephemeris kernels {expected_other_kernels} not found in "
+            f"metakernel files {other_kernels_found}"
+        )
+        return False
+
+    # If no ephemeris kernels are requested, we can return True.
+    if not expected_ephemeris:
+        return True
```
greglucas, line 658: This condition can be brought up to the top so you don't even need to iterate through anything else.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -337,13 +338,123 @@ def index_spin_file(s3_key: Path):
...
+        # Update existing entries with None values
+        for pointing_entry in (
+            session.query(models.PointingTable)
+            .filter(
+                (models.PointingTable.pointing_end_utc.is_(None))
+                | (models.PointingTable.repoint_start_utc.is_(None))
+                | (models.PointingTable.repoint_end_utc.is_(None))
+            )
+            .all()
+        ):
```
greglucas, line 385: If we get some values updated after the fact by a new version bump, this won't handle that because that entry would already be in the pointing table. I think our pointing table should be the latest/greatest, so I'm wondering if we can get rid of this branch checking for None entirely and just do an upsert on every single pointing entry every time we get a file? The database will always be a replica of the latest file then.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -337,13 +338,123 @@ def index_spin_file(s3_key: Path):
...
+                    repoint_start_utc = pd.to_datetime(
+                        next_row.iloc[0]["repoint_start_utc"]
+                    )
+                    repoint_end_utc = pd.to_datetime(
+                        next_row.iloc[0]["repoint_end_utc"]
+                    )
+                else:
+                    repoint_start_utc = None
+                    repoint_end_utc = None
+                    pointing_end_utc = None
```
tmplummer, line 437: This is repeated code from above that would be nice to have in a function. I am not very experienced with DBs, but it really seems like there should be some clever way that we could just set the `pointing_id`, `repoint_start_utc` and `repoint_end_utc` values from the repoint table and then use some DB magic with SQLAlchemy that auto-populates the pointing start/end values by using something like a lambda function. Functionality would be the same as here, so I won't block this PR but maybe writing a ticket to try to simplify this function would be good.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: This is remain from approach I took before meeting with Bryan. I will undo this.

tmplummer (commented):
From a high level, this looks good. I am happy with all of the pieces that we discussed in person but there seems like there is more to this PR that I don't understand. I am getting quite lost in all the various changes that have gotten put into this PR. The changes in the tests is overwhelming and so I didn't review those.

Is there a BUG ticket for the needed change to the `datetime_to_et` function?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: I kept this one here because there could be a case where we don't expect ephemeris kernel but we expect other kernel.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
tech3371: let me address this in a new PR.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: Let's revisit this in upcoming PR.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: I think I want to keep this same as other to not confuse with other functions from imap-data-access.

tests/lambda_endpoints/test_spice_metakernel_api.py
```diff
@@ -274,6 +275,7 @@ def test_metakernel_filtered_file_types(session):
     assert result["body"] == "No files found."
 
 
+@pytest.mark.skip(reason="Need to fix the spiceypy.datetime2et() call")
```
tmplummer, line 278: Is there a ticket for this needed change?

tech3371: No. But Bryan and I discussed approach for this. But as I mentioned this PR become already too big. That's why, I want to revisit that in another PR.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
tech3371: Sure thing. I can improve this function in next PR.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: We are not using metakernel as dependency but we are calling metakernel API with list files argument to true. This does same check as others but now, it returns expected files as list only. Because of that, I need to this extra step check to make sure that we got files for kernel type requested.

tech3371:
> From a high level, this looks good. I am happy with all of the pieces that we discussed in person but there seems like there is more to this PR that I don't understand. I am getting quite lost in all the various changes that have gotten put into this PR. The changes in the tests is overwhelming and so I didn't review those.
> 
> Is there a BUG ticket for the needed change to the `datetime_to_et` function?

Sorry @subagonsouth for big PR. Yeah, it started simple but become complicated. One bug fix while adding this new feature raised few cases we missed before. Refactoring test was needed to separate multiple tests that was depending on one another's db record. Because of that It was becoming hard to update one test without breaking the other. That's why I updated the test to be independent and only add record needed for that specific test.

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: But the other similar tables here are called `ScienceFiles`, `AncillaryFiles`, and `SPICEFiles`. `SpinTable` is the only one with Table in the name, so I think this would be consistent with what I was expecting on the ScienceFiles and how that is stored?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
greglucas: Ahh, I see. Thanks for clarifying.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: done

greglucas (approved):
We should open tickets for the updates required after this so that they don't get lost. I think there were a few items in the comments that you said you would address later, but they are already buried pretty deep so it'd be helpful to just clarify them as issues to make sure someone is assigned them and we hit the ground running on updates asap.

tech3371:
> We should open tickets for the updates required after this so that they don't get lost. I think there were a few items in the comments that you said you would address later, but they are already buried pretty deep so it'd be helpful to just clarify them as issues to make sure someone is assigned them and we hit the ground running on updates asap.

Sounds good. @greglucas I plan to work on PR updates next.

PR #730, opened 2025-07-17
Duplicate job handling

### LINKED ISSUE #729: Add check in batch starter for duplicate processing runs
Duplicate processing runs are when a job for the same instrument, data level, descriptor, and start_date has the same dependency string from the last run. 

For example if there is a job that is about to be submitted for mag l1c norm-magi 20250101, v002

check to see if mag l1c norm-magi 20250101, v001 was submitted with the same dependencies. If so, this is an exact duplicate run and we can skip the v002 job.

lacoak21:
# Change Summary

closes #729
## Overview
Skip duplicate processing runs. 

If this is a reprocessing job, check the previous dependencies and if they are the same, skip processing. 

## Updated Files
   - Add code to download the previous versions cadence file
   - Add code to check for duplicate processing 

## Testing
Add tests to check that duplicate processing runs are skipped.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -294,7 +294,12 @@ def get_special_case_date_range(session, job_node, start_date):
 
 
 def duplicate_job(
-    instrument, data_level, descriptor, start_date, version, serialized_dependencies
+    instrument,
+    data_level,
+    descriptor,
+    start_date,
+    previous_version,
+    serialized_dependencies,
 ) -> bool:
     """Determine if the current job is a duplicate of the most recent job.
 
@@ -308,7 +313,7 @@ def duplicate_job(
         Data descriptor.
     start_date : str
         Start date.
-    version : str
+    previous_version : str
         The previous version of the job
     serialized_dependencies : str
         The serialized upstream dependencies of the job.
@@ -324,7 +329,7 @@ def duplicate_job(
         data_level=data_level,
         descriptor=descriptor,
         start_time=start_date,
-        version=version,
+        version=previous_version,
         extension="json",
     )
     previous_dependency_path = previous_dependency_file.construct_path()
@@ -332,7 +337,7 @@ def duplicate_job(
     relative_path = previous_dependency_path.relative_to(
         imap_data_access.config["DATA_DIR"]
     )
-    previous_dependency_str = download_dependency_file(relative_path)
+    previous_dependency_str = get_previous_dependency_str(relative_path)
     if previous_dependency_str is None:
         logger.error(
             f"Failed to download previous dependency file: {previous_dependency_path}. "
@@ -860,8 +865,8 @@ def bulk_reprocessing_event(session, events):
             submit_all_jobs(session, job, start_date, end_date)
 
 
-def download_dependency_file(dependency_file_path: Path):
-    """Download a JSON file containing a job's dependencies from S3.
+def get_previous_dependency_str(dependency_file_path: Path):
+    """Download a JSON file containing a dependencies from S3 and return its contents.
 
     Parameters
     ----------
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -294,8 +294,8 @@ def get_special_case_date_range(session, job_node, start_date):
 
 
 def duplicate_job(
-    instrument, data_level, descriptor, start_date, version, dependencies
-):
+    instrument, data_level, descriptor, start_date, version, serialized_dependencies
+) -> bool:
     """Determine if the current job is a duplicate of the most recent job.
 
     Parameters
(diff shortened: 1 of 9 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -293,6 +293,60 @@ def find_most_recent_start_date(dep: dict, date: datetime) -> datetime:
...
+    ----------
+    instrument : str
+        Instrument.
+    data_level : str
+        Data level.
+    descriptor : str
+        Data descriptor.
+    start_date : str
+        Start date.
+    version : str
```
tech3371: Can you change this to `previous_version`? This will help reduce confusion with versions from other functions.
```suggestion
    version : str
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -293,6 +293,60 @@ def find_most_recent_start_date(dep: dict, date: datetime) -> datetime:
...
+        start_time=start_date,
+        version=version,
+        extension="json",
+    )
+    previous_dependency_path = previous_dependency_file.construct_path()
+    # Get the dependency s3 path
+    relative_path = previous_dependency_path.relative_to(
+        imap_data_access.config["DATA_DIR"]
+    )
+    previous_dependency_str = download_dependency_file(relative_path)
```
tech3371: Can you use `imap_data_access.download` instead here? This will help with authentication and DATA_DIR setup and so on. I feel like we were doing couple different ways of querying and downloading. Some are using direct lambda call and some are using `imap_data_access`. This may be bigger than this PR. Something to think about. Ping @greglucas since he added the auth.

tech3371: Also, can you change `download_dependency_file` to something like `get_previous_dependency_str()` because I thought you were only downloading but you were reading the content of files as well. If you change to `imap_data_access.download()`. Then you can `json.loads(download_path)`. This can remove the need of `download_dependency_file()`.

tech3371 (commented):
I had couple change request. Otherwise, it looks good to me.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Yes I can change the function name. 

Also I remember on a previous MR or an in person discussion someone mentioned I should just send a request directly to the api endpoint instead of going through imap-data-access.upload so thats why I did the same for download. I cant remember why. @greglucas do you remember this? I can definitely use imap_data_access.download if I misremembered that discussion.

greglucas: 🤔 , I remember something like that too for why we didn't want to do it that way but can't think of it off the top of my head. I think that was for "query" where I was saying why would we query when we could just look at the DB ourselves.

One thing we will have to keep in mind is that the public endpoints will only give back publicly released files. So there are a few tasks here in the future:
* Add an API Key to be injected / used by batch starter lambda.
* Change the URL for imap-data-access to use the api-key routes.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -321,6 +380,33 @@ def try_to_submit_job(
     descriptor = job_info["descriptor"]
     start_date_str = datetime.datetime.strftime(start_date, "%Y%m%d")
 
+    # Search for any duplicate jobs that have the same exact dependencies for this
+    # instrument, data level, descriptor, and start date by checking the CRID.
+    # Only check for duplicates if this is a reprocessing job.
+    # We know this is a reprocessing job if the version is not "v001".
```
greglucas, line 386: I don't know if this is strictly true. If we get a half days downlink of data one day we will produce v001. Then we get the rest of that data and we will produce a packet file with `v002`, but it wasn't reprocessed.

tech3371: I think this is a general check for any version above `v001` that it's not a duplicate job by checking that upstream dependent files were not same. If not same, then it continues on to submit job. So, I think this works for v002.

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: Yeah, this change request that I made is outside of this PR scope. Can we test this in public release work? If current setup doesn't work, we can switch to imap-data-access functions since that's tested.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: That is true - the packet file could be v002 but the "version" variable this is referencing here is the "job to run" so it would only be jobs after l0 and in this case the job downstream from that v002 packet file would be a reprocessing.

PR #734, opened 2025-07-22
I-ALIRT - Coverage plots

laspsandoval:
# Change Summary

## Overview
Adds code that will be used to generate json files that the webteam will use to create coverage plots that show coverage for each ground station partner and DSS station. This code incorporates outages as well.

## New Files
- ialirt_coverage.py
   - Code that will be used to generate json files that the webteam will use to create coverage plots that show coverage for each ground station partner and DSS station. 

## Testing
test_ialirt_coverage.py

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
@@ -6,7 +6,6 @@ from datetime import datetime, timedelta, timezone
 from pathlib import Path
 
 import boto3
-import botocore
 import imap_data_access
 import requests
 import spiceypy
@@ -56,7 +55,7 @@ def get_dsn(download_dir: Path):
     dsn_path = dsn_files[0]
 
     download_path = imap_data_access.download(dsn_path["file_path"])
-    logger.info(f"Adding to {download_path} to calibration files.")
+    logger.info(f"Downloading to {download_path}.")
 
     # Placeholder
     # TODO: parse the file and return a populated dict once we know the file structure.
@@ -105,8 +104,8 @@ def get_latest_spice_kernels() -> ProcessingInputCollection:
     return dependency_inputs
 
 
-def download_spice_file(dependencies) -> list[Path]:
-    """Download SPICE kernel files.
+def setup_spice_file(dependencies) -> list[Path]:
+    """Download and furnish SPICE kernel files.
 
     Parameters
     ----------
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
```diff
@@ -0,0 +1,293 @@
...
+    metakernel_files = response.json()
+
+    logger.info(f"Found metakernel files: {metakernel_files}. Adding to collection.")
+    dependency_inputs.add(SPICEInput(*metakernel_files))
+
+    return dependency_inputs
+
+
+def download_spice_file(dependencies) -> list[Path]:
+    """Download SPICE kernel files.
```
maxinelasp: Since you're furnishing in this function, maybe the name should be `set_up_spice_file` or something that indicates that? Or at minimum mention that you're furnishing it in the doc

sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
```diff
@@ -0,0 +1,293 @@
...
+    dependencies.download_all_files()
+
+    spice_files = dependencies.get_file_paths(data_type=SPICESource.SPICE.value)
+    spiceypy.furnsh([str(file.resolve()) for file in spice_files])
+
+    return spice_files
+
+
+def get_latest_outage_file(bucket: str, region: str) -> str | None:
+    """Get the most recent outage file key from S3.
```
maxinelasp, line 134: Can you use the ancillary file tools to get this?

sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
```diff
@@ -0,0 +1,293 @@
...
+def generate_and_upload_30_days(bucket: str, region: str, outages: dict, dsn: dict):
+    """Upload new coverage json files to S3.
+
+    Parameters
+    ----------
+    bucket : str
+        The name of the S3 bucket.
+    region : str
+        The region in which the s3 bucket resides.
+    outages : dict
```
maxinelasp, line 212: Can you document the expected dictionary structures for both outages and dsn?

sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
laspsandoval: Yes. Good catch. So I was deciding if I want to put the outage file in the SDC s3 bucket or ialirt s3 bucket. Thinking about it more probably SDC s3 bucket is better. That way I don't have to create an upload api.

maxinelasp: Since this is in a lambda it's probably fine to just directly upload and download it, but it would also be nice to use some of the tools we already have if it makes sense to put it in the main bucket... probably would work fine either way after looking through the rest of your review tbh

maxinelasp (approved):
Looks good, thanks!

PR #741, opened 2025-07-25
I-ALiRT - Quick update based on new endpoint

laspsandoval:
# Change Summary

## Overview
Update based on new endpoint.

## Updated Files
- ialirt_ingest.py
   - Update based on new endpoint.
- stackbuilder.py
   - Pass in account name
   - Pass in account name

## Updated Files
- test_ialirt_ingest.py

### DIFF
--- a/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
+++ b/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
@@ -57,12 +57,20 @@ class IalirtIngestLambda(Construct):
         # Create DynamoDB Table
         self.algorithm_data_table = self.create_algorithm_dynamodb_table()
 
+        account_name = self.node.get_context("account_name")
+        # once we have the account_name, get that section out of cdk.json
+        account_config = self.node.get_context(account_name)
+        domain_name = account_config.get("domain_name", "no-domain-set")
+        # https://api.imap-mission.com
+        # https://api.dev.imap-mission.com
+        data_access_url = f"https://api.{domain_name}"
+
         # Create Lambda Function
         self.ialirt_ingest_lambda = self.create_lambda_function(
             ialirt_bucket,
             self.algorithm_data_table,
             docker_path,
-            account,
+            data_access_url,
         )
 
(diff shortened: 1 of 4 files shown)


sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
```diff
@@ -173,6 +178,7 @@ def create_lambda_function(
                 "ALGORITHM_TABLE": algorithm_data_table.table_name,
                 "S3_BUCKET": ialirt_bucket.bucket_name,
                 "EFS_SPICE_MOUNT_PATH": "/mnt/data",
+                "AWS_ACCOUNT": account,
```
greglucas: Set the API URL here and pass that in, rather than having the function creating the URL. It looks like you aren't using imap-data-access, but you could pass in `IMAP_DATA_ACCESS_URL=https://api.dev.imap-mission.com/` to try and be consistent across the project.

https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/d44aea29439a6cf978aec296996e15840657effe/imap_data_access/__init__.py#L57

Then in the function use it directly where needed as: `os.getenv() + "metakernel"`

laspsandoval: Will the final API be:
https://api.prod.imap-mission.com or https://api.imap-mission.com?

I thought it was the later, but I looked at processing_construct.py and it looks like it will be the former.

In other words can I do this?
 "IMAP_DATA_ACCESS_URL": f"https://api.{self.node.get_context('account_name')}.imap-mission.com",

greglucas: It will be the latter. I just updated that code to be this, so you should be able to follow a similar pattern here I think.
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/780/files#diff-e16e2a783f960ace87d961636836f51e85fce3c111d2a39047ebea52a6c99717R92

greglucas (approved):

PR #742, opened 2025-07-25
SPICE Indexer Intervals Calculation Change (and Metakernel API speed-up)

### LINKED ISSUE #720: BUG -metakernel APIs doesn't work with large content
### Description of the issue

_No response_

`https://api.dev.imap-mission.com/metakernel?spice-path=/mnt/data/imap/spice/&start_time=0&end_time=1000000000` or `https://api.dev.imap-mission.com/metakernel?spice-path=/mnt/data/imap/spice/&start_time=0&end_time=1000000000&list_files=True`

list of kernels or metakernel files content

{"message":"Service Unavailable"}

### Code Snippet:

```python
Code
```

_No response_

bryan-harter:
# Change Summary

## Overview
One major factor slowing down the metakernel API is that many of the CK files we have received so far contain single points of data (*in addition* to intervals that allow interpolation). So within the database, this shows up as hundreds of tiny intervals of size 0, and therefore the metakernel API searches through hundreds of gaps between these points. 
 
In reality, we should discard these single points of data, and not treat them as a valid intervals to store in the database, because:
1) It sounds like these won't event be expected in production, so if we do, it is an error.
2) SPICE routines that need the CK data do not interpolate between these points. For example, "pxform", used to transform vectors between frames, will not interpolate between two single points even if they are incredibly close. So these data points are functionally useless to many of our routines (unless we use specialty SPICE functions to pull out the exact data points). 

## Updated Files
   - Set coverage tolerance to 0
   - Discard any coverage of length 0

## Testing
The testing files we have does coincidentally contain a TON of these datapoints, and it is filtering these out correctly.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
@@ -37,7 +37,7 @@ MAXIMUM_J2000_INTERVAL = [
 COVERAGE_ANGULAR_VELOCITY_ONLY = False  # Only include segments with angular velocity?
 COVERAGE_SPICE_ARRAY_LENGTH = 10000  # Use an array size of 10000 for coverage calc
 COVERAGE_LEVEL = "INTERVAL"  # the granularity at which the coverage is examined
-COVERAGE_TOLERANCE = 50000.0  # Tolerance value expressed in ticks of the spacecraft.
+COVERAGE_TOLERANCE = 0.0  # Tolerance value expressed in ticks of the spacecraft.
 COVERAGE_TIME_SYSTEM = "TDB"  # Whether to use J2000 (TDB) or spacecraft clock (SCLK)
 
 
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -37,7 +37,7 @@
 COVERAGE_ANGULAR_VELOCITY_ONLY = False  # Only include segments with angular velocity?
 COVERAGE_SPICE_ARRAY_LENGTH = 10000  # Use an array size of 10000 for coverage calc
 COVERAGE_LEVEL = "INTERVAL"  # the granularity at which the coverage is examined
-COVERAGE_TOLERANCE = 50000.0  # Tolerance value expressed in ticks of the spacecraft.
+COVERAGE_TOLERANCE = 0.0  # Tolerance value expressed in ticks of the spacecraft.
```
tmplummer, line 40: I can't remember exactly what this `COVERAGE_TOLERANCE` was used for, but I don't see it in the updated code, so I just wanted to make sure that this change is intended.

tmplummer (approved):
Looks good to me. Just double check the one comment about COVERAGE_TOLERANCE being updated.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
bryan-harter: So I forgot to mention this is my summary, but I think I was using this tolerance incorrectly. I thought I had to manually check for gaps larger than 1 second in duration, which is what 50,000 ticks of the spacecraft is. Instead, SPICE "Type 2/3/4/5" data will tell you the exact range for valid intervals that it can interpolate over, and we don't really need to check for missing data to that granularity. 

What "Tolerance" was doing essentially was adding 1 second on each side of the coverage window, which is actually incorrect for our purposes, since there is no data there. This actually masked the fact that many of the "intervals" detected by the indexer were 0 seconds in duration. Instead they were showing up as 2 seconds in our database, because of the tolerance. So this change allows the intervals calculated with ckcov to show up as having the same start/end time, thereby allowing the later code to discard it.

PR #746, opened 2025-07-28
 API Endpoints

### LINKED ISSUE #745: API Endpoint Error Messages Update
Update the API endpoints to ensure all error messages are returning accurate information.

daralynnrhode:
# Normalizing the API endpoint messages 

## Overview

Updating the error/non-error messages for the various iterations of api calls that can be made. Currently, with most calls whether a backslash is included and/or ignored will impact the json error message returned when calling the api directly through the url. 

This should cause all the url possibilities to return the correct error message, but I have not updated the default message for the query call yet. 

closes #745
## New Files

- root_api.py

## Deleted Files

- 

## Updated Files

- sds_api_manager_construct.py
- api_gateway_construct.py


sds_data_manager/constructs/api_gateway_construct.py
```diff
@@ -224,6 +224,10 @@ def add_route(
         lambda_function : lambda_.Function
             Lambda function to trigger when this route is hit.
         """
+        # normalize root route
```
bourque, line 227: How many roots could a router route if a router could route roots?

sds_data_manager/constructs/sds_api_manager_construct.py
```diff
@@ -69,6 +99,20 @@ def __init__(
...
+            self,
+            id="RootAPILambda",
+            function_name="root-api-handler",
+            code=code,
+            handler="SDSCode.api_lambdas.root_api.lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_12,
+            timeout=cdk.Duration.seconds(10),
+            memory_size=128,
+            allow_public_subnet=True,
+            environment={"MESSAGE": "Welcome to the IMAP API 👋"},
```
maxinelasp: Take the emoji out - it might not render properly on all systems

sds_data_manager/lambda_code/SDSCode/api_lambdas/root_api.py
```diff
@@ -0,0 +1,31 @@
...
+
+    Returns
+    -------
+    dict
+        A JSON response with the message.
+    """
+    return {
+        "statusCode": 200,
+        "headers": {"Content-Type": "application/json"},
+        "body": '{"message": "Welcome to the IMAP API. '
```
maxinelasp, line 29: Before, you passed the message in as an environment variable, but here, you just write the message directly. I think either is fine (slight preference for doing it inside this lambda) but remove the env variable route if you're doing it this way

sds_data_manager/constructs/sds_api_manager_construct.py
```diff
@@ -9,6 +9,36 @@
 from .api_gateway_construct import ApiGateway
 
 
+def add_stable_route(api, base_path, http_method, lambda_function, prefix_list):
```
maxinelasp, line 12: This function is confusing to me (lots of string manipulation), so can you add more documentation? Specifically what the outcome of using this function is and why we use it.

maxinelasp (commented):
Just had some minor comments, but I like this approach

maxinelasp (approved):
Looks good, thanks for making those updates!

bourque (approved):

PR #747, opened 2025-07-28
I-ALiRT - Construct for Coverage Plots

### LINKED ISSUE #733: I-ALiRT - add coverage plots to the pipeline
### Summary

We have the lambda (ialirt_coverage.py), but we need to add a daily cron job, etc so that it works in the pipeline.

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Added code to create construct for coverage plots and also to update some existing lambda code.

## New Files
- Dockerfile.coverage
   - Add Dockerfile that is deployed upon build
- ialirt_coverage_construct.py
   - Add construct for coverage lambda and daily cron job

## Updated Files
- ialirt_coverage.py
   - Add structure of what the dsn file will look like when delivered and updated file naming based on 
recent conversations.
- stackbuilder.py - Add coverage construct to build
   - Add coverage construct to build

## Testing
test_ialirt_coverage_construct.py
test_ialirt_coverage.py


sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
```diff
@@ -49,17 +57,41 @@ def get_dsn(download_dir: Path):
...
+            los_dt = datetime.strptime(f"{year} {doy} {los}", "%Y %j %H:%M:%S")
+
+            # If LOS time is earlier than AOS, it must be the next day
+            if los_dt < aos_dt:
+                los_dt += timedelta(days=1)
+
+            start = aos_dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+            end = los_dt.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+            dsn_dict.setdefault(station, []).append((start, end))
```
maxinelasp, line 94: Does it make sense to have a list of expected stations somewhere, so we can raise an error if a station name changes? Or is it ok to have stations add/remove themselves from the file whenever?

maxinelasp (approved):
Looks good, one comment/suggestion

sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
laspsandoval: For now I think I know the DSN stations, but there is always a possibility that I have not accounted for one. I wanted to keep this here so it wouldn't fail or be silently skipped and I would be able to see that a new station appeared on the plots and add it to the list. I'll see if I can reach out to the right person to get all of the DSN stations that we expect. I've asked the POC and they didn't seem to have that information, but I'll look into it more so I can have a constant list.

PR #748, opened 2025-07-28
Add SPICE dependencies to Lo and Hi L3 ENA maps

pleasant-menlo:
# Change Summary

## Overview
Add spice files to deps for `h90-ena-h-sf-sp-full-hae-4deg-6mo` and `l090-ena-h-sf-sp-ram-hae-4deg-6mo`

## New Dependencies
N/A

## New Files
N/A

## Deleted Files
N/A

## Updated Files
<!--List each updated file and add bullets to describe the purpose/function of the updates.
dependencies_config.csv


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
```diff
@@ -373,6 +373,10 @@ hi, l1c, 45sensor-pset, hi, l2, h45-ena-h-hf-nsp-anti-4deg-1yr, HARD_NO_TRIGGER,
 
 hi, l2, h90-ena-h-sf-nsp-ram-hae-4deg-6mo, hi, l3, h90-ena-h-sf-sp-full-hae-4deg-6mo, HARD, DOWNSTREAM
 hi, l2, h90-ena-h-sf-nsp-anti-hae-4deg-6mo, hi, l3, h90-ena-h-sf-sp-full-hae-4deg-6mo, HARD, DOWNSTREAM
+science_frames, spice, historical, hi, l3, h90-ena-h-sf-sp-full-hae-4deg-6mo, HARD_NO_TRIGGER, DOWNSTREAM
```
tmplummer, line 376: I assume that the `science_frames` and `pointing_attitude` kernels are required to get the GLOWS data required for the survival probability correction into the right frame? Otherwise, I don't understand the need for these kernels, but that is ultimately for you all to know. These extra dependencies certainly won't hurt in this case.

tmplummer (approved):
The specification of the dependencies looks correct. If these are needed dependencies, then I approve.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
pleasant-menlo: These are used to convert the exposure-weighted GLOWS survival probability "pointing set" data from IMAP_DPS to ECLIPJ2000. We were not aware of the `hae_longitude` and `hae_latitude` variables in the L1c data when we built this; it looks like those could be used for the same purpose.

PR #758, opened 2025-07-31
Fix CRID tests

lacoak21:
# Change Summary

## Overview
Fix CRID tests by adding the correct dependencies.

## Updated Files
   - Specify byte order in to_bytes function 
   - Add the correct dependencies to the CRID tests. 

## Testing
Fix tests

### DIFF
--- a/tests/lambda_endpoints/test_dependency_api.py
+++ b/tests/lambda_endpoints/test_dependency_api.py
@@ -934,19 +934,43 @@ def test_get_cadence_jobs():
 #####################################
 
 
-@pytest.mark.skip(reason="CRID calculation is not behaving as expected")
 def test_calculate_crid(session):
     """Test CRID calculation."""
     _static_spice_files(session)
     records = [
         # File to build CRID for
         ScienceFiles(
-            file_path="/path/to/imap_swe_l1a_sci_20240102_v001.cdf",
+            file_path="/path/to/imap_swe_l1b_sci_20240101_v001.cdf",
+            instrument="swe",
+            data_level="l1b",
+            descriptor="sci",
+            start_date=datetime(2024, 1, 1),
+            version="v001",
+            extension="cdf",
+            ingestion_date=datetime.strptime(
+                "2024-01-25 23:35:26+00:00", "%Y-%m-%d %H:%M:%S%z"
+            ),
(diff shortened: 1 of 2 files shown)


tests/lambda_endpoints/test_dependency_api.py
```diff
@@ -994,25 +1018,22 @@ def test_calculate_crid(session):
...
-    print(f"Record: {record.file_path}")
     crid = calculate_crid(session, record)
     # The CRID associated with a file is made up of the filepath and the
     # Upstream file versions numbers packed into 2 bytes and sorted by the filename
-    # imap_swe_l1b_sci_20240102_v001.cdf has a total of 3 upstream dependency files:
-    # - imap_swe_l1a_sci_20240102_v001.cdf.cdf
-    # - imap_swe_l1b-in-flight-cal_20230102_v001.cdf
+    # imap_swe_l1b_sci_20240101_v001.cdf has a total of 4 upstream dependency files:
+    # - imap_swe_l1a_sci_20240101_v001.cdf.cdf
+    # - imap_swe_l0_raw_20240101_v001.pkts
```
tech3371, line 1030: I have been curious why we need l0 to kickoff l1b.

tests/lambda_endpoints/test_dependency_api.py
```diff
@@ -1117,6 +1184,8 @@ def test_new_crid(session):
             ),
         ),
     ]
+    session.add_all(records)
```
tech3371, line 1187: I always forgets to commit lol. Thank you for catching that!

tests/lambda_endpoints/test_dependency_api.py
lacoak21: The CRID is made up of all of the upstream versions recursively. So in this case, the l0 file will be included in the list of versions that construct the CRID. The point of this is that we can basically track any expected reprocessing of an upstream file. 

For example if the l0 file version changed we can expect that the l1a is currently being processed and therefore we should wait until that l1a is done and triggers the processing.

tech3371: I see. That didn't know that.

tech3371 (approved):

Issue #753, opened 2025-07-30
Repoint date range PR follow-up


tech3371:
Previous PR became too big. Will address feedback from last PR in this issue.

Feedback change requested are:

- [x] https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/728#discussion_r2243090963
   * repoint file date range being end_date -1 to end_date.

- [ ] Revisit and refactor index_pointing_data based on these two feedback https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/728#discussion_r2243130016 and https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/728#discussion_r2243142310

tech3371:
One request to be addressed from attached comment:
1. If we get some values updated after the fact by a new version bump, this won't handle that because that entry would already be in the pointing table. I think our pointing table should be the latest/greatest, so I'm wondering if we can get rid of this branch checking for None entirely and just do an upsert on every single pointing entry every time we get a file? The database will always be a replica of the latest file then.

bryan-harter:
Is this ticket in regards to the Repoint file indexer only updating the most recent value of the repoint, as opposed to the entire table? Do we know if these old repoint start/end times get updated frequently?

tech3371:
I believe the first one, updating the most recent value of the repoint. I didn't watch it closely. @tmplummer do you know?

tmplummer:
I can't really understand what this ticket is about.

I suppose it is possible that a new repoint table could change past values and thus the DB table would need to be updated. I don't think that will happen frequently, but is possible if some issue was discovered in how the MOC is generating the repoint table.

PR #759, opened 2025-07-31
MNT: updated index_pointing_data to address last PR feedback

tech3371:
# Change Summary
Addresses second point on #753 

## Updated Files

   - refactored index_pointing_data function

## Testing

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
@@ -371,56 +371,50 @@ def index_pointing_data(s3_key: str):
     repoint_file_path = download(s3_key)
     # Read CSV file using pandas
     repoint_df = pd.read_csv(repoint_file_path)
-    max_repoint_id = repoint_df["repoint_id"].max()
     repoint_records = []
 
-    for repoint_id in repoint_df["repoint_id"].values:
+    for i_row, repoint_id in enumerate(
+        repoint_df["repoint_id"].values[:-1].astype(int)
+    ):
         # Had to convert to match the type in the database
         repoint_id = int(repoint_id)  # noqa: PLW2901
-        # First find if next row exists
-        if repoint_id < max_repoint_id:
-            # If next row exists, calculate the pointing data
-            row_data = {
-                "pointing_id": repoint_id,
-                "pointing_start_utc": datetime.strptime(
-                    repoint_df.loc[
-                        repoint_df["repoint_id"] == repoint_id, "repoint_end_utc"
-                    ].values[0],
-                    "%Y-%m-%dT%H:%M:%S.%f",
-                ),
-                "pointing_end_utc": datetime.strptime(
-                    repoint_df.loc[
-                        repoint_df["repoint_id"] == repoint_id + 1, "repoint_end_utc"
-                    ].values[0],
-                    "%Y-%m-%dT%H:%M:%S.%f",
-                ),
-                "repoint_start_utc": datetime.strptime(
-                    repoint_df.loc[
-                        repoint_df["repoint_id"] == repoint_id + 1, "repoint_start_utc"
-                    ].values[0],
-                    "%Y-%m-%dT%H:%M:%S.%f",
-                ),
-                "repoint_end_utc": datetime.strptime(
-                    repoint_df.loc[
-                        repoint_df["repoint_id"] == repoint_id + 1, "repoint_end_utc"
-                    ].values[0],
-                    "%Y-%m-%dT%H:%M:%S.%f",
-                ),
-            }
-        else:
-            row_data = {
-                "pointing_id": repoint_id,
-                "pointing_start_utc": datetime.strptime(
-                    repoint_df.loc[
-                        repoint_df["repoint_id"] == repoint_id, "repoint_end_utc"
-                    ].values[0],
-                    "%Y-%m-%dT%H:%M:%S.%f",
-                ),
-                "pointing_end_utc": None,
-                "repoint_start_utc": None,
-                "repoint_end_utc": None,
-            }
+        # Since for loop stops at -1, we can assume that next row exists
+        # and should be able to calculate the pointing data
+        current_row = repoint_df.iloc[i_row]
+        next_row = repoint_df.iloc[i_row + 1]  # Get the next
+        row_data = {
+            "pointing_id": repoint_id,
+            "pointing_start_utc": datetime.strptime(
+                current_row["repoint_end_utc"],
+                "%Y-%m-%dT%H:%M:%S.%f",
+            ),
+            "pointing_end_utc": datetime.strptime(
+                next_row["repoint_end_utc"],
+                "%Y-%m-%dT%H:%M:%S.%f",
+            ),
+            "repoint_start_utc": datetime.strptime(
+                next_row["repoint_start_utc"],
+                "%Y-%m-%dT%H:%M:%S.%f",
+            ),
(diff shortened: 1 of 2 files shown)


greglucas (commented):
I have some suggestions to try and simplify the code.

* Get rid of max_repoint_id. You should be able to iterate over `range(repoint_id.values[-1])`, then add one final entry outside of the loop.
* Do the date conversions on the pandas columns directly before the loop so your columns are already in the format you need. `df["col"] = pd.to_datetime(col, format)`
* Don't do the `df.loc[]` lookups, use `df.iloc[irepoint]` instead so you avoid the large comparison every time through the loop and you can use the index variables directly then.

tech3371:
> I have some suggestions to try and simplify the code.
> 
> * Get rid of max_repoint_id. You should be able to iterate over `range(repoint_id.values[-1])`, then add one final entry outside of the loop.
> * Do the date conversions on the pandas columns directly before the loop so your columns are already in the format you need. `df["col"] = pd.to_datetime(col, format)`
> * Don't do the `df.loc[]` lookups, use `df.iloc[irepoint]` instead so you avoid the large comparison every time through the loop and you can use the index variables directly then.

@greglucas I remember now why I had to use datetime object. Because of this error
```
E           sqlalchemy.exc.StatementError: (builtins.TypeError) SQLite DateTime type only accepts Python datetime and date objects as input.
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -371,87 +371,72 @@ def index_pointing_data(s3_key: str):
...
-                logger.error(
-                    f"Error calculating pointing UTCs for repoint_id {repoint_id}: {e}"
-                )
-                continue
-
-            params = {
+    max_repoint_id = repoint_df["repoint_id"].max()
+    repoint_records = []
+
+    for repoint_id in repoint_df["repoint_id"].values:
```
greglucas: Iterate over all but the final row. Does astype(int) work here so you don't have to cast below?
```suggestion
    for i_row, repoint_id in enumerate(repoint_df["repoint_id"].values[:-1].astype(int)):
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -371,87 +371,72 @@ def index_pointing_data(s3_key: str):
...
+        # Had to convert to match the type in the database
+        repoint_id = int(repoint_id)  # noqa: PLW2901
+        # First find if next row exists
+        if repoint_id < max_repoint_id:
```
greglucas: Remove this if, because now you know you aren't going to the last value, but are stopping one short.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -371,87 +371,72 @@ def index_pointing_data(s3_key: str):
...
-                "pointing_start_utc": pointing_start_utc,
-                "pointing_end_utc": pointing_end_utc,
-                "repoint_start_utc": repoint_start_utc,
-                "repoint_end_utc": repoint_end_utc,
+                "pointing_start_utc": datetime.strptime(
+                    repoint_df.loc[
+                        repoint_df["repoint_id"] == repoint_id, "repoint_end_utc"
+                    ].values[0],
+                    "%Y-%m-%dT%H:%M:%S.%f",
+                ),
```
```suggestion
                "pointing_start_utc": datetime.strptime(
                    repoint_df.iloc[irow]["repoint_end_utc"].values[0],
                    "%Y-%m-%dT%H:%M:%S.%f",
                ),
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -371,87 +371,72 @@ def index_pointing_data(s3_key: str):
...
+                "pointing_end_utc": datetime.strptime(
+                    repoint_df.loc[
+                        repoint_df["repoint_id"] == repoint_id + 1, "repoint_end_utc"
+                    ].values[0],
+                    "%Y-%m-%dT%H:%M:%S.%f",
+                ),
```
```suggestion
                "pointing_end_utc": datetime.strptime(
                    repoint_df.iloc[i_row + 1]["repoint_end_utc"].values[0],
                    "%Y-%m-%dT%H:%M:%S.%f",
                ),
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -371,87 +371,72 @@ def index_pointing_data(s3_key: str):
...
+                "repoint_start_utc": datetime.strptime(
+                    repoint_df.loc[
+                        repoint_df["repoint_id"] == repoint_id + 1, "repoint_start_utc"
+                    ].values[0],
+                    "%Y-%m-%dT%H:%M:%S.%f",
+                ),
```
greglucas: For the pointing, do you want the repointing before the pointing started, or the one to finish it off? I would have assumed no "+1" here, but I have no idea which is preferred.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -371,87 +371,72 @@ def index_pointing_data(s3_key: str):
...
             }
+        else:
+            row_data = {
+                "pointing_id": repoint_id,
+                "pointing_start_utc": datetime.strptime(
+                    repoint_df.loc[
+                        repoint_df["repoint_id"] == repoint_id, "repoint_end_utc"
+                    ].values[0],
+                    "%Y-%m-%dT%H:%M:%S.%f",
+                ),
```
greglucas: Move this outside of the loop and do it at the end.

```suggestion
            row_data = {
                "pointing_id": repoint_id,
                "pointing_start_utc": datetime.strptime(
                    repoint_df.iloc[-1]["repoint_end_utc"].values[0],
                    "%Y-%m-%dT%H:%M:%S.%f",
                ),
```

greglucas (commented):
Just some minor nits to try and clean up the loop.

👍 to the comment about datetimes, thanks for checking on that!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -371,87 +371,66 @@ def index_pointing_data(s3_key: str):
     repoint_file_path = download(s3_key)
     # Read CSV file using pandas
     repoint_df = pd.read_csv(repoint_file_path)
+    repoint_records = []
+
+    for i_row, repoint_id in enumerate(
+        repoint_df["repoint_id"].values[:-1].astype(int)
+    ):
+        # Had to convert to match the type in the database
+        repoint_id = int(repoint_id)  # noqa: PLW2901
```
greglucas, line 378:
```suggestion
```

greglucas (approved):
Nice, quick incorporation!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
tech3371: same thing on this. SQL picky about type.
>E       sqlalchemy.exc.IntegrityError: (sqlite3.IntegrityError) datatype mismatch

greglucas: Yeah, that is annoying :) You can remove the `.astype()` above then.

PR #764, opened 2025-08-05
BUG cadence maps contain overlapping data

lacoak21:
# Change Summary

## Overview
I noticed this issue before SIT-4 when we were testing the cadence maps. When the jobs get triggered by and eventBridge the way batch_starter is currently calculating the date range will cause the same file to be added on a dependency for two concurrent maps. This is because our get_files() function is end and start date inclusive. This is a fix to bump the start date of the map by one to ensure that the maps have unique data. We also discussed changing the get_files() function to be end_date exclusive so maybe that is preferred here? I am open to suggestions. 

** NOTE @sdhoyt this will not impact the SIT-4 rerun filenames because we are manually supplying the start and end date just fyi. 

## Updated Files
   - Fix date range

## Testing
Fix cadence daterange test


tmplummer (approved):
Seems reasonable to me.

tech3371 (approved):
code changes looks good to me

PR #767, opened 2025-08-07
Add hash to dependency file

lacoak21:
# Change Summary

## Overview
Add a hash of the upstream dependencies to the dependency file descriptor. 

There is currently a bug in sds-data-manager that @maxinelasp discovered. The dependency file is uploaded to s3 in batch_starter.py for a new processing run with the same version as the file that is to be produced. The issue is that if that processing run fails, that dependency file is kept ( this is desired behavior because we need to be able to download this for debugging purposes). Since that file is never deleted, when that processing job rerun gets kicked off, it will try to write a file with the same name to s3 and cause the batch starter to error out. Greg suggested we add a hash to the dependency filename. The hash contains the contents of the file so this way we can:

- Reuse any dependency JSON files if the filename is exactly the same
         - A matching hash ensures that none of the upstream dependencies are changed
- Produce different files if the contents have changed while still matching them to the file version.
          - the version number will still be the same as the CDF file.
- Save any dependency files for debugging if processing has failed.
- BONUS! - In batch_starter.py we no longer need to download and extract the contents of a JSON file to determine if the job is a duplicate. We can just check to see if the hashes match.

## Updated Files
   - Add a hash to the descriptor of the dependency file 
   - Move uploading the cadence file to before adding an entry in the processing job table just in case the cadence file upload errors out
   - I was testing with IDEX l2b and realized that this trigger should be a HARD_NO_TRIGGER

## Testing
Fix expected calls.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -348,6 +348,10 @@ def duplicate_job(
     # If the file already exists, then we know that an exact duplicate job has been
     # run with the same dependencies, start date, descriptor, instrument, data level,
     # and version.
+    # TODO currently this will skip bulk reprocessing jobs that have been kicked off
+    #   Due to algorithm updates. The job might look like a duplicate but have new code.
+    #   We need to somehow track if we are reprocessing due to a manual trigger
+    #   or not.
     return _file_exists(s3_key_path_str)
 
 
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -1,6 +1,7 @@
 """Functions for supporting the batch starter component of the architecture."""
 
 import datetime
+import hashlib
 import json
 import logging
 import os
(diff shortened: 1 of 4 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -323,32 +325,34 @@ def duplicate_job(
     bool
         True if the job is a duplicate, False otherwise.
     """
+    # Generate the previous dependency file path based on the inputs.
+    # The descriptor should include a hash of the serialized dependencies.
+    dep_descriptor = (
+        f"{descriptor}-"
+        f"{hashlib.sha256(serialized_dependencies.encode('utf-8')).hexdigest()}"
```
greglucas: Rather than a really long digest, could we make it only the first 6 or 8 characters instead? I think that will give us plenty of options to avoid collisions still while making all the filename fields easier to read/reason about.

```suggestion
        f"{hashlib.sha256(serialized_dependencies.encode('utf-8')).hexdigest()[:8]}"
```

greglucas (approved):
Minor nit to reduce the length of the hash. But I like this approach 👍 Nice BONUS too :)

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Great - good idea. That will definitely be enough. We only need to cover collisions for a day.

PR #769, opened 2025-08-07
Update hit l1b dependencies

vmartinez-cu:
This PR updates dependencies for HIT L1B processing to match a recent change to HIT processing that split up the L1A counts product into two products. 

Now the l1a counts-standard product produces two l1b products and the l1a counts-sectored product produces one l1b product for a total of 3 l1b products.


vmartinez-cu:
I need to update a unit test to fix this failure

tech3371 (approved):
LGTM

PR #772, opened 2025-08-08
Add s3 tags to reprocessed files

lacoak21:
# Change Summary

## Overview

The issue:
We check for duplicate jobs by seeing if a dependency file already exists in S3. Since the dependency file contains a hash of the serialized dependencies, it’s unique to that job. 

Sometimes, though, we want to reprocess a job that has already been run with the same dependencies — for example, if the software has been updated. We need a way for batch_starter to tell whether a job is a manually triggered reprocessing job or not.

For example: 
imap-data-access reprocess --instrument idex

Solution: 
Add an S3 tag to the file if it has been reprocessed due to a manual reprocessing event. The changes here are to carry this information from the upload api lambda handler to the indexer -> to batch starter. 

The changes necessary for imap-data-access are in this [PR](https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/255)

## Updated Files
-sds_data_manager/constructs/sds_api_manager_construct.py
   - Add S3 tagging policy to lambda handler
- sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
   - Add tags to '_generate_signed_upload_response'
   - DO NOT skip duplicate jobs if in a reprocessing run.
   - query for s3 manually_reprocessed tag

## Testing
Fix expected try_to_submit job calls and add tests surrounding s3 tag

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
@@ -77,7 +77,7 @@ def _generate_signed_upload_response(s3_key_path, tags=None):
         Params={
             "Bucket": BUCKET_NAME,
             "Key": s3_key_path,
-            "Metadata": tags or dict(),
+            "Tagging": "&".join(f"{k}={v}" for k, v in (tags or {}).items()),
         },
         ExpiresIn=3600,
     )
@@ -96,6 +96,9 @@ def lambda_handler(event, context):
     event : dict
         Specifically looking at the event['pathParameters']['proxy'], which
         specifies the filename to upload.
+        If event['pathParameters']['manually_reprocessed'] == "true", the file will be
+        uploaded with the "manually_reprocessed=true" S3 tag. This tag indicates that
+        the file was reprocessed due to a manual reprocessing API request.
     context : None
         Currently not used
 
@@ -106,6 +109,12 @@ def lambda_handler(event, context):
 
     """
     path_params = event.get("pathParameters", {}).get("proxy", None)
+    manually_reprocessed = (
+        event.get("queryStringParameters", {})
+        .get("manually_reprocessed", "false")
+        .lower()
+        == "true"
+    )
     logger.info("Parsing path parameters=[%s] from event=[%s]", path_params, event)
 
     if not path_params:
@@ -152,5 +161,7 @@ def lambda_handler(event, context):
     s3_key_path_str = str(
         s3_key_path.relative_to(imap_data_access.config["DATA_DIR"]).as_posix()
     )
-
-    return _generate_signed_upload_response(s3_key_path_str)
+    s3tags = {"manually_reprocessed": "true"} if manually_reprocessed else None
+    if s3tags:
+        logger.info(f"Uploading file with S3 tags: {s3tags}.")
+    return _generate_signed_upload_response(s3_key_path_str, tags=s3tags)
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -433,11 +433,11 @@ def try_to_submit_job(
                     f"Skipping submission."
                 )
                 return
-        else:
-            logger.info(
-                "This is a manually triggered reprocessing job. Not checking"
-                "for duplicate jobs."
-            )
+    else:
+        logger.info(
+            "This is a manually triggered reprocessing job. Skipping check for "
+            "duplicate jobs."
+        )
 
     # Serialize the upstream dependencies and write them to a JSON file. The Imap
     # processing code will read the JSON file and deserialize the dependencies. This is
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -380,6 +380,7 @@ def try_to_submit_job(
     start_date: datetime,
     version: str,
     serialized_dependencies: str,
+    skip_duplicate_jobs: bool = True,
 ):
     """Try to submit a batch job with the given job information.
 
@@ -396,6 +397,11 @@ def try_to_submit_job(
     serialized_dependencies : str
         The serialized ProcessingInputCollection of the upstream
         dependencies.
+    skip_duplicate_jobs : bool
+        In normal reprocessing (due to a new file upload) duplicate jobs should be
+        skipped to avoid redundant products. In cases where a user has manually
+        triggered reprocessing, we want to reprocess even if the job looks like a
+        duplicate. Default is True.
     """
     instrument = job_info["data_source"]
     data_level = job_info["data_type"]
@@ -406,27 +412,32 @@ def try_to_submit_job(
     # instrument, data level, descriptor, and start date by checking the CRID.
     # Only check for duplicates if this is a reprocessing job.
     # We know this is a reprocessing job if the version is not "v001".
-    if version != "v001":
-        previous_version = f"v{int(version[1:]) - 1:03d}"
-        if duplicate_job(
-            instrument,
-            data_level,
-            descriptor,
-            start_date_str,
-            previous_version,
-            serialized_dependencies,
-        ):
+    if skip_duplicate_jobs:
+        if version != "v001":
+            previous_version = f"v{int(version[1:]) - 1:03d}"
+            if duplicate_job(
+                instrument,
+                data_level,
+                descriptor,
+                start_date_str,
+                previous_version,
+                serialized_dependencies,
+            ):
+                logger.info(
+                    f"This job is a duplicate of the previous one for: "
+                    f"{instrument=},"
+                    f" {data_level=},"
+                    f" {descriptor=},"
+                    f" {start_date_str=},"
+                    f" and {previous_version=}. "
+                    f"Skipping submission."
+                )
+                return
+        else:
             logger.info(
-                f"This job is a duplicate of the previous one for: "
-                f"{instrument=},"
-                f" {data_level=},"
-                f" {descriptor=},"
-                f" {start_date_str=},"
-                f" and {previous_version=}. "
-                f"Skipping submission."
+                "This is a manually triggered reprocessing job. Not checking"
+                "for duplicate jobs."
             )
-            return
-    # TODO: do we need a reprocessing column to indicate if this is a reprocessing job?
 
     # Serialize the upstream dependencies and write them to a JSON file. The Imap
     # processing code will read the JSON file and deserialize the dependencies. This is
@@ -516,6 +527,7 @@ def submit_all_jobs(
     end_date,
     calculate_crids=False,
     filter_dependencies=True,
+    skip_duplicate_jobs=True,
 ):
     """Submit all jobs for the given job and upstream dependencies.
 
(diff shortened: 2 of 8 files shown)


lacoak21:
I realized I should add some tests for a reprocessing event in batch starter. Will do that next week.

sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
```diff
@@ -96,6 +96,9 @@ def lambda_handler(event, context):
     event : dict
         Specifically looking at the event['pathParameters']['proxy'], which
         specifies the filename to upload.
+        If event['pathParameters']['manually_reprocessed'] == "true", the file will be
```
greglucas, line 99: It is queryString below?

sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
```diff
@@ -152,5 +161,7 @@ def lambda_handler(event, context):
     s3_key_path_str = str(
         s3_key_path.relative_to(imap_data_access.config["DATA_DIR"]).as_posix()
     )
-
-    return _generate_signed_upload_response(s3_key_path_str)
+    s3tags = {"manually_reprocessed": "true"} if manually_reprocessed else None
+    if s3tags:
+        logger.info(f"Uploading file with S3 tags: {s3tags}.")
```
greglucas, line 166: Do we need this log on every upload? I assume this was for debugging. Or maybe we should log the tags/path inside the generate_signed_upload_response call.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
```diff
@@ -20,6 +20,39 @@
 s3 = boto3.client("s3")
 
 
+def manually_reprocessed_file(file_path):
```
greglucas, line 23: It is helpful to add a qualifier if you are checking state. Something like "is it manually reprocessed"
```suggestion
def is_manually_reprocessed_file(file_path):
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -784,6 +803,25 @@ def s3_processing_event(session, events):
...
         filename = body["detail"]["object"]["key"]
+        # Determine if the file is a manually reprocessed file. E.g. the file was
+        # reprocessed due to a bulk reprocessing api event.
+        # This bool is necessary to avoid skipping duplicate jobs in a
+        # "bulk reprocessing" or "manual reprocessing" run. In normal processing or
+        # reprocessing due to a new file version, duplicate jobs are found by seeing if
+        # a dependency file already exists in S3. Since the dependency file contains a
+        # hash of the serialized dependencies, it's unique to that job. Sometimes,
+        # reprocessing is desired for a job that has already been run with the same
+        # dependencies — for example, if the software has been updated.
```
greglucas, line 814: I don't think I fully follow why this is necessary even with this nice documentation. A reprocessing run should always be vXXX + 1, the dependency files are set with that same version as this new file I thought, so it would have to make a new dependency file and would have a different version number (maybe the same hash though) than the previous dependency file that wasn't reprocessed.

greglucas (commented):
I haven't been following the conversations enough to know what is fully going on, so I just put a few high level comments in. It does seem like a lot of extra work for something that I feel like we should be handling by version+1 and not special casing reprocessing X vs reprocessing Y vs standard processing Z.

tech3371:
> I haven't been following the conversations enough to know what is fully going on, so I just put a few high level comments in. It does seem like a lot of extra work for something that I feel like we should be handling by version+1 and not special casing reprocessing X vs reprocessing Y vs standard processing Z.

@greglucas, I totally understand. I thought too. Luisa and I had conversation and reprocessing seems to touch many areas. Luisa thought about many ideas and I did too. But passing reprocessing tag around would be simpler and less complicated. We thought about impact on all three repos and it seems minimal to add this tag. This simplifies logic between these batch starter/dependency/indexer(s) lambda and decouples them. Also, it's very helpful for recovery db during disaster case.

EMM did this too.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -396,6 +397,11 @@ def try_to_submit_job(
     serialized_dependencies : str
         The serialized ProcessingInputCollection of the upstream
         dependencies.
+    skip_duplicate_jobs : bool
+        In normal reprocessing (due to a new file upload) duplicate jobs should be
+        skipped to avoid redundant products. In cases where a user has manually
+        triggered reprocessing, we want to reprocess even if the job looks like a
+        duplicate. Default is True.
```
tech3371, line 404: I liked this point you mentioned.
```suggestion
        duplicate. Eg. If algorithm got updated but the input remained same. Default is True.
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -538,6 +550,11 @@ def submit_all_jobs(
         not want to filter any dependencies out, for example, ULTRA l3
         "u90-ena-h-sf-sp-full-hae-4deg-3mo" needs all the psets in the collection.
         Default is set to True.
+    skip_duplicate_jobs : bool
+        In normal reprocessing (due to a new file upload) duplicate jobs should be
+        skipped to avoid redundant products. In cases where a user has manually
+        triggered reprocessing, we want to reprocess even if the job looks like a
+        duplicate. Default is True.
```
tech3371, line 557: can add same above comment here

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: Luisa can add more but I believe there are other case that we don't want to increment version. Eg. MAG soft dependency or GLOWS L3 case and may be more that I am not remembering. I think we need a pipeline drawing for dependency and batch start logics. It can be hard to connect through code.

tech3371 (commented):
I made some comments but I can make more next week. So far, looks great!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Yes the reprocessing will bump the version and create a new dependency file but this issue surfaced when we added handling for the duplicate mag case. 

A duplicate job is determined by having the same dependency file as the previous version. 

So basically for the mag l1c case, since we have the delay for l1b and now expect to have duplicate mag l1c jobs, we check to see if the dependency file with version - 1 exists. If it does that means that it is a duplicate and we skip. 

mag l1b norm-mago trigger -> mag_l1c_norm-mago_v001.CDF  - dep file mag_l1c_norm-mago-1234abc_v001.CDF
mag l1b burst-mago trigger -> mag_l1c_norm-mago_v002CDF - dep file mag_l1c_norm-mago-1234abc_v002.CDF

Since the two jobs are creating dependency files that are exact except the version has been bumped we skip the second one. 

This causes issues in reprocessing because the previous dependency file might be the same but we want to continue with reprocessing.

greglucas: This concerns me that all of this code is added to handle one unique case. Is "reprocessing" defined as any job above v001? i.e. we can't ever get a standard processing run past that.

I realize you all have talked way more about this than me, so I am just stating my concern and won't block/hold it up. We are complaining about GLOWS L3 complications handled externally, but here we are doing this to ourselves IMO.

lacoak21: No sorry I should have been more clear. When I said "reprocessing" I meant reprocessing that was manually triggered and not triggered by an upload of a new version of a dependent file. 

Reprocessing due to a new file would update the dependency file hash and therefore wouldn't get stuck at this step. 

I think menlo also mentioned that in l3 they will want the deduplication step as well so I hesitated to make it a special case. I can try and talk to menlo and get confirmation on whether they will need that step. I am definitely open to trying to fix this just for mag.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -1152,7 +1197,8 @@ def lambda_handler(events: dict, context):
                     {
                         "body": '{"detail": '
                         '{"object": {"key": '
-                        '"imap_swe_l1b-in-flight-cal_20240101_v001.cdf"}}'
+                        '"imap_swe_l1b-in-flight-cal_20240101_v001.cdf"},
+                        "manually_reprocessed": true'
```
tech3371, line 1201: I think this example should go in 4th bullet point.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
```diff
@@ -20,6 +20,39 @@
...
+
+    Returns
+    -------
+    manually_reprocessed : bool
+        True if the file was reprocessed due to a manual reprocessing event, False
+        otherwise.
+
+    """
+    # Create an S3 client
+    s3_client = boto3.client("s3")
```
tech3371, line 39: This is duplicate of line 20

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -156,6 +156,7 @@ def test_lambda_handler(session, s3_client):
             dt.datetime(2024, 1, 1, 0, 0),
             "v002",
             processing_input.serialize(),
+            True,
```
tech3371, line 159: Should these be false since these are not manually reprocessed?

tech3371: also, should that be optional parameter? If not manual reprocessing, we can skip passing that `--reprocessing` argument.

tech3371 (commented):
Couple comments. Otherwise, the logic looks good to me. Couple minor checks

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: @tech3371 do you have thoughts about skipping the s3 tag and adding special case handling for mag and potentially glows? It is a lot of updates for a rare situation. We could fix this by requiring mag l1c to be reprocessed by specifically supplying the instrument descriptor and level so batch starter will know it is a bulk reprocessing run. 

Im holding off on addressing other comments until we make a decision. 

@maxinelasp I am also tagging you here in case you have thoughts.

tech3371: I still prefer reprocess tag. That seems like the most simple solution.

greglucas (approved):
I think there are a few minor updates left to make based on the comments still, but approving to keep this moving while Tenzin is out. It sounds like you and Tenzin have worked this out in detail and have a good path forward!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: This is actually accurate because the tag is so we can tell if a file upload trigger is a reprocessed file.

sds_data_manager/lambda_code/SDSCode/api_lambdas/upload_api.py
lacoak21: Yes this was for debugging. OK good call - ill add it inside of generate_signed_upload_response

tech3371:
@lacoak21 Should we close this?

PR #775, opened 2025-08-08
I-ALiRT - Alarm added for missing packets

### LINKED ISSUE #761: I-ALiRT - Add Ingest Issue Alarm
### Summary

If packets are not ingested for 1 hr APL wants to know. So we need to add that feature.

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
The MOC requested that an alarm be set in case we do not receive packets. For now I set the alarm to check if a packet arrived within a day and if not send an email. 

## New Files
- ialirt_alarm_construct.py
   - Created SNS topic for CloudWatch alarm.

## Updated Files
- ialirt_bucket_construct.py
   - added metrics used by alarm
- stackbuilder
   - added construct
- deploy.yml
   - added ALARM_EMAIL to the deployment
- ialirt_alarm.py
   - lambda that resets alarm

## Testing
Deployed to personal account and worked as expected.

## Notes
Added email list in Github ALARM_EMAIL.


sds_data_manager/constructs/ialirt_alarm_construct.py
```diff
@@ -0,0 +1,141 @@
...
+        code : lambda_.Code
+            Lambda code bundle.
+        kwargs : dict
+            Keyword arguments.
+
+        """
+        super().__init__(scope, construct_id, **kwargs)
+
+        # Upon deployment we must do
+        # cdk deploy -c alarm_email=ops@example.com
```
greglucas: Can we change this to add the context to the json structure? Otherwise automated deploys from production won't work.
We can store the e-mail in secrets / parameter store if needed instead.

sds_data_manager/constructs/ialirt_alarm_construct.py
```diff
@@ -0,0 +1,141 @@
...
+            datapoints_to_alarm=1440,  # all must be quiet
+            treat_missing_data=cloudwatch.TreatMissingData.BREACHING,
+            comparison_operator=cloudwatch.ComparisonOperator.LESS_THAN_THRESHOLD,
+            alarm_description="Alarm when no packets have arrived.",
+        )
+        alarm.add_alarm_action(cloudwatch_actions.SnsAction(alarm_topic))
+
+        return alarm
+
+    def create_reset_alarm_lambda(self, code, alarm_name):
```
greglucas: You shouldn't need this. The alarm will reset itself once the condition is met again. So you will only get a notification every 24 hours at most I believe and it won't keep notifying you.

laspsandoval: I tried that. And the alarm did not reset. CloudWatch alarms only send notifications when they transition into the ALARM state. It stays in that ALARM state until it is reset.

sds_data_manager/constructs/ialirt_alarm_construct.py
laspsandoval: I put the email here:
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/settings/secrets/actions

And updated the deploy.yml. However, if a user doesn't add the email when they are deploying (for testing for example) then an error won't be thrown. I didn't want people to get stuck on this.

laspsandoval:
@greglucas talked to Jenn Rieter (Gina is out of town) and they don't have an alarm setup for this and I think didn't plan on doing it. I think we should move ahead with this.

sds_data_manager/constructs/ialirt_alarm_construct.py
greglucas: I would rather our automated deployments succeed than our developers struggling with the error. Can you use parameter store instead and that would fail if it doesn't have it available? It isn't a secret e-mail, it is maybe just one you don't want made public.

https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/b0941afd77e7ca33e79c866b28afe5d748ca1ed1/sds_data_manager/constructs/api_gateway_construct.py#L71-L77

sds_data_manager/constructs/ialirt_alarm_construct.py
greglucas: I don't think that is right. It should transition into an unalarmed state once it gets a signal again. The metric would be valid and then only alarm a day later again in the future.

laspsandoval: I went to the Cloudwatch metrics and it was in a continuous alarm state. I agree. It seems wrong, but I only started getting an email per day when I added the reset lambda.

greglucas: You may have been in a continuous alarm state because you weren't getting events again to reset the alarm? You don't want messages every day I don't think. You only want messages when an event happens (no data for the past 24-hours when we previously had some). Otherwise if there is a long 2-week outage you'll be getting messages every single day.

If you do want the spammy version, then I'd suggest changing the alarm/metric combination to just query every single day, keeping only a single metric value rather than 1440. i.e. do you want to just check that 00:00 UTC -> 23:59 UTC was empty, or do you want this to be a rolling check? How immediate do you want the notification?

```
put_metric = cloudwatch.Metric(
    namespace="AWS/S3",
    metric_name="PutRequests",
    period=Duration.hours(24),  # aggregate per day
    statistic="Sum",
    dimensions_map={
        "BucketName": ialirt_bucket.bucket_name,
        "FilterId": "PacketsPrefix",
    },
)

alarm = cloudwatch.Alarm(
    self,
    "IalirtNoPutsDay",
    metric=put_metric,
    threshold=1,
    evaluation_periods=1,
    datapoints_to_alarm=1,
    comparison_operator=cloudwatch.ComparisonOperator.LESS_THAN_THRESHOLD,
    treat_missing_data=cloudwatch.TreatMissingData.BREACHING,
    alarm_description="Alarm when no packets have arrived in the last day.",
)
```

Overall, I'm not in favor of the extra lambda function to reset the alarm.

laspsandoval: Thanks Greg. Sorry I'm just getting to this. I've been making certain the pipeline is working first. I like your idea and agree a reset lambda was not ideal.

laspsandoval: I'll try it out.

laspsandoval: Tested it out again and still same result. I did change the threshold, evaluation period, and datapoints to alarm. And then I set Duration.hours(24). But the lambda is still needed to reset the alarm.

greglucas: I still don't believe this, and ChatGPT doesn't either... I'm not sure what is going on. But the whole point of an alarm is that you would get notified whenever it changes state to ALARMED. @bmcclellan-cu you set up the alarms on H3lioviz stuff, do you know how that works and whether a lambda is needed to reset the alarm state?

laspsandoval: Ok. I changed it. When I put a packets file in s3 it reset. So that's good.

greglucas (approved):
This looks much simpler to me now. Thanks Laura!

PR #778, opened 2025-08-11
CI/REL: Add automated production release workflow

greglucas:
This adds an automated release to our production account whenever we merge into the "prod" branch. So we can have the dev branch follow dev and prod follow prod.


.github/workflows/deploy.yml
```diff
@@ -14,8 +15,20 @@ jobs:
...
+        id: set_account
+        run: |
+          if [[ "${GITHUB_REF##*/}" == "dev" ]]; then
+            echo "account_name=dev" >> $GITHUB_ENV
+            echo "role_arn=arn:aws:iam::449431850278:role/GitHubDeploy" >> $GITHUB_ENV
+          elif [[ "${GITHUB_REF##*/}" == "prod" ]]; then
+            echo "account_name=prod" >> $GITHUB_ENV
+            echo "role_arn=arn:aws:iam::593025701104:role/GitHubDeploy" >> $GITHUB_ENV
+          else
+            echo "Branch not configured for deployment." && exit 1
```
tech3371, line 28: Should we set 'prod' environment variable through here or CDK?
Eg. IMAP_DATA_ACCESS_URL, IMAP_API_KEY, IMAP_WEBPODA_TOKEN or etc. used by `imap-data-access`

greglucas: That won't have any effect here because this is just an environment for deploying into. We should update the CDK for those. Lets open up another ticket for that set of changes.

tech3371: That sounds great! Thank you!

tech3371 (approved):
LGTM. Thank you for doing this!

PR #780, opened 2025-08-11
MNT/FIX: Use account specific configurations in APIs

### LINKED ISSUE #779: Set the proper environment variables for each specific account
We need to update the environment variables that are specific to each account within the CDK.

Should we set 'prod' environment variable through here or CDK?
Eg. IMAP_DATA_ACCESS_URL, IMAP_API_KEY, IMAP_WEBPODA_TOKEN or etc. used by `imap-data-access`

_Originally posted by @tech3371 in https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/778#discussion_r2268130827_

greglucas:
It looks like there is only one place needed to update actually and it was close already, but we can use the cdk.json domain_name to set this explicitly. Let me know if I missed anything else though, I searched for `IMAP_DATA_` and `WEBPODA`

closes #779

### DIFF
--- a/sds_data_manager/constructs/processing_construct.py
+++ b/sds_data_manager/constructs/processing_construct.py
@@ -83,6 +83,13 @@ class ProcessingConstruct(Construct):
             removal_policy=cdk.RemovalPolicy.DESTROY,
         )
         # Create the job definition
+        account_name = self.node.get_context("account_name")
+        # once we have the account_name, get that section out of cdk.json
+        account_config = self.node.get_context(account_name)
+        domain_name = account_config.get("domain_name", "no-domain-set")
+        # https://api.imap-mission.com
+        # https://api.dev.imap-mission.com
+        data_access_url = f"https://api.{domain_name}"
         batch.EcsJobDefinition(
             self,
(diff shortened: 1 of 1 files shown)


sds_data_manager/constructs/processing_construct.py
```diff
@@ -83,6 +83,13 @@ def add_job(self, job_name: str):
             removal_policy=cdk.RemovalPolicy.DESTROY,
         )
         # Create the job definition
+        account_name = self.node.get_context("account_name")
+        # once we have the account_name, get that section out of cdk.json
+        account_config = self.node.get_context(account_name)
```
tech3371, line 88: what is difference between these two?

tech3371 (approved):
This looks great. I was thinking of another point related to public release. But this work looks great. Thank you for quick turnaround!

sds_data_manager/constructs/processing_construct.py
tech3371: Sorry. I just saw that we are doing two step. nvm!

PR #784, opened 2025-08-12
 Separate lambda for API Redirect

daralynnrhode:
# Pulling the Redirect code into a separate lambda

## Overview

After making the changes in the last PR, the https://api.dev.imap-mission.com link results in a `{"message":"Internal Server Error"}`. I am hoping to fix this by putting the redirect code into a separate lambda, similar to how all the other lambdas are invoked. 

## New Files

- landing_page_api.py

## Updated Files

- sds_api_manager_construct.py

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/landing_page_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/landing_page_api.py
@@ -1,0 +1,31 @@
+"""Define lambda to support the home page reroute of the API."""
+
+
+def lambda_handler(event, context):
+    """Entry point to the landing page redirect lambda.
+
+    Redirects incoming requests to the root API path ("/")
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/api_lambdas/landing_page_api.py
```diff
@@ -0,0 +1,31 @@
...
+        about the invocation, function, and runtime environment.
+
+    Returns
+    -------
+    dict
+        A dictionary containing the HTTP status code (302), headers with the
+        Location field set to the landing page URL, and no body
+    """
+    return {
+        "statusCode": 302,
```
bourque, line 26: I think this is the first time I've ever seen a 302 status code used in an API. Neat!

bourque (approved):
I don't have the best knowledge of how these lambda's work, but these code changes look good to me.

PR #787, opened 2025-08-13
Initial implementation for daily job triggering

pleasant-menlo:
# Change Summary

## Overview
This PR is meant to show what it might look like to trigger a processor on a daily cadence rather than via uploaded dependency files. 

These jobs are responsible for managing their own inputs and dependencies. For example, a Glows L3b job would compare the existing L3a files with the existing L3b files to determine which L3b files need to be made or reprocessed. If no changes need to be made, it would exit without producing any files.

Time-based jobs will include: 
- Glows
- ENA products that require multiple maps as input

Our idea is that Glows would be scheduled to run first, and maps would run at a later time, hopefully after Glows has completed.

## Updated Files

- instrument_lambdas.py
   - Add a loop to create EventBridge rules for scheduled Lambda triggering
- batch_starter.py
   - Define a set of processing jobs that will be run by each rule
   - Add a new handler that looks for events from the new scheduled rules and dispatches the appropriate jobs

## Testing

Added a test case for batch_starter. Missing testing for instrument_lambdas.


sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -229,3 +235,26 @@ def __init__(
...
             )
+
+        scheduled_rules = {
+            # Example:
+            # "glows": "cron(20 6 * * ? *)",
+            # "sp_maps": "cron(20 14 * * ? *)",
+        }
+
+        for name, schedule_expression in scheduled_rules.items():
+            aws_events.CfnRule(
```
greglucas: Rather than the level 1 construct `Cfn...` forms here, I think we'd want to use the level 2 constructs like: https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_events/Rule.html
Unless there is an explicit reason we need to dive down into the low-level Cfn formatting.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -65,6 +66,15 @@
     },
 ]
 
+SCHEDULED_JOBS = {
```
greglucas: Maybe extract this out into a `dependency_config.csv`-like file so we can manage it external to the code and it would be a list of these. Potentially even including the cron-string of what schedule you want to use.

greglucas (commented):
Overall, I think I'm 👍 on this.

Do we need/want to shove this into batch_starter, or should it be a separate lambda that gets triggered so we don't get all of the potential "don't kick off this job paths" associated with batch_starter. I'd be worried that we might trigger this every day but every day batch starter would say it doesn't have what it needs, so it continues on and never actually does trigger the job you want.

i.e. I'm wondering if we've abstracted enough components out of batch_starter for you or if we will need to refactor that. The benefit of batch_starter is that it finds dependencies for you, but you aren't using that currently, so perhaps it is easier to avoid it altogether.

lacoak21:
> Overall, I think I'm 👍 on this.
> 
> Do we need/want to shove this into batch_starter, or should it be a separate lambda that gets triggered so we don't get all of the potential "don't kick off this job paths" associated with batch_starter. 

I agree with this. It would be easy to have another simple lambda. I think since you are directly calling "try_to_submit_job" it wouldn't hit any of the "don't kick off this job paths" due to the dependencies but it might be easier for debugging and clarity. 

I like this solution for GLOWS l3d.

pleasant-menlo:
We made some adjustments including:
 - moving the scheduled jobs to its own lambda
 - creating a scheduled_job_config.csv that contains the cron cadence and the job to run
 - Now using the aws Level 2 constructs for setting up the Rule

Let us know if there is anything else we should do. We took our best guess at writing the stuff that interacts with the AWS SDK, so let us know if anything looks wrong there, otherwise we'll find out when we try to run it.

We plan to inject more dependency information into the scheduled jobs at some point, so the processor does not have to be responsible for figuring out which spice kernels, repointing files, and ancillary files to use.

pleasant-menlo:
@greglucas @lacoak21 Hi all, we believe this PR is ready for review at this point. We are now passing the latest repointing file into all processors invoked by scheduled jobs, as GLOWS requires this as a dependency, and there is currently no way to query for the latest version of it.

tests/infrastructure/test_scheduled_job_config_reader.py
```diff
@@ -0,0 +1,108 @@
...
+import textwrap
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+from sds_data_manager.constructs import scheduled_job_config_reader
+
+
+def test_scheduled_job_config_reader():
```
greglucas: You can let pytest take care of the cleanup for you with their provided fixture: https://docs.pytest.org/en/stable/how-to/tmp_path.html
```suggestion
def test_scheduled_job_config_reader(tmp_path):
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/scheduled_job.py
```diff
@@ -0,0 +1,73 @@
...
+    """
+    triggered_jobs = read_scheduled_job_config()[events["scheduled"]]
+
+    processing_inputs = []
+    try:
+        min_python_date = dt.datetime(1, 1, 1)
+        latest_repoint_file_name = dependency.get_latest_repoint_file(min_python_date)
+        processing_inputs.append(RepointInput(latest_repoint_file_name))
+    except ValueError:
+        pass
```
greglucas: Is there a way we can specify the dependencies somehow? Otherwise, every scheduled job is going to require only this repointing file. Maybe that is OK because you only have one needed dependency and we don't anticipate growing this much if any more? (I'm sure someone else will want a cron job eventually...)

I'm sort of wondering if we can do something like read the dependency_config.csv file here and you'd still specify your dependencies over in that file, it would just get triggered on a schedule rather than through other events. This may go in the face of what I said previously though and trying to avoid batch starter... Sorry it has been a while!

sds_data_manager/constructs/scheduled_jobs_config.csv
```diff
@@ -0,0 +1 @@
+cron(20 6 * * ? *), glows, l3b, ion-rate-profile
```
greglucas: Could you add a human-readable schedule of what this cron means as a comment line above this? I don't want to have to look all of them up all the time, so it'd be good to put it for easy reference right above this.

greglucas (commented):
overall looks good. I'll try and test this out later this afternoon / tomorrow.

I have one concern about the finding of the repointing file for all scheduled jobs and what to do if you want/need other dependencies or no dependencies at all (just kick my job off and let it do its thing)

sds_data_manager/constructs/scheduled_job_config_reader.py
```diff
@@ -0,0 +1,53 @@
...
+                continue
+
+            contents = [item.strip() for item in line.split(",")]
+            if len(contents) != 4:
+                raise ValueError(
+                    f"Each scheduled job should have {header}\nCurrent line: {line}"
+                )
+            [schedule, instrument, data_level, descriptor] = contents
+
+            if instrument not in VALID_INSTRUMENTS:
```
lacoak21: Might also be worth checking that the schedule is a valid Schedule.expression() here.

pleasant-menlo: Hi Luisa, we looked into this. Unfortunately, AWS does not make this easy, you can still construct these objects with invalid strings and there is no aws api for validating as far as we can tell. We found [this](https://pypi.org/project/aws-cron-expression-validator/) python package that does full validation. We could also keep the validation more minimal. Do you have a preference?

lacoak21: OK If you already checked then thats fine to keep it more minimal in my opinion. I would avoid adding another package.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/scheduled_job.py
pleasant-menlo: Hi Greg, we agree this is not ideal. However, due to current time restraints we would like to avoid implementing a dependency system. 

For more context, the GLOWS processor has many ancillary and data dependencies. It is able to gather its other dependencies by querying the IMAP API. We could not use this approach for the repointing, because there is currently no API for find the latest repointing file version.  

If there was an API for querying the latest repointing file, this hack of passing it in here could be taken out. Is this something that could be done soon?

Would you recommend handling the repointing file differently before merging this PR? or could it be merged and then updated later when a repointing API is available? 

Thanks for collaborating with us on this

pleasant-menlo:
We just pushed another commit removing batch starter triggering past L3a from the dependency_config.csv

sds_data_manager/constructs/scheduled_jobs_config.csv
```diff
@@ -1 +1,4 @@
+# cron schedule expressions are "cron(minutes hours day-of-month month day-of-week year)"
+# For more information on available values see:
+# https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-scheduled-rule-pattern.html
```
lacoak21: Thats helpful 👍

sds_data_manager/constructs/scheduled_job_config_reader.py
```diff
@@ -41,6 +42,9 @@ def read_scheduled_job_config() -> dict[str, list[dict]]:
                 raise ValueError(f"Invalid instrument: {instrument}")
             if data_level not in VALID_DATALEVELS:
                 raise ValueError(f"Invalid data level: {data_level}")
+            regex = re.compile(r"^cron\(\S* \S* \S* \S* \S* \S*\)$")
```
lacoak21: Thanks!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
```diff
@@ -108,108 +108,6 @@ glows, ancillary, time-dep-bckgrd, glows, l3a, hist, HARD, DOWNSTREAM
 glows, ancillary, map-of-extra-helio-bckgrd, glows, l3a, hist, HARD, DOWNSTREAM
 glows, ancillary, pipeline-settings, glows, l3a, hist, HARD, DOWNSTREAM
 
-glows, l3a, hist, glows, l3b, ion-rate-profile, HARD, DOWNSTREAM
```
lacoak21, line 111: Survival maps triggered from glows l3e will still be triggered by batch starter right?

pleasant-menlo: Hi Luisa, we have been planning to move away from having L3 maps triggered by the batch starter as well. Our understanding of the batch starter map triggering is that every time a maps duration has elapsed a lambda will be  triggered that grabs all the input psets that are currently available for the map time range. 

Unfortunately for maps that depend on GLOWS L3e, there is a large delay between when GLOWS receives data and when it is able to produce L3e files, and its not easy to predict what this delay is as whether or not L3e gets processed depends on many factors. 

We are planning to use a scheduled job to trigger these maps. That job will query to determine what L3e files are available, determine what maps are able to be processed, and then process the maps.

lacoak21: Ok The SDC put in a fix a while ago to trigger survival maps off the arrival of glows l3e data and determine the correct time range to use. If you look at batch_starter.py these jobs are called "SPECIAL_CASE_JOBS". We will need to remove that logic if it will no longer be used. I can make a ticket for this.

lacoak21:
I just tested this in AWS and it successfully submitted the glows job.  The one thing I had to change to get it working is to move `scheduled_jobs_config.csv` and `scheduled_job_config_reader.py` to the pipeline_lambdas folder because only the SDSCode folder is available to the lambda. 

I would also suggest adding more logs as to why a job didnt get kicked off. 

The actual batch job errored out with: 

```
File "/imap_l3_processing/glows/l3bc/glows_initializer_ancillary_dependencies.py", line 40, in fetch_dependencies
pipeline_settings_path = download(pipeline_settings_dependency[0].name)
IndexError: list index out of range
```

lacoak21 (approved):

pleasant-menlo:
Hi Luisa, Thanks for adding logs, and approving. We did have one concern about using the try_submit_job just looking this over today. It looks like try_submit adds the job to a database of processing jobs, and this can fail if the job is already in that table (in an in_progress or completed state). We are not sure that we are making our jobs look unique enough such that they will always pass this check.

lacoak21:
> Hi Luisa, Thanks for adding logs, and approving. We did have one concern about using the try_submit_job just looking this over today. It looks like try_submit adds the job to a database of processing jobs, and this can fail if the job is already in that table (in an in_progress or completed state). We are not sure that we are making our jobs look unique enough such that they will always pass this check.

Sorry I didnt see this before merging in. I agree that this may cause issues if you are trying to reprocess jobs because you dont bump versions. It should be fine if you dont need to reprocess though but there may need to be another branch that bumps versions to make the database input unique.

greglucas:
> Sorry I didnt see this before merging in. I agree that this may cause issues if you are trying to reprocess jobs because you dont bump versions. It should be fine if you dont need to reprocess though but there may need to be another branch that bumps versions to make the database input unique.

I agree. Just to elaborate a bit, I think we'll be fine because the processing table has a unique constraint on it that includes a lot of information, including start_date.

https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/1dc486635cb925924650de7f2a663f1d3a7f652c/sds_data_manager/lambda_code/SDSCode/database/models.py#L97-L110

However, now as I'm typing I just went in and checked and you're supplying a fixed start_date... Can we change that to be the current_date of the run? That way you would get your unique job every day. @pleasant-menlo 

https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/787/files#diff-2b7ea97aa79ef80e35d73ac4498e9a85f308eb317f23fadf1317826575cec5e6R51

PR #790, opened 2025-08-13
Refactor duplicate job handling to avoid s3 tag 

lacoak21:
# Change Summary

## Overview
This PR handles checking for duplicate jobs while allowing for reprocessing jobs to be submitted. This PR is in lieu of [this](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/772). 

It works by first, querying for any jobs that are already in progress or succeeded. If so, and the job is in progress, batch starter skips jobs that are exact duplicates and otherwise bumps the version and continues if it is not ( the dependencies have changed). This is far simpler than the s3 tag solution and even reduces code.

This has been tested for the mag l1c case in S3 and ensures that reprocessing works using this command: 
imap-data-access reprocess --instrument idex --data-level l1a --descriptor all --start-date 20251016 --end-date 20251018

This PR also addresses a small spice querying bug i noticed when testing these changes in aws. 

## Updated Files
   - Add a new column to the processingJob table 
   - This is necessary because although the dependency information gets populated in the container_command column, that isnt done until later in the indexer when the job succeeds or fails.
   - Handle duplicate jobs in more succinct way

## Testing
Update expected versions

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -82,7 +82,6 @@ class ProcessingJob(Base):
     descriptor = Column(String, nullable=False)
     start_date = Column(DateTime, nullable=False)
     version = Column(String(8), nullable=False)
-    dependency_file = Column(String, nullable=False)
     # TODO:
     # Didn't make it required field yet. Revisit this
     # post discussion
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -82,6 +82,7 @@ class ProcessingJob(Base):
     descriptor = Column(String, nullable=False)
     start_date = Column(DateTime, nullable=False)
     version = Column(String(8), nullable=False)
+    dependency_file = Column(String, nullable=False)
     # TODO:
     # Didn't make it required field yet. Revisit this
     # post discussion
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -21,7 +21,6 @@ from sqlalchemy import func
 from sqlalchemy.exc import IntegrityError
 
 from ..api_lambdas import upload_api
-from ..api_lambdas.upload_api import _file_exists
 from ..database import database as db
 from ..database import models
 from . import dependency
@@ -206,11 +205,10 @@ def determine_job_version(
                 # Return the current max version and this job will not proceed if
                 # everything else is the same.
                 return max_version
-            else:
-                logger.info(
-                    f"Job with id: {max_version_record.id} is in progress, but the "
-                    f"dependencies have changed. Bumping version number."
-                )
+            logger.info(
+                f"Job with id: {max_version_record.id} is in progress, but the "
+                f"dependencies have changed. Bumping version number."
+            )
     else:
         max_version = None
     # If the descriptor is "all", we should only check the processing job table. The
@@ -324,63 +322,6 @@ def get_special_case_date_range(session, job_node, start_date):
     return new_start_date, new_end_date
 
 
-def duplicate_job(
-    instrument,
-    data_level,
-    descriptor,
-    start_date,
-    previous_version,
-    serialized_dependencies,
-) -> bool:
-    """Determine if the current job is a duplicate of the most recent job.
-
-    Parameters
-    ----------
-    instrument : str
-        Instrument.
-    data_level : str
-        Data level.
-    descriptor : str
-        Data descriptor.
-    start_date : str
-        Start date.
-    previous_version : str
-        The previous version of the job
-    serialized_dependencies : str
-        The serialized upstream dependencies of the job.
-
-    Returns
-    -------
-    bool
-        True if the job is a duplicate, False otherwise.
-    """
-    # Generate the previous dependency file path based on the inputs.
-    # The descriptor should include a hash of the serialized dependencies.
-    dep_descriptor = f"{descriptor}-{dependency_hash(serialized_dependencies)}"
-    previous_dependency_file = DependencyFilePath.generate_from_inputs(
-        instrument=instrument,
-        data_level=data_level,
-        descriptor=dep_descriptor,
-        start_time=start_date,
-        version=previous_version,
-        extension="json",
-    )
-    previous_dep_path = previous_dependency_file.construct_path()
-    # Strip off the data directory to get the upload path + name
-    # Must be posix style for the URL
-    s3_key_path_str = str(
-        previous_dep_path.relative_to(imap_data_access.config["DATA_DIR"]).as_posix()
-    )
-    # If the file already exists, then we know that an exact duplicate job has been
-    # run with the same dependencies, start date, descriptor, instrument, data level,
-    # and version.
-    # TODO currently this will skip bulk reprocessing jobs that have been kicked off
-    #   Due to algorithm updates. The job might look like a duplicate but have new code.
-    #   We need to somehow track if we are reprocessing due to a manual trigger
-    #   or not.
-    return _file_exists(s3_key_path_str)
-
-
 def dependency_hash(serialized_dependencies):
     """Generate a hash for the serialized dependencies. Use only the first 8 characters.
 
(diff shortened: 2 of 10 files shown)


sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -82,6 +82,7 @@ class ProcessingJob(Base):
     descriptor = Column(String, nullable=False)
     start_date = Column(DateTime, nullable=False)
     version = Column(String(8), nullable=False)
+    dependency_file = Column(String, nullable=False)
```
greglucas: Do we need to add this new column, or can you get away with comparing the "container_command" which I think _contains_ this dependency file in it?

It might still be nice to have this though... Thoughts?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -185,11 +187,30 @@ def filter_conditions(table):
...
-    max_version = (
-        session.query(func.max(models.ProcessingJob.version)).filter(
-            *filter_conditions(models.ProcessingJob)
-        )
-    ).scalar()
+    max_version_record = (
+        session.query(models.ProcessingJob)
+        .filter(*filter_conditions(models.ProcessingJob))
+        .order_by(models.ProcessingJob.version.desc())
+        .first()
```
greglucas, line 194: If we are ordering by version, don't we want the last() record? i.e. I thought these would be "v001", "v002", ...

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -154,7 +154,7 @@ def test_lambda_handler(session, s3_client):
             session,
             {"data_source": "swe", "data_type": "l1a", "descriptor": "sci"},
             dt.datetime(2024, 1, 1, 0, 0),
-            "v002",
+            "v001",
```
greglucas, line 157: Do we actually want these to be v001 now? Were we incorrectly setting these versions before, or what was going on that made this go down a version?

greglucas (approved):
I like this approach a lot. It would be good to get either Tenzin or Maxine's eyes on it too since you are all much more familiar with this area of the code than me. From a high-level though 👍  It seems easier to reason about to me and means that everything is self-contained to "I'm bumping the version based on what we either have (ScienceFiles table) or what is currently in-progress (Processing table) if I'm understanding it right. To summarize, you're going through and getting the maximum version which is the largest in-progress or complete version with the same exact dependencies as before. Then incrementing that value by 1 or setting a default of v001.

lacoak21:
> I like this approach a lot. It would be good to get either Tenzin or Maxine's eyes on it too since you are all much more familiar with this area of the code than me. From a high-level though 👍 It seems easier to reason about to me and means that everything is self-contained to "I'm bumping the version based on what we either have (ScienceFiles table) or what is currently in-progress (Processing table) if I'm understanding it right. To summarize, you're going through and getting the maximum version which is the largest in-progress or complete version with the same exact dependencies as before. Then incrementing that value by 1 or setting a default of v001.

Yes this is going through and getting the max version that is either in progress or complete. IF the job is in progress and the dependencies are the same, we dont bump the version and when we submit that job to the ProcessingJob table it will be a duplicate so processing will not continue. If the dependencies are different, this means we can bump the version and continue. 

I will wait for either Maxine or Tenzin to take a look 👍

sds_data_manager/lambda_code/SDSCode/database/models.py
lacoak21: Yeah unfortunately the container command column is only populated in the indexer after the job has finished so if the first job is in progress, that column will be empty when we try to read it in "determine_job_version". We can update it to submit it in batch starter but im wondering if there is a reason why we did it this way that I dont understand.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Version is being sorted by descending order ( .desc()) so I think we want first(). I can double check this is functioning this way.

tests/lambda_endpoints/test_batch_starter.py
lacoak21: This was because I was calling the lambda handler twice with the same inputs. Determine job version saw that the first job was "inprogress" and bumped the version. Now we only bump the version if the dependency inputs are different. I think I should probably consolidate these tests to only call it once because we are testing this functionality already.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
greglucas: This is correct then. I did not realize what "desc" did, I thought description. You are good here!

sds_data_manager/lambda_code/SDSCode/database/models.py
greglucas: I don't think there is a reason. If you think it would be better to do it in batch_starter immediately and it is available there, then I'd say feel free to make that update if you think it is preferable. If this is better / more ideal, then by all means keep it here.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -185,11 +187,30 @@ def filter_conditions(table):
...
+            dep_file = max_version_record.dependency_file
+            if current_dependency_hash in dep_file:
+                # Return the current max version and this job will not proceed if
+                # everything else is the same.
+                return max_version
+            logger.info(
+                f"Job with id: {max_version_record.id} is in progress, but the "
+                f"dependencies have changed. Bumping version number."
+            )
+    else:
```
maxinelasp, line 212: Very nice. So is this depending on the idea that reprocessing will result in changed dependencies every time, so we don't need to have a special case if we check the dependency hash?

maxinelasp (approved):
I like the change, nice work!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -402,32 +366,6 @@ def try_to_submit_job(
...
-                f"This job is a duplicate of the previous one for: "
-                f"{instrument=},"
-                f" {data_level=},"
-                f" {descriptor=},"
-                f" {start_date_str=},"
-                f" and {previous_version=}. "
-                f"Skipping submission."
-            )
-            return
-    # TODO: do we need a reprocessing column to indicate if this is a reprocessing job?
```
tech3371, line 429: Nice! Lots of simplification!

tech3371 (approved):
This changes looks good to me. Much more simpler for sure. Just curious if that worked with 3mo or 6mo reprocessing?

I traced the bulk reprocessing code and unless it's a special job, we call submit all jobs. I am concerns with what would we do with that. We can kickoff more than one job instead of 3months data with one job. This is outside of this PR but mentioning as an FYI.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: It actually IS checking the dependency hash to see if the job is a true duplicate - meaning if the same exact job is in progress and the dependency file hasnt changed it wont submit.

lacoak21:
> This changes looks good to me. Much more simpler for sure. Just curious if that worked with 3mo or 6mo reprocessing?
> 
> I traced the bulk reprocessing code and unless it's a special job, we call submit all jobs. I am concerns with what would we do with that. We can kickoff more than one job instead of 3months data with one job. This is outside of this PR but mentioning as an FYI.

This is not addressing cadence reprocessing at all. I have a separate PR to handle that and I should be ready to submit it tomorrow!

tech3371:
> > This changes looks good to me. Much more simpler for sure. Just curious if that worked with 3mo or 6mo reprocessing?
> > I traced the bulk reprocessing code and unless it's a special job, we call submit all jobs. I am concerns with what would we do with that. We can kickoff more than one job instead of 3months data with one job. This is outside of this PR but mentioning as an FYI.
> 
> This is not addressing cadence reprocessing at all. I have a separate PR to handle that and I should be ready to submit it tomorrow!

Perfect!

PR #791, opened 2025-08-18
Fix initial time kernel indexing 

lacoak21:
# Change Summary

## Overview
Updates `MAXIMUM_SCLK_INTERVAL` and `MAXIMUM_J2000_INTERVAL` to match `MAXIMUM_DATETIME_INTERVAL`. These constants are used during the initial indexing of the spacecraft time and leapsecond kernel. The default minimum j2000 time used to be a date in 2023 which was causing issues. 

after furnishing the correct time kernels, I used spiceypy.utc2et to convert the datetimes to J2000 and then spiceypy.sce2s to convert from J2000 to sclk @bryan-harter is that what you did initially? 

After this is approved I will reindex the time kernels 

## Updated Files
   - Fix all the maximum interval constants to match the max datetime interval


tmplummer (approved):
LGTM, I double checked the minimum J2000 value.

tech3371 (approved):
LGTM

greglucas:
@lacoak21 can you run the reindex with these new values?

lacoak21:
> @lacoak21 can you run the reindex with these new values?

Yes! WIll do that now.

PR #792, opened 2025-08-19
Cadence bulk reprocessing

lacoak21:
# Change Summary

## Overview
Add handling for cadence reprocessing. 

## Updated Files
   - Add function to handle getting the start_date and stop_dates for map jobs. 
   - Update function to return all cadence jobs if none specified 
-

## Testing
Add test to make sure cadence job gets reprocessed.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -1005,7 +1005,7 @@ def cadence_reprocessing_event(session, job, start_date, end_date):
             node
             for node in DEPENDENCY_CONFIG.get_cadence_jobs(cadence_str)
             if node["data_source"] == job["data_source"]
-            and ["data_type"] == job["data_type"]
+            and node["data_type"] == job["data_type"]
         ]
     else:
         cadence_str = job["descriptor"].split("-")[-1]
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -7,6 +7,7 @@ import logging
 import os
 from enum import Enum
 from pathlib import Path
+from typing import Optional
 
 import boto3
 import imap_data_access
@@ -30,6 +31,9 @@ from .dependency import DependencyConfig, get_jobs
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
+# date format
+DT_FORMAT = "%Y%m%d"
+DEPENDENCY_CONFIG = DependencyConfig()
 # Create a batch client
 BATCH_CLIENT = boto3.client("batch", region_name="us-west-2")
 # Define the retry strategy for batch jobs
@@ -66,7 +70,7 @@ SPECIAL_CASE_JOBS = [
 
 
 def cadence_to_datetime_range(
-    cadence: str, as_str: bool = False
+    cadence: str, start_date: Optional[datetime] = None, as_str: Optional[bool] = False
 ) -> tuple[datetime, datetime] | tuple[str, str]:
     """Convert the cadence to a datetime range.
 
@@ -74,6 +78,9 @@ def cadence_to_datetime_range(
     ----------
     cadence : str
         The cadence string (e.g. "1mo", "3mo", "6mo", "1yr").
+    start_date : datetime, optional
+        The start date for the cadence. This is used to calculate the end date. If
+        not provided, the end date will be set to today. Default is None.
     as_str : bool
         If True, return the start and end dates as strings. Default is False.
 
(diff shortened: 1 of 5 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -918,6 +923,11 @@ def bulk_reprocessing_event(session, events):
     for job in potential_jobs:
         if job in SPECIAL_CASE_JOBS:
             handle_special_case_reprocessing_jobs(session, job, start_date, end_date)
+        elif (
+            job in DEPENDENCY_CONFIG.get_cadence_jobs()
+            or job["descriptor"] in CadenceDays.valid_cadence_str()
+        ):
+            cadence_reprocessing_event(session, job, start_date, end_date)
```
tech3371, line 930: That's a good approach.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -975,62 +985,163 @@ def upload_dependency_file(dependency_file_path: Path, serialized_dependencies:
...
+            cadence_processing_event(
+                session,
+                events=None,
+                job=job_node,
+                start_date=start_date,
+                end_date=end_date,
+            )
+
+
+def cadence_processing_event(
```
tech3371, line 1062: I am glad you separate these two out. That helps a lot with tracking in the future.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -975,62 +985,163 @@ def upload_dependency_file(dependency_file_path: Path, serialized_dependencies:
...
+                    table.descriptor == job_node["descriptor"],
+                    table.start_date
+                    >= datetime.datetime.strptime(start_date, "%Y%m%d"),
+                    table.start_date <= datetime.datetime.strptime(end_date, "%Y%m%d"),
+                    table.status.in_(
+                        [models.Status.INPROGRESS.value, models.Status.SUCCEEDED.value]
+                    ),
+                )
+            ).all()
+        ]
```
tech3371, line 1032: I see, we are getting start_date. for a second I didn't know that. Can we rename this variable to `processed_start_dates` since this is start date of already processed or job in progress dates that we are reusing. Can you also mention that in a comment? Could be helpful in the future

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -975,62 +985,163 @@ def upload_dependency_file(dependency_file_path: Path, serialized_dependencies:
...
+
+        if not dates:
```
tech3371: If no dates, are we using dates passed in? Can we mention that?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -975,62 +985,163 @@ def upload_dependency_file(dependency_file_path: Path, serialized_dependencies:
...
+            continue
+        logger.info(
+            f"Handling cadence reprocessing. Found {len(dates)} files to reprocess."
+        )
+        for date in list(set(dates)):
+            # For each file to reprocess we need to determine the correct
+            # start and end date to use for the map job.
+            start_date, end_date = cadence_to_datetime_range(
```
tech3371, line 1050: In this for loop, we are only using dates from the db, should we do something about what I mention in above comments, about using input dates if no db records.

I see why we wouldn't overwrite with passed in record. At this point, you are assuming that there should be one procesed job at this point this time when we are requesting reprocessing. If that, then we should return in above `if not dates` logic.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -384,11 +385,11 @@ def get_cadence_jobs(self, cadence: str) -> list:
         """
         # Cadence jobs are only at data level l2 and contain either "1mo", "3mo", "6mo",
         # or "1yr" strings as the last part of the descriptor.
-
+        cadences = [cadence] if cadence else ["1mo", "3mo", "6mo", "1yr"]
```
tech3371: can we get this from this [class](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/75fd45742d300147539af01802741f2da4d417fc/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py#L110).valid_cadence_str()?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -384,11 +385,11 @@ def get_cadence_jobs(self, cadence: str) -> list:
...
         return [
-            node
-            for node in self.get_all_nodes("DOWNSTREAM")
-            if node[1] in ["l2", "l2b"] and cadence == node[2].split("-")[-1]
+            {"data_source": ds, "data_type": dt, "descriptor": des}
+            for ds, dt, des in self.get_all_nodes("DOWNSTREAM")
```
tech3371: can we spell out `ds` to data_source and so on?

tech3371 (commented):
The logics looks good to me but would be helpful if you can test on AWS. I had couple minor suggestions

lacoak21:
> The logics looks good to me but would be helpful if you can test on AWS. I had couple minor suggestions

Yes this has all been tested on s3!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: I was getting a circular error import. I can add it to __init__.py

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: If there are no dates that means there were no cadence files produced in that date range and therefore no reprocessing will occure. I am hesitant to use the dates passed in because that might make a map that is no longer a 3mo map at the correct day. This could confuse people. 

This is for reprocessing cadence jobs and not kicking them off (v001) if none exist. Maybe I can update the above query to check jobs that have failed instead. We want to rely on the cadence event bridge events to schedule the jobs with the correct date range.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: yes I dont think we should continue if there are no jobs already processed. I dont think we should return here though because even if there were no jobs for ultra, l2, 6mo, there might be jobs for lo, l2. 6mo. If we return, we would skip checking for the rest of the jobs. By continuing here, we just skip ultra l2 6mo and try to reprocess the lo ones.

tech3371: I see. Can you add some of the comments in the code so that someone doesn't make same misunderstanding as me?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: Thank you for doing that! That helps keep us consistent.

tech3371 (approved):

lacoak21:
pre-commit.ci autofix

PR #795, opened 2025-08-19
FIX: Add APIGateway CORS headers allow list

greglucas:
We don't care what headers get sent our way, so keep it wide open.

This causes an error in the current dev deployment website because they add an "Authorization Header"


lacoak21 (approved):
Makes sense to me.

PR #796, opened 2025-08-19
Updates to the start date and to pass repoint into CLI call

maxinelasp:
# Change Summary
These changes are to clean up and unify the way start_date is handled in sds-data-manager and to pass repoint all the way through to the CLI call correctly.

NOTE: This adds a column to the database, so that will need to be manually updated after merging.

## Updated Files

  - Added repointing column to ProcessingJob model
  - Updated unique constraint to include repointing field
  - Added repointing parameter to try_to_submit_job function
  - Added repointing field to ProcessingJob creation
  - Added repointing to dependency file path generation
  - Added repointing extraction from science files in job processing loop
  - Updated test expectations to include repoint=None parameter in function calls
  - Updated date format expectations from datetime objects to strings
  - Updated version number expectations for test cases

##  Testing

Updated existing unit tests to handle new repointing parameter. All tests maintain existing functionality while
supporting the new repointing feature.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -82,6 +82,7 @@ class ProcessingJob(Base):
     descriptor = Column(String, nullable=False)
     start_date = Column(DateTime, nullable=False)
     version = Column(String(8), nullable=False)
+    repointing = Column(Integer, nullable=True)
     # TODO:
     # Didn't make it required field yet. Revisit this
     # post discussion
@@ -96,6 +97,7 @@ class ProcessingJob(Base):
         # Partial unique index to ensure only one INPROGRESS or COMPLETED for a record
         # We do want to allow multiple FAILED records
         # NOTE: This does not work with sqllite (testing) DBs, only postgres
+        # if any columns are null, they are ignored (such as repointing)
         Index(
             "idx_unique_status",
             "instrument",
@@ -103,6 +105,7 @@ class ProcessingJob(Base):
             "descriptor",
             "start_date",
             "version",
+            "repointing",
             unique=True,
             postgresql_where=and_(status.in_(["INPROGRESS", "SUCCEEDED"])),
         ),
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -120,6 +120,7 @@ class ProcessingJob(Base):
             "descriptor": self.descriptor,
             "start_date": self.start_date.isoformat() if self.start_date else None,
             "version": self.version,
+            "repointing": self.repointing,
             # These parameters could be None when the batch job is in progress
             "job_definition": self.job_definition if self.job_definition else None,
             "job_log_stream_id": self.job_log_stream_id
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -5,6 +5,7 @@ import hashlib
 import json
 import logging
 import os
+import re
 from enum import Enum
 from pathlib import Path
 from typing import Optional
@@ -204,9 +205,16 @@ def determine_job_version(
         # we should bump the version number and continue with processing.
         if max_version_record.status == models.Status.INPROGRESS:
             command = max_version_record.container_command
-            if current_dependency_hash in command:
+            # Extract the dependency hash from the existing job's command
+            dependency_pattern = r"--dependency\s+\S*-([a-f0-9]{8})_\d{8}_v\d{3}\.json"
+            match = re.search(dependency_pattern, command)
+            if match and match.group(1) == current_dependency_hash:
                 # Return the current max version and this job will not proceed if
                 # everything else is the same.
+                logger.info(
+                    "Job already completed or in progress: duplicate dependency hash "
+                    "found"
+                )
                 return max_version
             logger.info(
                 f"Job with id: {max_version_record.id} is in progress, but the "
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -7,16 +7,19 @@ import logging
 import os
 from enum import Enum
 from pathlib import Path
+from typing import Optional
 
 import boto3
 import imap_data_access
 import requests
 from imap_data_access import (
+    VALID_INSTRUMENTS,
     AncillaryFilePath,
     DependencyFilePath,
     ScienceFilePath,
     SPICEFilePath,
 )
+from imap_data_access.processing_input import ProcessingInputType
 from sqlalchemy import func
 from sqlalchemy.exc import IntegrityError
 
@@ -377,9 +380,10 @@ def dependency_hash(serialized_dependencies):
 def try_to_submit_job(
     session: db.Session,
     job_info: dict,
-    start_date: datetime,
+    start_date: str,
     version: str,
     serialized_dependencies: str,
+    repoint: Optional[int] = None,
 ):
     """Try to submit a batch job with the given job information.
 
@@ -389,18 +393,20 @@ def try_to_submit_job(
         Database session.
     job_info : dict
         Dictionary containing components with dates and versions appended.
-    start_date : datetime
-        Start date of the data.
+    start_date : str
+        Start date of the data in the format 'YYYYMMDD'.
     version : str
         Version of the job.
     serialized_dependencies : str
         The serialized ProcessingInputCollection of the upstream
         dependencies.
+    repoint : int, optional
+        The repointing number for the job, if applicable. Default is None. Should
+        be just an integer, no "repoint" prefix.
     """
     instrument = job_info["data_source"]
     data_level = job_info["data_type"]
     descriptor = job_info["descriptor"]
-    start_date_str = datetime.datetime.strftime(start_date, "%Y%m%d")
 
     # Search for any duplicate jobs that have the same exact dependencies for this
     # instrument, data level, descriptor, and start date by checking the CRID.
@@ -412,7 +418,7 @@ def try_to_submit_job(
             instrument,
             data_level,
             descriptor,
-            start_date_str,
+            start_date,
             previous_version,
             serialized_dependencies,
         ):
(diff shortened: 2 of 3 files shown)


maxinelasp:
So unfortunately I couldn’t get it working fully before the end of the day (I’ll continue trying) but the merge with your recent changes made some tests fail and I’m having trouble figuring out where I went wrong with the merge. @lacoak21, would it be possible to have you take a quick look at the two failing batch_starter tests to see if you can figure out where the versioning/deduplication is going wrong? I am having trouble figuring it out... Thank you!!

As a reminder I'll be out for the rest of the week, but I wanted to put this up now to give everyone a chance to review.

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -96,13 +97,15 @@ class ProcessingJob(Base):
...
         # NOTE: This does not work with sqllite (testing) DBs, only postgres
+        # if any columns are null, they are ignored (such as repointing)
         Index(
             "idx_unique_status",
             "instrument",
             "data_level",
             "descriptor",
             "start_date",
             "version",
+            "repointing",
```
tech3371, line 108: Do you want to add this to the `to_dict()` function too?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -775,25 +795,31 @@ def s3_processing_event(session, events):
...
         calculate_crids = isinstance(file_obj, ScienceFilePath)
         for job in potential_jobs + potential_soft_jobs:
+            if job["data_source"] not in VALID_INSTRUMENTS:
+                raise ValueError(
+                    f"Unable to submit job for invalid instrument {job['data_source']}."
+                    f" Downstream dependencies must be science files."
+                )
+
             job.pop("relationship")
-            if job in SPECIAL_CASE_JOBS:
```
tech3371, line 785: I personally find previous logic easier to follow. Can we undo this change?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -537,8 +551,14 @@ def submit_all_jobs(
...
     num_jobs = len(primary_science.imap_file_paths)
     logger.info(f"Found {num_jobs} jobs to process.")
-    for filepath in primary_science.imap_file_paths:
-        job_start_date = datetime.datetime.strptime(filepath.start_date, "%Y%m%d")
+    for filename in primary_science.filename_list:
+        science_file = ScienceFilePath(filename)
+        start_date, end_date = determine_date_range(session, science_file)
+
+        # Get the repointing number from the same spot as the time range
+        # Is either a number or None
```
```suggestion
        # Get the repointing number from the science file object
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -567,16 +587,17 @@ def submit_all_jobs(
...
             current_dependency_hash=dependency_hash(serialized_deps),
         )
         try_to_submit_job(
             session,
             job_node,
-            job_start_date,
+            start_date,
             job_version,
             serialized_deps,
+            repoint=job_repointing,
```
tech3371, line 606: This looks good to me

tech3371 (commented):
Repoint changes looks good to me. I am going to ask @lacoak21 to review the other code changes in the batch starter.

Regarding the time changes, can we do that in another PR? I also get confused with what is str format and what is date format. We can do that in another PR to reduce confusion later on.

maxinelasp:
@tech3371 the start time changes are necessary in this PR because I needed to clean up the way time worked in order to access it where needed in the repointing calls. It would be challenging to separate the two out. 

If it would help, I can add more comments to the changes describing why they were necessary. The start_date as it currently exists in batch_starter is pretty confusing, which is why I renamed some variables and re-typed others.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -341,9 +344,10 @@ def dependency_hash(serialized_dependencies):
 def try_to_submit_job(
     session: db.Session,
     job_info: dict,
-    start_date: datetime,
+    start_date: str,
```
maxinelasp, line 354: I made this a string because generally start_date is called as a string, and most places in this function it's used as a string.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -412,8 +422,9 @@ def try_to_submit_job(
         instrument=instrument,
         data_level=data_level,
         descriptor=descriptor,
-        start_date=start_date,
+        start_date=datetime.datetime.strptime(start_date, "%Y%m%d"),
```
maxinelasp, line 432: the only time start_date is NOT a string is in database calls. I think it makes more sense to treat it as a string everywhere and convert only at the database steps.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -450,8 +461,8 @@ def try_to_submit_job(
 def submit_all_jobs(
     session,
     job_node,
-    start_date,
-    end_date,
+    trigger_start_date,
```
maxinelasp, line 471: This is not the start date of the _job_ but rather the start date of the triggering file. The only change here is to make it a string (as talked about above) and to rename it to be more clear that this is distinct from the "start_date" in try_to_submit_jobs and elsewhere.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -537,8 +551,14 @@ def submit_all_jobs(
     primary_science = primary_science_inputs[0]
     num_jobs = len(primary_science.imap_file_paths)
     logger.info(f"Found {num_jobs} jobs to process.")
-    for filepath in primary_science.imap_file_paths:
-        job_start_date = datetime.datetime.strptime(filepath.start_date, "%Y%m%d")
+    for filename in primary_science.filename_list:
+        science_file = ScienceFilePath(filename)
+        start_date, end_date = determine_date_range(session, science_file)
```
maxinelasp, line 563: Now there isn't a collision between "start_date" and "trigger_start_date". Before, there was a bug where one function was expecting the start date of the input file ("trigger_start_date") but what was getting passed in was the start date of the job "start_date". Moving the logic here removes that bug and renaming the variables makes it more clear what the start date is referring to.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -567,16 +587,17 @@ def submit_all_jobs(
             session=session,
             instrument=job_node["data_source"],
             descriptor=job_node["descriptor"],
-            start_date=job_start_date,
+            start_date=datetime.datetime.strptime(start_date, "%Y%m%d"),
```
maxinelasp, line 596: Again, this is the only spot in this function that needs anything except a string, so we can convert it here.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
maxinelasp: Yeah, this ended up being a weird change. I'll change it back.

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -624,7 +626,7 @@ def test_lambda_handler_missing_dependency_for_start_date(session, caplog):
     # Check that the expected message was logged.
     expected_log = (
         "Skipping job submission for {'data_source': 'mag', 'data_type': "
-        "'l2', 'descriptor': 'norm-srf'} with start_date: 20250117 because"
+        "'l2', 'descriptor': 'norm-srf'} with start_date: 20250418 because"
```
maxinelasp, line 629: Here is the bug I mentioned before. We do expect that job to have a start date of 20250418 even though the triggering file starts at 20250117

tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -1485,8 +1489,8 @@ def test_idex_l2b(session):
         mock_submit.assert_called_with(
             session,
             {"data_source": "idex", "data_type": "l2b", "descriptor": "all-1mo"},
-            dt.datetime(2023, 1, 9, 0, 0),
-            "v001",
+            "20230109",
+            "v002",
```
maxinelasp, line 1565: I feel pretty confident that this version number should be v002 instead of v001, but I was surprised at the change. I think the test before was incorrect, but would appreciate a second look.

lacoak21: Ill take a look now.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Good distinction.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -537,8 +551,14 @@ def submit_all_jobs(
     primary_science = primary_science_inputs[0]
     num_jobs = len(primary_science.imap_file_paths)
     logger.info(f"Found {num_jobs} jobs to process.")
-    for filepath in primary_science.imap_file_paths:
-        job_start_date = datetime.datetime.strptime(filepath.start_date, "%Y%m%d")
+    for filename in primary_science.filename_list:
```
lacoak21, line 561: Why did you change from primary_science.imap_file_paths to primary_science.filename_list? I see you convert it to the scienceFilePath below.

tests/lambda_endpoints/test_batch_starter.py
lacoak21: Oh interesting. This only passes with 'v001' for me. I checked out your branch. It also _should_ be __v001__ because it is the exact same job with the same dependencies so the version did not get bumped and in try to submit job it will exit due to a duplicate processing job

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
maxinelasp: imap_file_paths returns an ImapFilePath list, but I only want specifically ScienceFilePaths as later steps use ScienceFilePath specific attributes and that will break if, for example, an AncillaryFilePath is in there. Instead, it will obviously break on the conversion from an ancillary filepath to a ScienceFilePath and it will be clear what went wrong, and there is no odds of something slipping through.

lacoak21 (approved):
LGTM ! Thanks for making the start date handling cleaner.

tech3371 (approved):
we need this for launch prep SIT and will test details and fix as needed in future PR.

PR #802, opened 2025-08-21
I-ALiRT - adding cors

laspsandoval:
# Change Summary

## Overview
The webteam needs CORS permissions assigned to s3. The reason for this is that we use a presigned url. This is what happens:

1. We hit the I-ALiRT API
2. API Gateway calls download_api.py lambda
3. That lambda does not send back a file directly. Instead it sends back a presigned url for the file in s3.
4.  At this point the browser is not talking to the api gateway anymore it is talking directly to s3.

## Updated Files
- ialirt_bucket_construct.py
   - Add CORS permissions

### DIFF
--- a/sds_data_manager/constructs/ialirt_bucket_construct.py
+++ b/sds_data_manager/constructs/ialirt_bucket_construct.py
@@ -44,4 +44,16 @@ class IAlirtBucketConstruct(Construct):
             removal_policy=RemovalPolicy.DESTROY,
             auto_delete_objects=True,
             block_public_access=s3.BlockPublicAccess.BLOCK_ALL,
+            cors=[
+                s3.CorsRule(
+                    allowed_methods=[
+                        s3.HttpMethods.GET,  # For downloading JSON files
+                        s3.HttpMethods.HEAD,  # For metadata requests
+                        s3.HttpMethods.OPTIONS,  # Required for presigned URLs
+                    ],
+                    allowed_origins=["*"],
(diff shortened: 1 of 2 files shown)


greglucas (approved):
Looks good to me!

Care to open up a separate PR with the same fix for the SDC bucket (or include here if you want)? I think you are just the first one to encounter this with the frontend actually getting files. I'm pretty sure you're right that we'll need to do this for quicklook or downloading data from the SDC side too.

sds_data_manager/constructs/ialirt_bucket_construct.py
```diff
@@ -44,4 +44,16 @@ def __init__(
             removal_policy=RemovalPolicy.DESTROY,
             auto_delete_objects=True,
             block_public_access=s3.BlockPublicAccess.BLOCK_ALL,
+            cors=[
+                s3.CorsRule(
+                    allowed_methods=[
+                        s3.HttpMethods.GET,  # For downloading JSON files
+                        s3.HttpMethods.HEAD,  # For metadata requests
+                        s3.HttpMethods.OPTIONS,  # Required for presigned URLs
```
greglucas: There isn't an options field?
https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_s3/CorsRule.html
https://docs.aws.amazon.com/cdk/api/v2/python/aws_cdk.aws_s3/HttpMethods.html
```suggestion
```

greglucas (approved):

PR #805, opened 2025-08-22
CRID calculation bug fix

lacoak21:
# Change Summary

## Overview
While getting the upstream versions, the crid calculation was not skipping over dependencies that are not science files. 

## Updated Files
-sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
   - Add check to only get science file inputs


tech3371 (approved):
interesting, what was not valid instrument?

PR #810, opened 2025-08-27
Create pointing schedule lambda

torimarbois:
# Change Summary

## Overview
Create a lambda function that generates and uploads the ground station pointing schedule text files.

## New Files

- ialirt_pointing_schedule.py
   - Wraps the `generate_text_files()` function from imap_processing to upload pointing schedules to s3
- test_ialirt_pointing_schedule.py
   - Tests the pointing_schedule lambda

## Testing

New tests found in test_ialirt_pointing_schedule.py for both the lambda handler and the generate_and_upload_schedule function.

## Note
This PR goes hand-in-hand with the [imap_processing PR that adds the generate_text_files() function](https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/2045). I will wait until that PR is merged to test & merge this one.

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
@@ -17,15 +17,6 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 
-KERNELS = [
-    "planetary_ephemeris",  # e.g., de440s.bsp
-    "planetary_constants",  # e.g. pck00011.tpc
-    "leapseconds",  # e.g., naif0012.tls
-    "spacecraft_ephemeris",  # e.g., imap_spk_demo.bsp
-    "earth_attitude",  # e.g., earth_latest_high_prec.bpc
-]
-
-
 def generate_and_upload_schedule(bucket: str, region: str, station: str, day: str):
     """Generate and upload pointing schedule files to S3.
 
@@ -76,7 +67,8 @@ def lambda_handler(event, context):
             "planetary_ephemeris",  # e.g., de440s.bsp
             "planetary_constants",  # e.g. pck00011.tpc
             "leapseconds",  # e.g., naif0012.tls
-            "spacecraft_ephemeris",  # e.g., imap_spk_demo.bsp
+            "ephemeris_predicted",  # e.g., imap_spk_demo.bsp
+            "ephemeris_90days",  # e.g., imap_spk_demo.bsp
             "earth_attitude",  # e.g., earth_latest_high_prec.bpc
         ]
     )
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
@@ -1,0 +1,89 @@
+"""Lambda to generate and upload antenna pointing schedule files to s3."""
+
+import json
+import logging
+import os
+from datetime import datetime, timedelta, timezone
+
+import boto3
+import imap_processing.ialirt.constants
+
+from sds_data_manager.lambda_code.IAlirtCode.ialirt_coverage import (
+    get_latest_spice_kernels,
+    setup_spice_file,
+)
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+KERNELS = {
+    "planetary_ephemeris",  # e.g., de440s.bsp
+    "planetary_constants",  # e.g. pck00011.tpc
+    "leapseconds",  # e.g., naif0012.tls
+    "spacecraft_ephemeris",  # e.g., imap_spk_demo.bsp
+    "earth_attitude",  # e.g., earth_latest_high_prec.bpc
+}
+
+
+def generate_and_upload_schedule(bucket: str, region: str, station: str, day: str):
+    """Generate and upload pointing schedule files to S3.
+
+    Parameters
+    ----------
+    bucket : str
+        The name of the S3 bucket.
+    region : str
+        The region in which the s3 bucket resides.
+    station : str
+        The ground station for which the pointing schedule file should be generated.
+    day : str
+        The day for which to generate a pointing schedule, in ISO format.
+        Ex: "20250811".
+    """
+    file_name = f"{day}_{station}.txt"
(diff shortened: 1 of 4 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
```diff
@@ -0,0 +1,89 @@
...
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+KERNELS = {
+    "planetary_ephemeris",  # e.g., de440s.bsp
+    "planetary_constants",  # e.g. pck00011.tpc
+    "leapseconds",  # e.g., naif0012.tls
+    "spacecraft_ephemeris",  # e.g., imap_spk_demo.bsp
```
torimarbois: Should this be the `ephemeris_predicted` category instead of `spacecraft_ephemeris`?

sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
```diff
@@ -0,0 +1,89 @@
...
+from sds_data_manager.lambda_code.IAlirtCode.ialirt_coverage import (
+    get_latest_spice_kernels,
+    setup_spice_file,
+)
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+KERNELS = {
```
greglucas: Does this need to be a set, or can it be a list of kernels? Where does it get used?
```suggestion
KERNELS = [
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
```diff
@@ -0,0 +1,89 @@
...
+        The ground station for which the pointing schedule file should be generated.
+    day : str
+        The day for which to generate a pointing schedule, in ISO format.
+        Ex: "20250811".
+    """
+    file_name = f"{day}_{station}.txt"
+    s3_path = f"pointing_schedules/{station}/{day}/{file_name}"
+    local_file_path = ".."
+
+    # generate_text_files(station, day, local_file_path)
```
greglucas:
```suggestion
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
```diff
@@ -0,0 +1,89 @@
...
+            s3_client.put_object(Bucket=bucket, Key=s3_path, Body=f)
+        print(
+            f"Pointing schedule '{day}_{station}.txt' uploaded to "
+            f"s3://{bucket}/{s3_path}"
+        )
+    except Exception as e:
+        print(f"Error uploading file: {e}")
+
+    # Remove the file after it's been uploaded to s3
+    os.remove(f"{local_file_path}/{day}_{station}.txt")
```
greglucas: You are opening up an already open filehandle here which can be unpredictable with whether you are reading/writing from various locations on disk. `os.remove()` at the end is dangerous if there is an exception that occurs too which might leave a dangling file around that you don't want.

I think this could be done a little more succinctly with an in-memory io.BytesIO() object and then you avoid any need of the local filesystem which can also be odd on Lambdas.

Something like:
```python
raw_bytes = b"".join(...)
# or if you need io.BytesIO()
x = io.BytesIO()
x.writelines(...)
# Go back to the beginning of the memory location
x.seek(0)

s3_client.put_object(..., Body=raw_bytes)
```

torimarbois: This is tricky because although this code as-is shows the lines being written here in the lambda, it's actually happening inside of the `generate_text_files()` function. Do you have suggestions for how to wrap the file-writing logic in a similar way, but keeping it in the separate function from `imap_processing`?

torimarbois: With Laura's help, I changed the logic of the text file function to return a list of strings rather than writing the file from within the function. This should mean that I can use the io.BytesIO() suggestion.

sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
laspsandoval: Yes. It should be these

   "ephemeris_predicted",
    "ephemeris_90days",

sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
```diff
@@ -0,0 +1,89 @@
...
+    with open(f"{local_file_path}/{file_name}", "w") as file:
+        file.writelines([f"Station: {station}\n", "Target: IMAP\n"])
+
+    try:
+        s3_client = boto3.client("s3", region_name=region)
+
+        with open(f"{local_file_path}/{file_name}", "rb") as f:
+            # Upload the file content to S3
+            s3_client.put_object(Bucket=bucket, Key=s3_path, Body=f)
+        print(
```
laspsandoval: Add a log instead of printing

sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
```diff
@@ -0,0 +1,89 @@
...
+            f"Pointing schedule '{day}_{station}.txt' uploaded to "
+            f"s3://{bucket}/{s3_path}"
+        )
+    except Exception as e:
+        print(f"Error uploading file: {e}")
+
+    # Remove the file after it's been uploaded to s3
```
laspsandoval: Since you are moving to an in-memory upload use:
lines = generate_text_lines(station, day)
s3_client.put_object(
    Bucket=bucket,
    Key=s3_path,
    Body="".join(lines),
    ContentType="text/plain",
)

And then remove:
os.remove(...)

laspsandoval (commented):
A few comments. Looking good.

sds_data_manager/lambda_code/IAlirtCode/ialirt_pointing_schedule.py
torimarbois: Do these two categories correspond to the one file `imap_spk_demo.bsp`?

laspsandoval (approved):
Nice work Tori!

PR #822, opened 2025-09-05
FIX: Update the packet_downloader_lambda trigger for new location

greglucas:
We updated the spice location to be in /imap/spice instead of the root /spice, so we need to update that here. Import imap-data-access to get the location from there, since that is what is used to populate the filepaths.


maxinelasp (approved):

greglucas:
Not sure why I didn't think of this before, but I think we could easily just remove the suffix portion. This is the only file landing in the `/repoint/` subfolder, so lets just trigger whenever anything arrives and this will handle the future when the suffix changes on us.

PR #825, opened 2025-09-08
I-ALiRT - bug fix for dynamodb query

laspsandoval:
# Change Summary

## Overview
Realized that I was excluding some parameters. This is the fix.

## Updated Files
- ialirt_db_query_api.py

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api.py
@@ -28,15 +28,19 @@ def process_item_types(item: dict) -> dict:
     result = {}
 
     for key, value in item.items():
-        # Vectors fields
-        if key in {"mag_B_GSE", "mag_B_GSM", "mag_B_RTN"} and isinstance(value, list):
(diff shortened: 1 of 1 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api.py
```diff
@@ -28,15 +28,19 @@ def process_item_types(item: dict) -> dict:
...
+            if isinstance(value, list):
+                result[key] = [int(v) if v % 1 == 0 else float(v) for v in value]
+
+            # Scalar fields
+            elif isinstance(value, Decimal):
+                result[key] = int(value) if value % 1 == 0 else float(value)
+
+            else:
+                result[key] = value
+        except Exception as e:
```
greglucas: Why is this a bare exception? We should try and narrow the scope a bit to the actual ValueError / KeyError that is raised instead.

I'm not sure from the description of the PR what the underlying issue is and why this is the fix. Theoretically, you should be able to see what exceptions are being raised here and handle them above instead of defaulting back to the generic `result[key] = value` case.

greglucas (approved):

PR #836, opened 2025-09-10
ENH: Restrict APIs to not show unreleased files

### LINKED ISSUE #786: ENH: add handler for public and private API
This can effect all API lambdas. We need to add handler and how to give response back based on which API user hit. Eg. public, we need to return all public data and if authentication, then we need to return all data including non-released data.

tech3371:
# Change Summary
closes #786 

## Overview

filters unreleased files from return list based on which user it is. APIs that got effected by this are:
* download
* query

APIs that public won't have access to are:
* upload
* batch jobs logs
* batch job status
* reprocess

APIs that is made public right away are:
* SPICE query API
* metakernel API

## Updated Files
updated few files related to above needs


sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
```diff
@@ -112,6 +159,16 @@ def lambda_handler(event, context):
             ),
         )
 
+    is_auth_user = is_authenticated_path(event)
+    file_is_released = is_released(filepath)
+
+    if not is_auth_user and not file_is_released:
```
greglucas: We don't want to query the DB unnecessarily, so if we are on the authenticated path, we don't care to look up whether it is released or not. If you put these into the if-block itself it won't evaluate the second condition and short-circuit to avoid the unnecessary query.
```suggestion
    if not is_authenticated_path(event) and not is_released(filepath):
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
```diff
@@ -0,0 +1,97 @@
...
+    bool
+        True if the path is authenticated, False otherwise
+    """
+    # Get the routeKey and rawPath, defaulting to empty strings if not present
+    auth_api_endpoint = ["api-key", "authorized"]
+
+    # Safely extract route_key and raw_path
+    route_key = ""
+    if event.get("routeKey") and "/" in event.get("routeKey"):
+        route_key = event.get("routeKey").split("/")[1]
```
greglucas: You can use the default return of get to shorten these up.
```suggestion
    rk = event.get("routeKey", "")
    route_key = rk.split("/")[1] if "/" in rk else ""
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+    raw_path = ""
+    if (
+        event.get("rawPath")
+        and "/" in event.get("rawPath")
+        and len(event.get("rawPath").split("/")) > 1
+    ):
+        raw_path = event.get("rawPath").split("/")[1]
```
greglucas:
```suggestion
    rp = event.get("rawPath", "")
    raw_path = rp.split("/")[1] if "/" in rp else ""
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+    logger.info(f"Route Key: {route_key}, Raw Path: {raw_path}")
+    # Check if either contains authentication indicators
+    is_auth_user = route_key in auth_api_endpoint or raw_path in auth_api_endpoint
```
greglucas: Why do we need both paths to check? I think route_key should be enough. In fact, I think this entire function should basically be a one-liner, otherwise there is too much going on.

```python
return event.get("routeKey", "").startswith(("api-key", "authorized"))
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # Log the authentication determination
+    if is_auth_user:
+        logger.info("Returning all files, including unreleased ones.")
+        return search_results
+
+    # Filter out unreleased files for non-authenticated paths
+    released_files = [
+        result for result in search_results if result.get("released") is True
+    ]
```
greglucas: We should add a filter to the DB query rather than here.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -1273,10 +1276,20 @@ def lambda_handler(events: dict, context):
         Lambda context object
     """
     logger.info(f"Events: {events}")
+    is_auth_user = is_authenticated_path(events)
```
greglucas: Should we be allowed to enter this path from a non-authenticated pathway? i.e. we should remove the route if we don't want this to be allowed. Then we know whoever entered here is authenticated.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -1273,10 +1276,20 @@ def lambda_handler(events: dict, context):
...
+    is_reprocessing = (
+        api_event.get("reprocessing")
+        if (api_event := events.get("queryStringParameters"))
+        else False
+    )
```
greglucas: Use the default get return to chain these together.
```suggestion
    is_reprocessing = events.get("queryStringParameters", {}).get("reprocessing", False)
```

greglucas:
In the query API, I'd add something like this around line 55, so that you add this to the DB query itself so the DB is doing the filtering.

```python
query = select(model.__table__)
if not is_authenticated_user():
    query = query.filter("release" == True)
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
tech3371: that's what I had originally but AI changes

greglucas: See lower down for an even simpler update to everything to make this a one-line function.

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
tech3371: I didn't want to touch other code in rush we are in. That's why I added that final filtering.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: I don't follow completely here

greglucas: I don't think a non-authenticated user should ever be allowed to trigger batch_starter? i.e. I don't know why we need to check if we are an auth user in batch_starter, we should assume everyone is authorized that has made it this far.

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
greglucas: I'm not sure I follow. I think it would be simpler to add it to a DB query filter search. See my high level comment for a suggestion instead of doing it here.

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
tech3371: we can't do that since routeKey has 'GET /sth'. I am looking at workaround

greglucas: Use rawPath then. Whichever one doesn't have the GET in front of it and is just the `/api-key/download/xyz`

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: oh ok. So only check for auth user. That won't be good for other cases such as evenbridge and other automated actions

greglucas: I think I'm not following. That should make it easier because we are saying they always have access as long as they can trigger batch_starter.

The restriction comes in on the API endpoints, and we don't make whatever endpoint it is that can trigger this available publicly. That is our restriction, not within the batch starter code itself.

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
tech3371: no worries, I got confused. will simplify based on your feedback

sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
```diff
@@ -112,6 +159,13 @@ def lambda_handler(event, context):
             ),
         )
 
+    if not is_authenticated_user(event) and not is_released(filepath):
+        return http_response(
+            status_code=403,
+            body="The requested file is not released yet. Please contact the "
+            "SDC team for more information.",
```
greglucas: I wouldn't tell anyone to contact us, they can reach out themselves if they want. Also include the filepath for help debugging potentially.
```suggestion
            body=f"The file {filepath} is not a part of a public release yet."
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
```diff
@@ -52,6 +53,9 @@ def lambda_handler(event, context):
 
     # select the given table for the query
     query = select(model.__table__)
+    if not is_authenticated_user(event):
+        query = query.filter(model.released)
```
greglucas, line 60: It looks like this is the correct spelling on the tables. There are other "release" without the "d" in places that should be fixed up I think.

sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
```diff
@@ -6,11 +6,58 @@
...
+                table = models.ScienceFiles
+            else:
+                return False
+        else:
+            return False
+
+        with db.Session() as session:
+            file_record = session.query(table).filter(table.file_path == s3_key).first()
+            if file_record and hasattr(file_record, "release"):
+                return file_record.release
```
greglucas: Here is one with the "release" when it should be "released" I believe.

```suggestion
            if file_record:
                return getattr(file_record, "released", False)
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
greglucas, line 60: It looks like this is the correct spelling on the tables. There are other "release" without the "d" in places that should be fixed up I think.

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+    Returns
+    -------
+    bool
+        True if the path is authenticated, False otherwise
+    """
+    rk = event.get("routeKey", "")
+    route_key = rk.split("/")[1] if "/" in rk else ""
+    logger.error(f"Route Key: {route_key}")
+    return route_key in ("authorized", "api-key")
```
greglucas: I find this easier to read than a split personally.

The logger.error should be changed to a different log level, but I'd suggest just removing it altogether because this will be called a lot I think.
```suggestion
    return event.get("rawPath", "").startswith(("/authorized", "/api-key"))
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -1273,10 +1276,16 @@ def lambda_handler(events: dict, context):
         Lambda context object
     """
     logger.info(f"Events: {events}")
+    is_auth_user = is_authenticated_user(events)
+    is_reprocessing = events.get("queryStringParameters", {}).get("reprocessing", False)
+
+    if is_reprocessing and not is_auth_user:
+        logger.error("Unauthorized reprocessing attempt.")
+        return "Unauthorized reprocessing attempt."
```
greglucas: I'm not sure this path can ever happen? You've updated the "/reprocessing" API endpoints to only be authorized. That is what I meant by getting rid of this spot.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -1273,10 +1276,16 @@ def lambda_handler(events: dict, context):
...
 
     with db.Session() as session:
         api_event = events.get("queryStringParameters")
-        if api_event and api_event.get("reprocessing"):
+        if api_event and is_reprocessing and is_auth_user:
```
greglucas: This was already checked above? Maybe move the api_event above this even if you need it here as well.

(If you keep these code blocks, again I think they can be removed if you want)

greglucas (commented):
Looks good to me! Just some minor comments, particularly the release vs released spelling to check on that.

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
tech3371: Sounds good. But I tried that starts with but note that the routeKey doesn't start with /authorized but with something like this 'GET /authorized'. That's why I still had to split by '/'.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
tech3371: I see. That's true.

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
greglucas: I switched it to "rawPath" which does I think.

sds_data_manager/lambda_code/SDSCode/api_lambdas/utils.py
```diff
@@ -0,0 +1,29 @@
...
+        The API Gateway event object
+
+    Returns
+    -------
+    bool
+        True if the path is authenticated, False otherwise
+    """
+    rk = event.get("routeKey", "")
+    route_key = rk.split("/")[1] if "/" in rk else ""
+    return route_key in ("authorized", "api-key")
```
lacoak21: Can you use the always_private_routes list from the construct? Or put them in a common place for both places to reference?

sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
```diff
@@ -6,11 +6,58 @@
...
+    -------
+    bool
+        True if the file is released, False otherwise.
+    """
+    try:
+        filename = os.path.basename(s3_key)
+        file_obj = imap_data_access.file_validation.generate_imap_file_path(filename)
+
+        # Determine which table to use based on file type
+        if hasattr(file_obj, "__class__") and hasattr(file_obj.__class__, "__name__"):
```
lacoak21: You could also use isinstance() here.

lacoak21 (approved):

sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
tech3371: true!

PR #842, opened 2025-09-11
Skip spice query for initial upstream look

lacoak21:
# Change Summary

## Overview
Skip checking for SPICE in the initial upstream database call. The spice files will be collected further downstream. This is pointless, a waste of processing, adds logs that clutter the stream, and caused a bug that I noticed during integration tests. For example an ULTRA l1b ancillary file v002 was uploaded and the l1b job was not kick off because the initial upstream dependency call uses the start and end time of the ancillary file. The start was 202501 and the end was 20250910. There were no SPIN files for this time range so the job got skipped. Batch starter needs to first find the upstream science files excluding spice first, then once it has the start dates of the jobs to process, it will make a second call to get all file types (it was already making this second call anyway). 

TLDR: You cant find which SPICE files to use for a job from an ancillary start and end date so skip checking spice files until you know the science start date!! This will cause skipped jobs. 

## Updated Files
   - Skip checking spice files for initial db query for upstream files. 
   - Add bool to skip checking spice

## Testing
Add test to ensure that spice is skipped when needed.


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -869,95 +873,104 @@ def get_upstream_dependency_inputs(
...
-                        "start_time": start_time,
-                        "end_time": end_time,
-                        "list_files": "True",
-                        "file_types": combined_kernel_sources,
-                        # TODO: revisit this after SIT-4
-                        # "require_coverage": "True",
-                    }
-                },
-                None,
+        if get_spice:
```
lacoak21, line 876: All I did here was add this check and indent everything.

greglucas: There is a lot of indentation going on here :) I'm wondering if this extremely long function should be broken up into individual helper functions later to try and break up the logic a bit and make it easier to follow.

greglucas (approved):

tech3371:
I will review this after I test the private auth. Sorry for the delay!

lacoak21:
> I will review this after I test the private auth. Sorry for the delay!

No problem!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Good call  - I can make a ticket for that.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -501,8 +501,17 @@ def submit_all_jobs(
...
+    # Make initial query for upstream dependency files.
+    # These dependencies will be used to determine the start dates of the jobs to
+    # submit.
+    # If we are filtering dependencies, then we do not need to get spice files because
+    # there will be a second query for upstream dependencies for each potential file
+    # To process.
+    if filter_dependencies:
+        get_spice = False
+    else:
+        get_spice = True
```
tech3371, line 515: I see. So we are querying for SPICE before submitting job?

tech3371 (approved):
I am not aware of impact of this given so many things going to this week :). But current changes looks good to me and if something comes up, we can fix as we go.

Thank you for doing this so quick!

PR #843, opened 2025-09-11
MNT: Add IMAP_DATA_ACCESS_URL to job definitions

### LINKED ISSUE #736: Confirm api calls to upload and download are working in batch starter

greglucas:
# Change Summary
closes #736 

## Overview

We need to set the `IMAP_DATA_ACCESS_URL` environment variable whenever jobs/tasks are going to be pointing at our HTTP APIs. There are two things going on here:
* dev jobs need to point at our dev urls and production at our production urls
* Anything that needs to access unreleased data or upload (batch jobs) needs to use the api-key routes. For that, we inject the SSM parameter into the container at runtime.

I tried to search for all of the places where this is used. It would be good for someone else to try and sanity check me or mention any other places they see this being used.

### DIFF
--- a/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
+++ b/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
@@ -23,6 +23,7 @@ class IalirtIngestLambda(Construct):
         vpc: ec2.Vpc,
         efs_access_point: efs.AccessPoint,
         docker_path: str = "sds_data_manager/lambda_code",
+        data_access_url: str = "",
         **kwargs,
     ) -> None:
         """IalirtIngestLambda Stack.
@@ -41,6 +42,10 @@ class IalirtIngestLambda(Construct):
             EFS access point to mount inside the Lambda function.
         docker_path : str
             Path to the Dockerfile.
+        data_access_url : str, optional
+            The data access URL to use for this job, by default the empty string.
+            You should set this to the appropriate API endpoint, e.g.
+            https://api.dev.imap-mission.com
         kwargs : dict
             Keyword arguments.
 
@@ -53,14 +58,6 @@ class IalirtIngestLambda(Construct):
 
         # Create DynamoDB Table
         self.algorithm_data_table = self.create_algorithm_dynamodb_table()
-
-        account_name = self.node.get_context("account_name")
-        # once we have the account_name, get that section out of cdk.json
-        account_config = self.node.get_context(account_name)
-        domain_name = account_config.get("domain_name", "no-domain-set")
-        # https://api.imap-mission.com
-        # https://api.dev.imap-mission.com
-        data_access_url = f"https://api.{domain_name}"
 
         # Create Lambda Function
         self.ialirt_ingest_lambda = self.create_lambda_function(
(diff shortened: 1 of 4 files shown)


sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
```diff
@@ -54,14 +59,6 @@ def __init__(
         # Create DynamoDB Table
         self.algorithm_data_table = self.create_algorithm_dynamodb_table()
 
-        account_name = self.node.get_context("account_name")
```
laspsandoval, line 57: Looks good.

laspsandoval (approved):

sds_data_manager/constructs/indexer_lambda_construct.py
```diff
@@ -231,6 +236,7 @@ def __init__(
             security_groups=[rds_security_group],
             environment={
                 "IMAP_DATA_DIR": "/tmp",  # noqa: S108
+                "IMAP_DATA_ACCESS_URL": data_access_url,
```
lacoak21, line 239: Is this needed in the indexer lambda?

lacoak21 (approved):
This looks good to me too. I just went through the constructs and stack builder. I think you are all set here.

tech3371 (approved):
same. changes looks good to me. Thank you for taking this on!

sds_data_manager/constructs/processing_construct.py
```diff
@@ -44,6 +45,25 @@ def __init__(
...
+        # Then add that API Key to SSM as a separate parameter the ECS task can look up
+        #
+        #  aws ssm put-parameter --name /imap-sdc/batch-jobs/api-key --value <the-key> \
+        #    --type SecureString
+        #
+        self.batch_api_key_param = ssm.StringParameter.from_string_parameter_name(
+            self,
+            "BatchApiKeyParam",
+            # Location in parameter store where the batch API key is stored
+            "/imap-sdc/batch-jobs/api-key",
```
tech3371: I tried to test this PR but got this error.
```
❌  SDCStack failed: ValidationError: Parameters [/imap-sdc/batch-jobs/api-key] referenced by template have types not supported by CloudFormation.
```
and I ran this accordingly
```
aws ssm put-parameter --name /imap-sdc/batch-jobs/api-key --value <the-key> --type SecureString --overwrite
``` 
but that didn't fix it. I don't know enough with the key. sorry

greglucas: I don't know on this one either. It synthesized locally for me.

We might need to switch over to secretsmanager instead.

greglucas:
@tech3371 , I switched it over to read from a secure string parameter now. Maybe that helped? I can test this tomorrow too.

sds_data_manager/constructs/indexer_lambda_construct.py
greglucas: From what I saw, the indexer code only used direct boto3 calls to download/reference anything, so I don't think so. Let me know if you saw anything in particular though.

lacoak21: Ah ok I thought this was the regular indexer and was wondering why it was needed. Ignore this

PR #848, opened 2025-09-11
I-ALiRT - api privacy and various bugfixes

laspsandoval:
# Change Summary

## Overview
Restricts some directory access; may add more restrictions later. Adds in new URL and changes the Nexus repo name based on what is being deployed.

## Updated Files
- ialirt_coverage_construct.py
- ialirt_coverage.py
- stackbuilder.py

Comments written into the code

### DIFF
--- a/sds_data_manager/constructs/ialirt_api_manager_construct.py
+++ b/sds_data_manager/constructs/ialirt_api_manager_construct.py
@@ -133,6 +133,19 @@ class IalirtApiManager(Construct):
 
         download_api.add_to_role_policy(s3_read_policy)
 
+        # Deny access to logs/ and raw_records/ for the download_api lambda role
+        data_bucket.add_to_resource_policy(
+            iam.PolicyStatement(
+                effect=iam.Effect.DENY,
+                actions=["s3:GetObject"],
+                principals=[iam.ArnPrincipal(download_api.role.role_arn)],
+                resources=[
+                    f"{data_bucket.bucket_arn}/logs/*",
+                    f"{data_bucket.bucket_arn}/raw_records/*",
+                ],
+            )
+        )
+
         # {proxy+} is used to allow for any pathParams after /ialirt-download/
         api.add_route(
             route="/ialirt-download/{proxy+}",
--- a/sds_data_manager/constructs/ialirt_coverage_construct.py
+++ b/sds_data_manager/constructs/ialirt_coverage_construct.py
@@ -37,6 +37,7 @@ class IalirtCoverageConstruct(Construct):
         """
         super().__init__(scope, construct_id, **kwargs)
 
+        # TODO: change once new code is merged in.
         account_name = self.node.get_context("account_name")
         # once we have the account_name, get that section out of cdk.json
         account_config = self.node.get_context(account_name)
--- a/sds_data_manager/constructs/ialirt_coverage_construct.py
+++ b/sds_data_manager/constructs/ialirt_coverage_construct.py
@@ -37,14 +37,25 @@ class IalirtCoverageConstruct(Construct):
         """
         super().__init__(scope, construct_id, **kwargs)
 
+        account_name = self.node.get_context("account_name")
+        # once we have the account_name, get that section out of cdk.json
+        account_config = self.node.get_context(account_name)
+        domain_name = account_config.get("domain_name", "no-domain-set")
+        # https://api.imap-mission.com
+        # https://api.dev.imap-mission.com
+        data_access_url = f"https://api.{domain_name}"
+
         # Create Lambda Function
-        ialirt_coverage_lambda = self.create_coverage_lambda(ialirt_bucket, docker_path)
+        ialirt_coverage_lambda = self.create_coverage_lambda(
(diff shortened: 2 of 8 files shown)


sds_data_manager/constructs/ialirt_api_manager_construct.py
```diff
@@ -133,6 +133,19 @@ def __init__(
 
         download_api.add_to_role_policy(s3_read_policy)
 
+        # Deny access to logs/ and raw_records/ for the download_api lambda role
```
laspsandoval, line 136: This remove access to these directories via the API.

sds_data_manager/constructs/ialirt_coverage_construct.py
```diff
@@ -37,14 +37,25 @@ def __init__(
...
+        # once we have the account_name, get that section out of cdk.json
+        account_config = self.node.get_context(account_name)
+        domain_name = account_config.get("domain_name", "no-domain-set")
+        # https://api.imap-mission.com
+        # https://api.dev.imap-mission.com
+        data_access_url = f"https://api.{domain_name}"
+
         # Create Lambda Function
-        ialirt_coverage_lambda = self.create_coverage_lambda(ialirt_bucket, docker_path)
+        ialirt_coverage_lambda = self.create_coverage_lambda(
```
laspsandoval, line 46: Pass in the data_access_url. Note that I think this will change once Greg merges.

sds_data_manager/constructs/ialirt_processing_construct.py
```diff
@@ -65,6 +68,12 @@ def __init__(
         # Add autoscaling for each container
         self.add_autoscaling()
 
+        # Determine the latest tag based on the account name
```
laspsandoval, line 66: Melissa Mantey from the POC wanted to have two images in Nexus. So this is here per her request.

sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
```diff
@@ -297,6 +297,7 @@ def lambda_handler(event, context):
 
     bucket = os.environ.get("S3_BUCKET")
     region = os.environ.get("AWS_REGION")
+    url = os.environ.get("IMAP_DATA_ACCESS_URL")
```
laspsandoval, line 300: Pass this in instead of using hard-coded string.

sds_data_manager/constructs/ialirt_coverage_construct.py
```diff
@@ -37,14 +37,26 @@ def __init__(
         """
         super().__init__(scope, construct_id, **kwargs)
 
+        # TODO: change once new code is merged in.
```
greglucas: I just merged the other code in, so you should be able to use the other pathway now.

sds_data_manager/constructs/ialirt_processing_construct.py
greglucas: Might be better to name the tag "prod" you want on production and then you can keep going with "latest" for latest. You don't really want "latest" pointing to prod since that is the default, so these should be swapped likely, but maybe better just to rename your production one to make it obvious.

sds_data_manager/constructs/ialirt_api_manager_construct.py
```diff
@@ -133,6 +133,19 @@ def __init__(
...
+        data_bucket.add_to_resource_policy(
+            iam.PolicyStatement(
+                effect=iam.Effect.DENY,
+                actions=["s3:GetObject"],
+                principals=[iam.ArnPrincipal(download_api.role.role_arn)],
```
greglucas, line 141: Do you need to disallow query too?

sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
```diff
@@ -91,13 +91,15 @@ def get_dsn(download_dir: Path):
...
-def get_latest_spice_kernels(kernels: list[str]) -> ProcessingInputCollection:
+def get_latest_spice_kernels(kernels: list[str], url: str) -> ProcessingInputCollection:
     """Query the SPICE metakernel API for latest SPICE kernel filenames.
 
     Parameters
     ----------
     kernels : list[str]
         List of SPICE kernel categories to collect.
+    url: str
+        AWS account name.
```
greglucas: URL to download the kernels from.

sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
greglucas: Do you need to pass this in / load it? It should get automatically used if it is an environment variable and loaded by imap-data-access.

i.e. I think you only need to set it on the lambda as an environment variable and then it should "just work".

greglucas (approved):

sds_data_manager/constructs/ialirt_api_manager_construct.py
laspsandoval: I'm fine with folks seeing the filenames as long as they can't download them.

PR #850, opened 2025-09-12
Revert "Initial implementation for daily job triggering"

lacoak21:
Reverts IMAP-Science-Operations-Center/sds-data-manager#787

There is a new concern about using try_to_submit_job that I didnt see before I merged in changes.


pleasant-menlo:
From the original PR: We did have one concern about using the try_submit_job just looking this over today. It looks like try_submit adds the job to a database of processing jobs, and this can fail if the job is already in that table (in an in_progress or completed state). We are not sure that we are making our jobs look unique enough such that they will always pass this check.

greglucas:
I think we should leave the current version in and make updates to it if necessary. It isn't a catastrophic failure that needs reverting IMO.

PR #851, opened 2025-09-12
Revised Schedule Job Triggering: Now submitting each job with a unique start_date

pleasant-menlo:
# Change Summary
Updated scheduled job lambda so it passes the current time as the start date of the scheduled job. Made a version of try_submit_job that takes a datetime, so multiple jobs can be submitted on the same day, and they will be recognized as unique.

The original scheduled job PR #787 

## Updated Files
- lambda_code/SDSCode/pipeline_lambdas/scheduled_job.py
- tests/lambda_endpoints/test_scheduled_job.py

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/scheduled_job.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/scheduled_job.py
@@ -13,7 +13,6 @@ from . import (
     batch_starter,
     dependency,
 )
-from .batch_starter import dependency_hash
 from .scheduled_job_config_reader import read_scheduled_job_config
 
 logger = logging.getLogger(__name__)
@@ -97,7 +96,9 @@ def try_to_submit_job(
     # release
     # The descriptor should include a hash of the serialized dependencies.
     # This makes it unique for this file and set of dependencies.
-    dep_descriptor = f"{descriptor}-{dependency_hash(serialized_dependencies)}"
+    dep_descriptor = (
+        f"{descriptor}-{batch_starter.dependency_hash(serialized_dependencies)}"
+    )
     dependency_file = DependencyFilePath.generate_from_inputs(
         instrument=instrument,
         data_level=data_level,
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/scheduled_job.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/scheduled_job.py
@@ -2,13 +2,10 @@
 
 import datetime as dt
 import logging
-from typing import Optional
 
-from imap_data_access import DependencyFilePath, ProcessingInputCollection, RepointInput
-from sqlalchemy.exc import IntegrityError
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/scheduled_job.py
```diff
@@ -45,15 +48,140 @@ def scheduled_processing_event(session, events):
     processing_input_collection = ProcessingInputCollection(*processing_inputs)
 
     for job in triggered_jobs:
-        batch_starter.try_to_submit_job(
+        try_to_submit_job(
             session,
             job,
-            "20000101",
+            dt.datetime.now(),
```
greglucas: You are right on the border with your 6:00 and the UTC / local time jumping between dates, lets try and stick with just UTC if OK.
```suggestion
   dt.datetime.now(dt.timezone.utc).strftime("%Y%m%d")
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/scheduled_job.py
```diff
@@ -45,15 +48,140 @@ def scheduled_processing_event(session, events):
...
             "v001",
             processing_input_collection.serialize(),
         )
 
 
+def try_to_submit_job(
```
greglucas: Why did you need to bring this in? Could you elaborate.

greglucas (commented):
I'm not sure I 100% follow why we need to create a new very similar `try_to_submit_job` function here?

I actually think it is a benefit in some sense to only allow one of these jobs to succeed / be in-progress on any given day so that we don't get a rogue cron-job kicking off every minute or something on accident.

I think we can address the more-than-one job per day issue later if it does become a problem IMO.

pleasant-menlo:
Hi Greg, that makes sense. Part of the motivation for allowing multiple jobs to succeed within a day was for testing. We were imagining we would test by submitting an event to the scheduled job event queue or firing the lambda, and could imagine wanted to do that more than once a day.

We only created the new try_submit_job to allow us to pass a datetime instead of the "%Y%m%d" that wouldn't allow for multiple jobs on the same day, we could imagine refactoring the try_to_submit_job to take either.

greglucas:
>We only created the new try_submit_job to allow us to pass a datetime instead of the "%Y%m%d" that wouldn't allow for multiple jobs on the same day, we could imagine refactoring the try_to_submit_job to take either.

I would still prefer to leave it alone and use what we already have. If you are testing, could you bump the version instead somehow?

greglucas (approved):
Thanks for the patience! I like this approach and think it seems reasonable. Lets readdress in a follow-up PR if this doesn't quite work for you still.

pleasant-menlo:
Thanks for your feedback. I think we ended up with a good solution.

pleasant-menlo:
@greglucas Did you mean to merge this after approving?

greglucas:
Yes, sorry. We are testing right now and I didn't want to merge during a test. I will merge shortly.

PR #853, opened 2025-09-12
Spice download s3

greglucas:
# Change Summary

## Overview

Previously we used imap-data-access to download the SPICE files. This caused issues because the file was in s3 but wasn't indexed yet and therefore wasn't "released". We can get the file directly from s3 instead, so create the function here.


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -76,7 +125,14 @@ def furnish_best_spice_file(kernel_type: str):
     kernel_filename = json.loads(metakernel_response["body"])[0]
     logger.info(f"Furnishing the latest {kernel_type} kernel: {kernel_filename}")
     # Download the latest kernel file
-    highest_version_spice_file = download(kernel_filename)
+    # Convert this into an s3 key
+    # Relative to our base directory to trim off the innitial path
```
tech3371:
```suggestion
    # Relative to our base directory to trim off the initial path
```

tech3371 (approved):

PR #858, opened 2025-09-15
FIX/ENH: Add authenticated routes to rest of API endpoints

greglucas:
We should expose all endpoints to authenticated versions. This was causing issues with the SPICE queries.


pleasant-menlo (approved):
Looks good!

lacoak21 (approved):
Looks good 👍

PR #860, opened 2025-09-15
api updates

laspsandoval:
# Change Summary

## Overview
Added necessary change for API restrictions for I-ALiRT

## Updated Files
- stackbuilder.py
   - pass in account name
   - add stable route function
- download_api.py
   - now always allows spice kernels

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
@@ -29,8 +29,9 @@ def is_released(s3_key):
     bool
         True if the file is released, False otherwise.
     """
-    if "test_data" in s3_key:
-        # Default to released for any test data files for our CI testing
+    if any(string in s3_key for string in ["test_data", "spice"]):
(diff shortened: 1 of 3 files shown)


sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
```diff
@@ -29,8 +29,9 @@ def is_released(s3_key):
     bool
         True if the file is released, False otherwise.
     """
-    if "test_data" in s3_key:
-        # Default to released for any test data files for our CI testing
+    if any(string in s3_key for string in ["test_data", "spice"]):
+        # Default to released for any test data files or
+        # SPICE kernels.
```
greglucas: Can you remove this change? Now that we have the downloads accessing the DB, this should be working again and this will allow us to make spice private if needed later on.

laspsandoval: done

greglucas (approved):

PR #868, opened 2025-09-15
MNT: Add HEAD route to the download API endpoint

### LINKED ISSUE #864: Allow HEAD requests on the APIs
Internal website team request: Can we update the API to accept HEAD requests? If it makes any difference, this would only be needed for calls to /download/. A HEAD request seems to be one way (maybe the only way?) that the javascript could see the redirect URL without requesting the whole file. The HEAD response would need to include the same CORS-friendly headers as GET.

greglucas:
# Change Summary

## Overview

The frontend team has requested a HEAD request option to the download endpoint. This adds that route in without doing any different handling on our end.

closes #864

### DIFF
--- a/sds_data_manager/constructs/api_gateway_construct.py
+++ b/sds_data_manager/constructs/api_gateway_construct.py
@@ -155,11 +155,7 @@ class ApiGateway(Construct):
             cors_preflight={
                 "allow_headers": ["*"],
                 "allow_origins": ["*"],
-                "allow_methods": [
-                    apigwv2.CorsHttpMethod.GET,
-                    apigwv2.CorsHttpMethod.POST,
-                    apigwv2.CorsHttpMethod.OPTIONS,
-                ],
+                "allow_methods": [apigwv2.CorsHttpMethod.ANY],
             },
         )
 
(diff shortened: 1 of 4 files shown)


sds_data_manager/constructs/api_gateway_construct.py
```diff
@@ -155,11 +155,7 @@ def __init__(
             cors_preflight={
                 "allow_headers": ["*"],
                 "allow_origins": ["*"],
-                "allow_methods": [
-                    apigwv2.CorsHttpMethod.GET,
-                    apigwv2.CorsHttpMethod.POST,
-                    apigwv2.CorsHttpMethod.OPTIONS,
-                ],
+                "allow_methods": [apigwv2.CorsHttpMethod.ANY],
             },
```
laspsandoval, line 159: Can you change in ialirt too?

laspsandoval (commented):
Could you update ialirt too?

sds_data_manager/constructs/api_gateway_construct.py
greglucas: Can you tell me where this needs to be updated? I think this does get propagated to ialirt as well if you are using this same construct. I did push a change for the `/ialirt-download` routes, do you have anywhere else that needs updating?

laspsandoval: I-ALiRT uses this construct so we are ok!

laspsandoval (approved):
THanks Greg!

PR #877, opened 2025-09-19
Add scope to authorizer - NOT NEEDED PRIOR TO LAUNCH

laspsandoval:
# Change Summary

## Overview
A heads up: this is not needed prior to launch, but will be needed shortly afterwards when scientists start to review their data.

Many scientists will need to view the DynamoDB for I-ALiRT, but not need (and we won't want them) to have credentials for other things, such as uploading files. In the future we also may want to separate out credentials for downloading vs uploading, for example. This PR addresses the need for different permissions, but using the same api-key.

## Updated Files
- lambda_api_key_authorizer.py
   - handles path of lambda and compares the scope of the credentials to determine if the call is authorized.
- manage_api_keys.py
   - adds the ability to create a scope in which different permissions scopes are set

### DIFF
--- a/sds_data_manager/lambda_code/authorization/lambda_api_key_authorizer.py
+++ b/sds_data_manager/lambda_code/authorization/lambda_api_key_authorizer.py
@@ -1,8 +1,12 @@
 """Authorization for API Keys within the SDS."""
 
 import json
+import logging
 
 import boto3
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
 
 # Retrieve the API keys from AWS Systems Manager Parameter Store
 # Do this outside of the lambda handler body to cache the keys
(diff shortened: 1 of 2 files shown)


greglucas:
I like this approach and I think it might help other use cases like the upload/download stuff like you mention in the description, so 👍 from me.

One thing we should also consider is how to handle the migration of existing API Keys and if we can also add a way to modify the scope of someone and add some instructions for how to do that.

laspsandoval:
> I like this approach and I think it might help other use cases like the upload/download stuff like you mention in the description, so 👍 from me.
> 
> One thing we should also consider is how to handle the migration of existing API Keys and if we can also add a way to modify the scope of someone and add some instructions for how to do that.

Agreed, I created this issue.
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/890

sds_data_manager/lambda_code/authorization/lambda_api_key_authorizer.py
```diff
@@ -19,7 +23,15 @@
 
 def lambda_handler(event, context):
     """Get the API Key from the request header and check if it is valid."""
+    logger.info("Received event: %s", json.dumps(event))
```
greglucas: This function is triggered a _lot_. Do we want to log every event? I'm torn because I think it would just be too hard to parse this in general.

greglucas (approved):

PR #879, opened 2025-09-19
Update L1D dependencies for MAG

maxinelasp:
# Change Summary
Turns out MAG L1D does not need burst mode data and the team is expecting to have normal mode only data. 

The processing code handles this case gracefully, just need to fix the triggers to move to soft for L1B. 

Since there is a strong dependency between L1B and L1C we shouldn't see any race conditions with this one.

Sorry for the very last second update!


laspsandoval (approved):
Nice!

tech3371 (approved):
I think you did already but if you haven't, can you check if these in AWS pipeline to make sure it works since soft tiggers has not been tested a lot.

PR #898, opened 2025-10-01
I-ALiRT - Pointing schedule completion

laspsandoval:
# Change Summary

## Overview
Create a construct to run the lambda that creates the pointing schedule for the ground stations.

## New Files
- ialirt_pointing_schedule_construct.py
   - Construct to run the lambda
   - Put it on a cron job
- Dockerfile.pointing
   - Dockerfile for the construct

## Updated Files
- stackbuilder.py
   - Deploy pointing schedule construct
- ialirt_pointing_schedule.py
   - Update environment

### DIFF
--- a/sds_data_manager/constructs/ialirt_pointing_schedule_construct.py
+++ b/sds_data_manager/constructs/ialirt_pointing_schedule_construct.py
@@ -1,0 +1,95 @@
+"""Cron job to created pointing schedule."""
+
+from aws_cdk import Duration, RemovalPolicy, aws_s3
+from aws_cdk import aws_events as events
+from aws_cdk import aws_events_targets as targets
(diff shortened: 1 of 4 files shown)


sds_data_manager/constructs/ialirt_pointing_schedule_construct.py
```diff
@@ -0,0 +1,95 @@
+"""Cron job to created pointing schedule."""
```
torimarbois: This is a super nit, but I think it would be better to refer to _**pointing schedules**_ as a plural thing throughout this file, rather than a singular schedule, because the nature of the job is to create multiple schedules.

torimarbois (approved):
Looks good to me!

PR #899, opened 2025-10-01
BUG fix: fix spacecraft reprocessing

lacoak21:
# Change Summary

## Overview
Spice queries were being skipped for spacecraft reprocessing jobs because they are treated as a "special case" in the submit_all_jobs function. Normally, reprocessing jobs first query upstream science jobs to determine which jobs to reprocess, so spice is excluded from the initial query and handled later downstream. Spacecraft pointing-attitude jobs have no upstream dependencies. In this case, the initial query incorrectly skipped spice and then attempted to submit the job with no dependencies.

To fix this, set filter_dependencies = False when the job is a spacecraft pointing-attitude job to make sure spice is collected in the initial dependency query. 

## Updated Files
   - filter_dependencies boolean should be false when the reprocessing job is a "spacecraft attitude-pointing" job


lacoak21:
This was tested successfully[ on dev](https://us-west-2.console.aws.amazon.com/cloudwatch/home?region=us-west-2#logsV2:log-groups/log-group/$252Faws$252Flambda$252FBatchStarterLambda/log-events/2025$252F10$252F01$252F$255B$2524LATEST$255D8db10f9d64f840b89668f6c897724c80)

tech3371 (approved):
That was so quick!!

tech3371:
now to deploy to prod, you would need to open another PR to sync dev to prod branch.

PR #900, opened 2025-10-01
Update api-key instructions and increase tier size

laspsandoval:
# Change Summary

## Overview
We ran out of space when adding names to parameter store so this increases our tier size and also updates instructions based on recent changes.

## Updated Files
- README.md
   - update instructions
- manage_api_keys.py
   - increase tier for Param Store

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,9 +128,9 @@ For example from the command line with the proper AWS credentials:
 
 ```bash
 python sds_data_manager/lambda_code/authorization/manage_api_keys.py list
-python sds_data_manager/lambda_code/authorization/manage_api_keys.py add <owner> <email>
+python sds_data_manager/lambda_code/authorization/manage_api_keys.py add <owner> <email> <scope>
 python sds_data_manager/lambda_code/authorization/manage_api_keys.py remove <key>
 AWS_PROFILE=imap-sdc-dev AWS_DEFAULT_REGION=us-west-2 \
   python sds_data_manager/lambda_code/authorization/manage_api_keys.py \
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -128,9 +128,9 @@ For example from the command line with the proper AWS credentials:
 
 ```bash
 python sds_data_manager/lambda_code/authorization/manage_api_keys.py list
-python sds_data_manager/lambda_code/authorization/manage_api_keys.py add <owner> <email>
+python sds_data_manager/lambda_code/authorization/manage_api_keys.py add <owner> <email> <scope>
 python sds_data_manager/lambda_code/authorization/manage_api_keys.py remove <key>
 AWS_PROFILE=imap-sdc-dev AWS_DEFAULT_REGION=us-west-2 \
   python sds_data_manager/lambda_code/authorization/manage_api_keys.py \
-      add "First Last" "user@example.com"
+      add "First Last" "user@example.com" "full"
```
lacoak21, line 135: It looks like the add_key function doesn't handle this parameter yet?

greglucas (changes requested):
I don't think this is the way to go here, I think it would just buy us a little bit of room is all. I think I made a bad choice initially with doing this in an SSM parameter store.

I think we should switch to a small DynamoDB table for this instead where we store all API Keys and other fields like "scope" in there and then query the DyanmoDB table within the lambda authorizer.

I can try and work on this later this week, but won't be able to do it immediately.

laspsandoval:
Closing since we are moving in a different direction.

PR #904, opened 2025-10-02
MNT/FIX: Switch API key storage from SSM Parameter to DynamoDB

### LINKED ISSUE #902: API Key lookup is running out of space in Parameter Store
The current place we store API Keys is in SSM Parameter store and then we look these up and verify them in our lambda authorizer. This has a limit of 4kB in item size, so we have run out of space to add new users.

To fix this, we should look at a different way of managing the API Keys. A small DynamoDB table seems like a reasonable approach and will deliver fast lookups still. We can also add the scopes and other necessary fields very easily that way too.

greglucas:
We are running out of space in SSM Parameter when adding many API Keys to a single parameter. Switching to a DynamoDB table allows us fast lookups still along with much larger storage capability.

closes #902

### DIFF
--- a/sds_data_manager/constructs/api_gateway_construct.py
+++ b/sds_data_manager/constructs/api_gateway_construct.py
@@ -9,14 +9,14 @@ https://ialirt.prod.imap-mission.com/ialirt-log-query
 
 from typing import Optional
 
-from aws_cdk import Duration, aws_sns
+from aws_cdk import Duration, RemovalPolicy, aws_sns
 from aws_cdk import aws_apigatewayv2 as apigwv2
 from aws_cdk import aws_apigatewayv2_authorizers as apigwv2_authorizers
 from aws_cdk import aws_apigatewayv2_integrations as apigwv2_integrations
 from aws_cdk import aws_certificatemanager as acm
 from aws_cdk import aws_cloudwatch as cloudwatch
 from aws_cdk import aws_cloudwatch_actions as cloudwatch_actions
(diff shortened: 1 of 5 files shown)


sds_data_manager/constructs/api_gateway_construct.py
```diff
@@ -159,6 +168,28 @@ def __init__(
...
+        Returns
+        -------
+        dynamodb.Table
+            The created DynamoDB table for API keys
+        """
+        return dynamodb.Table(
+            self,
+            "ApiKeysTable",
+            table_name="imap-sdc-api-keys",
+            partition_key=dynamodb.Attribute(
```
laspsandoval, line 186: Just an idea you can take or leave. Sort key by name or email?

laspsandoval (approved):
Quick question: does the documentation need to change at all? 

Nice job Greg. Thanks for working on this.

sds_data_manager/constructs/api_gateway_construct.py
greglucas: I think that is an interesting idea, but I'm not sure it gets us a lot if everything is already unique because the API Key is the main index so you can't really sort or get multiple entries of anything. Is there another use-case you have in mind?

greglucas:
> Quick question: does the documentation need to change at all?

I don't think so because all of the API Key management is handled via the scripting under the hood. So now the `manage_api_keys` script makes calls to dynamo instead of ssm is all.

PR #909, opened 2025-10-03
I-ALiRT - Db query modifications

laspsandoval:
# Change Summary

## Overview
Improve the api used for DynamoDB responses.

## New Files
- ialirt_db_query_api_formatted.py
   - The same as ialirt_db_query_api.py but with the following suggestions by Greg: 
   - Return arrays of values instead of individual items
   - A user doesn't need ttj2000ns, last_modified, met, etc. 
   - met_in_utc isn't super user friendly
   - Trim the floats to a few decimal places. 
   - make a new endpoint so you can keep the one that is currently there and deploy a new one alongside so there is no downtime for the LaTiS folks. 

## Updated Files
   - add new endpoint to construct

## Tests
test_ialirt_db_query_api_formatted.py

### DIFF
--- a/sds_data_manager/constructs/ialirt_api_manager_construct.py
+++ b/sds_data_manager/constructs/ialirt_api_manager_construct.py
@@ -232,7 +232,7 @@ class IalirtApiManager(Construct):
         )
 
         # Grant the lambda function read/write permissions on the DynamoDB table.
-        algorithm_table.grant_read_write_data(ialirt_db_query_handler)
+        algorithm_table.grant_read_data(ialirt_db_query_handler)
 
         add_stable_route(
             api,
@@ -259,11 +259,11 @@ class IalirtApiManager(Construct):
         )
 
         # Grant the lambda function read/write permissions on the DynamoDB table.
-        algorithm_table.grant_read_write_data(ialirt_db_query_formatted_handler)
+        algorithm_table.grant_read_data(ialirt_db_query_formatted_handler)
 
         add_stable_route(
             api,
-            "/ialirt-db",
+            "/space-weather",
             "GET",
             ialirt_db_query_formatted_handler,
             restricted_route_prefixes,
--- a/sds_data_manager/constructs/ialirt_api_manager_construct.py
+++ b/sds_data_manager/constructs/ialirt_api_manager_construct.py
@@ -241,3 +241,30 @@ class IalirtApiManager(Construct):
             ialirt_db_query_handler,
             restricted_route_prefixes,
         )
+
+        ialirt_db_query_formatted_handler = lambda_.Function(
+            self,
+            "IAlirtDbQueryApiFormattedHandler",
+            function_name="ialirt-db-query-formatted-handler",
+            code=code,
+            handler="IAlirtCode.ialirt_db_query_api_formatted.lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_12,
+            timeout=cdk.Duration.minutes(1),
+            memory_size=1000,
+            environment={
+                "ALGORITHM_TABLE": algorithm_table.table_name,
+                "REGION": env.region,
+            },
+            layers=layers,
+        )
+
+        # Grant the lambda function read/write permissions on the DynamoDB table.
+        algorithm_table.grant_read_write_data(ialirt_db_query_formatted_handler)
+
+        add_stable_route(
+            api,
+            "/ialirt-db",
+            "GET",
+            ialirt_db_query_formatted_handler,
+            restricted_route_prefixes,
+        )
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
@@ -69,7 +69,7 @@ def lambda_handler(event, context):  # noqa: PLR0912, PLR0915
 
     Example of result:
     -----------------
-    result = {'he_omni_high_en': [0, None],
+    result = {'he_omni_high_en': [0, "null"],
     'B_GSE': [[-6.382, -1.353, -5.045],
     [-2.058, 3.792, -3.989]],
     'time_tag_utc': ['2025-10-02T07:07:13', '2025-10-02T07:07:17'], ...}
@@ -92,8 +92,8 @@ def lambda_handler(event, context):  # noqa: PLR0912, PLR0915
     query_kwargs = {"KeyConditionExpression": key_expr}
 
     allowed_params = {
-        "time_start",
-        "time_end",
+        "met_start",
+        "met_end",
         "utc_start",
         "utc_end",
         "last_modified_start",
@@ -109,7 +109,7 @@ def lambda_handler(event, context):  # noqa: PLR0912, PLR0915
             ),
         }
 
-    time_prefixes = {"time", "utc", "last_modified"}
+    time_prefixes = {"met", "utc", "last_modified"}
     used_time_prefixes = {
         param.split("_start")[0].split("_end")[0]
         for param in params
@@ -120,20 +120,17 @@ def lambda_handler(event, context):  # noqa: PLR0912, PLR0915
         return {
             "statusCode": 400,
             "body": json.dumps(
-                {
-                    "message": "Cannot query multiple time keys "
-                    "(time, utc, last_modified)"
-                }
+                {"message": "Cannot query multiple time keys (met, utc, last_modified)"}
             ),
         }
 
     if (
-        ("time_start" in params and "time_end" in params)
+        ("met_start" in params and "met_end" in params)
         or ("utc_start" in params and "utc_end" in params)
         or ("last_modified_start" in params and "last_modified_end" in params)
     ):
-        if "time_start" in params:
-            params_key = "time"
+        if "met_start" in params:
+            params_key = "met"
             time_key = "met"
         elif "utc_start" in params:
             params_key = "utc"
@@ -144,12 +141,12 @@ def lambda_handler(event, context):  # noqa: PLR0912, PLR0915
 
         start_value = (
             int(params[f"{params_key}_start"])
-            if params_key == "time"
+            if params_key == "met"
             else params[f"{params_key}_start"]
         )
         end_value = (
             int(params[f"{params_key}_end"])
-            if params_key == "time"
+            if params_key == "met"
             else params[f"{params_key}_end"]
         )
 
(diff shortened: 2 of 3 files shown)


sds_data_manager/constructs/ialirt_api_manager_construct.py
```diff
@@ -213,3 +213,30 @@ def __init__(
...
+            memory_size=1000,
+            environment={
+                "ALGORITHM_TABLE": algorithm_table.table_name,
+                "REGION": env.region,
+            },
+            layers=layers,
+        )
+
+        # Grant the lambda function read/write permissions on the DynamoDB table.
+        algorithm_table.grant_read_write_data(ialirt_db_query_formatted_handler)
```
greglucas: This lambda should only need to read

sds_data_manager/constructs/ialirt_api_manager_construct.py
```diff
@@ -213,3 +213,30 @@ def __init__(
...
+
+        add_stable_route(
+            api,
+            "/ialirt-db",
```
greglucas: Don't use "db" in an endpoint name, that is an implementation detail that doesn't need to be leaked out to the public. You already have `ialirt` in the subdomain, so I don't think you need that here either. Try and think of what a descriptive request would look like here if you were a naive user just wanting to get space weather data I'd say. Maybe just the generic `space-weather` or something similar? ChatGPT might have some good suggestions.
```suggestion
            "/space-weather",
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
```diff
@@ -0,0 +1,196 @@
...
+        return {
+            "statusCode": 400,
+            "body": json.dumps(
+                {"message": "Cannot query by end time without start time"}
+            ),
+        }
+
+    query_kwargs["KeyConditionExpression"] = key_expr
+
+    response = table.query(**query_kwargs)
```
greglucas, line 220: If you are running into limits with the query size within DynamoDB itself, we should use a paginator.
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb/paginator/Query.html

Alternatively / in addition, it might be smart to put some limits on how large of a query a user can do so you aren't doing full table scans all the time.

sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
```diff
@@ -0,0 +1,196 @@
...
+
+import boto3
+from boto3.dynamodb.conditions import Key
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+
+def process_item_types(item: dict) -> dict:
+    """Convert Decimal values to int/float for known fields.
```
greglucas, line 17: Add a note about truncating to 3 decimal places to reduce response size.

sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
```diff
@@ -0,0 +1,196 @@
...
+    query_kwargs = {"KeyConditionExpression": key_expr}
+
+    allowed_params = {
+        "met_start",
+        "met_end",
+        "met_in_utc_start",
+        "met_in_utc_end",
+        "last_modified_start",
+        "last_modified_end",
+    }
```
greglucas, line 102: Remove MET, that is an implementation detail and we don't necessarily want to leak that to the public. A public person only wants to enter with "human time" in mind. Then behind the scenes you can do all of the conversions necessary to make it into a MET if that is what is in the DB, but a user doesn't need to know the mapping of the DB, the API should be a human interface not a DB interface.
```suggestion
    allowed_params = {
        "start_time",
        "end_time",
        "last_modified_start",
        "last_modified_end",
    }
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
```diff
@@ -0,0 +1,196 @@
...
+            result[key] = int(value) if value % 1 == 0 else round(float(value), 3)
+
+        else:
+            result[key] = value
+
+    return result
+
+
+def lambda_handler(event, context):  # noqa: PLR0912, PLR0915
+    """Create metadata and add it to the database.
```
greglucas: Add it to the database? I thought this was exposing the API to the users, so only a reader.

I'm still unclear what the formatted response will look like, could you add a comment here indicating a quick example of what that would look like so that people will know what the expected response format is. Whether it is per-item, an array of entries and how those are formatted etc.

tests/lambda_endpoints/test_ialirt_db_query_api_formatted.py
```diff
@@ -0,0 +1,274 @@
...
+
+    assert processed_items == [
+        {
+            "apid": 478,
+            "met": 123456789,
+            "ttj2000ns": 123456789000000,
+            "mag_B_GSE": [0.0, 0.1, 0.2],
+            "mag_B_magnitude": 0.22,
+            "met_in_utc": "2025-06-20T08:00:00",
+        }
```
greglucas, line 286: Is this what is returned to the user? Could you strip out and rename some of these fields too.

See here, where the public interface is simplified. They give `bx_gsm` as their label, and a generic `bt`. `mag_B_magnitude` is a bit confusing because mag could mean magnitude.

But note how they display an array of values so the labels aren't duplicated multiple times and it looks like they avoid the array nesting and return individual components at the top level to flatten things out. I think we should try and keep it simple like this.

https://services.swpc.noaa.gov/products/solar-wind/mag-2-hour.json

greglucas (commented):
I think there is still a bit more work to be done here to make it a human interface rather than an interface to the machine/database definition, I've tried to add some comments regarding what I mean, but happy to hop on a call if that would help too.

sds_data_manager/constructs/ialirt_api_manager_construct.py
laspsandoval: Oh wow. Good catch.

sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
laspsandoval: Wrong docstring. Corrected.

Done.

tests/lambda_endpoints/test_ialirt_db_query_api_formatted.py
laspsandoval: Ok. I'll aim for something more like this. I would like to keep the instrument prefixes. I think that if we document each parameter well on the website the prefix will be useful to at least some users.

greglucas: I somewhat disagree with that because we shouldn't need to know what instrument took a measurement, as an end-user I only care about the temperature, the magnetic field etc... I don't care that it was swapi, codice. That may be useful for detailed/advanced users, but we should be aiming for the scientific community that doesn't necessarily want to know the underlying details.

sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
laspsandoval: I'm looking at the documentation and I think lambda functions is invoked synchronously (like through API Gateway) the entire JSON response payload cannot exceed 6 MB.

https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html?utm_source=chatgpt.com

The WebTeam implemented pagination on their end. It looks like 1MB limit is already implemented for DynamoDB so I don't think that I need additional limits.

I'm adding LastEvaluated to the response so if pagination is required the user will have that information.

tests/lambda_endpoints/test_ialirt_db_query_api_formatted.py
laspsandoval: I think the scientific community will absolutely want to know which instrument we are using. And I think it will make it simpler for scientists to select a single instrument and look at only that data.

greglucas: If you go up a directory from that link I sent, there are no instruments listed. It is "mag" and "plasma" for magnetic field values and in-situ plasma values.
https://services.swpc.noaa.gov/products/solar-wind/

Those are measured by different spacecraft and different instruments, but the SWPC API doesn't expose that information, you have to go to the individual instrument/spacecraft sites to get that level of information if you want it.

If you want to have a separate API endpoint or options to filter by a specific instrument then that seems fine, but I still don't think it is generally desired in API design to expose extra details to the end users. It could be additional metadata included in the response to describe the data products with an "info" field or something like that.

laspsandoval: I see what you are saying. Changed.

sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
```diff
@@ -0,0 +1,231 @@
...
+        The JSON formatted document with the data required for the
+        lambda function to process
+    context : LambdaContext
+        This object provides methods and properties that provide
+        information about the invocation, function,
+        and runtime environment.
+
+    Example of result:
+    -----------------
+    result = {'he_omni_high_en': [0, None],
```
greglucas:
```suggestion
    result = {'he_omni_high_en': [0, "null"],
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
```diff
@@ -0,0 +1,231 @@
...
+        }
+
+    if (
+        ("time_start" in params and "time_end" in params)
+        or ("utc_start" in params and "utc_end" in params)
+        or ("last_modified_start" in params and "last_modified_end" in params)
+    ):
+        if "time_start" in params:
+            params_key = "time"
+            time_key = "met"
```
greglucas: I would leave this as what you had before if you do want to allow met queries. "met_start". Because time_start to me means the human start time.

sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api_formatted.py
```diff
@@ -0,0 +1,231 @@
...
+            if params_key == "time"
+            else params[f"{params_key}_start"]
+        )
+        end_value = (
+            int(params[f"{params_key}_end"])
+            if params_key == "time"
+            else params[f"{params_key}_end"]
+        )
+
+        key_expr &= Key(time_key).between(start_value, end_value)
```
greglucas, line 170: Rather than special casing this between, can you do the less than key expression for the end value?

```
if end in params:
    key_expr &= Key(time_key).lte(end_value)

if start in params
    key_expr &= Key(time_key).gte(start_value)
else:
    # limit the query automatically by calculating the range for a user and/or raising if the range is too large?
    start = end_value - 1-day

greglucas (commented):
Minor suggestion about the time ranges and how we are calculating those and whether we should default to something and/or limit that and provide a nice message to the user that their selected range is too large.

greglucas (approved):

PR #910, opened 2025-10-03
Change how start date is calculated for repoint table triggers

### LINKED ISSUE #903: Update repoint start date in batch starter
### Requirements:

After reviewing recent failures of the pointing attitude job with flight data, I believe that the start date for the pointing kernel job needs to be updated. Currently, in the batch starter code, the start date is just end date minus one day.
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/dev/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py#L703

I think that we need to get the start date by querying the DB for repoint files and use the end date from the previous repoint file as the start date.

tmplummer:
# Change Summary

## Overview
This updates how the start_date is calculated for repoint table files in batch starter.
I identified an issue with the current implementation where we would miss a day when generating the pointing attitude kernel. This implementation is intended to make sure that the current and previous attitude history kernels are included in the dependencies for the job that gets triggered.

## Updated Files
   - New function `calculate_repoint_table_date_range` that queries the DB and gets the end date from the previous repoint file as the start date for the new file.
   - Directly test that the correct start/end date are returned from `determine_date_range` for repoint files
   - Update expected job parameters to agree with latest change.

Closes: #903

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -681,6 +681,56 @@ def calculate_pointing_date_range(session, pointing_id):
     return start_date, end_date
 
 
+def calculate_repoint_table_date_range(session, file_obj):
+    """Calculate the date range for a repoint-table.
+
+    The end date can easily be gotten from the filename. In order to determine
+    the start date, we query the database and use the end date from the previous
+    repoint table.
+
+    NOTE:
+    Repoint file is used to kick off the pointing_attitude job only.
+    This date range is used to query attitude kernel file(s). If
+    other jobs become dependent on triggering off of the repoint file,
+    please revisit this logic.
+
+
+    Parameters
+    ----------
+    session : sqlalchemy.orm.Session
+        Database session.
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -681,6 +681,56 @@ def calculate_pointing_date_range(session, pointing_id):
...
 
 
+def calculate_repoint_table_date_range(session, file_obj):
+    """Calculate the date range for a repoint-table.
+
+    The end date can easily be gotten from the filename. In order to determine
+    the start date, we query the database and use the end date from the previous
+    repoint table.
+
+    NOTE:
```
greglucas:
```suggestion
    Notes
    -----
```

greglucas (approved):

greglucas:
I'm not sure if this is relevant here, but I've noticed we are getting 2 repoint files basically every time. One for the "new pointing" and one `v2` update closing the previous pointing. So we get two simultaneous files which may kick off two jobs for you is my guess.

PR #912, opened 2025-10-06
I-ALiRT - add realtime construct

laspsandoval:
# Change Summary

## Overview
Add realtime construct so we can use the api.

## Updated Files
   - add realtime construct so we can use the api.

### DIFF
--- a/sds_data_manager/constructs/ialirt_api_manager_construct.py
+++ b/sds_data_manager/constructs/ialirt_api_manager_construct.py
@@ -125,6 +125,33 @@ class IalirtApiManager(Construct):
             lambda_function=packets_query_api_lambda,
         )
 
+        # realtime query API lambda
+        realtime_query_api_lambda = lambda_.Function(
+            self,
+            id="IAlirtCodeRealtimeQueryAPILambda",
+            function_name="ialirt-realtime-query-api-handler",
+            code=code,
+            handler="IAlirtCode.ialirt_realtime_query_api.lambda_handler",
+            runtime=lambda_.Runtime.PYTHON_3_12,
+            timeout=cdk.Duration.minutes(1),
+            memory_size=1000,
+            allow_public_subnet=True,
+            vpc=vpc,
+            environment={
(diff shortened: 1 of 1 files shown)


sds_data_manager/constructs/ialirt_api_manager_construct.py
```diff
@@ -125,6 +125,33 @@ def __init__(
...
+                "S3_BUCKET": data_bucket.bucket_name,
+                "REGION": env.region,
+            },
+            layers=layers,
+        )
+
+        realtime_query_api_lambda.add_to_role_policy(s3_read_policy)
+
+        api.add_route(
+            route="/ialirt-realtime-query",
```
greglucas, line 151: ialirt is in the subdomain so you probably don't need it here.
```suggestion
            route="/realtime-query",
```

greglucas (commented):
What is `realtime-query`, what does it do? Is there a better description for this endpoint that would tell someone what to expect from it. Is it pertaining to data, ground stations, etc.

sds_data_manager/constructs/ialirt_api_manager_construct.py
laspsandoval: Every other api route starts with ialirt-. I see your point that it is redundant. However, I do think it will be more confusing if one of them doesn't have that prefix and the others do.

laspsandoval:
> What is `realtime-query`, what does it do? Is there a better description for this endpoint that would tell someone what to expect from it. Is it pertaining to data, ground stations, etc.

This is just a tool that I created for the Webteam so that they have the latest realtime filename. The realtime file contains tcp connection information and packet creation information that can be plotted for the ground stations so that they are able to see diagnostic information.

sds_data_manager/constructs/ialirt_api_manager_construct.py
lacoak21: I agree with keeping "ialirt" for continuity

lacoak21 (approved):

PR #915, opened 2025-10-06
Packet download cron based

greglucas:
The event based approach was too simplistic and didn't account for
multiple files arriving. For now, people want their data sooner, so
run the data retrieval every 6-hours and create packet files as
needed during those periods only if there was data received on the
ground. This could lead to more version bumps than necessary, but
tries to strike a reasonable position of not having to wait too
long to get packets either.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
@@ -2,12 +2,14 @@
 
 import logging
 import os
+from datetime import datetime, timedelta, timezone
 from pathlib import Path
 
 import boto3
 import imap_data_access
 from imap_data_access import SPICEFilePath, webpoda
 
+S3_BUCKET = os.environ["S3_BUCKET"]
 S3_CLIENT = boto3.client("s3")
 # We need to access the webpoda-api-key
 SECRETS_MANAGER = boto3.client("secretsmanager")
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
```diff
@@ -2,139 +2,131 @@
...
 S3_CLIENT = boto3.client("s3")
 # We need to access the webpoda-api-key
 SECRETS_MANAGER = boto3.client("secretsmanager")
 # We need to access the IMAP API key from SSM
 SSM_CLIENT = boto3.client("ssm")
 
 # Logger setup
-# Set default logging level to INFO, to also capture INFO for the underlying downloaders
-logging.basicConfig(level=logging.INFO)
+if len(logging.getLogger().handlers) > 0:
```
lacoak21, line 20: Why did you have to add this?

lacoak21 (approved):
LGTM!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
greglucas: The INFO logs weren't getting logged for some reason. I think the root logger within AWS might be getting set up before this call was taking place previously or something like that. So, these few lines get the root logging handler and update that if it is present already (this case), or if running locally we fall back to the other version. This was a recommendation from some AWS stack overflow thing I read. It did give me the logs I was looking for in dev.

PR #917, opened 2025-10-07
release update to repoint table start date to prod

tmplummer:
# Change Summary

Release change to how the repoint table start date is calculated to prod.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -681,6 +681,56 @@ def calculate_pointing_date_range(session, pointing_id):
     return start_date, end_date
 
 
+def calculate_repoint_table_date_range(session, file_obj):
+    """Calculate the date range for a repoint-table.
+
+    The end date can easily be gotten from the filename. In order to determine
+    the start date, we query the database and use the end date from the previous
+    repoint table.
+
+    NOTE:
+    Repoint file is used to kick off the pointing_attitude job only.
+    This date range is used to query attitude kernel file(s). If
+    other jobs become dependent on triggering off of the repoint file,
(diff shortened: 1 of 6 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -681,6 +681,57 @@ def calculate_pointing_date_range(session, pointing_id):
...
+    tuple
+        A tuple containing the start date and end date in the format YYYYMMDD.
+    """
+    # Query the pointing table to find the pointing information.
+    end_date = file_obj.spice_metadata["end_date"]
+    previous_entries = (
+        session.query(models.RepointFiles)
+        .filter(models.RepointFiles.end_date <= end_date)
+        .order_by(models.RepointFiles.end_date, models.RepointFiles.version)
+        .all()
```
greglucas, line 717: Can you descend the ordering and then get the first entry instead? Then update the logic below.
```suggestion
        .order_by(models.RepointFiles.end_date.desc(), models.RepointFiles.version.desc())
        .first()
```

greglucas (approved):
Minor suggestion to improve the query and only grab one entry instead of all of them, but that shouldn't hold this up.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
greglucas: Just realized that this is already in "dev". Was reviewing something that already went in 😂

tmplummer: My sqlalchemy is pretty weak. It looks like I should have ordered descending and then used `.limit(2)` to just get the first two results. I can write a ticket for that change.

PR #921, opened 2025-10-07
I-ALiRT - Rsync alarm

### LINKED ISSUE #844: I-ALiRT Rsync Alarm
### Summary

Using logs, make rsync alarm

Awesome, thanks! The line would contain command failed: rsync
nd if you could put me on the email list for these alarms that would be great

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
The POC would like to know if any of their logs contain an rsync failure. This PR creates a lambda and construct that searches new logs for failtures.

## New Files
- ialirt_rsync_alarm_construct.py
   - Creates sns topics that kicks off the lambda upon arrival of a new log
   - Creates an alarm based on the metric created in the lambda
- ialirt_rsync_alarm.py
   - Creates the metric based on contents of log file
- stackbuilder.py
   - Call new construct

## Testing
test_ialirt_rsync_alarm.py

Deployed and confirmed functionality


sds_data_manager/constructs/ialirt_rsync_alarm_construct.py
```diff
@@ -0,0 +1,177 @@
...
+        ialirt_rsync_lambda = self.create_realtime_lambda(ialirt_bucket, code)
+        # Create Event Rule
+        self.create_event_rule(ialirt_bucket, ialirt_rsync_lambda)
+        # Parameter store lookup.
+        # Note: this must be run once for each account:
+        # aws ssm put-parameter --name /imap/ialirt/rsync_alarm_email
+        # --value ialirt@example.com --type String --overwrite
+        rsync_alarm_email = ssm.StringParameter.value_for_string_parameter(
+            self, "/imap/ialirt/rsync_alarm_email"
+        )
```
greglucas: We already have `/imap/ialirt/alarm_email` in a separate location, do we need another one here? Can you just send a different message to the same "alarm e-mail" instead. These ssm parameters are easy to lose track of in new accounts, so it'd be nice to consolidate where we can.

sds_data_manager/constructs/ialirt_rsync_alarm_construct.py
```diff
@@ -0,0 +1,177 @@
...
+        self.setup_monitoring(rsync_alarm_email)
+
+    def create_realtime_lambda(
```
greglucas: What does "realtime lambda" mean here? Do you mean rsync_log_reader_lamber? It would be nice to expand the name a bit here.

sds_data_manager/constructs/ialirt_rsync_alarm_construct.py
```diff
@@ -0,0 +1,177 @@
...
+        )
+
+        lambda_role.add_to_policy(
+            iam.PolicyStatement(
+                actions=["cloudwatch:PutMetricData"],
+                resources=["*"],
+            )
+        )
+
+        s3_read_write_policy = iam.PolicyStatement(
```
greglucas: Only a "read" policy it looks like?

sds_data_manager/lambda_code/IAlirtCode/ialirt_rsync_alarm.py
```diff
@@ -0,0 +1,101 @@
...
+        True if any log file contains the string 'command failed: rsync';
+        False otherwise.
+    """
+    obj = s3_client.get_object(Bucket=bucket, Key=key)
+    body = obj["Body"]
+    for line in body.iter_lines():
+        if b"command failed: rsync" in line:
+            logger.warning(f"Found rsync failure in {key}")
+            return True
+    return False
```
greglucas: I don't even think you need to iterate here. I think you can just say `if b"command failed: rsync" in obj["Body"]`.

greglucas (commented):
This looks very similar to the other alarm/metric we already have, so I'm wondering if we can re-use any of that infrastructure and publish to the same "alarm" topic and just send a different message or something if that is needed.

The other thing I'm a bit unclear about here is why we need the metric/alarm at all. Could you just have the lambda triggered on every log file arrival in s3, that lambda would read the contents of the s3 object like you already have and then send a message to the topic if it found the rsync error. Then you can avoid the polling/cloudwatch alarm metrics stuff I think.

s3-log-arrival -> lambda-to-read-logs -> write-to-topic-if-error

sds_data_manager/constructs/ialirt_rsync_alarm_construct.py
laspsandoval: Typo. It is now updated.

laspsandoval:
> This looks very similar to the other alarm/metric we already have, so I'm wondering if we can re-use any of that infrastructure and publish to the same "alarm" topic and just send a different message or something if that is needed.
> 
> The other thing I'm a bit unclear about here is why we need the metric/alarm at all. Could you just have the lambda triggered on every log file arrival in s3, that lambda would read the contents of the s3 object like you already have and then send a message to the topic if it found the rsync error. Then you can avoid the polling/cloudwatch alarm metrics stuff I think.
> 
> s3-log-arrival -> lambda-to-read-logs -> write-to-topic-if-error

Good ideas. Check it out now.

sds_data_manager/lambda_code/IAlirtCode/ialirt_rsync_alarm.py
```diff
@@ -0,0 +1,99 @@
...
+
+    bucket = event["detail"]["bucket"]["name"]
+    region = event["region"]
+    key = event["detail"]["object"]["key"]
+
+    s3_client = boto3.client(
+        "s3",
+        region_name=region,
+        config=botocore.client.Config(signature_version="s3v4"),
+    )
```
greglucas, line 92: You can move this up if you want to cache the boto3 initialization work to speed up cold-starts. Probably not a huge issue with this lambda though.

greglucas (approved):
Nice, I like this approach and think it ties your alarming together nicely.

sds_data_manager/lambda_code/IAlirtCode/ialirt_rsync_alarm.py
laspsandoval: Nope. Not an issue. Good for other lambdas, though.

PR #931, opened 2025-10-14
Fix to versioning logic in batch starter

lacoak21:
# Change Summary

## Overview
Tim noticed that pointing-attitude jobs had version gaps. This happened because the function determine_job_version by default used the max version found in the processing job table (bumped by one) as the version for the job. In some cases, although previous jobs have succeeded, they may not produce any files, thus causing version gaps. 

In the case of the pointing-attitude job, there is no way to determine the kernel name that will be produced so we should stick to checking the processing job table for the max version. Tim is going to update the code to fail if no kernels are produced which will solve the version gap issue. 

There is still a logic gap in the determine_job_version function for science jobs. By default we should use the max version from the science table not the processing job table UNLESS the descriptor is "all" because the files produced will not have that as the descriptor. In that case we should use the max version from the processing job table. This will fix most of the version gaps for science files as well. 

*** NOTE *** jobs with descriptor == "all" may still have version gaps if the job "succeeds" but does not produce any files. 

## Updated Files
   - Use max version from science files table and not the processing job table.


greglucas (approved):
This seems good to me at first glance. It would be nice to add some tests of what you expect to happen with the various situations with the processing jobs table and science files table having skipped/out-of-sync versions in these cases.

lacoak21:
> This seems good to me at first glance. It would be nice to add some tests of what you expect to happen with the various situations with the processing jobs table and science files table having skipped/out-of-sync versions in these cases.

Sounds good - Ill do some tests on dev later today or tomorrow 👍

greglucas:
> Sounds good - Ill do some tests on dev later today or tomorrow 👍

I was actually wondering if you can do this with the test database by inserting jobs and science files at different version levels to make assertions against.

lacoak21:
> > Sounds good - Ill do some tests on dev later today or tomorrow 👍
> 
> I was actually wondering if you can do this with the test database by inserting jobs and science files at different version levels to make assertions against.

Oh yes definitely.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -180,43 +180,52 @@ def filter_conditions(table):
...
         .order_by(models.ProcessingJob.version.desc())
         .first()
     )
     if max_version_record:
-        max_version = max_version_record.version
-        # If there is a job already in progress, determine whether the current job
-        # is a duplicate of the in-progress job by checking the dependency file hash.
-        # If the hashes are different, then we know the dependencies have changed and
-        # we should bump the version number and continue with processing.
+        max_version_proc = max_version_record.version
```
tech3371: Can you rename this? I don't know what `proc` means. I think it means processing_table_max_version, is that right?
```suggestion
        max_version_proc = max_version_record.version
```

tech3371 (approved):
Minor change requests. Otherwise, looks good to me

PR #932, opened 2025-10-14
Correctly get the coverage range of pointing attitude files

### LINKED ISSUE #655: Use SPICE to get exact coverage for pointing attitude kernel when indexing
### Algorithm Description:

Update how `file_coverage` start and end date/times are populated when indexing the Pointing attitude kernel.
The code should be updated to get the exact date/times using spice just like when getting coverage for a s/c attitude kernel.

Current implementation is located near sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py (line 243).

### Related Issues/PRs:
#633

tmplummer:
# Change Summary

## Overview
Update the code that gets CK kernel coverage to work for the DPS kernel.

## Updated Files
   - Get exact coverage of DPS kernel by using same `get_coverage_dictionary` function as is used for other kernels
   - Get the object covered by CK using spiceypy
   - Raise ValueError when CK does not cover exactly one object
   - Add CK specific test coverage for `get_coverage_dictionary` function.

Closes: #655


greglucas (approved):
Is this basically saying "don't special case this anymore" and the important change is removing the pointing_attitude elif block?

tmplummer:
@greglucas 
> Is this basically saying "don't special case this anymore" and the important change is removing the pointing_attitude elif block?

Sorry, I could have described this better in the overview. When this was implemented, I think @tech3371 tried to replicate what was in the `get_coverage_dictionary` function for extracting the coverage of a CK. The problem was that that code was hard-coded to get the coverage of the IMAP_SPACECRAFT object. I would say that the main change is that I made the `get_coverage_dictionary` general for any CK that provides coverage for a single object by getting that object directly from the CK using the `spiceypy.ckobj` function. This change removes the requirement to special case the `pointing_attitude` kernels. It also gets the correct start and end time for the kernel.

The reason that I was reminded that this ticket existed and needed to be addressed was that I used my coverage plot to plot the coverage of the pointing_attitude files and noticed that the coverage in the DB lines up with day boundaries rather than pointing boundaries.

Figure 1(50)

PR #933, opened 2025-10-15
REL: Add workflow_dispatch trigger for deploying to production

greglucas:
This automates the release process to deploy to production from any given state and adds a tag with the timestamp to the repository indicating when something was previously deployed for easy rollbacks if we need to revert to a previous state.

The new workflow to release to production will be to go to the workflow_dispatch under actions and trigger this and choose what you want to deploy to production. Only people with repository write permissions can deploy.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -4,33 +4,73 @@ on:
   push:
     branches:
       - dev
-      - prod
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: "Environment to deploy to"
+        required: true
+        type: choice
+        options:
+          - dev
+          - prod
+        default: prod
+      ref:
+        description: "Optional git ref (commit SHA, branch, or tag) to deploy (for rollback)"
+        required: false
+        default: ""
+
+permissions:
+  id-token: write
+  contents: write  # needed for tagging and release creation
+  actions: read
 
 jobs:
   cdk-deploy:
     runs-on: ubuntu-latest
-    # These permissions are needed to interact with GitHub's OIDC Token endpoint.
-    permissions:
-      id-token: write
-      contents: read
 
     steps:
-      - name: Set deployment account and role
+      # ----------------------------------------------------
+      # 🧭 Determine environment + AWS role + tagging behavior
+      # ----------------------------------------------------
+      - name: Determine deployment target
         id: set_account
         run: |
-          if [[ "${GITHUB_REF##*/}" == "dev" ]]; then
-            echo "account_name=dev" >> $GITHUB_ENV
-            echo "role_arn=arn:aws:iam::449431850278:role/GitHubDeploy" >> $GITHUB_ENV
-          elif [[ "${GITHUB_REF##*/}" == "prod" ]]; then
-            echo "account_name=prod" >> $GITHUB_ENV
-            echo "role_arn=arn:aws:iam::593025701104:role/GitHubDeploy" >> $GITHUB_ENV
+          if [[ "${GITHUB_EVENT_NAME}" == "push" ]]; then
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -4,33 +4,73 @@ on:
...
+      - name: Determine deployment target
         id: set_account
         run: |
-          if [[ "${GITHUB_REF##*/}" == "dev" ]]; then
-            echo "account_name=dev" >> $GITHUB_ENV
-            echo "role_arn=arn:aws:iam::449431850278:role/GitHubDeploy" >> $GITHUB_ENV
-          elif [[ "${GITHUB_REF##*/}" == "prod" ]]; then
-            echo "account_name=prod" >> $GITHUB_ENV
-            echo "role_arn=arn:aws:iam::593025701104:role/GitHubDeploy" >> $GITHUB_ENV
+          if [[ "${GITHUB_EVENT_NAME}" == "push" ]]; then
```
tech3371, line 38: should we deploy to dev on merge too?

.github/workflows/deploy.yml
```diff
@@ -4,33 +4,73 @@ on:
...
+            elif [[ "${{ github.event.inputs.environment }}" == "prod" ]]; then
+              echo "account_name=prod" >> $GITHUB_ENV
+              echo "role_arn=arn:aws:iam::593025701104:role/GitHubDeploy" >> $GITHUB_ENV
+              echo "should_tag=true" >> $GITHUB_ENV
+            else
+              echo "Invalid environment specified." && exit 1
+            fi
           else
-            echo "Branch not configured for deployment." && exit 1
+            echo "Unsupported trigger type" && exit 1
```
tech3371, line 59: can we say something like only dev and prod is supported?

tech3371 (approved):
Nice!

.github/workflows/deploy.yml
```diff
@@ -4,33 +4,73 @@ on:
...
+          elif [[ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" ]]; then
+            if [[ "${{ github.event.inputs.environment }}" == "dev" ]]; then
+              echo "account_name=dev" >> $GITHUB_ENV
+              echo "role_arn=arn:aws:iam::449431850278:role/GitHubDeploy" >> $GITHUB_ENV
+              echo "should_tag=false" >> $GITHUB_ENV
+            elif [[ "${{ github.event.inputs.environment }}" == "prod" ]]; then
+              echo "account_name=prod" >> $GITHUB_ENV
+              echo "role_arn=arn:aws:iam::593025701104:role/GitHubDeploy" >> $GITHUB_ENV
+              echo "should_tag=true" >> $GITHUB_ENV
+            else
```
tech3371, line 55: I am wondering if we can specify a specific forked working branch too for development and testing. That would be a nice to have.

.github/workflows/deploy.yml
greglucas: The push is whenever something is merged to dev.

.github/workflows/deploy.yml
greglucas: That is already the case and won't get to this spot unless the workflow is changed. I don't think it is worth keeping a comment in sync with the actual triggers.

.github/workflows/deploy.yml
greglucas: No, that would be somewhat dangerous in my opinion to open this up to forks. If you have permissions to push a branch to this repository then you can deploy from that specific working branch to dev.

tech3371: Yeah, I was thinking too but worth asking!

PR #934, opened 2025-10-15
ENH: Add repoint table query endpoint

greglucas:
# Change Summary

## Overview

This adds an API endpoint to query for repointing files. It is really similar to the spin table, so I tried to keep the logic the same and re-use the lambda. Unfortunately there is a slight difference in the spin-table having a `start_date`, but the repoint-table not having that key so some slight avoidance of that was necessary.


tech3371 (approved):
Looks great! You are on roll!!

PR #935, opened 2025-10-15
I-ALiRT - Spice metadata

laspsandoval:
# Change Summary

## Overview
We need a way to track when SPICE kernels are used. This PR creates an entry in the database for each processing job as a way to track which SPICE kernels are used.

## Updated Files
- ialirt_ingest.py
   - Add SPICE kernels to the DynamoDB

## Testing
- test_ialirt_ingest.py

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
@@ -487,9 +487,6 @@ def lambda_handler(event, context):
     logger.info("Retrieved filename: %s", filename)
     dependency_inputs = get_latest_spice_kernels(url)
 
-    # Insert kernel metadata every minute.
-    insert_kernels(dependency_inputs, algorithm_table)
-
     logger.info("dependency_inputs: %s", dependency_inputs)
     download_spice_file(dependency_inputs)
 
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -445,6 +486,10 @@ def lambda_handler(event, context):
     filename = os.path.basename(s3_filepath)
     logger.info("Retrieved filename: %s", filename)
     dependency_inputs = get_latest_spice_kernels(url)
+
+    # Insert kernel metadata every minute.
+    insert_kernels(dependency_inputs, algorithm_table)
```
greglucas: Shouldn't we only do this if we are running the algorithm? I figured this would only be valuable if we want to tie specific processing data back to the kernels being used.

Does this also get output to the user? I think this should be there for information purposes but not exposed in the data query.

laspsandoval: Since the MAG algorithm will always be executed (since their data product has a 4 second cadence) then we will need this metadata every minute and it doesn't necessarily need to be associated with their algorithm.

I need to make changes about what is output to the user for this and other items (apid, for example). I also need to go back and make changes so that this information is available in the archive cdf.

greglucas: I'm suggesting to put it down at l.512 only in the `if filenames:` branch because otherwise we'll be putting it in even when we haven't received packets from ground stations I think.

laspsandoval: Got it. Ok.

greglucas (approved):
While looking at that, maybe it would make sense to also move the kernel downloading and that after the check too. So basically change the initial check to `if not filenames: logger.info(); return` higher up in the function and then only after you know you have some data then get the kernels and run the processing.

But this code looks good, that can be future improvement.

PR #942, opened 2025-10-16
ENH: increase timeout on spice query and batch starter lambda

tech3371:
# Change Summary

## Overview

Increase timeout on those lambda since it can take longer than 1min


greglucas:
😮 I'm mildly against this. Do we know what is taking the time? I hate APIs that take longer than a few seconds and we are going from 1 minute to 5!

lacoak21:
> 😮 I'm mildly against this. Do we know what is taking the time? I hate APIs that take longer than a few seconds and we are going from 1 minute to 5!

It was timing out on the spice metakernel api call. I was trying to reprocess the pointing-attitude job with this command:

` imap-data-access --api-key <key> --url https://api.imap-mission.com/api-key reprocess --data-level l1a --instrument spacecraft --descriptor pointing-attitude --start-date 20250901 --end-date 20251014`

tech3371:
Greg, do you want to try shorter than 5min? and then try again in the future?

greglucas (approved):

greglucas:
I think it is fine to get over hurdles, but we need a ticket to improve these in the future I think.

I'm guessing this is improvement in the code which might take a little while to debug and figure out. But I'm almost positive this should be doable in a few seconds (maybe I'm naive, but I don't see how it should take minutes)

greglucas:
> It was timing out on the spice metakernel api call. I was trying to reprocess the pointing-attitude job with this command:

> imap-data-access --api-key <key> --url https://api.imap-mission.com/api-key reprocess --data-level l1a --instrument spacecraft --descriptor pointing-attitude --start-date 20250901 --end-date 20251014

@lacoak21 are these API calls from local or remote? Is this another area that we should not be calling an external API but rather the functions we already have available through local imports.

greglucas:
Also, you can drop this now with version 0.37 :) `--url https://api.imap-mission.com/api-key`

lacoak21:
I just checked and it is directly calling the metakernel lambda handler

PR #946, opened 2025-10-17
I-ALiRT - Quick latency fix

laspsandoval:
# Change Summary

## Overview
Note: DB changes to come.

## Updated Files
- ialirt_db_query_api.py
   - Added logging to measure time.
   - Increased lambda memory and removed unnecessary layers

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api.py
@@ -3,6 +3,7 @@
 import json
 import logging
 import os
+import time
 from decimal import Decimal
 
 import boto3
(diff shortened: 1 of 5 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_db_query_api.py
```diff
@@ -67,12 +68,16 @@ def lambda_handler(event, context):  # noqa: PLR0912
         and runtime environment.
 
     """
+    t0 = time.perf_counter()
     table_name = os.environ.get("ALGORITHM_TABLE")
     region = os.environ.get("AWS_DEFAULT_REGION", "us-west-2")
     dynamodb = boto3.resource("dynamodb", region_name=region)
     table = dynamodb.Table(table_name)
```
greglucas: These should be moved up into the execution environment (right after the imports) and not within a function call, then it can be cached by the lambda startup.

greglucas (approved):

PR #947, opened 2025-10-17
Api authorizer performance

### LINKED ISSUE #945: Authorization lambdas can be slow
It looks like on cold starts of the lambda we get a 2.5s latency just from the authorization portion of a request. A quick fix would be to increase the memory of the authorizer lambda to get a little bit more performance, but we should look into other performance increases here too as that seems like an unacceptable latency on initial interaction with the SDC.

Image

greglucas:
# Change Summary

## Overview

Improve performance of authorization of API Keys

closes #945


sds_data_manager/constructs/api_gateway_construct.py
```diff
@@ -124,6 +125,9 @@ def __init__(
             handler=self.api_key_authorizer_lambda,
             response_types=[apigwv2_authorizers.HttpLambdaResponseType.SIMPLE],
             identity_source=["$request.header.x-api-key"],
+            # Cache multiple requests with the same API key so we don't
```
laspsandoval, line 128: Very nice.

sds_data_manager/constructs/api_gateway_construct.py
```diff
@@ -105,6 +105,7 @@ def __init__(
             handler="lambda_api_key_authorizer.lambda_handler",
             code=lambda_.Code.from_asset("sds_data_manager/lambda_code/authorization"),
             timeout=Duration.seconds(10),
+            memory_size=512,  # Increase memory for larger network
```
laspsandoval, line 108: How did you come up with this number?

laspsandoval (approved):
The caching will help a lot. Thanks Greg.

sds_data_manager/constructs/api_gateway_construct.py
greglucas: ChatGPT told me this is a common number. I've used it other places too to get more throughput (1k seems excessive and probably wouldn't be worth the cost). We could go to 256 as well probably. Might be something we play with if needed.

greglucas:
Agreed, the caching is the big win I think

PR #948, opened 2025-10-17
CI: Update container build and automate push to prod

greglucas:
# Change Summary

## Overview

This updates the build containers script to take a specific instrument from a list of choices rather than requiring typing and rearranges a few things.

It also adds a production deployment step where we pull the dev container and promote it into the production account.

We also parallelize all of the above per instrument in a separate runner.

Iterated on this with Claude.


tmplummer (approved):
I'm not very fluent in github actions, but this looks like some great quality of life improvements!

PR #949, opened 2025-10-17
I-ALiRT - Dynamodb addition

laspsandoval:
# Change Summary

## Overview
Add a parallel db table w/ new partition key and sort key.

## Updated Files
   - Add new db table w/ partition key = instrument and sort key = met_in_utc
- ialirt_ingest.py
   - Add new partition key “instrument” to kernel ingest
   - Update the insert function
   - Add capability to insert into new table (after the old table just in case there are any errors)

## Testing
- conftest.py
   - Add new database setup
- test_ialirt_ingest_lambda_construct.py
   - test_ialirt_ingest.py

## Notes
We will need to update our query apis and also archive api once this is approved and we begin to transition to the new database.

### DIFF
--- a/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
+++ b/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
@@ -157,28 +157,14 @@ class IalirtIngestLambda(Construct):
                 name="instrument",
                 type=ddb.AttributeType.STRING,
             ),
-            # Sort key (SK) = met_in_utc.
+            # Sort key (SK) = time_utc.
             sort_key=ddb.Attribute(
-                name="met_in_utc",
+                name="time_utc",
                 type=ddb.AttributeType.STRING,
             ),
             billing_mode=ddb.BillingMode.PAY_PER_REQUEST,  # On-Demand capacity mode.
         )
 
-        # Add a GSI for last modified timestamp.
-        self.data_table.add_global_secondary_index(
-            index_name="last_modified",
-            # Partition key (PK) = instrument.
-            partition_key=ddb.Attribute(
-                name="instrument", type=ddb.AttributeType.STRING
-            ),
-            # Sort key (SK) = last modified.
-            sort_key=ddb.Attribute(
-                name="last_modified",
-                type=ddb.AttributeType.STRING,
-            ),
-            projection_type=ddb.ProjectionType.ALL,
-        )
         return self.data_table
 
     def create_lambda_function(
--- a/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
+++ b/sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
@@ -62,11 +62,13 @@ class IalirtIngestLambda(Construct):
 
         # Create DynamoDB Table
         self.algorithm_data_table = self.create_algorithm_dynamodb_table()
+        self.data_table = self.create_table()
 
         # Create Lambda Function
         self.ialirt_ingest_lambda = self.create_lambda_function(
             ialirt_bucket,
             self.algorithm_data_table,
+            self.data_table,
             docker_path,
             data_access_url,
         )
@@ -133,10 +135,57 @@ class IalirtIngestLambda(Construct):
         )
         return self.algorithm_data_table
 
+    def create_table(self) -> aws_dynamodb.Table:
+        """Create and return the algorithm data product table."""
+        removal_policy = (
+            RemovalPolicy.RETAIN
+            if self.account_name == "prod"
+            else RemovalPolicy.DESTROY
+        )
+        point_in_time_recovery = True if self.account_name == "prod" else False
+
+        self.data_table = ddb.Table(
+            self,
+            "IalirtDataTable",
+            table_name="ialirt-data-table",
+            # RemovalPolicy.RETAIN to keep the table after stack deletion.
+            removal_policy=removal_policy,
+            # Restore data to any point in time within the last 35 days.
+            point_in_time_recovery=point_in_time_recovery,
+            # Partition key (PK) = instrument.
+            partition_key=ddb.Attribute(
+                name="instrument",
+                type=ddb.AttributeType.STRING,
+            ),
+            # Sort key (SK) = met_in_utc.
+            sort_key=ddb.Attribute(
+                name="met_in_utc",
+                type=ddb.AttributeType.STRING,
+            ),
+            billing_mode=ddb.BillingMode.PAY_PER_REQUEST,  # On-Demand capacity mode.
+        )
+
+        # Add a GSI for last modified timestamp.
+        self.data_table.add_global_secondary_index(
+            index_name="last_modified",
+            # Partition key (PK) = instrument.
+            partition_key=ddb.Attribute(
+                name="instrument", type=ddb.AttributeType.STRING
+            ),
+            # Sort key (SK) = last modified.
+            sort_key=ddb.Attribute(
+                name="last_modified",
+                type=ddb.AttributeType.STRING,
+            ),
+            projection_type=ddb.ProjectionType.ALL,
+        )
+        return self.data_table
+
     def create_lambda_function(
         self,
         ialirt_bucket: aws_s3.Bucket,
         algorithm_data_table: aws_dynamodb.Table,
+        data_table: aws_dynamodb.Table,
         docker_path: str,
         data_access_url: str,
     ) -> lambda_.DockerImageFunction:
@@ -187,6 +236,7 @@ class IalirtIngestLambda(Construct):
             ),
             environment={
                 "ALGORITHM_TABLE": algorithm_data_table.table_name,
+                "DATA_TABLE": data_table.table_name,
                 "S3_BUCKET": ialirt_bucket.bucket_name,
                 "EFS_SPICE_MOUNT_PATH": "/mnt/data",
                 "IMAP_DATA_ACCESS_URL": data_access_url,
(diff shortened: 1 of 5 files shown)


sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
```diff
@@ -133,10 +135,57 @@ def create_algorithm_dynamodb_table(self) -> aws_dynamodb.Table:
...
+            # Restore data to any point in time within the last 35 days.
+            point_in_time_recovery=point_in_time_recovery,
+            # Partition key (PK) = instrument.
+            partition_key=ddb.Attribute(
+                name="instrument",
+                type=ddb.AttributeType.STRING,
+            ),
+            # Sort key (SK) = met_in_utc.
+            sort_key=ddb.Attribute(
+                name="met_in_utc",
```
greglucas: If this is going to be storing the measurements I think we should avoid MET because sometimes it came down in multiple packets.
```suggestion
                name="time_utc",
```

sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
```diff
@@ -133,10 +135,57 @@ def create_algorithm_dynamodb_table(self) -> aws_dynamodb.Table:
...
+                type=ddb.AttributeType.STRING,
+            ),
+            billing_mode=ddb.BillingMode.PAY_PER_REQUEST,  # On-Demand capacity mode.
+        )
+
+        # Add a GSI for last modified timestamp.
```
greglucas: Will anything ever be modified in I-ALiRT? I thought we were only going to be producing what we were producing. I don't think our users should be querying by modified times, but rather the only entry-point should be "time_utc".

As a side point, I kind of assume that if we do modify things it will be a modification of everything so then we just go back to the normal time being of interest. Maybe we can add this in later if it does become necessary?

tests/infrastructure/test_ialirt_ingest_lambda_construct.py
```diff
@@ -30,6 +30,66 @@ def populate_algorithm_table(setup_dynamodb):
...
 
 
+@pytest.fixture
+def populate_data_table(setup_data_table):
+    """Populate DynamoDB table."""
+    data_table = setup_data_table["data_table"]
+    items = [
+        {
+            "instrument": "mag",
+            "met_in_utc": "2021-01-01T00:00:00.000000001",
```
greglucas: Another thing that will help here is if we can truncate the times to seconds to reduce storage space, I'm not sure where this would come in up above though. I don't think we have any resolutions more than every 4 seconds here.

greglucas (commented):
Nice, I think this is a great approach to do it if I'm understanding correctly! You are creating a new table so that we can be inserting into both side-by-side in production. Then we'll test things out and maybe run a migration of previous items into the new table and only remove the original table if this approach is useful.

One thing that might be nice here is to try and be a bit more restrictive about what we are putting into each partition insert, so picking and choosing the instrument data appropriately based on the table name (the new table doesn't need an APID column for instance and we want mag_hk to be a separate instrument). I'm a bit confused if the HIT instrument insert is going to be inserted at a 4-second cadence or just one item in this new approach.

sds_data_manager/constructs/ialirt_ingest_lambda_construct.py
laspsandoval: Agreed. I was back on forth on this one.

laspsandoval:
I made some pretty significant updates to ialirt_ingest.py. The rest is mostly the same (besides responding to the PR comments)

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -421,6 +428,122 @@ def insert_data(data: list[dict], algorithm_table, instrument: str):
...
+    )
+    hk_response = data_table.query(
+        KeyConditionExpression=Key("instrument").eq(f"{instrument}_hk")
+        & Key("time_utc").between(min_time, max_time)
+    )
+
+    existing_items = {item["time_utc"]: item for item in response.get("Items", [])}
+    existing_hk_items = {
+        item["time_utc"]: item for item in hk_response.get("Items", [])
+    }
```
greglucas: It looks like you are only looking at the time to see if there is an entry already in the table and ignoring it if so and not using the saved existing item. In that case I think you can simplify this to create a set of items to lookup in, maybe you can also have DynamoDB only filter to the time_utc for you by adding a FilterKey

```python
existing_science_times = set(item["time_utc"] for item in response.get("Items", []))

for data in science_data:
    if data["time_utc"] in existing_science_times:
        continue
    data_table.put_item(Item=data)
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -421,6 +428,122 @@ def insert_data(data: list[dict], algorithm_table, instrument: str):
...
+
+    # Query existing items.
+    response = data_table.query(
+        KeyConditionExpression=Key("instrument").eq(instrument)
+        & Key("time_utc").between(min_time, max_time)
+    )
+    hk_response = data_table.query(
+        KeyConditionExpression=Key("instrument").eq(f"{instrument}_hk")
+        & Key("time_utc").between(min_time, max_time)
+    )
```
greglucas: Thinking about this, do we care if we just call the PutItem regardless? We overwrite the existing item and don't deal with the extra hassle of figuring out if there are entries already.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -421,6 +428,122 @@ def insert_data(data: list[dict], algorithm_table, instrument: str):
...
+    }
+
+    # Insert science data
+    for raw in science_data:
+        time = raw["time_utc"]
+        existing = existing_items.get(time)
+
+        if not existing:
+            data_table.put_item(Item=raw)
+        logger.info(f"Inserted {instrument.upper()}.")
```
greglucas: This will log a lot of "instrument", but not any detail about what was inserted. Maybe outside the loop you could put in something like "inserted X entries for {instrument}" so you aren't logging so much and give some idea of what was inserted.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -421,6 +428,122 @@ def insert_data(data: list[dict], algorithm_table, instrument: str):
...
+    # Calculate the spacecraft position and velocity in GSE/GSM coordinates.
+    et = str_to_et(min_time)
+    gsm_state = imap_state(
+        [et], ref_frame=SpiceFrame.IMAP_GSM, observer=SpiceBody.EARTH
+    )
+    gse_state = imap_state(
+        [et], ref_frame=SpiceFrame.IMAP_GSE, observer=SpiceBody.EARTH
+    )
+    geolocation = {
+        "instrument": "geolocation",
```
greglucas: what about calling this "spacecraft", "sc_pos_vel" or something similar? geolocation is usually something on Earth, but we are outside of Earth's orbit.

tests/lambda_endpoints/test_ialirt_ingest.py
```diff
@@ -242,6 +246,170 @@ def test_insert_data(
...
+            "met_in_utc": "2021-01-01T00:00:01",
+            "ttj2000ns": 759175836184000001,
+            "hit_data": Decimal("2.0"),
+        },
+    ]
+
+    science_data, hk_data = reformat_data(test_data)
+
+    assert all("apid" not in d for d in science_data)
+    assert all("met" not in d for d in science_data)
```
greglucas: You only have 2 items you're inserting, could you assert against the actual expectation?

tests/lambda_endpoints/test_ialirt_ingest.py
```diff
@@ -242,6 +246,170 @@ def test_insert_data(
...
+        Key={"instrument": "geolocation", "time_utc": "2021-01-01T00:00:00"}
+    )["Item"]
+
+    # Not updated
+    assert item1["mag_data"] == Decimal("0.0")
+    assert item2["mag_data"] == Decimal("0.0")
+
+    # New item should be inserted
+    assert item3["mag_data"] == Decimal("5.0")
+    assert item4["mag_hk_status"]["hk1v5_danger"] is True
```
greglucas, line 421: Are there different "mag_hk" groupings, or can you bring the "hk1v5_danger" up a level and into the "mag_hk" instrument itself?

greglucas (commented):
Sorry for the slow review here. A few more nitpicks / questions, but getting close!

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: You know what, I don't think that we do care. Good point.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: Yes I caught that just now.

tests/lambda_endpoints/test_ialirt_ingest.py
laspsandoval: Yes I thought about this and played around with it and decided to leave this as is for now. MAG has already set up processing based on the structure of the current database. I would have to change imap_processing so that it would insert the data in a different way. It just became complicated and I opted for simple changes initially.

greglucas (approved):

tests/lambda_endpoints/test_ialirt_ingest.py
greglucas: 👍 I wouldn't worry about what anyone else has at this point. You are trying to set this up the best way possible and we are before operational use still IMO. We shouldn't go changing things to change them, but if there is a reason you want to do something, don't hold it up because someone else would need to adapt. It will be much harder in a few months, so now is the time to do what you think is best.

But, I had no issue either way here, so this is still good this way to me.

PR #955, opened 2025-10-20
Get end time of repoint table from Pointing table

### LINKED ISSUE #954: Query Pointing table to get repoint table end date when indexing
#### Background:
The current implementation just uses the year_doy from the filename as the repoint table end date when indexing the file This date gets truncated to the start of that DOY. This potentially causes data to get missed that we could process because the repoint table likely includes data after the start of that DOY.

#### Suggested Fix
Query the Pointing table to get the end time. The end time should be the greater of either the last Pointing end time or the last pointing start time. Tenzin says that the Pointing table is updated before the repoint table gets indexed.

We need to double check that the Pointing table will record the start time of an incomplete Pointing.

tmplummer:
Add test coverage to make sure end_time of repoint table gets set correctly on ingest

# Change Summary

## Overview
From the ticket: 
### Background:

The current implementation just uses the year_doy from the filename as the repoint table end date when indexing the file This date gets truncated to the start of that DOY. This potentially causes data to get missed that we could process because the repoint table likely includes data after the start of that DOY.
### Suggested Fix

Query the Pointing table to get the end time. The end time should be the greater of either the last Pointing end time or the last pointing start time. Tenzin says that the Pointing table is updated before the repoint table gets indexed.

## Updated Files
   - Fix comment
   - Get repoint file end_date from Pointing table which will have already been updated before repoint ingest.
   - Add some explicit test coverage for the logic used to get the repoint file end_date

Closes: #954

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -615,7 +615,7 @@ def calculate_repoint_table_date_range(session, file_obj):
     tuple
         A tuple containing the start date and end date in the format YYYYMMDD.
     """
-    # Query the pointing table to find the pointing information.
+    # Query the repoint table to get the exact date/time.
     end_date = file_obj.spice_metadata["end_date"]
     previous_entries = (
         session.query(models.RepointFiles)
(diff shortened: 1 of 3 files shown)


greglucas (approved):
Wow, that was fast! I didn't realize we were already connecting to the DB here, so this makes sense to use the pointing table now that we have that info readily available. Nice tests too!

tech3371 (approved):
Looks good. Thank you for doing this!

PR #956, opened 2025-10-20
BUG: Filter science dependencies by repoint number if ENA/glows

lacoak21:
# Change Summary

## Overview
Jobs were getting kicked off with dependencies that contained multiple files of different repoint ids. This caused processing job failures for at least HI and ULTRA.

Fix: Search for science files that have the same repointing number for glows, hi, ultra, and lo

## Updated Files
   - Add repointing number for the upstream file query.
   - Pass through the repoint number and add functionality to filter science files by it. 

## Testing
test to make sure the new parameter in the function works as expected

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -26,7 +26,7 @@ from sqlalchemy.exc import IntegrityError
 from ..api_lambdas import upload_api
 from ..database import database as db
 from ..database import models
-from . import VALID_CADENCE_STRS, dependency
+from . import REPOINT_DEPENDENT_INSTRUMENTS, VALID_CADENCE_STRS, dependency
 from .dependency import DependencyConfig
 
 # Logger setup
@@ -49,8 +49,6 @@ BATCH_JOB_RETRY_STRATEGY = {
 }
 # Create an sqs client
 SQS_CLIENT = boto3.client("sqs", region_name="us-west-2")
-
-REPOINT_DEPENDENT_INSTRUMENTS = ["glows", "hi", "lo", "ultra"]
 
 
 def cadence_to_datetime_range(
@@ -486,11 +484,7 @@ def submit_all_jobs(
         start_date, end_date = determine_date_range(session, science_file)
 
         # Get the repointing number from the science file object
-        job_repointing = (
-            science_file.repointing
-            if job_node["data_source"] in REPOINT_DEPENDENT_INSTRUMENTS
-            else None
-        )
+        job_repointing = science_file.repointing
 
         # If there is only one file to process, then we can use upstream dependencies
         # that have already been queried.
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -50,6 +50,8 @@ BATCH_JOB_RETRY_STRATEGY = {
 # Create an sqs client
 SQS_CLIENT = boto3.client("sqs", region_name="us-west-2")
 
+REPOINT_DEPENDENT_INSTRUMENTS = ["glows", "hi", "lo", "ultra"]
+
 
 def cadence_to_datetime_range(
     cadence: str,
@@ -484,7 +486,11 @@ def submit_all_jobs(
         start_date, end_date = determine_date_range(session, science_file)
 
         # Get the repointing number from the science file object
-        job_repointing = science_file.repointing
+        job_repointing = (
+            science_file.repointing
+            if job_node["data_source"] in REPOINT_DEPENDENT_INSTRUMENTS
+            else None
+        )
 
         # If there is only one file to process, then we can use upstream dependencies
         # that have already been queried.
@@ -499,6 +505,7 @@ def submit_all_jobs(
                 relationship="ALL",
                 start_date=start_date,
                 end_date=end_date,
+                repoint=job_repointing,
                 calculate_crids=False,
                 get_spice=True,
             )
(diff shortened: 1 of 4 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -50,6 +50,8 @@
 # Create an sqs client
 SQS_CLIENT = boto3.client("sqs", region_name="us-west-2")
 
+REPOINT_DEPENDENT_INSTRUMENTS = ["glows", "hi", "lo", "ultra"]
```
lacoak21: Any better name ideas?

greglucas: That seems good to me. Here is what we do for the packet_downloader which is named similarly.
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/aea9d339b324700bad92645c4080bb38d3c21156/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py#L119

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -484,7 +486,11 @@ def submit_all_jobs(
         start_date, end_date = determine_date_range(session, science_file)
 
         # Get the repointing number from the science file object
-        job_repointing = science_file.repointing
+        job_repointing = (
+            science_file.repointing
+            if job_node["data_source"] in REPOINT_DEPENDENT_INSTRUMENTS
+            else None
+        )
```
greglucas: Isn't this None in the `science_file.repointing` attribute if an instrument isn't dependent on repointing? i.e. None is the default I thought.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -1078,6 +1084,10 @@ def get_files(
                 models.ScienceFiles.start_date <= end_date,
             )
         )
+        # If repoint is provided, filter by repointing number
+        if repoint is not None:
+            type_specific_conditions.append(table.repointing == repoint)
```
greglucas, line 1092: Do we need this as an alternative to the time, not in addition to it? i.e. we only do the time or we only do the repointing number.

tests/lambda_endpoints/test_dependency_api.py
```diff
@@ -186,6 +187,74 @@ def test_missing_required_params():
...
+                repointing=48,
+                ingestion_date=datetime.strptime(
+                    "2024-01-25 23:35:26+00:00", "%Y-%m-%d %H:%M:%S%z"
+                ),
+            ),
+        ]
+    )
+    session.commit()
+    # Get upstream files for glows l1a all. This should return None because although
+    # the glows l0 raw file exists, the spice files do not exist in the database.
```
greglucas: Great comments in the test for our future selves!

lacoak21: Oh whoops that is actually a left over one and needs updating 😆

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Hm good question. I dont think I am the person to answer that. @subagonsouth or @tech3371 any thoughts?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
lacoak21: Yes it is already None. It does feel redundant here but @tech3371 recommended checking it against that list of instruments similarly to what she did when she calculating the date ranges for these instruments.

greglucas: My personal preference would probably be to leave it as-is and let it fall out properly. We don't currently have any repointing data on in-situ's, but this effectively prevents that. Maybe there is a data product in the future where someone wants to change an in-situ product to line up with repointings.

tech3371: Checking it here may be redundant. But I do think we need that at later steps in dependency.py. So passing it should be enough.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
tech3371: Here we do need to filter by repoint id since repoint date range can cover more than one science files. Because of this, we need to get all dependency for that specific science file we are looking up for and want to kickoff job. This will be even more true in longer reprocessing date range.

tmplummer: I would think that filtering by time and repoint would give the same results as filtering only by repoint.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -1078,6 +1084,10 @@ def get_files(
                 models.ScienceFiles.start_date <= end_date,
             )
         )
+        # If repoint is provided, filter by repointing number
+        if repoint is not None:
```
tmplummer: Does this also need to check that the dependency being filtered has a repoint field. I don't think that we currently have this situation, but I think there was talk at one point of having Hi depend on a non-ena instrument for determining good times.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: Ok I think you're right. So in that case I will leave this as is?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
greglucas: I think this might have been where @tech3371 was thinking we would do that check. I forgot that this was actually for "upstream" science files that we are searching for. I agree then that we should add the condition on being a repointing_instrument here for the situation @subagonsouth mentions.

lacoak21: Thats a great point. 

So something like : 

if repoint is not None:
    type_specific_conditions.append(
        and_(table.repointing.is_not(None), table.repointing == repoint)
    )

greglucas: I was thinking you could do that upon entrance and not need to add it to the query because you already know instrument at this point?

```python
if repoint is not None and instrument in REPOINT_DEPENDENT_INSTRUMENTS:
    ...
```

tech3371: That logic looks good to me.

lacoak21: Ok makes sense!

greglucas (approved):

PR #962, opened 2025-10-22
BUG: only return data available in both DB and S3

tech3371:
# Change Summary

## Overview

Bug fix in query API that returns file that are not both in DB or S3


lacoak21:
Shouldn't we be concerned when files records exist in the db but are not in s3?

tech3371:
> Shouldn't we be concerned when files records exist in the db but are not in s3?

I think so. I am concerned about that but this feels like a bug to me.

lacoak21:
> I think so. I am concerned about that but this feels like a bug to me.

Ok I see what you mean. The query should truly reflect whats in s3.

lacoak21 (approved):

tech3371:
Let me merge this and tests. My test environment both local and Git is down. Will open another PR if this didn't work.

greglucas:
I don't like this because it is adding a lot of HEAD_OBJECT checks unnecessarily. We should add a synchronizer or run the synchronizer whenever we do these updates.

PR #963, opened 2025-10-22
Add logging statement in the query API

lacoak21:
# Change Summary

## Overview
Fix query api

## Updated Files
- sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
   - Fix the query api to check for files in s3 

## Testing
fix tests


tech3371 (approved):

tech3371:
@pleasant-menlo This will fixed mis-alignment between DB and S3 bucket. Now it will check that it exists in both

Issue #958, opened 2025-10-21
BUG - repoint end date in batch starter not coming from DB


tmplummer:
### Description of the issue

I updated code to correctly calculate the end_date of a Repoint table based on data that is already inserted into the Pointing table. I now see that batch starter is just using the YYYY_DOY from the repoint table filename as end date. It should be getting the end date from the repoint table instead.

### Code Snippet:

```python
Code
```

lacoak21:
I added a temporary solution to this problem https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/964
I seems like it is reading the actual start and stop times from the table but like Tim mentioned, it is only specific up to the day.

bryan-harter:
We get this from the DB now!

PR #964, opened 2025-10-23
Fix SPICE metakernel API query for ENA instrument

lacoak21:
# Change Summary

## Overview
For ENA instruments, the start and end dates are set from the pointing start and end times and are floored in batch_starter.py. This means the pointing end was getting cut off. Therefore, in get_upstream_dependency_inputs in dependency.py for ENA instruments that have a repointing number, we add one day to the end date to ensure all necessary SPICE files are retrieved. 

## Updated Files
   - Add 24 hours to ENA pointing dependent instruments

## Testing
Add a test: This actually mocks the exact scenario that is causing issues with ULTRA l1b. The test proves that this solves the bug!

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
@@ -839,6 +839,7 @@ def get_upstream_dependency_inputs(
     dependencies: list,
     start_date: datetime,
     end_date: datetime,
+    data_source: str,
     repoint: Optional[int] = None,
     calculate_crids: bool = False,
     get_spice: bool = True,
@@ -857,6 +858,8 @@ def get_upstream_dependency_inputs(
         Start date to find dependent files with.
     end_date : datetime
         End date to find dependent files with.
+    data_source : str
+        Data source of the current processing job.
     repoint : int, optional
         If provided, will be used to filter files by repoint number.
     calculate_crids : bool
@@ -939,7 +942,12 @@ def get_upstream_dependency_inputs(
                     start_date.strftime("%Y%m%d")
                 )
                 # TODO revisit setting end_time after SIT-4. Should be handled upstream
-                add_24_hrs = True if end_date == start_date else False
+                if end_date == start_date or (
+                    data_source in REPOINT_DEPENDENT_INSTRUMENTS and repoint is not None
+                ):
+                    add_24_hrs = True
+                else:
+                    add_24_hrs = False
                 end_time = yyyymmdd_to_seconds_since_j2000(
                     end_date.strftime("%Y%m%d"), add_24_hrs
                 )
@@ -1260,6 +1268,7 @@ def get_jobs(
         start_date=start_date,
         end_date=end_date,
         repoint=repoint,
+        data_source=data_source,
         calculate_crids=calculate_crids,
         get_spice=get_spice,
     )
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
@@ -839,7 +839,6 @@ def get_upstream_dependency_inputs(
     dependencies: list,
     start_date: datetime,
     end_date: datetime,
-    data_source: str,
     repoint: Optional[int] = None,
     calculate_crids: bool = False,
     get_spice: bool = True,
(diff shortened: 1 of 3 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -939,7 +942,12 @@ def yyyymmdd_to_seconds_since_j2000(
                     start_date.strftime("%Y%m%d")
                 )
                 # TODO revisit setting end_time after SIT-4. Should be handled upstream
```
greglucas, line 941: Gotta love finding these notes to ourselves 😂

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
```diff
@@ -939,7 +942,12 @@ def yyyymmdd_to_seconds_since_j2000(
...
-                add_24_hrs = True if end_date == start_date else False
+                if end_date == start_date or (
+                    data_source in REPOINT_DEPENDENT_INSTRUMENTS and repoint is not None
```
greglucas: Don't we want this regardless of the data_source? Just apply it if any repoint is not None gets passed in... Then you don't need the data_source pass-through everywhere.

lacoak21: True. I guess i was adding it as a safety precaution but now that I think about it, the sole fact that it has the repoint number means we should be adding this time bump.

tmplummer:
I'm ignorant as to how the dependency manager decides if it has the correct SPICE coverage. My concern with this approach is that we are still not looking of the exact coverage that we need for a pointing. This likely means that we may have cases where we can run processing but since we are looking for a larger coverage range than we require, the dependency manager will think that it doesn't have the correct SPICE coverage and will skip a job.

The even bigger problem could be that then when the kernel comes that provides the missing coverage from the above scenario, that job that got skipped won't be identified as a job to run based on that new kernel.

The fix in this PR can act as a temporary fix, but I think the only way to have a robust fix is to change from the YYYYMMDD strings for specifying start_date and end_date to allowing hh:mm:ss so that we can get really precise about the needed coverage.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py
lacoak21: 🤣

lacoak21:
> I'm ignorant as to how the dependency manager decides if it has the correct SPICE coverage. My concern with this approach is that we are still not looking of the exact coverage that we need for a pointing. This likely means that we may have cases where we can run processing but since we are looking for a larger coverage range than we require, the dependency manager will think that it doesn't have the correct SPICE coverage and will skip a job.
> 
> The even bigger problem could be that then when the kernel comes that provides the missing coverage from the above scenario, that job that got skipped won't be identified as a job to run based on that new kernel.
> 
> The fix in this PR can act as a temporary fix, but I think the only way to have a robust fix is to change from the YYYYMMDD strings for specifying start_date and end_date to allowing hh:mm:ss so that we can get really precise about the needed coverage.

Agreed - I think that is the best future solution. I am hoping to get a temporary fix in so I can at least start producing some l1b data for ULTRA. I started working to clean this up but i think I would need a larger chunk of time because I started realizing just how many things the start and end date touch. What if, as a better temporary fix, we just call the calculate_pointing_date_range again right before the metakernel api query? I know that is redundant, but at least we would have the correct query times. Then sometime next week I can add the refactor? 

I definitely understand the argument of just telling the ULTRA team to wait more and do this the proper way first if thats what people would prefer.

tmplummer:
> What if, as a better temporary fix, we just call the calculate_pointing_date_range again right before the metakernel api query? I know that is redundant, but at least we would have the correct query times. Then sometime next week I can add the refactor?

I am happy with this as a temporary fix. I just wanted to make sure it was temporary.

tmplummer (approved):
Looks good as a temporary fix.

lacoak21:
Oh got it. Sounds good 👍.

tmplummer:
For visibility... I wrote this bug ticket when I realized there is still an issue in how the repoint time range is being calculated: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/958

tech3371 (approved):

PR #967, opened 2025-10-28
Add Hi L3 Spectral Index products to dependency config

pleasant-menlo:
Added Hi Spectral Index products to dependency config

All other Hi L3 products will be generated via the scheduled job workflow.


tmplummer (approved):
LGTM

PR #969, opened 2025-10-28
MNT: Add Lo ancillary flux files to dependency config

greglucas:
# Change Summary

## Overview

* added the `esa-eta-fit-factors` ancillary files to the l2 map dependencies for Lo.
* uploaded the `esa-eta-fit-factors` ancillary file to production so it is ready to go.


tmplummer (approved):
LGTM

PR #977, opened 2025-10-30
Revert s3-db query response out of sync updates

greglucas:
We don't want to do a HEAD request on every file we are querying. If the database and s3 are out of sync, we have bigger issues and we should be running the synchronizer lambda instead.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
@@ -3,11 +3,8 @@
 import datetime
 import json
 import logging
-import os
 from collections import namedtuple
 
-import boto3
-import botocore
 from sqlalchemy import func, select
 
 from ..api_lambdas.utils import is_authenticated_user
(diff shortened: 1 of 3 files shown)


sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
```diff
@@ -134,28 +131,6 @@ def lambda_handler(event, context):  # noqa: PLR0912 PLR0915
...
-        # check if object exists
-        try:
-            s3_client.head_object(Bucket=bucket, Key=s3_key)
-            existing_files.append(result)
-        except Exception as e:
-            logger.error(
-                f"File not found in S3: {s3_key} but exists in DB - error: {e}"
-            )
-
-    search_results = existing_files
```
tech3371, line 158: Instead of filtering, should we just log for ourselves? The logs are nice to look at since some people may not look at db.

greglucas: IMO, we should not be querying both the DB and s3, the point of the DB is to represent what is in s3 and keep track of it. If we are going to look at s3 then we should just query s3 directly the DB doesn't matter anymore. We get charged per HEAD request, so this would be a lot I think because it is going through every file individually per query.

tech3371: true. sounds good!

tech3371 (approved):

PR #978, opened 2025-10-30
FIX: Update synchronizer to work and skip bad files

greglucas:
There are some files that can't be turned into imap-data-access file types, so skip those for now and don't synchronize the DB with those items.

I tested this in dev to see what the issues were and this got around all of them. Not super robust, but I think this at least gets us the science and ancillary files synchronized.


sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
```diff
@@ -45,7 +46,8 @@ def lambda_handler(event, context):
     paginator = client.get_paginator("list_objects_v2")
     prefix = "imap/"
     s3_files_dict = {}
-    # TODO search spice/ folder
+    # TODO search spice, dependency, and quicklook folders
```
tech3371, line 49: we don't index dependency and quicklook. So, we just skip them all the time? and yeah, we need to add SPICE.

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/database/synchronizer.py
greglucas: I think we have plans to do quicklook? Maybe not though. But those can come later.

PR #982, opened 2025-10-31
ULTRA static dead time ratio LUTs 

lacoak21:
# Change Summary

## Overview
ULTRA is currently not operating in a mode that produces sectored rates data.  The sectored rates are needed for calculating the exposure time in l1c. They will not be produced during commissioning so the ultra team asked me to use these static dead time ratios temporarily so they can start producing psets.

## Updated Files
   - Add dead time luts


lacoak21:
I also realized we were missing the spin table.

greglucas (approved):
Do you have the code to make use of these in the `imap_processing` container?

Have you uploaded the 45/90 files to the ancillary directories?

lacoak21:
> Do you have the code to make use of these in the `imap_processing` container?
> 
> Have you uploaded the 45/90 files to the ancillary directories?

Yes i uploaded the ancillary files to prod and am about to put up a PR for imap_processing. I just wanted to create the pset locally to make sure there is nothing else that is needed

greglucas:
> I just wanted to create the pset locally to make sure there is nothing else that is needed

As an FYI. For that, I've been modifying a local dependency file with my own injections of data as needed rather than waiting for batch_starter updates.

lacoak21:
> As an FYI. For that, I've been modifying a local dependency file with my own injections of data as needed rather than waiting for batch_starter updates.

Thats what I have been doing too. This is where having that dependency.json api would be amazing.

PR #983, opened 2025-11-03
FIX: Always set SPICE files to be released

greglucas:
# Change Summary

Avoid any database query and just download these SPICE files.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
@@ -38,7 +38,10 @@ def is_released(s3_key):
 
         # Determine which table to use based on file type using isinstance
         if isinstance(file_obj, imap_data_access.file_validation.SPICEFilePath):
-            table = models.SPICEFiles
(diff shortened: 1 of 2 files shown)


tech3371 (approved):
Looks great!

sds_data_manager/lambda_code/SDSCode/api_lambdas/download_api.py
```diff
@@ -38,7 +38,10 @@ def is_released(s3_key):
 
         # Determine which table to use based on file type using isinstance
         if isinstance(file_obj, imap_data_access.file_validation.SPICEFilePath):
-            table = models.SPICEFiles
+            # NOTE: the spin and repoint files are not in the SPICE table, they
+            #       are in separate tables, so there would have to be other
+            #       logic here to separate out those file types if we wanted that.
+            return True  # SPICE files are always released
```
tech3371, line 44: Can you add this at line 164?
```
    if not is_released(filepath) and not is_authenticated_user(event):
```

tech3371 (changes requested):

tech3371 (approved):
nvm. Logic is good since we won't even go below if/else statement

PR #984, opened 2025-11-04
Increase batch job retries

lacoak21:
# Change Summary

## Overview
Increase the number of tries for the batch jobs. If a job gets terminated due to a spot task interruption. Max retries is 10. 

I looked into "hibernating" the instance but unfortunately that only is available for ec2 spot compute environments and not fargate spot. 

A lot of people in aws discussion boards have been asking for a fargate spot to fargate on-demand fallback policy but seems like that wont happen ( too easy maybe :)). 

It seems like the general recommendation is just to switch to on-demand or increase retries and make sure you are using many availability zones which our vpc already has. 

*** another thing to note fargate spot interruptions can send termination notices so eventually we should configure out jobs to stop processing when we get that 2 min warning. Hopefully this will stop jobs getting terminated right in the middle of uploading files. https://repost.aws/knowledge-center/fargate-spot-termination-notice

## Updated Files
   - increase retries to the max value.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -38,7 +38,7 @@ DEPENDENCY_CONFIG = DependencyConfig()
 BATCH_CLIENT = boto3.client("batch", region_name="us-west-2")
 # Define the retry strategy for batch jobs
 BATCH_JOB_RETRY_STRATEGY = {
-    "attempts": 2,
+    "attempts": 10,
     "evaluateOnExit": [
         {
             "onStatusReason": "Your Spot Task was interrupted.",


tech3371:
> # Change Summary
> ## Overview
> Increase the number of tries for the batch jobs. If a job gets terminated due to a spot task interruption. Max retries is 10.
> 
> I looked into "hibernating" the instance but unfortunately that only is available for ec2 spot compute environments and not fargate spot.
> 
> A lot of people in aws discussion boards have been asking for a fargate spot to fargate on-demand fallback policy but seems like that wont happen ( too easy maybe :)).
> 
> It seems like the general recommendation is just to switch to on-demand or increase retries and make sure you are using many availability zones which our vpc already has.
> 
> *** another thing to note fargate spot interruptions can send termination notices so eventually we should configure out jobs to stop processing when we get that 2 min warning. Hopefully this will stop jobs getting terminated right in the middle of uploading files. https://repost.aws/knowledge-center/fargate-spot-termination-notice
> 
> ## Updated Files
> * sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
>   
>   * increase retries to the max value.

So this is a billing related issue, meaning our batch was using more resource than spot instance? Should we increase instance size or something instead?

greglucas (approved):
What do we see in our processing logs for these when this happens? Do we get the logs from the latest batch job try, or only the first one that failed and need to update some of the pointers somehow?

I am curious about the SIGTERM and whether we can catch that in our processing. Specifically, I would think we could check for that before uploading files and if it is set, then don't continue on uploading. So at least we wouldn't have the "in the middle of uploads it failed" scenario.

PR #985, opened 2025-11-04
MNT/ENH: Add repoint file kickoff for ENA imagers

greglucas:
This adds back the repointing job trigger with a few modifications:
* Only runs on the "latest" file that gets delivered (avoids duplicate potential jobs if we get 3 repointing files delivered at the same time)
* Downloads packets between the last repointing file more than one hour ago and this current one (all packet ingest times during that period)
* Separates out the insitu logic from the repointing-based logic

The ENA teams need repoint files to produce their files, so we want to add a different trigger based on repoint files arriving for them. Sometimes we get delivered several repoint files in quick succession and we don't want to kick off each of those jobs every time so we put in a check on only running the job with the latest repoint file in the bucket and flooring the times to the nearest hour for what data we are querying between.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
@@ -23,6 +23,9 @@ else:
     logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+# ENA imagers group by repointing
+REPOINTING_INSTRUMENTS = {"glows", "hi", "lo", "ultra"}
+
 
 def setup_environment():
     """Get secrets and set necessary config values."""
@@ -51,9 +54,15 @@ def setup_environment():
     imap_data_access.config["DATA_DIR"] = Path("/tmp")  # noqa: S108
 
 
-def get_latest_repoint_file():
+def get_latest_repoint_file_and_query_times():
     """Retrieve the latest repointing file from S3.
 
+    Additionally, determine the time range to query for packets based upon
+    the latest repointing file and the previous repointing file that is
+    outside of 1-hour before the latest repointing file.
+
+    Notes
+    -----
     If we want to connect to the database later, we could use the RepointingFile
     database table to get this value instead.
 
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
```diff
@@ -105,8 +138,28 @@ def lambda_handler(event, context):
     """
     setup_environment()
 
-    repointing_key = get_latest_repoint_file()
-    imap_data_access.config["DATA_DIR"]
+    # If we were triggered by an s3 repointing event, download ENA data
```
tech3371: can we update function document now that you have updated to trigger on repoint file ingestion?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
```diff
@@ -105,8 +138,28 @@ def lambda_handler(event, context):
...
+    if "Records" in event:
+        event_repoint_key = event["Records"][0]["s3"]["object"]["key"]
+        download_ena_data(event_repoint_key)
+    else:
+        download_insitu_data()
+
+    return {"statusCode": 200, "body": "Packet downloader finished."}
+
+
+def download_ena_data(event_repoint_key):
```
tech3371, line 122: little preference, can you move these function above `lambda_handler`?

tech3371 (approved):
looks good. Minor comments.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
greglucas: Updated the docstring when I moved this.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/packet_downloader.py
greglucas: Yep, just pushed a commit with the rearrangement.

greglucas:
I'm going to merge this so we can try it out 🤞

PR #987, opened 2025-11-06
I-ALiRT - Improvements for DynamoDB API

### LINKED ISSUE #944: I-ALiRT - Add last_eval key for SWPC
### Summary

In ialirt_db_query_api.py:

    items = response.get("Items", [])
    processed_items = [process_item_types(item) for item in items]

    result = {
        "items": processed_items,
    }

    # Append LastEvaluatedKey if more data is available (pagination)
    last_evaluated_key = response.get("LastEvaluatedKey")
    if last_evaluated_key:
        result["last_evaluated_key"] = process_item_types(last_evaluated_key)

    return {"statusCode": 200, "body": json.dumps(result)}

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

### LINKED ISSUE #957: I-ALiRT API Suggestions by MAG
### Summary

Hi Laura
 
We have noticed you have added SPICE to the JSON of the I-ALiRT data - very cool.
 

Image

 
Just to let you know that this has broken 2 bits of code that we use to pull I-ALiRT data, and there is nothing like a version number we can use to work out how to decode different versions or a type to tell the records araprt easily. We have fixed our stuff but is this the final version of the data? Having 2 different record types with different schemas in a JSON array does make it much harder for any API data consumers to decode and parse the data, especially if there is not a discriminator column that is common to both document types allowing you to filter them.
 
Maybe this is how it will stay and that is ok! But if you are still planning changes or a "v2" I might suggest some options which make it easier to code a client for the data:
Add a common dicriminator field like type="science" or type="spice" to all document types output by the API. Docs with the same type have the same basic shape. this is easy to add now, breaks nothing and helps immediately
Make all items in any array have the same shape.Breaking change though
Return a HTTP header, or a field in the response/payload somewhere that has a version number in it so that the API minimally tells the client a version number so that new document schema means the version number changes. More on that below but HTTP header is non breaking
 
An example ideal ialirt JSON doc. I really like the way the [JSON API](https://nam10.safelinks.protection.outlook.com/?url=https%3A%2F%2Fjsonapi.org%2F&data=05%7C02%7Claura.sandoval%40lasp.colorado.edu%7Cd1d765c62b7e44e7559c08de10981387%7Cb1ddd7831dbc446b878fa7b03478ea9e%7C1%7C0%7C638966442557150239%7CUnknown%7CTWFpbGZsb3d8eyJFbXB0eU1hcGkiOnRydWUsIlYiOiIwLjAuMDAwMCIsIlAiOiJXaW4zMiIsIkFOIjoiTWFpbCIsIldUIjoyfQ%3D%3D%7C0%7C%7C%7C&sdata=qW6DSioMxOxUqEafEGyFWlgkyJA5YisPvwkpP8PVcY8%3D&reserved=0) spec defines JSON docs as it makes it really easy for API consumers (and clients can be generated automatically) but here is some changes that you could make that might make it easier now and with more room for tweaks in future. So my suggestion is not full compliant JSON API as that is a lot of work at this point, and would be breaking changes now. But i think not huge effort and would really help us users.
 
Return a doc with a "data" array and not an array at the root - this means you have plenty of places to add extra information, this can include links to other records (kinda solving tim's thing about getting the latest data with a link to it) . Also it adds included data like the SPICE that is not the main record. You can then easily link different records to the different SPICE. You can also then add any keys you like to the root doc to convey metadata without breaking the data array that holds the vital stuff. 
 

Image

 
So this is lots of feedback - and I know lots of big changes. You may want some/none of them and thats ok! Just throwing out some ideas. But better to make changes sooner than later if they are to the main record structure.
 
Lastly if there are any more schema changes we would love an email or heads up before reelase if that is possible.
 
Thanks for the API with our data in!
 
Alastair

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
After the redesign of the DynamoDB and also feedback from various users I am striving to create an api that:

1. Is compatible with the current API since folks are already using it
2. Takes into account really great recommendations from the MAG team: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/957
3. Aligns with our goals of lower latency
4. Creates a last_evaluated_key so that pagination is simplified for users: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/944

## Updated Files
- README.md
   - minor update to example
- insert_kernels.py
   - Fixed bug so that new database table the SPICE kernels will be ingested because previously they did not contain the necessary partition and sort key.
   - I also realized that I do not currently have a way to associate the science data with the proper spice kernel. So this is updated.
- ialirt_data_query_api.py
   - This is the API. It should have the following structure below when used. It includes the ability to paginate. It also provides the user with a more structure view of the data (separating metadata from actual data).

```
{
    "statusCode": 200,
    "headers": {
        "Content-Type": "application/json"
    },
    "body": {
        "meta": {
            "count": 2,
            "type": "science",
            "instrument": "mag",
            "has_more": True,
            "last_evaluated_key": '{"instrument": "mag", "time_utc": "2025-11-06T16:20:01"}'
        },
        "links": {
            "self": "https://ialirt.imap-mission.com/api-key/ialirt-db-query?instrument=mag",
            "next": (
                "https://ialirt.imap-mission.com/api-key/ialirt-db-query?"
                "instrument=mag&last_evaluated_key=%7B%22instrument%22%3A%20%22mag%22%2C"
                "%20%22time_utc%22%3A%20%222025-11-06T16%3A20%3A01%22%7D"
            )
        },
        "data": [
            {"instrument": "mag", "time_utc": "2025-11-06T16:20:00", "data": "item0"},
            {"instrument": "mag", "time_utc": "2025-11-06T16:20:01", "data": "item1"}
        ]
    }
}
```

## Testing
- test_ialirt_data_query_api.py
- test_insert_kernels.py

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
@@ -13,7 +13,7 @@ from boto3.dynamodb.conditions import Key
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
-table_name = os.environ.get("ALGORITHM_TABLE")
+table_name = os.environ.get("DATA_TABLE")
 region = os.environ.get("AWS_DEFAULT_REGION", "us-west-2")
 dynamodb = boto3.resource("dynamodb", region_name=region)
 table = dynamodb.Table(table_name)
@@ -45,7 +45,7 @@ def _error(code, message):
     }
 
 
-def build_next_url(event, last_evaluated_key, query_params):
+def build_next_url(event, last_evaluated_key):
     """Build the next URL for pagination.
 
     Parameters
@@ -183,7 +183,7 @@ def lambda_handler(event, context):
     encoded_lek = json.dumps(last_evaluated_key) if last_evaluated_key else None
     has_more = last_evaluated_key is not None
 
-    next_url = build_next_url(event, last_evaluated_key, params)
+    next_url = build_next_url(event, last_evaluated_key)
 
     json_body = json.dumps(
         {
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
@@ -174,12 +174,7 @@ def lambda_handler(event, context):
         items.extend(response.get("Items", []))
         query_time_total += t2 - t1
 
-    last_evaluated_key = response.get("LastEvaluatedKey")
-
     t3 = time.perf_counter()
-
-    encoded_lek = json.dumps(last_evaluated_key) if last_evaluated_key else None
-    has_more = last_evaluated_key is not None
 
     json_body = json.dumps(
         {
@@ -187,8 +182,6 @@ def lambda_handler(event, context):
                 "count": len(items),
                 "type": meta_type,
                 "instrument": meta_instrument,
-                "has_more": has_more,
-                "last_evaluated_key": encoded_lek,
             },
             "data": items,
         },
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
@@ -5,6 +5,7 @@ import logging
 import os
 import time
 from datetime import datetime, timedelta, timezone
+from urllib.parse import urlencode, quote_plus
 
 import boto3
 from boto3.dynamodb.conditions import Key
@@ -44,6 +45,36 @@ def _error(code, message):
     }
 
 
+def build_next_url(event, last_evaluated_key, query_params):
+    """Build the next URL for pagination.
+
+    Parameters
+    ----------
+    event : dict
+        The Lambda event object.
+    last_evaluated_key : dict
+        The last evaluated key from DynamoDB query.
+
+    Returns
+    -------
+    next_url : str
+        The URL for the next page of results.
+    """
+    query_params = event.get("queryStringParameters", {})
+    if last_evaluated_key:
+        query_params["last_evaluated_key"] = quote_plus(json.dumps(last_evaluated_key))
+
+    query_string = urlencode(query_params)
+
+    headers = event.get("headers", {})
+    host = headers.get("host", "")
+    path = event.get("requestContext", {}).get("path", "")
+
+    next_url = f"https://{host}{path}?{query_string}"
+
+    return next_url
+
+
 def lambda_handler(event, context):
     """Create metadata and add it to the database.
 
@@ -79,12 +110,14 @@ def lambda_handler(event, context):
         return _error(400, f"Unexpected parameters: {', '.join(unexpected)}")
 
     if not params.get("instrument"):
-        logger.info("No instrument specified, defaulting to all instruments")
         meta_instrument = "all"
         meta_type = "science"
-    elif params["instrument"] == "spice" or params["instrument"].endswith("hk"):
-        meta_instrument = "none"
-        meta_type = params["instrument"]
+    elif params["instrument"] == "spice":
+        meta_instrument = "spice"
+        meta_type = "spice"
+    elif params["instrument"].endswith("hk"):
+        meta_instrument = params["instrument"]
+        meta_type = "hk"
     else:
         meta_instrument = params["instrument"]
         meta_type = "science"
@@ -107,9 +140,6 @@ def lambda_handler(event, context):
     for instrument in instruments:
         key_expr = Key("instrument").eq(instrument)
         query_kwargs = {"KeyConditionExpression": key_expr}
-
-        if params.get("last_evaluated_key"):
-            query_kwargs["ExclusiveStartKey"] = json.loads(params["last_evaluated_key"])
 
         if any(
             param in params
@@ -136,6 +166,10 @@ def lambda_handler(event, context):
             query_kwargs["KeyConditionExpression"] &= Key("time_utc").between(
                 one_minute_ago.isoformat(), now.isoformat()
             )
+
+        if params.get("last_evaluated_key") and len(instruments) == 1:
+            query_kwargs["ExclusiveStartKey"] = json.loads(params["last_evaluated_key"])
+
         t1 = time.perf_counter()
         response = table.query(**query_kwargs)
         t2 = time.perf_counter()
@@ -145,18 +179,29 @@ def lambda_handler(event, context):
     last_evaluated_key = response.get("LastEvaluatedKey")
 
     t3 = time.perf_counter()
+
+    encoded_lek = json.dumps(last_evaluated_key) if last_evaluated_key else None
+    has_more = last_evaluated_key is not None
+
+    next_url = build_next_url(event, last_evaluated_key, params)
+
     json_body = json.dumps(
         {
             "meta": {
                 "count": len(items),
                 "type": meta_type,
                 "instrument": meta_instrument,
-                "last_evaluated_key": last_evaluated_key,
+                "has_more": has_more,
+                "last_evaluated_key": encoded_lek,
             },
+            "links": {
+                "next": next_url
+            } if has_more else {},
             "data": items,
         },
         default=str,
     )
+
     t4 = time.perf_counter()
 
     logger.info(
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
@@ -5,7 +5,7 @@ import logging
 import os
 import time
 from datetime import datetime, timedelta, timezone
-from urllib.parse import quote_plus, unquote_plus, urlencode
+from urllib.parse import unquote_plus
 
 import boto3
 from boto3.dynamodb.conditions import Key
@@ -40,12 +40,22 @@ def apply_time_filters(params: dict, query_kwargs: dict) -> Key:
     end = params.get("time_utc_end") or params.get("met_in_utc_end")
 
     if start and end:
-        key_expr &= Key("time_utc").between(start, end)
+        start_dt = datetime.fromisoformat(start)
+        end_dt = datetime.fromisoformat(end)
+        if end_dt - start_dt > timedelta(days=1):
+            return _error(400, "Start and end time cannot exceed 1 day apart.")
     elif start:
-        key_expr &= Key("time_utc").gte(start)
-    else:
-        return _error(400, "End time provided without start time")
+        # Calculate end to be 1 day later.
+        start_dt = datetime.fromisoformat(start)
+        end_dt = start_dt + timedelta(days=1)
+        end = end_dt.strftime("%Y-%m-%dT%H:%M:%S")
+    elif end:
+        # Calculate start to be 1 day earlier.
+        end_dt = datetime.fromisoformat(end)
+        start_dt = end_dt - timedelta(days=1)
+        start = start_dt.strftime("%Y-%m-%dT%H:%M:%S")
 
+    key_expr &= Key("time_utc").between(start, end)
     query_kwargs["KeyConditionExpression"] = key_expr
 
     return key_expr
@@ -73,37 +83,7 @@ def _error(code: int, message: str) -> dict:
     }
 
 
-def build_next_url(event: dict, last_evaluated_key: dict) -> str:
-    """Build the next URL for pagination.
-
-    Parameters
-    ----------
-    event : dict
-        The Lambda event object.
-    last_evaluated_key : dict
-        The last evaluated key from DynamoDB query.
-
-    Returns
-    -------
-    next_url : str
-        The URL for the next page of results.
-    """
-    query_params = event.get("queryStringParameters") or {}
-    if last_evaluated_key:
-        query_params["last_evaluated_key"] = quote_plus(json.dumps(last_evaluated_key))
-
-    query_string = urlencode(query_params)
-
-    headers = event.get("headers", {})
-    host = headers.get("host", "")
-    path = event.get("requestContext", {}).get("path", "")
-
-    next_url = f"https://{host}{path}?{query_string}"
-
-    return next_url
-
-
-def lambda_handler(event, context):  # noqa: PLR0915
+def lambda_handler(event, context):
     """Create metadata and add it to the database.
 
     This function is an event handler for s3 ingest bucket.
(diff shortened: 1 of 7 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,249 @@
...
+    -------
+    key_expr : Key
+        The updated key expression with time filters applied.
+    """
+    key_expr = query_kwargs["KeyConditionExpression"]
+
+    start = params.get("time_utc_start") or params.get("met_in_utc_start")
+    end = params.get("time_utc_end") or params.get("met_in_utc_end")
+
+    if start and end:
```
greglucas, line 51: Do you want to do any validation on the times? Something like `datetime.strptime(start, "YYYY-mm-ddTHH:MM:SS")` or whatever format you are expecting. Otherwise I'm uncertain whether you could get some weird situations with someone saying time_utc_start="1" to essentially do a full table scan because it would return results. I feel like I am always struggling with datetimes and formatting expectations with APIs, so a nice error message back saying what you expect a user to submit can be really helpful.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,249 @@
...
+        key_expr &= Key("time_utc").between(start, end)
+    elif start:
+        key_expr &= Key("time_utc").gte(start)
+    else:
+        return _error(400, "End time provided without start time")
+
+    query_kwargs["KeyConditionExpression"] = key_expr
+
+    return key_expr
```
greglucas: return the dictionary that has been modified here instead? Otherwise it has been modified at a distance.
```suggestion
    return query_kwargs
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,249 @@
...
+        The Lambda event object.
+    last_evaluated_key : dict
+        The last evaluated key from DynamoDB query.
+
+    Returns
+    -------
+    next_url : str
+        The URL for the next page of results.
+    """
+    query_params = event.get("queryStringParameters") or {}
```
greglucas:
```suggestion
    query_params = event.get("queryStringParameters", {})
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,249 @@
...
+        The error dictionary.
+    """
+    return {
+        "statusCode": code,
+        "body": json.dumps({"message": message}),
+        "headers": {"Content-Type": "application/json"},
+    }
+
+
+def build_next_url(event: dict, last_evaluated_key: dict) -> str:
```
greglucas: To me this seems like extra effort that could be hard to maintain. It seems like you are trying to build the subsequent URLs for users rather than having them manager the dates/times on their end. I guess I feel like we should try and limit the request to an amount of data you know you can return and put it on someone else to manage the subsequent requests. Like "No more than 1-day of data at a time", and then a user can decide how to chunk up and make their queries to satisfy that requirement. We don't want to manage their queries for them.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,249 @@
...
+
+    # Get instrument or default to all.
+    requested_instrument = params.get("instrument")
+    instruments = (
+        [requested_instrument]
+        if requested_instrument
+        else ["hit", "mag", "codice_lo", "codice_hi", "swapi", "swe"]
+    )
+
+    # Pagination only allowed for one instrument
```
greglucas: Oof, yeah this is another reason that this would add complexity IMO and we should drop the capability.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,249 @@
...
+                "time_utc_start",
+                "time_utc_end",
+                "met_in_utc_start",
+                "met_in_utc_end",
+            )
+        ):
+            result = apply_time_filters(params, query_kwargs)
+            # Checks if there was an error.
+            if isinstance(result, dict):
+                return result
```
greglucas: Oh interesting, I see now why you were returning only the key. I guess another option is to not return anything and just return None and do `if is not None` here instead of instance check.

This is a bit misleading currently with the different return styles for error / success states.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,249 @@
...
+        else:
+            # Get latest 1 minute if not specified.
+            logger.info(
+                "No time range specified, defaulting to last "
+                "1 minute for instrument: %s",
```
greglucas: one hour? 1-minute seems awfully short to me.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,249 @@
...
+
+        if params.get("last_evaluated_key"):
+            raw_last_evaluated_key = params["last_evaluated_key"]
+            query_kwargs["ExclusiveStartKey"] = json.loads(
+                unquote_plus(raw_last_evaluated_key)
+            )
+
+        t1 = time.perf_counter()
+        response = table.query(**query_kwargs)
+        t2 = time.perf_counter()
```
greglucas, line 191: Are we planning on taking these out eventually?

It might be nice for the time being to put in logs for individual query times temporarily to see what it looks like as we go along too. Something like:
`logger.info(f"Querying {instrument} table between times {start} - {end} took {t2-t1}s")`

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -540,22 +564,37 @@ def insert_kernels(dependency_inputs, algorithm_table):
...
-        "met": int(met),
-        "met_in_utc": met_to_utc(met).split(".")[0],
-        "ttj2000ns": int(met_to_ttj2000ns(met)),
-        "instrument": "spice",
-        "last_modified": last_modified_utc,
-        "spice_kernels": spice_kernels,
-    }
+    # Will return the same kernel_set_id for the same set of kernels.
+    payload = json.dumps(spice_kernels, sort_keys=True).encode("utf-8")
+    kernel_set_id = hashlib.md5(payload).hexdigest()[:16]  # noqa: S324
```
greglucas: Do we really need the kernel_set_id put in with all of the instruments too? I feel like someone could back this out in other ways if they wanted. Like we know every 1-minute of MAG data is processed with the set of kernels corresponding to this SPICE time.

I'm not sure it makes much of a difference either way, but just wanted to throw this out there because it seems like it adds some complexity and table size without gaining a lot IMO.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -540,22 +564,37 @@ def insert_kernels(dependency_inputs, algorithm_table):
...
+            "spice_kernels": spice_kernels,
+            "kernel_set_id": kernel_set_id,
+        }
+    else:
+        kernel_item = {
+            "instrument": "spice",
+            "time_utc": met_to_utc(met).split(".")[0],
+            "spice_kernels": spice_kernels,
+            "ttj2000ns": int(met_to_ttj2000ns(met)),
+            "kernel_set_id": kernel_set_id,
```
greglucas: Thinking about this now actually. Maybe the "kernel_set_id" is equivalent to "spice kernels at this time" so you could tell someone to go and get that specific PK+SK which is actually a fast and efficient query if needed. Searching for a "kernel_set_id" would need a full scan.

greglucas (commented):
Overall this looks nice! My two minor quibbles are with the spice referencing and the last-evaluated key.
* I would remove the LastEvaluatedKey. NOAA has added a lot of scope creep with that IMO. I think that can be managed pretty easily externally. We can/should probably look at limiting the query range someone can request to give them a nicer error message if they are requesting all data or something like that.
* I'd remove the SPICE reference and let people search for the corresponding times themselves instead, OR put the SPICE met-time in the instrument reference so it is an easy query and not a scan to find the set-id.

Do you know how much data this version will allow you to query at once and how it compares to the previous API? I'm curious how much better (if at all) it is.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
laspsandoval: If you recall the conversation that we had via Slack with Doug Lindholm, I do think that it would be desirable for the webteam to have the Last Evaluated feature. 

"LaTiS has to page requests to the ialirt service due to the limited number of results it can return (DynamoDB limitation?) The ialirt service does not preserve the the paging info from DynamoDB, so LaTiS just keeps reading batches until it gets an empty one. Thus the earlier start times before there is data will return an empty result and LaTiS will stop trying. Yet another reason why direct access to the DynamoDB may simplify some things.I recall considering this problem. I might be able to at least keep trying to page until I hit the upper time bound, or current time."

laspsandoval: But I see what you are saying. I think you are right. I'll see if I can set a time limit of 1 day for now and see how that goes. It really depends on the instrument. for how much data they should be able to get.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
laspsandoval: True.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
laspsandoval: Removed this function.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
laspsandoval: Nice idea.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: That was what I thought initially, but after considering it more, I don't think it would be very easy. The reason is that the kernels use the insert time as their sort key (time_utc). They have to because they don't have a spacecraft packet creation time, which is what the instruments use.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
laspsandoval: Interesting idea. Let me work on it and tell me what you think.

laspsandoval:
> Overall this looks nice! My two minor quibbles are with the spice referencing and the last-evaluated key.
> 
> * I would remove the LastEvaluatedKey. NOAA has added a lot of scope creep with that IMO. I think that can be managed pretty easily externally. We can/should probably look at limiting the query range someone can request to give them a nicer error message if they are requesting all data or something like that.
> * I'd remove the SPICE reference and let people search for the corresponding times themselves instead, OR put the SPICE met-time in the instrument reference so it is an easy query and not a scan to find the set-id.
> 
> Do you know how much data this version will allow you to query at once and how it compares to the previous API? I'm curious how much better (if at all) it is.

It should be very similar to the other api for the MAG data. Except now at least the mag hk will be separate. And we have also reduced the number of decimals stores for MAG (in imap_processing). 

The other instruments should have far better performance.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,245 @@
...
+
+    Returns
+    -------
+    datetime | None
+        A `datetime` object if the timestamp is valid,
+        or `None` if the format is invalid.
+    """
+    try:
+        return datetime.fromisoformat(ts)
+    except Exception:
```
greglucas: There is a lot of `return None` / `if X is None` checks throughout. Would it be easier to follow by raising your own Exception here that you can then catch rather than having different return types?

```python
class BadTimeException(Exception):
    pass
```
In your main code function something like this:
```
try:
    ...
except BadTimeException:
    return _error(...)
```

Catch the smallest scope which is ValueError here.
```suggestion
    except ValueError:
        raise BadTimeException from e
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,245 @@
...
+        ):
+            result = apply_time_filters(params, query_kwargs)
+            if result is None:
+                return _error(400, "Invalid time format: expected YYYY-MM-DDTHH:MM:SS")
+            query_kwargs, range_start, range_end = result
+            # Checks if the max time range is exceeded.
+            if abs(
+                datetime.fromisoformat(range_end) - datetime.fromisoformat(range_start)
+            ) > timedelta(days=1):
+                return _error(400, "Start and end time cannot exceed 1 day apart.")
```
greglucas: Can you put this check in the `apply_time_filters` call as well so all time handling is in one location? You could maybe raise the `BadTimeException` with the message you want to deliver to the end user depending on the case that fails and then reformat that into the json response.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -0,0 +1,245 @@
...
+                "No time range specified, defaulting to last 1 hour for instrument: %s",
+                instrument,
+            )
+            now = datetime.now(timezone.utc)
+            one_hour_ago = now - timedelta(hours=1)
+            query_kwargs["KeyConditionExpression"] &= Key("time_utc").between(
+                one_hour_ago.isoformat(), now.isoformat()
+            )
+            range_start = one_hour_ago.isoformat()
+            range_end = now.isoformat()
```
greglucas: This too I think can go into the final clause of apply_time_filters to say if there was no start/end use the default code you have here.

greglucas (approved):
Nice, I think this looks good Laura! I have some minor suggestions on the time handling, but it is really just reorganization/flow logic and not substantial here.

Approving so we can get this in and test it out, I'm excited to see what the performance is like on this.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
laspsandoval: Nice idea. I like this better.

PR #988, opened 2025-11-06
Adding SPICE dependencies for MAG L1D/L2

maxinelasp:
# Change Summary
Adding new SPICE dependencies for MAG L1D and L2 to address issues with frame transformations to DSRF, GSE, GSM, and RTN

Tagging @mfacchinelli for visibility

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
@@ -785,6 +785,9 @@ pointing_attitude, spice, historical, mag, l1d, norm-srf, HARD, DOWNSTREAM
 spin, spin, historical, mag, l1d, norm-srf, HARD, DOWNSTREAM
 leapseconds, spice, historical, mag, l1d, norm-srf, HARD_NO_TRIGGER, DOWNSTREAM
 spacecraft_clock, spice, historical, mag, l1d, norm-srf, HARD_NO_TRIGGER, DOWNSTREAM
+ephemeris_reconstructed, spice, historical, mag, l1d, norm-srf, HARD, DOWNSTREAM
+ephemeris_predicted, spice, historical, mag, l1d, norm-srf, HARD_NO_TRIGGER, DOWNSTREAM
+planetary_ephemeris, spice, historical, mag, l1d, norm-srf, HARD_NO_TRIGGER, DOWNSTREAM
 
 # L1C/L1B -> L2
 # MAG L2 and L1D look pretty similar.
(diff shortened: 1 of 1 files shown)


tmplummer (approved):
LGTM. Did you test to make sure that these new dependencies are sufficient for the SPICE needs?

mfacchinelli:
Thanks @maxinelasp!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
```diff
@@ -785,6 +785,9 @@ pointing_attitude, spice, historical, mag, l1d, norm-srf, HARD, DOWNSTREAM
 spin, spin, historical, mag, l1d, norm-srf, HARD, DOWNSTREAM
 leapseconds, spice, historical, mag, l1d, norm-srf, HARD_NO_TRIGGER, DOWNSTREAM
 spacecraft_clock, spice, historical, mag, l1d, norm-srf, HARD_NO_TRIGGER, DOWNSTREAM
+ephemeris_reconstructed, spice, historical, mag, l1d, norm-srf, HARD, DOWNSTREAM
+ephemeris_predicted, spice, historical, mag, l1d, norm-srf, HARD_NO_TRIGGER, DOWNSTREAM
```
tech3371, line 789: I feel like you don't need predict since you are looking for reconstructured kernel. But I haven't looked closely.

tech3371 (approved):

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
maxinelasp: Tim told me I could use both - I want to have the latest kernel at all times, and he said predict was more frequent.

tmplummer: Right. The `ephemeris_reconstructed` only comes once per week. So, if you only use reconstructed, processing will be waiting for that kernel delivery. Using both will allow for processing to run as soon as the other dependencies are present... Oh, but I think that you need to change the dependency on the `ephemeris_reconstructed` to SOFT?

I'm confused now. Hi L1C has the following:
```
ephemeris_reconstructed, spice, historical, hi, l1c, 90sensor-pset, HARD_NO_TRIGGER, DOWNSTREAM
ephemeris_predicted, spice, best, hi, l1c, 90sensor-pset, HARD_NO_TRIGGER, DOWNSTREAM
ephemeris_90days, spice, best, hi, l1c, 90sensor-pset, HARD_NO_TRIGGER, DOWNSTREAM
```
This seems to work for Hi without waiting for the ephemeris_reconstructed kernel.

maxinelasp: Yes, I used the Hi settings as an example. Should I update it to SOFT? your logic makes sense to me... it looks like currently, this runs only with the RECON files, even though both are available: `Found metakernel files: ['naif020.tls', 'imap_011.tf', 'imap_science_108.tf', 'imap_sclk_008.tsc', 'de440.bsp', 'imap_recon_20250415_20260415_v01.bsp', 'imap_2025_105_2026_105_01.ah.bc', 'imap_dps_2025_105_2026_105_009.ah.bc']. Adding to collection.`

tmplummer: I guess that someone with more knowledge about the dependency types will need to weigh in. I just looked at a Hi L1C job from this morning and I see the following spice dependencies:
```
['naif0012.tls', 'imap_120.tf', 'imap_science_100.tf', 'imap_sclk_0036.tsc', 'de440.bsp', 'imap_pred_od009_20251104_20251216_v01.bsp', 'imap_2025_308_2025_309_001.ah.bc', 'imap_2025_309_2025_310_001.ah.bc', 'imap_2025_310_2025_310_001.ah.bc', 'imap_dps_2025_308_2025_309_001.ah.bc', 'imap_dps_2025_307_2025_309_001.ah.bc', 'imap_dps_2025_309_2025_310_001.ah.bc', 'imap_dps_2025_308_2025_310_001.ah.bc']
```
So, it happily ran with only the predicted ephemeris file: `imap_pred_od009_20251104_20251216_v01.bsp`. To me that means that there is either a bug and that job shouldn't have been triggered, or it is implemented such that any ephemeris satisfies dependencies. Either way, you should be good as implemented. If it is a bug, then we will need to adjust all instruments ephemeris dependencies when that bug is fixed.

PR #995, opened 2025-11-11
MNT: Keep processing I-ALiRT instruments catching errors

### LINKED ISSUE #993: I-ALiRT processing shouldn't fail on one instrument
If one instrument fails processing, we shouldn't raise immediately. This means that other instruments that may have had data are not getting run. We can add a try/except to catch errors and then log / re-raise outside the loop if needed.

greglucas:
We want to keep processing as we go through the instrument loop so that one instrument doesn't stop another one from processing. We accumulate the errors and raise at the end if there are any logging all relevant errors during the process.

closes #993


laspsandoval (approved):
Nice. Thanks Greg.

PR #997, opened 2025-11-12
FIX: I-ALiRT codice ancillary lookup naming

greglucas:
codicelo is not a valid instrument, we only have one codice on the main mission, so we need to trim off the lo/hi distinction that I-ALiRT uses before querying.


greglucas:
Closing in favor of #998

PR #1003, opened 2025-11-13
BUG: SPICE trigger dependencies

lacoak21:
# Change Summary

## Overview
Tenzin noticed a bug in swe l2:

`There is another thing going on with SWE L2 processing. Sometimes, job gets more than one science files. I think that’s happening because spin file kicked off the batch starter and found few files waiting.`

This was happening because the dependencies should be "filtered" if the trigger file was a SPICE, Spin, Ancillary or Repoint file. Basically anything but a Science file. Science files have an exact date associated with them. The rest can have a range and therefore, can trigger multiple jobs. The flow should be: 

1. SPICE, Spin, Ancillary or Repoint  triggered batch starter
2. get upstream science files in that date range
3. For each upstream science file "filter" the dependencies for only ones needed for the start date of that science file

This will also help reduce the large metakernel queries that are timing out. 

## Updated Files
   - spacecraft_pointing_attitude_job :small helper function for repeated code
   - Filter dependencies for all file triggers except science

## Testing
Add a test for a spice event

### DIFF
--- a/tests/lambda_endpoints/test_batch_starter.py
+++ b/tests/lambda_endpoints/test_batch_starter.py
@@ -11,6 +11,7 @@ from unittest.mock import Mock, call, patch
 
 import imap_data_access
 import pytest
+from imap_data_access import SpinInput
 from imap_data_access.processing_input import (
     ProcessingInputCollection,
     ScienceInput,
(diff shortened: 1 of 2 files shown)


tests/lambda_endpoints/test_batch_starter.py
```diff
@@ -333,6 +334,120 @@ def test_lambda_handler_multiple_events(session, s3_client):
...
+
+    context = {"context": "sample_context"}
+    with (
+        patch.object(batch_starter, "BATCH_CLIENT", Mock()) as mock_batch_client,
+        patch.object(batch_starter, "generate_queue_url", return_value=False),
+    ):
+        lambda_handler(events, context)
+        # There should be 2 different jobs submitted for the two swe l1b files
+        assert mock_batch_client.submit_job.call_count == 2
+        # Assert_called_with only works on the last call
```
greglucas, line 400: No need to do this, but just as an FYI there is `mock_calls` that you can get to look at multiple calls and how each one was submitted.

https://github.com/IMAP-Science-Operations-Center/imap-data-access/blob/d8962f4b1e079773e509b3ad08a71fad19d7bd7e/tests/test_io.py#L434-L447

greglucas (approved):
Looks right to me.

> This will also help reduce the large metakernel queries that are timing out.

How will this help metakernel queries? That uses batch_starter? I figured it was only using DB queries directly itself.

tests/lambda_endpoints/test_batch_starter.py
lacoak21: Oh wow I swear I tried finding this in the documentation! I should have known there was a way

lacoak21:
> 
> How will this help metakernel queries? That uses batch_starter? I figured it was only using DB queries directly itself.

Yeah this lambda handler does just query the db but it is causing timeouts if the date range is too large : https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/b0b4bb85fae753533723e8ffb045589c0d5220b6/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency.py#L949C16-L961C18

greglucas:
Ahh, OK I see. You are saying this will help reduce the failures in this function due to the metakernel queries.

I thought you were saying this would speed up the metakernel queries themselves.

lacoak21:
> Ahh, OK I see. You are saying this will help reduce the failures in this function due to the metakernel queries.
> 
> I thought you were saying this would speed up the metakernel queries themselves.

Ohh I see now. I wish it did!

tech3371:
Thank you for getting this in!

PR #1004, opened 2025-11-14
FIX: Insert into table requires time_utc sort key to be present

greglucas:
# Change Summary

## Overview

The algorithms have their time key reformatted in the process_algorithms function when inserting into the different tables, but the SPICE keys didn't get swapped. We need to swap the SPICE key as well, depending on the table name we are inserting into. I think this is the only thing missing as we have the "instrument": "spice" already in there.


laspsandoval:
Yep. I have that in my other PR. Approved! Thanks.

laspsandoval (approved):

PR #1009, opened 2025-11-18
CI: Add L3 container promotion from dev to prod

### LINKED ISSUE #952: Add automated promotion of L3 containers to production
The L3 containers get built and pushed to dev from a separate repository. After #948 we have a way to automatically promote containers from dev to prod. But, we need an extra flag/option to indicate whether it should include the L3 repositories or not. This would look like `{instrument}-l3-repo` instead of `{instrument}-repo`. I think it could be a checkbox/flag on the actions workflow state and then that would add the extra `-l3` string into the repo names.

greglucas:
Adds the "-l3" to the repository name if we are promoting L3 containers which can be selected by a user when deploying.

closes #952

### DIFF
--- a/.github/workflows/promote_containers_dev_to_prod.yml
+++ b/.github/workflows/promote_containers_dev_to_prod.yml
@@ -31,6 +31,11 @@ on:
         required: false
         type: string
         default: 'latest'
+      l3_data:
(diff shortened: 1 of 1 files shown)


tech3371 (approved):
Thank you for getting this in so quick!!

PR #1015, opened 2025-12-01
Make SWAPI L3a PUI and Proton trigger on SPICE changes

pleasant-menlo:
Change ephemeris_reconstructed, attitude_history, and pointing_attitude to be a HARD dependency for SWAPI L3a proton-sw and pui-he.


lacoak21 (approved):

lacoak21:
LGTM

greglucas:
@pleasant-menlo another thing that stuck out to me is that the `ephemeris_reconstructed` I think gets delivered ~weekly so it might lag significantly behind other products. You might want to put an `ephemeris_predicted` in with a `HARD_NO_TRIGGER` here too which I think some other teams are doing, so that you will use the predicted ephemeris the first go-around, and then reprocess once the reconstructed data arrives.

There was some e-mails going around a long time ago where teams aren't too concerned with the difference between predicted/reconstructed ephemeris, they are going to be "close enough". The attitude is what the teams really care about.

PR #1017, opened 2025-12-02
I-ALiRT - data query api updates

laspsandoval:
# Change Summary

## Overview
The webteam provided some feedback on the new api:
1. Remove the query time limit
2. Make it to where a string value is not returned (this was corrected in the old api)

## Updated Files
- ialirt_data_query_api.py
   - Remove the query time limit
   - Make it to where a string value is not returned (this was corrected in the old api)

## Testing
- test_ialirt_data_query_api.py

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
@@ -5,6 +5,7 @@ import logging
 import os
 import time
 from datetime import datetime, timedelta, timezone
+from decimal import Decimal
 
 import boto3
 from boto3.dynamodb.conditions import Key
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -64,9 +103,6 @@ def apply_time_filters(params: dict, query_kwargs: dict) -> tuple:
         start_dt = now - timedelta(hours=1)
         end_dt = now
 
-    if abs(end_dt - start_dt) > timedelta(days=1):
-        raise BadTimeError("Start and end time cannot exceed 1 day apart.")
```
greglucas, line 68: What happens when you request too much data? I worry that we would give the user back some cryptic message here that doesn't help with identifying how much data a user can expect.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -207,7 +244,6 @@ def lambda_handler(event, context):
             },
             "data": items,
         },
-        default=str,
```
greglucas, line 210: Rather than handling the list/dict cases in the process_item_types function call, what if you used a custom encoder instead? So then each individual type would be encoded for you and you wouldn't have to worry about the list/dict cases additionally. Then I don't even think you'd need to worry about when/where to call the formatting either, just let the json dumping take care of the serialization.

```python
class DecimalEncoder(json.JSONEncoder):
    def default(self, obj):
        if isinstance(obj, Decimal):
            return f"{obj:.3f}" # Format Decimal to 3 decimal places
        # put case for float here as well?

        # Let the base class default method raise the TypeError for other types
        return super().default(self, obj)

json.dumps(..., cls=DecimalEncoder)
```

laspsandoval: I like this way better.

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -24,6 +25,22 @@ class BadTimeError(Exception):
...
+
+    def default(self, obj):
+        """Override JSON encoding for Decimal values."""
+        if isinstance(obj, Decimal):
+            # - If the Decimal is an integer, return int
+            # - Otherwise, float rounded to 3 decimal places
+            if obj % 1 == 0:
+                return int(obj)
+            return round(float(obj), 3)
+
```
greglucas, line 39: Do you also want to do this for float data types as well to round them, or is everything coming in as Decimal so you'll never have raw floats?

```suggestion
            return round(float(obj), 3)
    if isinstance(obj, float):
        return super().default(round(obj, 3))
```

greglucas (approved):

PR #1022, opened 2025-12-05
ENH: Update processing job information with a local command

### LINKED ISSUE #1021: Add local run command to API output
When getting the processing job details, it would be helpful to display a command a local user could use. Right now we have to add `imap_cli` as a command and remove the `--upload-to-sdc`. We can do that for the user in the response from the API.

See the internal IMAPSDCWEB-45 ticket for more discussion.

greglucas:
# Change Summary

Adds a `job_command` to the output. I'm not sure if people would prefer we replace the "container_command" here or leave it. I kind of prefer just removing it from the API response (still in the processing jobs DB table), so that we never show the `--upload-to-sdc` to external users.

closes #1021


lacoak21:
LGTM - I cant think of a reason for not replacing the container command (for the query) without the upload part.

lacoak21 (approved):

PR #1026, opened 2025-12-09
I-ALiRT - allow packets and archive to be downloaded by users with api key

laspsandoval:
# Change Summary

## Overview
Allow for packets and archive directories in s3 to be downloaded.

## Updated Files
- download_api.py
   - Allow for packets and archive directories in s3 to be downloaded.


laspsandoval:
Nevermind. I think its fine.

PR #1028, opened 2025-12-09
1025 organize s3 processing event workflow

### LINKED ISSUE #1025: Organize s3_processing_event workflow
### Algorithm Description:

Looking through batch_starter trying to identify where to make changes to allow for the Hi goodtimes dependency on multiple repointings, I found the multiple calls to `get_jobs` confusing. It seems to me that calling `get_jobs` with no start/end date is really wanting to just call `get_dependencies`. I am going to try to clean up and clarify the workflow in this ticket.

tmplummer:
# Change Summary

## Overview
I was getting confused about what the dependencies.get_jobs() function did and noticed that when called without star/end times, it was essentially a wrapper around get_dependencies so I decided move the logging and error handling into get_dependencies and call that directly when what is desired are dependencies from the dependency.json.

Those more knowledgeable about the design of this code are free to reject this change. I just thought it made the workflow more clear to have different functions for the two purposes that get_jobs was serving.

## Updated Files
   - Call dependency.get_dependencies() when querying for info from dependency json file.
   - Move logging/error handling into get_dependency()
   - Remove dual function of get_jobs() to provide dependency info.
   - Update tests to call get_dependency() instead of get_jobs in appropriate places.
   - Update tests to call get_dependency() instead of get_jobs in appropriate places.

Closes: #1025


maxinelasp (approved):
Good change!

lacoak21 (approved):
Agreed! This makes way more sense

PR #1035, opened 2025-12-12
I-ALiRT - Conceal certain HIT products and set public release date

laspsandoval:
# Change Summary

## Overview
Conceal certain HIT products and set public release date.

## Updated Files
- Minor docstring change
   - api_gateway_construct.py
- lambda_api_key_authorizer.py
   - Remove restrictions to database so that they are managed now by the lambda based on release date
- ialirt_data_query_api.py
   - Filters data products that the HIT team does not want to be released based on the scope of the api-key. Adds a release date so that users without an api key cannot see prior data.

## Testing
test_ialirt_data_query_api.py

### DIFF
--- a/sds_data_manager/constructs/api_gateway_construct.py
+++ b/sds_data_manager/constructs/api_gateway_construct.py
@@ -3,8 +3,8 @@
 Sets up api gateway, creates routes, and creates methods that are linked to the
 lambda function.
 
-An example of the format of the url: https://api.prod.imap-mission.com/query
-https://ialirt.prod.imap-mission.com/ialirt-log-query
+An example of the format of the url: https://api.imap-mission.com/query
+https://ialirt.imap-mission.com/ialirt-log-query
 """
 
 from typing import Optional
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
@@ -18,6 +18,18 @@ region = os.environ.get("AWS_DEFAULT_REGION", "us-west-2")
 dynamodb = boto3.resource("dynamodb", region_name=region)
 table = dynamodb.Table(table_name)
 
+FULL_SCOPES = {
+    "full",
+    "ialirt_scientist",
+}
+
+RESTRICTED_FIELDS = {
+    "hit_h_a_side_low_en",
+    "hit_h_a_side_med_en",
+    "hit_h_b_side_low_en",
+    "hit_h_b_side_med_en",
+}
+
 
 class BadTimeError(Exception):
     """Raised when the time filters are invalid."""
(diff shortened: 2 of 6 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -132,6 +151,35 @@ def _error(code: int, message: str) -> dict:
...
+    # If caller has full HIT access, do nothing
+    if scope in FULL_SCOPES:
+        return items
+
+    filtered_items = [
+        {k: v for k, v in item.items() if k not in RESTRICTED_FIELDS}
+        if item.get("instrument") == "hit"
+        else item
+        for item in items
+    ]
```
greglucas, line 175: Filtering items by API key or not reads a bit weird to me here. I wonder if these should instead be in a `hit-extra` PK or similar instead kind of like MAG has an extra spot for the other hk fields only the instrument team wants access to. Speaking of which, are those MAG fields only accessible by users with API Key scope?

I guess what I'm getting at is that the API Key seems to need to be passed in a lot of places here, where I think it could be at the top-level where you say "API Key gets you access to these extra `hit-hk`, `mag-hk`, `hit-extra` PKs and the rejection is immediate rather than flowing through all these routines for a filtering method. Not that the same request will show different values depending on whether a user is logged in or not.

laspsandoval: I see what you are saying. I think that the HIT team might release these products eventually, though.

greglucas: OK, my preference would be to only apply this to HIT then and not all instruments where you are looping/checking each individual item for a MAG request.

Something like outside of this function you would have a check based on the queried instrument:

```python
if instrument == "hit":
    items = filter_hit_items_by_scope()
```

sds_data_manager/constructs/api_gateway_construct.py
```diff
@@ -3,8 +3,8 @@
 Sets up api gateway, creates routes, and creates methods that are linked to the
 lambda function.
 
-An example of the format of the url: https://api.prod.imap-mission.com/query
-https://ialirt.prod.imap-mission.com/ialirt-log-query
+An example of the format of the url: https://api.imap-mission.com/query
```
lacoak21, line 6: Good catch

sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -84,6 +100,9 @@ def apply_time_filters(params: dict, query_kwargs: dict) -> tuple:
     start = start_dt.strftime("%Y-%m-%dT%H:%M:%S")
     end = end_dt.strftime("%Y-%m-%dT%H:%M:%S")
 
+    if not has_api_key and start < PUBLIC_CUTOFF_UTC:
```
lacoak21, line 103: Would it be worth logging a warning instead and setting the start time to public_cutoff_utc? I guess the user will know how to update the query so might not be worth it.

lacoak21 (approved):
Looks good!

PR #1038, opened 2025-12-17
I-ALiRT - Archive cadence

laspsandoval:
# Change Summary

## Overview
Pass in new database, handle pagination, update time to be exactly at midnight

## Updated Files
- stackbuilder.py
   - pass in new database
- ialirt_archive_construct.py
   - pass in new database
- ialirt_archive.py
   - pass in new database, handle pagination, update time to be exactly at midnight

## Testing
- test_ialirt_archive.py
- test_ialirt_archive_construct.py

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
@@ -3,7 +3,7 @@
 import json
 import logging
 import os
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, time, timedelta, timezone
 from pathlib import Path
 
 import boto3
@@ -14,6 +14,52 @@ from imap_processing.ialirt.utils.create_xarray import create_xarray_from_record
 
 logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
+
+
+INSTRUMENTS = ["mag", "codice_lo", "codice_hi", "hit", "swe", "swapi", "spacecraft"]
+
+
+def query_instrument(
+    data_table, instrument: str, start_iso: str, end_iso: str
+) -> list[dict]:
+    """Query database and handles pagination.
+
+    Parameters
+    ----------
+    data_table : ddb.Table
+        Algorithm database table.
+    instrument : str
+        Instrument name.
+    start_iso : str
+        Start date of query.
+    end_iso : str
+        End date of query.
+
+    Returns
+    -------
+    items : list
+        Items queried for the instrument.
+    """
+    items: list[dict] = []
+    last_key = None
+
+    while True:
+        kwargs = dict(
+            KeyConditionExpression=(
+                Key("instrument").eq(instrument)
+                & Key("time_utc").between(start_iso, end_iso)
+            ),
+        )
+        if last_key:
+            kwargs["ExclusiveStartKey"] = last_key
+
+        resp = data_table.query(**kwargs)
+        items.extend(resp.get("Items", []))
+        last_key = resp.get("LastEvaluatedKey")
+        if not last_key:
+            break
+
+    return items
 
 
 def lambda_handler(event, context):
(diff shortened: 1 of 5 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
```diff
@@ -32,43 +78,38 @@ def lambda_handler(event, context):
...
-        logger.info("No new data to process.")
-        return response
+    all_items = []
+    for inst in INSTRUMENTS:
+        inst_items = query_instrument(data_table, inst, start_iso, end_iso)
+        logger.info("%s: %d items", inst, len(inst_items))
+        all_items.extend(inst_items)
 
-    dataset = create_xarray_from_records(response["Items"])
+    dataset = create_xarray_from_records(all_items)
```
greglucas, line 121: Does this handle an empty dataset? If we don't get any ialirt packets for a day, would this still create all the expected fields with empty fill values, or do we need to exit out earlier above this so we don't write an empty file?

sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
```diff
@@ -32,43 +78,38 @@ def lambda_handler(event, context):
...
     dataset.attrs["Data_version"] = "000"
-    dataset.attrs["Start_date"] = yesterday.strftime("%Y%m%d")
+    dataset.attrs["Start_date"] = prior_day.strftime("%Y%m%d")
```
greglucas: `seven_days_ago`? prior_day sounds like yesterday to me still.

sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
```diff
@@ -16,6 +16,52 @@
 logger.setLevel(logging.INFO)
 
 
+INSTRUMENTS = ["mag", "codice_lo", "codice_hi", "hit", "swe", "swapi", "spacecraft"]
+
+
+def query_instrument(
```
greglucas, line 22: Don't we already have this in the main API-route? I feel like we should be using the same query function in both places so we don't get out of sync.

laspsandoval: ialirt_ingest.py queries S3 (packet filenames).

ialirt_data_query_api.py queries DynamoDB, but does not paginate.

This archive lambda needs to pull all items in a time window (1 day), so pagination is required.

sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
laspsandoval: Yes. Addressed that empty dataset now.

sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
```diff
@@ -32,43 +78,45 @@ def lambda_handler(event, context):
...
-            Key("apid").eq(478) & Key("last_modified").between(start_iso, end_iso)
-        ),
-    )
-
-    if not response["Items"]:
-        logger.info("No new data to process.")
-        return response
-
-    dataset = create_xarray_from_records(response["Items"])
+    target_date = (now - timedelta(days=7)).date()
```
lacoak21, line 94: Why is this now a week previous? Maybe add a comment?

lacoak21 (approved):
Just one question and comment suggestion

sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
laspsandoval: Ok. I'll add a comment. The reason is that the scientists are now talking about saving the I-ALiRT data to the solid state record onboard. This means that we will get all I-ALiRT data regardless of there being an immediate contact. But we will get that data during normal DSN (3 times per week). So I set this lambda to kick off nightly and grab data from a week ago instead of a day ago.

PR #1045, opened 2025-12-23
Update cadence job first trigger date

lacoak21:
# Change Summary

## Overview
Currently the first map job is scheduled to trigger on jan 1st. It should actually be three months after the start of phase E. I am not sure exactly what that date is, but for now this PR pushes the start date out until we can determine the correct date. 

## Updated Files
-sds_data_manager/constructs/instrument_lambdas.py
   - Update the first cadence job start date


tmplummer (approved):
Nice catch on not letting this run on the first. Will you please write an issue to capture this TODO?

PR #1050, opened 2026-01-07
Add new ancillary dependencies for SWAPI L3a

pleasant-menlo:
# Change Summary
Added new ancillary dependencies "helium-inflow-vector" and "hydrogen-inflow-vector" for all SWAPI L3a processing. The ancillaries are not used in proton-sw, and alpha-sw, but will cause the processor the fail if they are not present, so we have marked those as HARD_NO_TRIGGER. The files are required and used in calculations for the pui-he product.

## Updated Files
- dependency_config.csv
  - Added new ancillary dependencies "helium-inflow-vector" and "hydrogen-inflow-vector" for all SWAPI L3a processing


tech3371 (approved):
LGTM

PR #1058, opened 2026-01-14
Spice file monitoring

maxinelasp:
# Change Summary
This sets up an email notification system for missing SPICE files. If a file is 7 days stale, it will send an email to the SDC list. This number is configurable if we want a different span of days... I wasn't really sure what was reasonable. This should hopefully give us more visibility into missing SPICE files automatically.

## Overview

- Create a new lambda function to check the DB and emit CW metrics for files, which automatically triggers once a day
- Set up an alarm/notification system in CDK, including automatically pointing to the IMAP SDC email list

This should be pretty separate from the rest of the system, so overall a minor change. This makes three different alarms to read and alarm based on the metric for the SPICE data age.


tmplummer (commented):
This is great! We will probably want to add more files such as: repoint, predicted_ephemeris, and maybe others.
I think that the threshold for the existing kernels could be set to 4 days. This comes from the fact that we are supposed to have 3 contacts/week plus a bit of wiggle room.

Does there need to be any test coverage for lambda code?

maxinelasp:
Ok, I can add repoint and predicted_ephemeris (are they both in the normal SPICE table?) and change the threshold down. IMO this does not need a test since it's a nice notification service and not critical infra. I ran this in dev and it worked great. If the metric stops getting reported for whatever reason we'll get notified with an alarm email.

tech3371:
I will review this tomorrow if no one gets to this. Had a busy day today.

sds_data_manager/constructs/spice_monitoring_construct.py
```diff
@@ -0,0 +1,320 @@
...
+                },
+            )
+        )
+
+        return monitoring_lambda
+
+    def _create_schedule_rule(self):
+        """Create EventBridge rule to run Lambda daily."""
+        # Run daily at 10:00 AM UTC
+        schedule_rule = events.Rule(
```
sdhoyt, line 220: Curious why this looks like it's scheduled every 10 hours, but the metric is created once per day? Or is this the 10th hour of each day?

sdhoyt (approved):
I'm not the best person to review, but it looked good to me

sds_data_manager/constructs/spice_monitoring_construct.py
maxinelasp: 10th hour of each day!

PR #1066, opened 2026-01-16
ENH: adding features to support thruster API

### LINKED ISSUE #911: BUG - Thruster files not available via the SPICE query API
### Description of the issue

The thruster/small forces files are not queryable via the SPICE query API. I would expect them to be available under the `file_type` thruster, as in the expected SPICE file list.

https://api.imap-mission.com/spice-query?start_time=799242457&end_time=899242957&type=thruster

No files are returned.

https://api.imap-mission.com/spice-query?start_time=799242457&end_time=899242957

No thruster/small forces files are returned in this batch, and I would expect to see some.

The available thruster files should be listed in the query API, so as to be downloadable via `imap-data-access`

No small forces/thruster files are listed in the query API

### Code Snippet:

```python
Code
```

_No response_

tech3371:
# Change Summary
closes #911 

## Overview

Thruster fire files were not getting index to SPICE table because SPICE is unable to get date range coverage. That is because thruster fire files are text files. Because of that reason, we need to store and setup API for thruster in the same manner as spin and repoint data. 

That's what I have done in this PR. I have tested these changes in dev account (https://api.dev.imap-mission.com/small-forces-table) and it works as expected. Once this PR is merge, we will look into indexing thruster fire files in production.

## Updated Files

- added new table and API for thruster
- updated existing spin and repoint API handler to handle thruster as well
- updated SPICE index to index thruster to its own table now

## Testing

- added tests for thruster files


tech3371:
pre-commit.ci autofix

sds_data_manager/lambda_code/SDSCode/api_lambdas/non_spice_table_api.py
```diff
@@ -97,7 +99,7 @@ def lambda_handler(event, context):  # noqa: PLR0912
                         table.version,
                         table.ingestion_date,
                         row_number,
-                    ).alias("latest_spin_files")
+                    ).alias("latest_files")
```
greglucas: Is this alias used at all? I don't see it updated anywhere else.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/spice_indexer.py
```diff
@@ -440,6 +440,33 @@ def index_repoint_file(s3_key):
...
 
+def index_thruster_file(s3_key):
+    """Insert thruster file metadata into thruster database table.
+
+    Parameters
+    ----------
+    s3_key: str
+        S3 path of the thruster file.
+    """
+    logger.info(f"Indexing {s3_key} to ThrusterFiles table")
```
greglucas: You are already logging this outside of the function when you call it. We should only keep one or the other log, not both.

tests/lambda_endpoints/test_spice_indexer_lambda.py
```diff
@@ -587,3 +615,60 @@ def test_index_repoint_file_multiple_versions(
...
+    index_thruster_file("imap/spice/thruster/imap_2025_100_2025_110_hist_02.sff")
+
+    # Verify second version was indexed
+    thruster_v02 = (
+        session.query(models.ThrusterFiles)
+        .filter_by(file_path="imap/spice/thruster/imap_2025_100_2025_110_hist_02.sff")
+        .first()
+    )
+    assert thruster_v02 is not None
+    assert thruster_v02.version == "02"
```
greglucas: This would fail with an AttributeError if thruster_v02 was None, so just let this line handle that check.
```suggestion
    assert thruster_v02.version == "02"
```

tests/lambda_endpoints/test_spin_api.py
```diff
@@ -225,3 +254,42 @@ def test_repoint_table(repoint_db):
...
+        assert "version" in result
+        assert "ingestion_date" in result
+
+    # Both files should be returned since they overlap the date range
+    assert len(results) == 2
+
+    # Check results contain expected values
+    file_paths = [result["file_path"] for result in results]
+    assert "imap/spice/thruster/imap_2025_100_2025_110_hist_01.sff" in file_paths
+    assert "imap/spice/thruster/imap_2025_100_2025_110_hist_02.sff" in file_paths
```
greglucas:
```suggestion
    file_paths = [result["file_path"] for result in results]
    assert ["imap/spice/thruster/imap_2025_100_2025_110_hist_01.sff", "imap/spice/thruster/imap_2025_100_2025_110_hist_02.sff"] == file_paths
```

greglucas (approved):
A few minor nits, otherwise looks good to me.

* Do we want to call this `SmallForcesFile` since that is what SFF stands for? Rather than ThrusterTable. I don't know if there are other forces or not, but something to consider in naming things.
* I think this is fine as a separate table, but I am wondering if we are expecting any more of these and if we should try to put this into a more collective table where a column could be "file type" or something instead since this seems so close to the spin table files.

sds_data_manager/lambda_code/SDSCode/api_lambdas/non_spice_table_api.py
tech3371: I think you are right. It's not used.

tech3371:
> * Do we want to call this `SmallForcesFile` since that is what SFF stands for? Rather than ThrusterTable. I don't know if there are other forces or not, but something to consider in naming things.
> * I think this is fine as a separate table, but I am wondering if we are expecting any more of these and if we should try to put this into a more collective table where a column could be "file type" or something instead since this seems so close to the spin table files.

First one sounds great. For the second, I like keeping things separate so that we can debug or trace easier.

tech3371:
pre-commit.ci autofix

tech3371:
I addressed the feedback and going to merge this.

PR #1067, opened 2026-01-19
I-ALiRT - update logs

laspsandoval:
# Change Summary

## Overview
I need to improve visibility in the logs. The MAG data, in particular, is making it difficult to see what is happening. Therefore I just updated what I am able to see. Also, I need visibility on what packets are being used.

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
@@ -266,7 +266,7 @@ def parse_packets(filenames: list, bucket: str, download_dir: Path, apid=478):
     return combined
 
 
-def process_algorithms(  # noqa: PLR0912, PLR0915
+def process_algorithms(  # noqa: PLR0915
     combined: xr.Dataset, algorithm_table, table_name, kernel_set_key
 ):
     """Process the algorithms and insert data, as needed.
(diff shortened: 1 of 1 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -362,7 +362,10 @@ def process_algorithms(  # noqa: PLR0915
                 logger.info("Processing HIT.")
                 result = process_func(combined)
 
-            logger.info("%s result: %s", instrument, result)
+            if not result:
+                logger.info("Empty result for: %s", instrument)
+            else:
+                logger.info("Data populated for: %s", instrument)
```
greglucas: We shouldn't add conditional logic to format log output. I think this could have information about the result in a minified form. So just print how many results were processed, which can be zero if you want to search for that case.

```suggestion
            logger.info("[%s] results populated for [%s]", len(result), instrument)
```

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -680,7 +683,7 @@ def lambda_handler(event, context):
 
     if filenames:
         logger.info("Found %d files to process", len(filenames))
-        logger.info("Parsing packets.")
+        logger.info(f"Parsing packets: {filenames}")
```
greglucas:
```suggestion
        logger.info(f"Parsing packet files: {filenames}")
```

greglucas (approved):
Minor suggestions.

laspsandoval:
pre-commit.ci autofix

PR #1070, opened 2026-01-20
I-ALiRT - update mag hk credentials

### LINKED ISSUE #892: I-ALiRT - Separate credentials for MAG HK
### Summary

Modify the api to only allow queries for science parameters.
One question here though. Could you have this all in the same table and then just have the API handle whether the values are delivered or not?
i.e. we could put these behind the authentication mechanism and they would see them through the team login, but not the public data request.

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
# Change Summary

## Overview
Quick update so that only users with correct credentials can access mag hk.

## Updated Files
- ialirt_data_query_api.
   - Quick update so that only users with correct credentials can access mag hk.

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
@@ -224,11 +224,10 @@ def lambda_handler(event, context):
         meta_instrument = "spice"
         meta_type = "spice"
     elif params["instrument"].endswith("hk"):
-        if scope in FULL_SCOPES:
-            meta_instrument = params["instrument"]
-            meta_type = "hk"
-        else:
+        if scope not in FULL_SCOPES:
             return _error(403, "Unauthorized for HK access.")
+        meta_instrument = params["instrument"]
+        meta_type = "hk"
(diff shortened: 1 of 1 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_data_query_api.py
```diff
@@ -224,8 +224,11 @@ def lambda_handler(event, context):
         meta_instrument = "spice"
         meta_type = "spice"
     elif params["instrument"].endswith("hk"):
-        meta_instrument = params["instrument"]
-        meta_type = "hk"
+        if scope in FULL_SCOPES:
+            meta_instrument = params["instrument"]
+            meta_type = "hk"
+        else:
+            return _error(403, "Unauthorized for HK access.")
```
greglucas: Minor nit to prefer just the if path.

```suggestion
        if scope not in FULL_SCOPES:
            return _error(403, "Unauthorized for HK access.")
        meta_instrument = params["instrument"]
        meta_type = "hk"
```

greglucas (approved):

PR #1091, opened 2026-01-29
MNT: Update Lo l1b dependencies to do all-rates at once

greglucas:
Merging histrates and monitorrates together.

Renamed to `all-rates`. Do we want the hyphen or not?


sdhoyt (approved):
Thanks for updating this!

PR #1093, opened 2026-01-29
I-ALiRT - update permissions to allow coverage downloads

laspsandoval:
# Change Summary

## Overview
This is needed to be public for the website and to allow external folks to download it if needed.

## Updated Files
- lambda_api_key_authorizer
   - Update permissions for coverage files

### DIFF
--- a/sds_data_manager/lambda_code/authorization/lambda_api_key_authorizer.py
+++ b/sds_data_manager/lambda_code/authorization/lambda_api_key_authorizer.py
@@ -36,15 +36,14 @@ def lambda_handler(event, context):
     ):
         return {"isAuthorized": False}
 
+    # Public download except for logs and packets.
     if (
-        path.startswith("/ialirt-download")
-        and scope
-        not in (
-            "full",
-            "ialirt_external_partner",
-            "ialirt_scientist",
-        )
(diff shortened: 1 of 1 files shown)


sds_data_manager/lambda_code/authorization/lambda_api_key_authorizer.py
```diff
@@ -36,6 +36,10 @@ def lambda_handler(event, context):
     ):
         return {"isAuthorized": False}
 
+    # Allow I-ALiRT station coverage data to be downloaded.
+    if path.startswith("/ialirt-download/coverage"):
+        return {"isAuthorized": True}
+
     if path.startswith("/ialirt-download") and scope not in (
```
greglucas: Move the check into here `and not converage in str(path)` instead of a separate block?

I'm confused whether you want this to require an API Key still, or make it generally public, i.e. just return True without even needing to query for an API Key at all (move it way to the top).

laspsandoval: Opened up ialirt downloads except for packets and logs. That way I don't need to mess with this later.

greglucas (approved):

PR #1095, opened 2026-01-29
MNT: Add ephemeris dependency to lo pointing set

greglucas:
# Change Summary

Adding ephemeris. Do we want to have the reconstructed version as a hard dependency if that would cause a delay in 2 weeks? Or do we want to remove that and just go with the best available predicts which should be fine.


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
tmplummer:
```suggestion
lo, l1b, spin, lo, l1b, prostar, HARD, DOWNSTREAM
```

tmplummer (approved):

PR #1106, opened 2026-01-31
I-ALiRT - open database to public

laspsandoval:
# Change Summary

## Overview
This opens the database to the public (note the lambda takes care of credentials so HK, data before Feb 1, and certain HIT products are restricted still)


greglucas (approved):
Just a note that the title confused me because you are not opening up the DB for public access but rather the API endpoints. (LaTiS folks wanted direct access to Dynamo and I still agree with your decision here to not allow that for anyone)

PR #1107, opened 2026-02-03
I-ALiRT Data Version Update - prepare for public release

laspsandoval:
# Change Summary

## Overview
Andriy requested that we begin on version 001.

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
@@ -123,7 +123,7 @@ def lambda_handler(event, context):
         )
         return
     dataset = create_xarray_from_records(all_items)
-    dataset.attrs["Data_version"] = "000"
+    dataset.attrs["Data_version"] = "001"
     dataset.attrs["Start_date"] = seven_days_ago.strftime("%Y%m%d")
     test_data_path = write_cdf(dataset, istp=True)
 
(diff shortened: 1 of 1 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_archive.py
```diff
@@ -123,9 +123,11 @@ def lambda_handler(event, context):
         )
         return
     dataset = create_xarray_from_records(all_items)
-    dataset.attrs["Data_version"] = "000"
+    dataset.attrs["Data_version"] = "001"
     dataset.attrs["Start_date"] = seven_days_ago.strftime("%Y%m%d")
-    test_data_path = write_cdf(dataset, istp=True)
+    test_data_path = write_cdf(
+        dataset, istp=True, compression=None, auto_fix_depends=False
```
lacoak21, line 129: Did you mean to set compression=None as well?

laspsandoval: Yes. Andriy had said that he did not recommend any compression for the cdf. Do you know why we have compression?

tech3371: If passing in `compression=None` doesn't compress the epoch, then that's a good thing. It's something we have noticed in science files and haven't looked closely. If yours worked, Laura, we want that for science files too. But I meant that as future things we need and not related to this PR.

tech3371 (approved):
I will in the mean time to keep you going!

Issue #1059, opened 2026-01-15
Batch Starter Refactoring


tech3371:
This is a major task that will involve refactoring batch starter lambda logic into more simplified and structured look up. This is major because batch starter cover these major features of SDC infrastructure

- job kickoff
- dependency lookup
- event tracking
- cadence job kickoff
- reprocessing job handler

Current list of improvement requested to batch starter lambda
- [ ] Adding more precision to dependency query. Right now, it's yyyymmdd but it needs to be yyyymmddhhmmss. Also, move to string or datetime. But more precision.
- [ ] Refactor to use structure object similar to `imap-data-access`'s `<Science>FilePath` object in dependency query.
- [ ] Remove or reduce if/else statement using outcome from above bullet point discussion

maxinelasp:
MAG L1D would benefit from point 1.

tech3371:
If it's blocking you right now, can we look into fix that right now Maxine? Otherwise, we will be working on this refactoring soon but it will take a bit more time since we need to revisit design, discussion and implementation.

maxinelasp:
It's not blocking me, just something causing minor annoyances. But, it is not affecting processing at all and doesn't really have that much of an effect. So it doesn't need to be a top priority.

tech3371:
@bryan-harter This ticket captures all related ticket to Batch Starter refactoring and use cases that are waiting on this work.

bryan-harter:
I'm gonna go ahead and close this mega ticket, the remaining 5 tickets have various priorities and now just fall under standard development

PR #1115, opened 2026-02-06
Update cadence job triggering

### LINKED ISSUE #1057: Clean cadence event bridge rules
For cadence events, we need them to be exactly every x days so we were not able to do a cron job or something similar. We opted to schedule event bridge rules for each cadence event. 

Cloud formation only lets you deploy up to 500 resources. This solution created many resources causing us to approach our limit.

We were also missing the 1mo cadence jobs. 

I want to switch to a single event schedule for each cadence. This uses a rate pattern so we can specify exactly how often we need to schedule the jobs. I initially thought we weren't able to use this pattern because we needed the jobs to be scheduled every x.x days and the expression did not allow for decimals. To get around this, we can convert the number of days to minutes so we can be exact. There is no upper limit to the n minutes as long as it is an integer as I initially had thought.

lacoak21:
# Change Summary

closes #1057
## Overview
This PR simplifies the cadence job triggering. Before we created a single EventBridge rule for each of the cadence jobs. We were getting close to our resource limit for cloudFormation. Instead of individual rules we now create one schedule rule for each cadence. 

See issue for more details. 

## Updated Files
   - Create one Rule for each cadence type including 1mo. 
   - Create function to calculate the next run based off of the start date. If the date we are deploying is after the start date we need to handle that.

### DIFF
--- a/sds_data_manager/constructs/instrument_lambdas.py
+++ b/sds_data_manager/constructs/instrument_lambdas.py
@@ -167,14 +167,17 @@ class BatchStarterLambda(Construct):
         # expression. E.g., we cannot specify "rate(91.2 days)" for 3 months.
         # TODO First job date should be 3 months after phase e start date.
         #   TODO determine exact phase e start date.
-        first_job = datetime.datetime(2026, 5, 1)
+        first_map_jobs = datetime.datetime(2026, 5, 1)
+        # 1mo jobs are not map jobs. We want them to start earlier. E.g. IDEX l2b is
+        # a 1 month cadence job.
+        first_1mo_jobs = datetime.datetime(2026, 1, 1)
         today = datetime.datetime.now()
         cadence_strs = ["1mo", "3mo", "6mo", "1yr"]
         for cadence_str in cadence_strs:
             cadence = CadenceDays.str_lookup(cadence_str)
             # Calculate interval in minutes
             interval_minutes = int(cadence.value * 24 * 60)
-
+            first_job = first_1mo_jobs if cadence_str == "1mo" else first_map_jobs
             # Calculate the next run time based on the first job date and the cadence
             next_run = calculate_next_run(first_job, today, interval_minutes)
             scheduler.CfnSchedule(
--- a/sds_data_manager/constructs/instrument_lambdas.py
+++ b/sds_data_manager/constructs/instrument_lambdas.py
@@ -2,12 +2,16 @@
 
 import datetime
 
-import aws_cdk as cdk
-from aws_cdk import Duration, Environment, aws_events
+from aws_cdk import Duration, Environment
 from aws_cdk import aws_ec2 as ec2
-from aws_cdk import aws_iam as iam
+from aws_cdk import (
+    aws_iam as iam,
+)
 from aws_cdk import aws_lambda as lambda_
 from aws_cdk import aws_s3 as s3
+from aws_cdk import (
+    aws_scheduler as scheduler,
+)
 from aws_cdk import aws_secretsmanager as secrets
 from aws_cdk import aws_sqs as sqs
 from aws_cdk.aws_lambda_event_sources import SqsEventSource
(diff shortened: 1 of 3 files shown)


sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -137,103 +141,100 @@ def __init__(
...
+                start_date=start_date_str,
+                flexible_time_window=scheduler.CfnSchedule.FlexibleTimeWindowProperty(
+                    mode="OFF"
+                ),
+                target=scheduler.CfnSchedule.TargetProperty(
+                    arn=self.instrument_lambda.function_arn,
+                    role_arn=scheduler_role.role_arn,
+                    input=f'{{"cadence": "{label}"}}',
+                    retry_policy=scheduler.CfnSchedule.RetryPolicyProperty(
+                        maximum_retry_attempts=185
```
tech3371: Should we reduce this retry amount? Instead, monitor setup or notification may help us quicker to be aware and debug earlier.

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -137,103 +141,100 @@ def __init__(
...
+                    ),
+                ),
```
tech3371, line 217: Nice! Much cleaner!

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -137,103 +141,100 @@ def __init__(
...
+        return first_job
+    else:
+        # Calculate how many minutes have passed since the first job
+        delta = (today - first_job).total_seconds() / 60
+        # Calculate how many cadence events have passed since the first job
+        events_passed = int(delta // interval_minutes)
+        # Calculate the next run date
+        next_run = first_job + datetime.timedelta(
+            minutes=(events_passed + 1) * interval_minutes
+        )
```
tech3371, line 254: This looks good. I think it's true but just wanted to double check that we can trigger manually on batch starter lambda if any of those past cadence event has failed. Is that true?

tech3371 (approved):
Nice simplification!

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -137,103 +141,100 @@ def __init__(
...
         #    - 3 month map jobs (every 365.25 / 4 days)
         #    - 6 month map jobs (every 365.25 / 2 days)
         #    - 1 year map jobs (every 365.25 days)
-        # Note: each map job trigger will have its own eventBridge rule, because we need
-        # them to run every x days starting from the same date (t0).
+        # Note: We are defining the schedules to run at minute-level intervals because
+        # AWS EventBridge Scheduler does not allow for decimal values in the rate
+        # expression. E.g., we cannot specify "rate(91.2 days)" for 3 months.
         # TODO First job date should be 3 months after phase e start date.
         #   TODO determine exact phase e start date.
```
tech3371: remove now?
```suggestion

```

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -137,103 +141,100 @@ def __init__(
...
+        cadence_strs = CadenceDays.str_lookup()
+        for cadence_str in cadence_strs:
+            cadence = CadenceDays.str_lookup(cadence_str)
+            # Calculate interval in minutes
+            interval_minutes = int(cadence.value * 24 * 60)
+            first_job = first_1mo_jobs if cadence_str == "1mo" else first_map_jobs
+            # Calculate the next run time based on the first job date and the cadence
+            next_run = calculate_next_run(first_job, today, interval_minutes)
+            # Format date as yyyy-MM-ddTHH:mm:ss.SSSZ (with milliseconds)
+            start_date_str = next_run.strftime("%Y-%m-%dT%H:%M:%S.000Z")
```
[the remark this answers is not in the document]
lacoak21: @tech3371 do you see an issue here? I wasnt sure how to avoid calculating each start date everytime we deploy.

sds_data_manager/constructs/instrument_lambdas.py
lacoak21: No, after SIT4 we removed the start_date" and "end_date" parameters. So when we trigger batch starter, it will only kickoff maps from whatever day it is currently to minus whatever cadence it is. Its the same behavior we had before.

tech3371: As you mentioned in our today's discussion, we will need to revisit this. Can you create child ticket under this parent ticket, https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/1059?

tech3371 (approved):
looks good

sds_data_manager/constructs/instrument_lambdas.py
tech3371: Can you note this monitor setup when you create ticket for this work, https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/1115#discussion_r2790302262?

sds_data_manager/constructs/instrument_lambdas.py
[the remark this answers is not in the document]
lacoak21: Ok after more thought I am ok with this because we should be getting minimum monthly diffs. If there was the potential for daily diffs I would have a bigger problem.

PR #1123, opened 2026-02-12
I-ALiRT - gap fix

laspsandoval:
# Change Summary

## Overview
When the lambda searched 5 minutes back it did not gather enough packets for CoDICE. It usually collected 4 packets, which is sometimes enough to process CoDICE, but not always. I changed it to 7 minutes to be safe.

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
@@ -193,14 +193,14 @@ def query_filenames(bucket: str, region: str, now: datetime):
         config=botocore.client.Config(signature_version="s3v4"),
     )
 
-    seven_minutes_ago = now - timedelta(minutes=7)
+    look_back_time = now - timedelta(minutes=7)
 
     # Account for any cases in which data spans a threshold since
     # s3 only uses prefixes for queries.
     # Example:
     # now = 2026-01-01T00:02:00Z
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -193,14 +193,14 @@ def query_filenames(bucket: str, region: str, now: datetime):
         config=botocore.client.Config(signature_version="s3v4"),
     )
 
-    five_minutes_ago = now - timedelta(minutes=5)
+    seven_minutes_ago = now - timedelta(minutes=7)
```
tmplummer: This is fine, but you might consider not naming this with a specific time delta. You could use `lookback_time` or something more generic.

sds_data_manager/lambda_code/IAlirtCode/ialirt_ingest.py
```diff
@@ -193,14 +193,14 @@ def query_filenames(bucket: str, region: str, now: datetime):
...
 
     # Account for any cases in which data spans a threshold since
     # s3 only uses prefixes for queries.
     # Example:
     # now = 2026-01-01T00:02:00Z
     # five_minutes_ago = 2025-12-31T23:57:00Z
```
tmplummer: Should this be updated?

tmplummer (approved):
Couple of suggestions. Take it or leave it depending on urgency.

PR #1128, opened 2026-02-18
Speedup for the metakernel

### LINKED ISSUE #1008: SPICE metakernel API timeout
- [ ] API is giving service error. may be because of timeout.
- [ ] Example call- https://api.imap-mission.com/metakernel?start_time=813326469&end_time=816091269&file_types=leapseconds,spacecraft_clock,imap_frames,science_frames,planetary_ephemeris,ephemeris_reconstructed,attitude_history&list_files=True
- [ ] Optimize query time

bryan-harter:
# Change Summary

## Overview

Sped up the metakernel API significantly, it was searching to fill WAY too many gaps before. It's not super noticeable if each file only has a few gaps, but when there are many gaps per file it was doing a lot of double searching. 

## File changes

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
- Instead of searching "gap_list" each time, it only searches "subgap_list", or the list of gaps that are within the scope of the time interval it is searching. 
- Stopped converting the gap times to integers. I'm not sure why that was there in the first place, because it causes some issues later on with file contents not being exactly equal to the gaps we're searching. This makes it look like some files partially fill some gaps by fractions of a second, when they actually don't. 
- The very last change in the file I think was just simply a bug, and that would also occasionally cause an extra interval to be searched unnecessarily 

## Testing

I didn't add any testing, the output is ultimately the same but it takes less time. In theory we could add a unit test that takes the JSON structure I was testing with and then time how long it takes to run or something.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -10,7 +10,8 @@
 		"ghcr.io/devcontainers/features/desktop-lite:1": {
 			"version": "latest"
 		}, "ghcr.io/devcontainers/features/docker-in-docker:2": {
-			"version": "latest"
+			"version": "latest",
+			"moby": false
 		}
 	},
 	"forwardPorts": [6080],
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
@@ -416,14 +416,22 @@ seconds since J2000.
             if file_interval_start > search_window_start:
                 # <----------- search window --------....
                 #       <----- file coverage --------....
+                if search_window_start < gap_start:
+                    start = gap_start
+                else:
+                    start = search_window_start
                 sub_gaps.extend(
-                    [[search_window_start, file_interval_start]]
+                    [[start, file_interval_start]]
                 )  # Gaps before interval
             if file_interval_end < search_window_end:
                 # ....--- search window --------------->
                 # ....-- file coverage -------->
+                if search_window_end > gap_end:
+                    end = gap_end
+                else:
+                    end = search_window_end
                 sub_gaps.extend(
-                    [[file_interval_end, search_window_end]]
+                    [[file_interval_end, end]]
                 )  # Gaps after interval
 
         return sub_gaps
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
@@ -199,6 +199,9 @@ seconds since J2000.
     def _remove_duplicates_from_sorted_file_list(self, type: str):
         """Remove any duplicate found in self.spice_files[type].
 
+           Loops through the list, and determines which files to
+           keep.
+
         Parameter
         ---------
         type: str
@@ -213,7 +216,7 @@ seconds since J2000.
                 continue
             file_names_to_keep.append(old_file_list[i]["file_name"])
             new_file_list.append(old_file_list[i])
-        
+
         self.spice_files[type] = new_file_list
 
     def _limitstring(self, dirstring, limit, sym):
@@ -309,7 +312,7 @@ seconds since J2000.
                 logger.debug(
                     "File did not cover time range, not adding to metakernal list."
                 )
-                subgap_list=[trange]
+                subgap_list = [trange]
             elif not subgap_list:
                 logger.debug(
                     "File was valid, and no further gaps were found. "
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
@@ -420,9 +420,7 @@ seconds since J2000.
                     start = gap_start
                 else:
                     start = search_window_start
-                sub_gaps.extend(
-                    [[start, file_interval_start]]
-                )  # Gaps before interval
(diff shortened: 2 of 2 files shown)


bryan-harter:
Note that I think there are still improvements to be made. While the metakernel search that david attempted to kick off the other day now finishes in a few seconds, a mission-long search like this: https://api.imap-mission.com/metakernel?start_time=813326469&end_time=816091269&file_types=pointing_attitude&list_files=True

still fails to finish.

.devcontainer/devcontainer.json
```diff
@@ -10,7 +10,8 @@
 		"ghcr.io/devcontainers/features/desktop-lite:1": {
 			"version": "latest"
 		}, "ghcr.io/devcontainers/features/docker-in-docker:2": {
-			"version": "latest"
+			"version": "latest",
+			"moby": false
```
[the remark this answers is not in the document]
bryan-harter: Copilot caught me sneaking in a change! Yes, docker in docker doesn't work with the latest version of ubuntu because something called "moby" was removed. But if you set moby:false, then docker-in-docker works again.

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
```diff
@@ -199,26 +199,25 @@ def contains_gaps(self):
...
-            for j in range(i + 1, len(file_list)):
-                if file_list[i]["file_name"] == file_list[j]["file_name"]:
-                    indicies_to_delete.append(j)
-        for i in sorted(set(indicies_to_delete), reverse=True):
-            del file_list[i]
-        self.spice_files[type] = file_list
+            file_names_to_keep.append(old_file_list[i]["file_name"])
+            new_file_list.append(old_file_list[i])
+
+        self.spice_files[type] = new_file_list
```
tech3371, line 220: is changes in this function a solution for this point?
>The very last change in the file I think was just simply a bug, and that would also occasionally cause an extra interval to be searched unnecessarily

tmplummer (commented):
This looks fine to me. For a quick fix.

Since the full mission query still fails, I think that we need further improvement. If the failure is only due to the gap issue with the DPS kernels, I suggest adding special handling for those where we ignore the internal gaps.

tmplummer (approved):
Forgot to hit approve

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
```diff
@@ -199,26 +199,25 @@ def contains_gaps(self):
...
-            logger.debug(
-                f"Searching for duplicates for file {file_list[i]['file_name']}"
-            )
-            for j in range(i + 1, len(file_list)):
-                if file_list[i]["file_name"] == file_list[j]["file_name"]:
-                    indicies_to_delete.append(j)
-        for i in sorted(set(indicies_to_delete), reverse=True):
-            del file_list[i]
-        self.spice_files[type] = file_list
+            file_names_to_keep.append(old_file_list[i]["file_name"])
```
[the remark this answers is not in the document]
bryan-harter: Unfortunately a set doesn't maintain order, and the order is important here.

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
bryan-harter: Nope this function is called after all the intervals have been searched, its job is simply to remove any matches. The bug that was causing extra intervals to be searched is the one down on line 391

bryan-harter:
Ok all these little tweaks I'm finding have actually now improved the DPS query by quite a bit. The query above actually finishes in just a second or two! However, a larger query now just *barely* finishes in time: https://api.imap-mission.com/metakernel?start_time=813326469&end_time=916091269&file_types=pointing_attitude&list_files=True

There probably are still even more improvements to be made for performance, I just haven't found them yet. 

However, another path to go down is to set the allowed gap size for DPS kernels to 600 seconds. In other words, any gap under 5 minutes we assume it is a repointing maneuver, and we do not try to fill it in. That will reduce the number of calculations drastically, in fact it should finish in basically under a second. The potential issue I see with that is it would allow other "real" gaps through. But this 5 minute gap fix would be really quick to implement, and might not even change much.

sds_data_manager/lambda_code/SDSCode/api_lambdas/metakernel.py
tech3371: I see

tech3371:
> Ok all these little tweaks I'm finding have actually now improved the DPS query by quite a bit. The query above actually finishes in just a second or two! However, a larger query now just _barely_ finishes in time: https://api.imap-mission.com/metakernel?start_time=813326469&end_time=916091269&file_types=pointing_attitude&list_files=True
> 
> There probably are still even more improvements to be made for performance, I just haven't found them yet.
> 
> However, another path to go down is to set the allowed gap size for DPS kernels to 600 seconds. In other words, any gap under 5 minutes we assume it is a repointing maneuver, and we do not try to fill it in. That will reduce the number of calculations drastically, in fact it should finish in basically under a second. The potential issue I see with that is it would allow other "real" gaps through. But this 5 minute gap fix would be really quick to implement, and might not even change much.

Thank you for getting this in so quick!!

Like Tim was saying, can we do a 5 min for dsp kernel and keep others to what you had originally? and instead of 5min constant for dps, we can make use of repoint files to get exact time repoint maneuver took if we want too.

PR #1131, opened 2026-02-19
Cadence Jobs one month start date

lacoak21:
# Change Summary

## Overview

The one month jobs are not map jobs and therefore should start a month after launch and follow that pattern moving forward.


tech3371 (approved):

lacoak21:
pre-commit.ci autofix

PR #1164, opened 2026-03-10
I-ALiRT - open port to UKSA

laspsandoval:
# Change Summary

## Overview

Add their port to what is allowed into the SG.

Add a way to view their schedule.

### DIFF
--- a/sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
+++ b/sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
@@ -18,6 +18,7 @@ from imap_data_access.processing_input import (
 from imap_processing.ialirt.generate_coverage import (
     format_coverage_summary,
     generate_coverage,
+    parse_uksa_schedule_xlsx,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
```diff
@@ -258,6 +299,8 @@ def generate_and_upload_30_days(bucket: str, region: str, outages: dict, dsn: di
         Dictionary containing outages data.
     dsn : dict
         Dictionary containing DSN data.
+    uksa : list, optional
```
lacoak21, line 302: why do we only have a list like this for UKSA?

lacoak21 (approved):
This looks good! The only question/ confusion I had was why get_uska is treated separately.

lacoak21:
> This looks good! The only question/ confusion I had was why get_uska is treated separately.

Im not going to hold back approval because it looks good I just might be misunderstanding something

sds_data_manager/lambda_code/IAlirtCode/ialirt_coverage.py
laspsandoval: UKSA is unique from Kiel. Kiel uses ephemeris kernels to decide on coverage. UKSA has an excel spreadsheet that they sent us that is used to create the coverage.

lacoak21: Ah got it 👍

PR #1166, opened 2026-03-10
Bugfix - make on-demand

laspsandoval:
# Change Summary

## Overview

Remove spot=True to go from using spot instances to on-demand.


maxinelasp (approved):
Thanks Laura!

PR #1170, opened 2026-03-11
I-ALiRT - EC2 bugfix

laspsandoval:
# Change Summary

## Overview

EC2 bugfix

## Testing

Deployed to dev and it fixed the issue

### DIFF
--- a/sds_data_manager/constructs/processing_construct.py
+++ b/sds_data_manager/constructs/processing_construct.py
@@ -69,12 +69,11 @@ class ProcessingConstruct(Construct):
         # Create compute environment
         compute_environment = batch.FargateComputeEnvironment(
             self,
-            "ProcessingComputeEnvironment-spot",
-            compute_environment_name="ProcessingComputeEnvironment-spot",
+            "ProcessingComputeEnvironment-ondemand",
+            compute_environment_name="ProcessingComputeEnvironment-ondemand",
             vpc=vpc,
             vpc_subnets=ec2.SubnetSelection(subnet_type=ec2.SubnetType.PUBLIC),
             spot=False,
-            replace_compute_environment=True,
         )
 
         # Create job queue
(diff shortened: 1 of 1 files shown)


sds_data_manager/constructs/processing_construct.py
```diff
@@ -69,8 +69,8 @@ def __init__(
         # Create compute environment
         compute_environment = batch.FargateComputeEnvironment(
             self,
-            "ProcessingComputeEnvironment-spot",
-            compute_environment_name="ProcessingComputeEnvironment-spot",
+            "ProcessingComputeEnvironment-ondemand",
+            compute_environment_name="ProcessingComputeEnvironment-ondemand",
```
maxinelasp, line 73: Is this compute environment name used anywhere? How does batch determine that this is the compute environment it uses?

laspsandoval: Should be here:

        # Create job queue
        self.job_queue = batch.JobQueue(
            self,
            "ProcessingJobQueue",
            job_queue_name="ProcessingJobQueue",
            compute_environments=[
                batch.OrderedComputeEnvironment(
                    compute_environment=compute_environment, order=1
                )
            ],
        )

laspsandoval: The name is not used anywhere. I checked.

tech3371: We need if/else check for dev/prod too. Dev should continue using spot.

maxinelasp: I don't think it matters that much if dev uses on-demand now that we aren't using it as much

maxinelasp (approved):

maxinelasp:
We should also probably remove the spot compute environment... not sure of the best way to do that.

PR #1175, opened 2026-03-11
CoDICE 3d distribution products use L2 lookup tables

pleasant-menlo:
# Change Summary

## Overview

Updated CoDICE Lo L3 distribution dependencies to use shared L2 geometric factor and efficiency ancillary files.

## File changes

dependency_config.csv


tmplummer (approved):
LGTM

PR #1180, opened 2026-03-12
Swapi remove proton ephemeris dependency

pleasant-menlo:
# Change Summary
- Remove SWAPI l3a proton-sw dependency on ephemeris SPICE kernels
- Make SWE trigger off predicted ephemeris
- Do not trigger SWAPI pui-he on reconstructed ephemeris

## Overview
We discovered SWAPI l3a proton does not require ephemeris data.

Previously SWE did not have to be triggered by ephemeris, because we knew it was available if SWAPI proton-sw was available. Since proton-sw no longer depends on ephemeris, it is important for SWE to directly trigger off ephemeris in case predicted ephemeris is the last SWE dependency to become available.

Predicted ephemeris should be sufficient for the pickup ion computation, so we made the SOFT_TRIGGER dependency on reconstructed ephemeris a SOFT_NO_TRIGGER to prevent excess versions of the file being produced.

## File changes
- dependency_config.csv


tech3371 (approved):

jtniehof:
Looks like this is working as expected and SWAPI is being triggered on new L2. Thanks!

PR #1187, opened 2026-03-23
IDEX l1b processing job date range bug fix

lacoak21:
# Change Summary

## Overview

Small bug fix. I was checking for an l1b trigger file when it should be an l1a file.


ahotasu (approved):
Looks good to me.

PR #1190, opened 2026-03-24
IDEX l1b filename bug fix

lacoak21:
# Change Summary

## Overview

I added a buffer to the idex l1b job start date but the buffer should have ONLY been added to the dependency query start date and not the start date that is used in the job file name. This caused the resulting l1b filename to have the incorrect start-date. 

## File changes

- fix the special case handling for idex l1b jobs. The start date should stay the same (keep the same date as the l1a file) but the query should collect spice files from earlier than the start date. The IDEX filenames are incorrect anyway I will be fixed in the future but this is a temporary fix to get the IDEX team data. 

## Testing

Add a test to ensure we are submitting the job with the correct filename start date.


lacoak21:
This has been tested in dev successfully.

tech3371 (approved):
If it works in dev, then change looks good to me.

PR #1196, opened 2026-03-25
Sync dev to dependency_refactor branch

tech3371:
# Change Summary


tech3371:
I like to look into different solution to track increment progress of refactoring work. Our PR merge default setting is 'squash and merge' and I want avoid doing that for future merge into dev. It will cause to wipe away commits in dev and merge them into one commit. I like to avoid that.

PR #1202, opened 2026-03-30
IDEX l1b dependency logic bug fix

lacoak21:
# Change Summary

## Overview

IDEX l1b bug fix. Last week I put in the fixes for the idex l1b issue but it only fixed it for reprocessing commands. This patch fixes it for both reprocessing AND from an l1a trigger. 

Both reprocessing and l1a triggers have been tested in dev. 

## File changes

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py

Always get new dependencies for idex l1b jobs using the new buffer start date.


laspsandoval (approved):
Make sense.

vmartinez-cu (approved):
LGTM, nice job!

Issue #1153, opened 2026-03-03
Task: Transition all instrument into their own new dependency config


tech3371:
### Summary
Create dependencies files for each instrument with new dependency config details

tech3371:
Now Checking all yaml dependency with latest csv file content for all instrument and all data level.

- [x] CoDICE
- [x] GLOWS
- [x] HIT
- [x] IDEX
- [x] MAG
- [x] SWAPI
- [x] SWE
- [x] IMAP-Ultra
- [x] IMAP-Hi
- [x] IMAP-Lo

tech3371:
I reviewed and transition using latest dependency file from yesterday. Opened PR as needed with fixes. In the future, we will need to keep it up to data through keep CSV and YAML match when new requests comes in or check again before refactor feature integration test in the future. This initial work is complete.

PR #1235, opened 2026-04-13
CoDICE: L3a minor updates to use inheritance

tech3371:
# Change Summary
related to #1153 

## Overview

went through dependency csv and no updated needed for CoDICE. Minor improvement only.


laspsandoval (approved):
Nice cleanup!

PR #1237, opened 2026-04-13
Hi: dep updates and improvments

tech3371:
# Change Summary
related to #1153 

## Overview

Double checking and correction based on latest dependency csv file in dev.


tmplummer (approved):
I think this looks good. I am having a little trouble joining the spice inheritance. My suggestion would be only to make a spice_basic common dependency that goes across levels/products.

tech3371:
> I think this looks good. I am having a little trouble joining the spice inheritance. My suggestion would be only to make a spice_basic common dependency that goes across levels/products.

Totally. I will check that dependency is correct for the time being and if you show an example of what you are looking for, we all can follow that as example.

tech3371:
@tmplummer I merged Hi's dependency. Can you open PR with your idea of inheritance? I like to see and know and use for others in the future.

PR #1245, opened 2026-04-14
Update versioning Logic in Batch Starter

### LINKED ISSUE #1244: BUG: Issue with batch starter versioning
There is a bug that both Tim and Sean found today that happens when a job doesnt produce a file but still succeeded. 

This can happen if there isnt any relevent data to process OR there is a failure in uploading a file. 

Either way we need to fix this.

lacoak21:
# Change Summary

closes #1244

## Overview

We only check the science files table for max version which incorrectly can submit a duplicate processing job if the job before it did not produce a file. 

For example:

processing job for lo, l1b , bgrates v001 shows it suceeded but produced no file. 

later if you want to reprocess, batch starter checks what the max version is for lo, l1b, bgrates, finds no file and tries to submit another job using version == v001. This causes a duplicate record error because that record already exists from before and shows it succeeded. 

The fix is to check BOTH the science files table and the processing job table and use the max version. This can cause version gaps if a previous run didnt produce a file but it is better than the job getting submitted.  

## File changes

- update comment and logic 

## Testing
update test to reflect new logic changes.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
@@ -272,19 +272,19 @@ def determine_job_version(
         )
     ).scalar()
 
-    # Step 5: By default, use the max version from the science files table unless
-    # it is a spacecraft "pointing-attitude" job. If a so, then use the max version
-    # from the processing jobs table. If the job is a spacecraft pointing-attitude job,
-    # it will produce a SPICE kernel and not a science file. There is no way to
-    # determine the filename of the kernel that will be produced, so we rely on the max
-    # version from the processing jobs table.
-    if instrument == "spacecraft" and descriptor == "pointing-attitude":
-        max_version = max_version_processing
-    else:
-        max_version = max_version_sci
+    # Step 5: Generally, we use the science files table as the source
+    # of truth for the max version. However, some jobs may complete without producing a
+    # science file (e.g. spacecraft "pointing-attitude" jobs produce a SPICE kernel
+    # instead) or we may just not produce a file due to lack of relevant data. To
+    # account for this, we take the max between the science files table
+    # and the processing jobs table. For spacecraft "pointing-attitude" jobs, the
+    # science files table will always be empty, so the max version will come solely from
+    # the processing jobs table.
+    versions = [v for v in [max_version_processing, max_version_sci] if v is not None]
+    max_version = max(int(v[1:]) for v in versions) if versions else None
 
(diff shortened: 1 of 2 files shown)


lacoak21:
Another alternative is to ONLY check the processing job table. 

I opted to check both to be safe because someone could theoretically update the version of the output file in cli.py or the processing job record could accidentally be deleted or edited or something. In a perfect world none of those things would happen but if they did this way we could still process jobs.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/batch_starter.py
```diff
@@ -272,19 +272,19 @@ def filter_conditions(table):
...
-    # it is a spacecraft "pointing-attitude" job. If a so, then use the max version
-    # from the processing jobs table. If the job is a spacecraft pointing-attitude job,
-    # it will produce a SPICE kernel and not a science file. There is no way to
-    # determine the filename of the kernel that will be produced, so we rely on the max
-    # version from the processing jobs table.
-    if instrument == "spacecraft" and descriptor == "pointing-attitude":
-        max_version = max_version_processing
-    else:
-        max_version = max_version_sci
+    # Step 5: Generally, we use the science files table as the source
```
tmplummer, line 275: Hmmm. I'm wondering if this is the right way to fix this. For the Hi example, I think it is a bug that the job succeeded when there was an error uploading a file. I don't know enough about the Lo job to know if that job should have also failed.

An alternative way to handle this would be to add the command or perhaps the dependency hash directly to the unique row restriction. That would allow a job with new dependencies to be re-run without skipping a version number.

This is probably splitting hairs, but I am trying to think of unintended consequences that this implementation might have. I think this will now allow a job with identical dependencies to run over and over because it now increments the version each time that job runs. It would be an odd situation where that job succeeds each time and produces no products, but who knows.

lacoak21: Yeah I was thinking about just adding the 'container_command' (which contains the dependency file and therefore the hash) but then i realized it might cause unintended consequences. Now I cant track down that line of thought but ill try to think of it again. 

The lo job didnt produce a file not due to a bug but because there was nothing to produce. IDEX has a case like this too. 

Yeah maybe this needs more thought. I think currently a job with identical dependencies can rerun and rerun anyway (incase there was a code update) 

I think this goes back to your point about tracking the githash as well because this is too complicated and a lot of the complication comes from needing to be able to reprocess when there has been a code update.

lacoak21:
Closed. We decided to add the git hash to simplify logic

PR #1246, opened 2026-04-14
MAG and SWE: dependency updates

tech3371:
# Change Summary
related to #1153 

## Overview

updated MAG and SWE dependency based on latest change in csv.


laspsandoval (approved):
LGTM

PR #1276, opened 2026-04-27
Upgrade from poetry 1.7 to poetry 2.3

maxineofficial:
Including conforming to PEP 621. Co-authored with Claude.

This is the same change as in imap_processing, with the following differences:

- some changes were required to pre-commit, as poetry export is now provided as a plugin
- all the dynamic versioning work is not needed, as we don't use that tool in this repo

Below is copied from imap_processing PR. 

# Change Summary
Now that we're expecting more external people to use our project, we should use the latest version of Poetry.

Benefits:
- Correct version of Poetry installs when you run pip install poetry
- Conforms with latest Python standards, making pip work better
- Better/cleaner organization
- Plugins are fixed - no longer do we need to run `poetry self add`, as the required plugins will install automatically
- More compatibility with venv 

Changes from Poetry 1 to Poetry 2:
**`poetry shell` is gone**

- Removed from core Poetry 2.x
- Use poetry env activate to print the activation command, then run it manually
- Or install the poetry-plugin-shell plugin to restore the old behavior: poetry self add poetry-plugin-shell
Lock file

**Lock file format changed — not backwards compatible with Poetry 1.x**
- poetry lock --no-update is gone; just use poetry lock
- poetry lock by default no longer upgrades pinned versions, just reconciles with pyproject.toml
- pyproject.toml validation is stricter

**Poetry 2.x fully adopts PEP 621 ([project] table) alongside [tool.poetry]**
- [tool.poetry] must now either have a version field or rely on a plugin via [tool.poetry.requires-plugins] for dynamic versioning
 - Plugin dependencies declared in [tool.poetry.requires-plugins] are automatically installed when running poetry install

## File changes

The main changes are in pyproject.toml. No dependencies were changed, this is just a reorganization. (note: dependencies were upgraded, as the lock file changed.) Once everyone upgrades to Poetry 2, the main changes will be that you no longer use `poetry shell`. This is documented. 

There were a few other updates removing some of the hacks around the dynamic versioning. We are now on the latest version of that plugin.

## Testing

I would appreciate if reviewers would clone this branch, install Poetry 2, and verify that everything installs successfully.


maxineofficial:
Even though I didn't upgrade ruff, it is still now failing on a bunch of files. I went ahead and made those fixes to get pre-commit to pass.

Edit: I did upgrade ruff by accident. It's downgraded in this PR, and upgraded in #1277

laspsandoval (approved):
This looks good to me. I did:
poetry install --all-extras

When we release imap_processing I would like to bump the version in pyproject.toml just to check that the requirements.txt will be updated properly.

README.md
```diff
@@ -13,51 +14,50 @@ Our goal with the project is that users will only need to modify the file config
 
 The code in this repository takes the form of an AWS CDK project. It provides the architecture for:
 
-1. An HTTPS API to upload files to an S3 bucket (*in development*)
+1. An HTTPS API to upload files to an S3 bucket (_in development_)
```
tech3371, line 17: can we remove in development now that we are in prod and operational?

tech3371 (approved):

README.md
maxineofficial: I'm going to actually update this README in a different PR

maxineofficial: I should have done it as part of the docs migration yesterday

PR #1283, opened 2026-04-29
Remove redundant unique=True parameter for primary keys

### LINKED ISSUE #1280: Remove redundant unique flag from primary keys
We need to remove the unique=True parameter from primary key columns because it's redundant — primary keys are already required to be unique by the SQL standard

lacoak21:
# Change Summary
closes #1280
## Overview

Remove the unique=True parameter from primary key columns because it's redundant — primary keys are already required to be unique by the SQL standard. 

This was a detected drift by the alembic tool for both dev and prod. When this is merged in, prod will be 100% synced with the code. 

## File changes

   - remove redundant unique keys

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/database/models.py
+++ b/sds_data_manager/lambda_code/SDSCode/database/models.py
@@ -141,7 +141,7 @@ class ScienceFileBase:
     Those two tables share many of the same columns.
     """
 
-    file_path = Column(String, nullable=False, primary_key=True, unique=True)
+    file_path = Column(String, nullable=False, primary_key=True)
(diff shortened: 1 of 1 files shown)


lacoak21:
This is the revision created for prod vs the schema in our code. 
Screenshot 2026-04-29 at 11 27 41 AM

As you can see, it is saying that prod needs unique constraint keys on all of the primary keys. That is unnecessary. When cdk deployed these tables, it was smart enough to realize that primary_key = True and unique = True is redundant but alembic is not, so it is easiest to update this for the sake of keeping the alembic updates simpler.

maxineofficial (approved):

PR #1285, opened 2026-04-29
Add L2 mag as a soft no trigger dep to pitch angle products

pleasant-menlo:
# Change Summary
Add MAG L2 as a soft no trigger dependency for Pitch Angle Distribution products (HIT, SWE, CoDICE). 

https://github.com/IMAP-Science-Operations-Center/imap_L3_processing/issues/13


tmplummer:
As part of the dependency refactor, we are going to be moving to a different format dependency file. Can you please also make the needed updates to the linked file. It is not yet being used, but this will help us keep it in sync with the dependency JSON.
https://github.com/pleasant-menlo/sds-data-manager/blob/7a0b4bcd48df574677791f5285fb5675743bcf02/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/dependencies/imap_codice_dependencies.yaml

pleasant-menlo:
@tmplummer Just added these dependencies to the new yaml format. Let us know if there's anything else it needs!

tmplummer (approved):
Thanks for adding to the future dependency files too! LGTM

PR #1290, opened 2026-04-30
Initialize database migration workflow

### LINKED ISSUE #1289: Set up alembic in sds-data-manager
initialize alembic

lacoak21:
# Change Summary

closes #1289
## Overview

Set up alembic for database migrations. 

## File changes

- add new files for alembic migrations


tech3371 (commented):
I don't know enough here. @maxinelasp or @bryan-harter, can you guys review this?

alembic/env.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Interpret the config file for Python logging.
+# This line sets up loggers basically.
+if config.config_file_name is not None:
+    fileConfig(config.config_file_name)
+
+# add your model's MetaData object here
+# for 'autogenerate' support
+# from myapp import mymodel
+# target_metadata = mymodel.Base.metadata
```
maxineofficial: Can you explain what these lines mean and why they are commented out here?

alembic/env.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def run_migrations_online() -> None:
+    """Run migrations in 'online' mode.
+
+    In this scenario we need to create an Engine
+    and associate a connection with the context.
+
+    """
+    connectable = engine_from_config(
```
maxineofficial, line 65: Can you put in some kind of check for prod and print out some big message here about how migrations are applying to prod? I'm picturing like, if I am running the code against a prod env var without realizing, it would be nice if there was a printuot like "THIS IS PROD BTW"

alembic/script.py.mako
```diff
@@ -0,0 +1,28 @@
+"""${message}.
```
maxineofficial, line 1: Again, is it possible/reasonable to distinguish dev vs prod in this template generator? Or is the output script the same for both?

alembic.ini
```diff
@@ -0,0 +1,149 @@
+# A generic, single database configuration.
```
maxineofficial, line 1: Are these all the default values, or did you change any?

maxineofficial (commented):
Broadly looks good, I had a few comments but no major concerns!

alembic/env.py
lacoak21: Im not sure if there is a way of telling its prod or not other than hardcoding the PROD rds db instance id in the code. Unfortunately there is no account id in the database url.

lacoak21: we might be able to tag the RDS instance with dev and prod and then check the tag each time. Not sure if thats overkill or not.

alembic/script.py.mako
lacoak21: There should only be one script that will work for both PROD and DEV.

alembic.ini
lacoak21: These are all default.

alembic/env.py
lacoak21: Oops I will delete those. That was the example.

alembic/env.py
lacoak21: I am not implement this because I dont want to have any rds ids hardcoded in case its sensitive.

laspsandoval:
How/when are migrations run in the deployment pipeline? Is there a Lambda, a pre-deploy step, a GitHub Actions job?

lacoak21:
> How/when are migrations run in the deployment pipeline? Is there a Lambda, a pre-deploy step, a GitHub Actions job?

This is manual for now because this PR is a blocker to a lot of work I have to complete in the next two weekish. 

Eventually im hoping it can be a github actions step but for right now I outlined manual instructions in these docs https://github.com/IMAP-Science-Operations-Center/imap_processing/pull/3112

alembic/README
```diff
@@ -0,0 +1 @@
+See https://imap-processing.readthedocs.io/en/latest/infrastructure/database-migration/index.html for more information
```
laspsandoval: I got a file not found for this url. Will you update this later?

laspsandoval (commented):
Just a couple of questions. Thanks!

alembic/README
lacoak21: Yes that PR hasnt been merged yet. (in imap_processing)

maxineofficial (approved):

PR #1310, opened 2026-05-06
CoDICE: remove invalid data products

### LINKED ISSUE #1232: CoDICE: Remove not valid products from dependency file
Emal from CoDICE team,
```
Hi All,

We have a new request from the CoDICE team regarding the production of some of the Lo products. 

Please do not produce the following products at all:
Lo_sw-angular
Lo_nsw-angular
Lo_nsw-species

We have determined that the use of the position variable (determined from a delay line circuit onboard) to bin counts into the sw or nsw products results in a large portion of sw counts ending up in the nsw products. Thus, the nsw products (and sw angular) are not representative of the true ion population. So, we have decided not to release these as is to the public to avoid misinterpretation of the data. 

We are working to develop a method to correct this issue on ground, with the possibility of producing a new L3 product of corrected intensities. For now, the best thing to is to simply not release those specific products. 
```

SDC Notes from discussion

**Don't produce these products in prod:**
Lo_sw-angular
Lo_nsw-angular
Lo_nsw-species

Questions:
When you say don’t produce, do you mean don’t produce at all at SDC from now?
>yes

How long will that pause last?
>May be indefinitely. Onboard processing to determine position is locked onboard. It’s hard coded. Due to that, data is binned incorrectly forever. Can correct on ground by using DE in L3.
Will it last until you have new algorithms to correct them?
If correction is found and finalized, it will happen in L3.
For your public release, we can definitely look into solution to hide it from public release.
Stop producing and then clean up and not release at all.

tech3371:
# Change Summary
closes #1232

## Overview

Removes data products that CoDICE team have verified invalid from in-flight data.


lacoak21 (approved):
LGTM

bryan-harter (approved):

PR #1311, opened 2026-05-07
Set pointing_attitude as HARD_NO_TRIGGER for pointing-attitude job to…

### LINKED ISSUE #1264: Investigate and update DPS kernel dependencies
I noticed that the recent spacecraft pointing attitude jobs are getting kicked off twice each contact. One of the times, is producing a DPS kernel with longer than expected coverage that does not match the coverage duration of the latest attitude history file.

For this ticket, investigate the issue, look at batch starter logs for those jobs and identify why this is happening. Either update the pointing attitude job dependencies or fix the problem another way.

@tmplummer suspects that the unexpected duration kernel is getting triggered by the *.ah file upload and changing the dependency type to HARD_NO_TRIGGER would fix this issue.

tmplummer:
… stop extended length pointing attitude kernel from being produced

# Change Summary

## Overview

Currently, after a contact, we are generating two pointing-attitude kernels. One, with the correct coverage, is kicked off by the repoint-table. The other, which covers two attitude_history kernels, is kicked off by the attitude_history kernel. This change removes job triggering from the attitude_history dependency.

Closes: #1264


tech3371 (approved):
Thank you Tim for updating Yaml content too!

PR #1318, opened 2026-05-08
Update SWE l3 mag dependency to fix typo in upstream data type.

pleasant-menlo:
# Change Summary

## Overview

We noticed that the SWE L3 dependency configuration mistakenly lists MAG L1d twice, instead of L1d and L2.

## File changes

imap_swe_dependencies.yaml, (l3, sci): Adjust the second (soft) MAG dependency to be L2.


tmplummer (approved):
Thanks for catching this!

PR #1322, opened 2026-05-10
ENH: transfered dependency node validation in DependencyNode module

### LINKED ISSUE #1227: Dependency Node validation improvement
> I will do that in upcoming PR to bring in other utils functions. Right now, this PR is nice standalone and small scope PR. I plan to move onto other common foundations next. I will bring this in that work. 

 _Originally posted by @tech3371 in [#1195](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/pull/1195/changes#r3061194765)_

tech3371:
# Change Summary
closes #1227 

## Overview

Refactor based on discussion from previous PR feedback and discussion with Tim Plummer.

## File changes

- No major code changes. Moved node validation code into DependencyNode.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
@@ -1,6 +1,6 @@
 """Common functions for pipeline lambdas."""
 
-from dataclasses import asdict, dataclass
+from dataclasses import asdict, dataclass, field
 from datetime import datetime
 from typing import Any, ClassVar
 
@@ -14,7 +14,34 @@ DATE_RANGE_OPTIONS = ("p", "h", "d", "l", *NEAREST_OPTIONS)
 
 @dataclass
 class DependencyNode:
-    """Shared between batch starter and dependency lambda."""
+    """Shared between batch starter and dependency lambda.
+
+    A valid node must have exactly 5 or 6 elements:
+        [
+            source,
+            data_type,
+            descriptor,
+            required,
+            kickoff_job,
+            Optional([past, future])
+        ]
+    If it includes past/future date ranges, it should follow the following format:
+        - p - pointing
+        - h - hourly
+        - d - days
+        - l - last_processed
+        - nd - nearest day
+        - np - nearest pointing
+
+        past and future should end with one of these options. Eg.
+            ["-3p", "3p"] means 3 pointing
+            ["-3d", "5d"] means 5 days
+            ["-2h", "2h"] means 2 hours
+            ["-1l"] means last processed
+            ["6np"] means nearest 6 pointing
+
+    Validation is performed for each field.
+    """
 
     _data_source_validator: ClassVar[DataSource] = DataSource()
     _data_type_validator: ClassVar[DataType] = DataType()
@@ -24,37 +51,10 @@ class DependencyNode:
     descriptor: str
     required: bool = True
     kickoff_job: bool = True
-    date_range: list = None
+    date_range: list = field(default_factory=list)
 
     def __post_init__(self):
-        """Validate all fields on construction.
-
-        A valid node must have exactly 5 or 6 elements:
-            [
-                source,
-                data_type,
-                descriptor,
-                required,
-                kickoff_job,
-                Optional([past, future])
-            ]
-        If it includes past/future date ranges, it should follow the following format:
-            - p - pointing
-            - h - hourly
-            - d - days
-            - l - last_processed
-            - nd - nearest day
-            - np - nearest pointing
-
-            past and future should end with one of these options. Eg.
-                ["-3p", "3p"] means 3 pointing
-                ["-3d", "5d"] means 5 days
-                ["-2h", "2h"] means 2 hours
-                ["1l"] means last processed
-                ["6np"] means nearest 6 pointing
-
-        Validation is performed for each field.
-        """
+        """Validate all fields on construction."""
         self._validate_source(self.source)
         self._validate_data_type(self.data_type)
         self._validate_descriptor(self.descriptor)
@@ -80,28 +80,25 @@ class DependencyNode:
         if not date_range:
             return
 
-        if not isinstance(date_range, (list)) or not (1 <= len(date_range) <= 2):
+        if not isinstance(date_range, list) or not (1 <= len(date_range) <= 2):
             raise ValueError(
                 "Date range must be a list of 1-2 elements [past] or [past, future], "
                 f"got {date_range}"
             )
 
         # Handle both single-element and two-element lists
-        past = date_range[0] if len(date_range) > 0 else None
+        past = date_range[0]
         future = date_range[1] if len(date_range) > 1 else None
 
-        if past is None and future is None:
-            return
+        is_nearest = past.endswith(NEAREST_OPTIONS)
 
-        is_nearest = past.endswith(NEAREST_OPTIONS) if past else False
-
-        # Validate past if provided
+        # Validate past
         if is_nearest:
             past_option = "np" if past.endswith("np") else "nd"
-            past_int = int(past[:-2]) if past[:-2] else None
+            past_int = int(past[:-2])
         else:
-            past_option = past[-1] if past else None
-            past_int = int(past[:-1]) if past else None
+            past_option = past[-1]
+            past_int = int(past[:-1])
 
         # Validate past option and its integer value
         if (past_option not in DATE_RANGE_OPTIONS) or (
@@ -114,15 +111,15 @@ class DependencyNode:
 
         # Validate future if provided
         if future is None:
-            return
+            return True
         elif future.endswith(NEAREST_OPTIONS):
             raise ValueError(
                 "Nearest need to be in this format, [<int><option>, ]. "
                 "Eg. ['6np',] or ['6nd',]"
             )
         else:
-            future_option = future[-1] if future else None
-            future_int = int(future[:-1]) if future else None
+            future_option = future[-1]
+            future_int = int(future[:-1])
 
         # Validate future option and integer value
         if (future_option not in DATE_RANGE_OPTIONS) or (future_int < 0):
@@ -173,13 +170,20 @@ class UpstreamDependencyNode(DependencyNode):
     reprocessing: bool = False
     repoint: int | None = None
 
-    # check in init funciton that use do pass in start_date and
-    # end_date for this node.
     def __post_init__(self):
-        """Validate that start_date and end_date are provided."""
-        if self.start_date is None or self.end_date is None:
+        """Validate all fields, including base fields and start_date/end_date."""
+        super().__post_init__()
+        self._validate_dates(self.start_date, self.end_date)
+
+    def _validate_dates(self, start_date: datetime, end_date: datetime) -> None:
+        """Validate start_date and end_date are datetime instances."""
+        if not isinstance(start_date, datetime):
             raise ValueError(
-                "start_date and end_date must be provided for UpstreamDependencyNode"
+                f"'start_date' must be a datetime instance, got {type(start_date)}"
+            )
+        if not isinstance(end_date, datetime):
+            raise ValueError(
+                f"'end_date' must be a datetime instance, got {type(end_date)}"
             )
 
 
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
@@ -2,14 +2,22 @@
 
 from dataclasses import asdict, dataclass
 from datetime import datetime
-from typing import Any
+from typing import Any, ClassVar
 
 from .. import VALID_CADENCE_STRS
+from ..dependency import DataSource, DataType
+
+# Date range validation constants
+NEAREST_OPTIONS = ("nd", "np")
+DATE_RANGE_OPTIONS = ("p", "h", "d", "l", *NEAREST_OPTIONS)
 
 
 @dataclass
 class DependencyNode:
     """Shared between batch starter and dependency lambda."""
+
+    _data_source_validator: ClassVar[DataSource] = DataSource()
+    _data_type_validator: ClassVar[DataType] = DataType()
 
     source: str
     data_type: str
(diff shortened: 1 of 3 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -2,22 +2,65 @@
...
+            - l - last_processed
+            - nd - nearest day
+            - np - nearest pointing
+
+            past and future should end with one of these options. Eg.
+                ["-3p", "3pm"] means 3 pointing
+                ["-3d", "5d"] means 5 days
+                ["-2h", "2h"] means 2 hours
+                ["1l"] means last processed
+                ["6np"] means nearest 6 pointing
```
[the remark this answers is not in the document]
tech3371: good catch on typo

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -27,6 +70,92 @@ def deserialize(cls, json_object: dict[str, Any]):
...
 
+    def _validate_boolean_fields(self, required: bool, kickoff_job: bool) -> None:
+        """Validate required and kickoff_job are booleans."""
+        if not isinstance(required, bool) or not isinstance(kickoff_job, bool):
+            raise ValueError("'required' and 'kickoff_job' must be boolean values")
+
+    def _validate_date_range(self, date_range) -> None:
+        """Validate date range format if provided."""
+        if not date_range:
+            return
```
tmplummer, line 81: Should this check for `is None` rather than just checking falsey?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -27,6 +70,92 @@ def deserialize(cls, json_object: dict[str, Any]):
...
+        if past is None and future is None:
+            return
+
+        is_nearest = past.endswith(NEAREST_OPTIONS) if past else False
+
+        # Validate past if provided
+        if is_nearest:
+            past_option = "np" if past.endswith("np") else "nd"
+            past_int = int(past[:-2]) if past[:-2] else None
+        else:
```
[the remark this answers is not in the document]
tmplummer: It might be worth using regex for this somewhat complex validation.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -27,6 +70,92 @@ def deserialize(cls, json_object: dict[str, Any]):
...
+        # TODO: validate descriptor once we finalize the descriptor list
+        # for each instrument and data type.
+        if not isinstance(descriptor, str) or not descriptor.strip():
+            raise ValueError(
+                f"Descriptor must be a non-empty string, got '{descriptor}'"
+            )
+
 
 @dataclass
 class UpstreamDependencyNode(DependencyNode):
```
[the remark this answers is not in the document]
tmplummer: I think this is a good comment. Can we and do we want to validate UpstreamDependencyNode as well? You can probably just remove the `__post_init__` method from this class.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -101,3 +230,46 @@ def get_cadence_duration(descriptor: str) -> str | None:
...
+            f"Node dict must contain keys {required_keys}, got {set(yaml_dict.keys())}"
+        )
+    return {
+        "source": yaml_dict["upstream_source"],
+        "data_type": yaml_dict["upstream_data_type"],
+        "descriptor": yaml_dict["upstream_descriptor"],
+        "required": yaml_dict.get("required", True),
+        "kickoff_job": yaml_dict.get("kickoff_job", True),
+        "date_range": yaml_dict.get("date_range", None),
+    }
```
tmplummer: I *think* that this is where you get validation for free by just doing `return UpstreamDependencyNode(**yaml_dict)`.

tmplummer: Oh, I see now that there is some translation in key names needed. This looks fine.

tmplummer (approved):
Cool, I like this change!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -2,22 +2,65 @@
...
+
     source: str
     data_type: str
     descriptor: str
     required: bool = True
     kickoff_job: bool = True
     date_range: list = None
 
+    def __post_init__(self):
+        """Validate all fields on construction.
```
maxineofficial: I think most of this docstring should be under the class definition

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -27,6 +70,92 @@ def deserialize(cls, json_object: dict[str, Any]):
...
+            return
+
+        if not isinstance(date_range, (list)) or not (1 <= len(date_range) <= 2):
+            raise ValueError(
+                "Date range must be a list of 1-2 elements [past] or [past, future], "
+                f"got {date_range}"
+            )
+
+        # Handle both single-element and two-element lists
+        past = date_range[0] if len(date_range) > 0 else None
```
maxineofficial: Probably don't need the if-else for this one, since it's checked above.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -27,6 +70,92 @@ def deserialize(cls, json_object: dict[str, Any]):
...
+        future = date_range[1] if len(date_range) > 1 else None
+
+        if past is None and future is None:
```
maxineofficial: Likewise, I don't think this check is needed. You could combine the is instance and `if not date_range` checks above, keep lines 90-94 the same, and remove the `if not isinstance(date_range, (list)) or not (1 <= len(date_range) <= 2):` check.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -27,6 +70,92 @@ def deserialize(cls, json_object: dict[str, Any]):
...
+        if not isinstance(required, bool) or not isinstance(kickoff_job, bool):
+            raise ValueError("'required' and 'kickoff_job' must be boolean values")
+
+    def _validate_date_range(self, date_range) -> None:
+        """Validate date range format if provided."""
+        if not date_range:
+            return
+
+        if not isinstance(date_range, (list)) or not (1 <= len(date_range) <= 2):
+            raise ValueError(
```
maxineofficial, line 84: It's confusing that sometimes, this function returns, and sometimes, it throws an error. I think it would make sense to have these validation functions either always throw, or return True/False.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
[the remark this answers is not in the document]
maxineofficial: I agree regex would make this cleaner and more reliable. You could just split the string into digits and characters, then check the characters against the options

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -2,22 +2,65 @@
...
+            [
+                source,
+                data_type,
+                descriptor,
+                required,
+                kickoff_job,
+                Optional([past, future])
+            ]
+        If it includes past/future date ranges, it should follow the following format:
+            - p - pointing
```
maxineofficial: I think it would be more reliable if these were all 2 character format strings.

could even just use "s"

ps
hs
ds
lp
nd
np

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -2,22 +2,65 @@
...
+            - d - days
+            - l - last_processed
+            - nd - nearest day
+            - np - nearest pointing
+
+            past and future should end with one of these options. Eg.
+                ["-3p", "3p"] means 3 pointing
+                ["-3d", "5d"] means 5 days
+                ["-2h", "2h"] means 2 hours
+                ["1l"] means last processed
```
maxineofficial: Why not negative here? Should keep this consistent

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
maxineofficial, line 173: This function is not needed if you just make start_date and end_date required attributes (ie remove `= None`)

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
```diff
@@ -101,3 +230,46 @@ def get_cadence_duration(descriptor: str) -> str | None:
         return cadence
 
     return None
+
+
+def format_upstream_node_input(yaml_dict: dict) -> dict:
```
maxineofficial: Why not just return a DependencyNode object here?

tests/lambda_endpoints/test_dependency_new.py
```diff
@@ -92,63 +89,54 @@ def test_validate_node_dict_valid():
...
-        is True
     )
 
 
 def test_validate_node_dict_with_date_range():
-    """Test validation of dict nodes with date range."""
-    config = DependencyConfigReader()
-    assert (
-        config.validate_node(
+    """Test that dict nodes with date range instantiate without error."""
```
maxineofficial, line 99: Can you add a test for all date range options (maybe parameterized)

maxineofficial (changes requested):

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
tech3371: I find extra character confusing

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
tech3371: good catch. last processed should have `-`

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
tech3371: I feel like this start_date and end_date required keeps coming back to me at every PR...lol. We all had same idea. But I can't make these start and end date required parameter if function I am inheriting class has mix of required and optional parameters.

DependencyNode has these optional and it's causing error when I am inheriting DependencyNode class. DependencyNode need these optional parameters
```
    required: bool = True
    kickoff_job: bool = True
    date_range: list = None
```

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
tech3371: This is middle man function that formats input. I can return Dependency Node if we want. No preference here!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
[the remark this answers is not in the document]
tech3371: I will try this.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
maxineofficial: Oh right, of course. Python... sigh. Perhaps this is a sign that `UpstreamDependencyNode` should not inherit from `DependencyNode`, but instead have `dependency_node: DependencyNode` as one of the attributes. 

Either way, could you add a comment in this post init function to explain why they aren't required?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
maxineofficial: But, it would remove all the branching logic where you first check if it's a "nearest" string, and then check for the other options. If you simplify the logic by using regex in _validate_date_range, then this comment isn't really relevant anymore

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
maxineofficial: I think it would make more sense and be more reliable to return DependencyNode instead of a dictionary. You already want people to turn around and put the dictionary into DependencyNode right away, so it might as well be enforced by the code.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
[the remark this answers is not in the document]
tech3371: If you guys don't mind, I like to skip this update for this PR. May be in the future. I like to move onto other pending work.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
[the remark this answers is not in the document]
tech3371: This worked nicely!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/utils.py
tech3371: I did what Tim and Copilot suggested and that gave behaviour we wanted still.

tech3371:
@maxinelasp I addressed most of your comment. Can you review again and unblock merge if you can? Appreciate it!

maxineofficial (approved):

PR #1331, opened 2026-05-12
HIT: custom data dependency updates

### LINKED ISSUE #1330: HIT: add custom l0 file descriptor to HIT dependency tree
We need to add HIT custom L0 descriptor to dependency tree, `pha-telemetry-corrected`.

tech3371:
# Change Summary
closes #1330 

## Overview

Way to trigger HIT pipeline when HIT team delivers us custom L0 data. These L0 data are not from POC or spacecraft.

I don't know how long we need to support it. If it's once, then we push this to prod and ingest HIT L0 file and let it process through and undo this change afterwards. I am excluding HIT L1B HK processing for this work since that's less critical. If we need it, then we can revisit but excluding from this PR to keep this moving fast.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
@@ -712,6 +712,11 @@ hit, l0, raw, hit, l1b, hk, HARD, DOWNSTREAM
 leapseconds, spice, historical, hit, l1b, hk, HARD_NO_TRIGGER, DOWNSTREAM
 spacecraft_clock, spice, historical, hit, l1b, hk, HARD_NO_TRIGGER, DOWNSTREAM
 
+# Custom L0 data delivered by HIT
+hit, l0, pha-telemetry-corrected, hit, l1a, pha-telemetry-corrected, HARD, DOWNSTREAM
+leapseconds, spice, historical, hit, l1a, pha-telemetry-corrected, HARD_NO_TRIGGER, DOWNSTREAM
+spacecraft_clock, spice, historical, hit, l1a, pha-telemetry-corrected, HARD_NO_TRIGGER, DOWNSTREAM
+
 hit, l1a, counts-standard, hit, l1b, standard-rates, HARD, DOWNSTREAM
 hit, l1a, counts-standard, hit, l1b, summed-rates, HARD, DOWNSTREAM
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_config.csv
```diff
@@ -712,6 +712,11 @@ hit, l0, raw, hit, l1b, hk, HARD, DOWNSTREAM
 leapseconds, spice, historical, hit, l1b, hk, HARD_NO_TRIGGER, DOWNSTREAM
 spacecraft_clock, spice, historical, hit, l1b, hk, HARD_NO_TRIGGER, DOWNSTREAM
 
+# Custom L0 data delivered by HIT
+hit, l0, pha-telemetry-corrected, hit, l1a, pha-telemetry-corrected, HARD, DOWNSTREAM
+leapseconds, spice, historical, hit, l1a, pha-telemetry-corrected, HARD_NO_TRIGGER, DOWNSTREAM
+spacecraft_clock, spice, historical, hit, l1a, pha-telemetry-corrected, HARD_NO_TRIGGER, DOWNSTREAM
+
```
vmartinez-cu, line 719: Will `pha-telemetry-corrected` be a descriptor in the output l1a files? Just double checking since l1b processing won't expect this descriptor. L1b expects `counts-standard` and `counts-sectored` for science data and L0 `raw` for housekeeping data.

l1a processing currently outputs four products:
        "imap_hit_l1a_counts-standard"
        "imap_hit_l1a_counts-sectored"
        "imap_hit_l1a_direct-events"
        "imap_hit_l1a_hk"

tech3371: Looking at your HIT code, I think it should produce these four l1a files still since we don't pass descriptor into the l1a code.

My idea is to keep l1a code just the way you had it. Right now, we pass `all` descriptors but we still produces those four products you listed. Similarly, we are passing `pha-telemetry-corrected` but we should produce same four products.

vmartinez-cu: Okay great! In that case, hit processing will need to be updated to ensure that the quality flag that gets set in l1a (triggered by the `pha-telemetry-corrected` descriptor) carries through to L2 since this descriptor won't be used in downstream processing

vmartinez-cu (approved):

Issue #1315, opened 2026-05-07
BUG - determine job version does not include repoint number


tmplummer:
### Description of the issue

On April 7 2026, the spacecraft did a TCM which caused the day to have three repoints instead of the nominal one. This resulted in the ENAs having multiple L0 files with that date, but different repoint numbers. When versioning jobs, the batch starter does not account for repoint number which means it just incremented version for each of the repoints even though they should have all been v001.

Listing Hi L1A files shows the issue:
```
❯ IMAP_API_KEY=$IMAP_PROD_API_KEY imap-data-access query --table science --instrument hi --start-date 20260407 --end-date 20260407 --data-level l1a --descriptor 45sensor-hk
Found [3] matching files in science table
|--------------------------------------------------------------------------------------------------------------------------------------------------------|
| Instrument | Data Level | Descriptor  | Start Date | Ingestion Date    | Version | Repointing | Filename                                               |
|--------------------------------------------------------------------------------------------------------------------------------------------------------|
| hi         | l1a        | 45sensor-hk | 20260407   | 20260408 01:36:50 | v001    | 209        | imap_hi_l1a_45sensor-hk_20260407-repoint00209_v001.cdf |
| hi         | l1a        | 45sensor-hk | 20260407   | 20260408 01:36:49 | v002    | 210        | imap_hi_l1a_45sensor-hk_20260407-repoint00210_v002.cdf |
| hi         | l1a        | 45sensor-hk | 20260407   | 20260410 02:25:57 | v003    | 211        | imap_hi_l1a_45sensor-hk_20260407-repoint00211_v003.cdf |
|--------------------------------------------------------------------------------------------------------------------------------------------------------|
```

### Code Snippet (If applicable)

```python
Code
```

bryan-harter:
Verified to work from latest internal SIT test!

PR #1332, opened 2026-05-13
Multiple repoints on the same day should each start at v001

vineetbansal:
Closes issue #1315 

I hope I'm understanding the issue correctly - no worries if I'm totally off - feel free to close without merging, or suggest changes (or tweak the branch yourself).

# Change Summary

## Overview

Added an `xfail` test to demonstrate the issue in the first commit, and fixed it in a subsequent commit.

## File changes

Added an argument `repointing` to `determine_job_version`, populated at the call site in `submit_all_jobs` (minus the `spacecraft_pointing_attitude_job` case which I don't think are repoint specific?), and the `cadence_processing_event` call site where my understanding is that jobs are resolved over a date range, not per pointing.

## Testing

Added a test `test_determine_job_version_per_repointing`, `xfail` at first to demonstrate the issue with a subsequent fix.


tmplummer (approved):
This looks like it fixes the ticket that I wrote. I think that there is still a blocker that will keep us from reprocessing these data on 4/7 because the repoint number does not flow through all the way from bulk reprocessing trigger, or s3 trigger to job submission. This may be something to not worry about since all this code is being re-worked. Hopefully, the Dagster partitioning will take care of all this for us.

PR #1334, opened 2026-05-13
I-ALiRT - vpn tunnel

### LINKED ISSUE #1339: I-ALiRT - Create VPN Tunnel
### Summary

_No response_

### Requirements

The more specific the better.
  - [ ]
  - [ ]

### Related tickets

_No response_

laspsandoval:
What we provide to NOAA:
  - IKE Gateway IPs 
  - BGP IP LASP
  - AWS-side ASN
  - I-ALiRT's EIP + port
  - LASP's advertised BGP prefixes

What NOAA provides to us:
  - Their border router public IPs (WASH + DENV) (stackbuilder.py:460)
  - Their ASN (ialirt_vpn_construct.py:124)
  - The PSK (stackbuilder.py:446)

 Neither side "provides" - needs to go back to NOAA as a platform limitation:
  - BGP MD5 string - AWS's managed Site-to-Site VPN doesn't support this.

**Notes:**

- NOAA has two border routers — one in McLean, VA (WASH) and one in Denver, CO (DENV) — each with a public IP. Two sites exist purely for redundancy: if one goes down, traffic reroutes through the other. Each NOAA router opens an IPSec (IKEv2) tunnel to AWS, authenticated with a shared secret (PSK).

**Why we used Transit Gateway:**

A Virtual Private Gateway (VGW) is the "default" way to terminate a Site-to-Site VPN. It attaches directly to a VPC and injects routes into that VPC's subnet route tables. A VGW cannot route decrypted VPN traffic to a NAT Gateway.

Why that's a problem here: I-ALiRT's actual endpoint is an EC2 instance in an Auto Scaling Group. Its private IP isn't stable. If the ASG replaces the instance, the private IP changes. What is stable is its Elastic IP, which is reached by going out through a NAT Gateway. So NOAA's decrypted traffic needs to land in the VPC and then get handed to the NAT Gateway so it can reach that Elastic IP - exactly the hop a VGW can't do.

A Transit Gateway can do that hop. A TGW attaches to the VPC and its traffic follows whatever the subnet's own route table says - including a NAT Gateway route. So terminating the VPN on a TGW instead of a VGW means the decrypted NOAA traffic gets funneled through the NAT Gateway, out to the Elastic IP, and hits the instance the same way everyone else's traffic does.

----------------------------------
Screenshot 2026-07-22 at 9 40 03 AM

**- Step 1: NOAA sends data**

The packet has:
- **Source IP**: NOAA's internal private IP
- **Destination IP**: the I-ALiRT EIP

**- Step 2: NOAA's border router encrypts the packet**

The router has a rule: Traffic destined for the I-ALiRT EIP goes through the VPN tunnel

The router wraps the original packet in an encrypted IPSec envelope. The outer envelope has:
- **Source IP**: NOAA border router public IP  (WASH) or  (DENV)
- **Destination IP**: AWS's auto-assigned VPN tunnel endpoint public IP

**- Step 3: Travels across public internet**

**- Step 4: AWS Transit Gateway decrypts the packet**

* Note: NOAA border router public IPs go into the CfnCustomerGateway, which is essentially AWS's record of "this is NOAA's router, trust it.

 AWS decrypts the envelope using the pre-shared key (Secrets Manager, /ialirt/noaa/noaa-vpn-psk).

The outer envelope is discarded. The original inner packet is restored:
- **Source IP**: NOAA's private internal IP (TBD Wallops/Suitland address)
- **Destination IP**: the I-ALiRT EIP

**- Step 5: Transit Gateway routes the packet into the VPC**

  TGW route table sends it through the VPC attachment, into a private subnet. Packet is unchanged.

**- Step 6: Private subnet routes the packet to the NAT Gateway**

Default route (0.0.0.0/0) sends it there, since the EIP isn't inside the VPC's own address range. The NAT Gateway rewrites the source IP to its own EIP (NOAA's original source IP is gone at
  this point):
  - Source IP: NAT Gateway's own EIP
  - Destination IP: the I-ALiRT EIP

**- Step 7: NAT Gateway hands off to Internet Gateway, which translates EIP → private IP**

**- Step 8: Security group check**

  SG sees source = NAT Gateway's own EIP (not NOAA's real IP), checked against the allow-list.

**- Step 9: Container receives packet**

----------------------------------
**Testing**
To validate the NOAA N-Wave Site-to-Site VPN connection ahead of the real NOAA onboarding, I stood up an EC2 instance running strongSwan as a stand-in for NOAA's border router, configured
  with the IKEv2/AES-256-CBC/SHA2-256/DH group14 parameters specified in the ICD, and deployed the IalirtStack CDK infrastructure — which routes the NOAA VPN connection through a Transit
  Gateway to a NAT Gateway and I-ALiRT's stable Elastic IP in a sandbox AWS account for testing. This architecture was tested end-to-end on the WASH site's Tunnel 1, with a TCP connection successfully established from the strongSwan test host through the IPsec tunnel, BGP-derived routing, NAT Gateway, and security group to I-ALiRT's Elastic IP, confirming actual data packets traverse the full path.

**Detailed reproducible instructions**

 0. Prerequisites: A running EC2 acting as the test "NOAA router" with strongSwan + FRR already installed (we reused an existing one: dev account).
  
 1. Confirm the stack deployed the VPN resources
  aws cloudformation describe-stack-resources --stack-name IalirtStack \
    --profile lcs --region us-west-2
  Look for AWS::EC2::VPNConnection, AWS::EC2::CustomerGateway, AWS::EC2::TransitGateway* entries.

  2. Look up the tunnel endpoint you're bringing up (e.g. WASH)
  aws ec2 describe-vpn-connections --vpn-connection-ids <wash-vpn-id> \
    --region us-west-2 --profile lcs \
    --query 'VpnConnections[].Options.TunnelOptions[].{OutsideIp:OutsideIpAddress,InsideCidr:TunnelInsideCidr}' \
    --output json
  Take Tunnel 1's OutsideIp and InsideCidr.

  3. Point strongSwan at that tunnel
  On the test EC2, edit /etc/swanctl/conf.d/noaa-aws.conf: set remote_addrs, the remote { id = ... } block, and the secrets.ike-tunnel1.id-2 entry all to the OutsideIp from step 2.

  4. Point FRR's BGP config at the AWS-side inside IP
  Edit /etc/frr/frr.conf: router bgp <NOAA ASN, e.g. 64583>, neighbor <AWS-side inside IP, e.g. x.x.x.x/30's .1> remote-as 64512 (AWS's amazon_side_asn), and a network statement for whatever
  prefix you want to test-advertise.

  5. Bring the tunnel up
  sudo swanctl --load-all
  sudo swanctl --initiate --child aws-tunnel1

  6. Verify BGP came up
  sudo vtysh -c "show bgp summary"
  Expect ESTABLISHED against the AWS-side inside IP.

  7. Confirm from the AWS side
  aws ec2 describe-vpn-connections --vpn-connection-ids <wash-vpn-id> \
    --region us-west-2 --profile lcs --query 'VpnConnections[].VgwTelemetry' --output json
  Expect Status: UP and a nonzero AcceptedRouteCount on that tunnel's OutsideIpAddress.

  8. Find I-ALiRT's current Elastic IP (dynamically Lambda-assigned per instance, not static — always look it up fresh)
  aws ec2 describe-instances \
    --filters "Name=tag:aws:cloudformation:stack-name,Values=IalirtStack" "Name=instance-state-name,Values=running" \
    --region us-west-2 --profile lcs \
    --query 'Reservations[].Instances[].{InstanceId:InstanceId,PublicIp:PublicIpAddress}' --output json

  9. Send test traffic, explicitly sourced from the test EC2's private IP
  echo "test" | nc -v -s <test-ec2-private-ip> <ialirt-eip-from-previous-step> 7563 -i 1
  Expect "Connection ... succeeded!".

### DIFF
--- a/sds_data_manager/constructs/ialirt_vpn_construct.py
+++ b/sds_data_manager/constructs/ialirt_vpn_construct.py
@@ -29,8 +29,9 @@ class IalirtVpnConstruct(Construct):
             The Virtual Private Gateway to attach the VPN connections to.
         psk : str
             Pre-shared key for IKE authentication. Pass a CDK token from
-            ``secret_value_from_json(...).to_string()`` so the value is resolved
-            by CloudFormation at deploy time and never appears in the template.
+            ``secret_value_from_json(...).unsafe_unwrap()`` so the value is
+            resolved by CloudFormation at deploy time and never appears in
+            the template.
         wash_ip : str
             NOAA border router public IP at McLean, VA (WASH), retrieved from SSM.
         denv_ip : str
--- a/sds_data_manager/constructs/ialirt_vpn_construct.py
+++ b/sds_data_manager/constructs/ialirt_vpn_construct.py
@@ -11,7 +11,7 @@ class IalirtVpnConstruct(Construct):
         self,
         scope: Construct,
         construct_id: str,
-        vpn_gateway: ec2.CfnVPNGateway,
+        transit_gateway_id: str,
         psk: str,
         wash_ip: str,
         denv_ip: str,
@@ -25,8 +25,13 @@ class IalirtVpnConstruct(Construct):
             Parent construct.
         construct_id : str
             A unique string identifier for this construct.
-        vpn_gateway : ec2.CfnVPNGateway
-            The Virtual Private Gateway to attach the VPN connections to.
+        transit_gateway_id : str
+            The Transit Gateway to attach the VPN connections to. A Transit
+            Gateway is used (rather than a Virtual Private Gateway) because a
+            VGW cannot route decrypted VPN traffic to a NAT Gateway, and a NAT
+            Gateway is required so NOAA's traffic reaches I-ALiRT via its
+            stable Elastic IP instead of an EC2 private IP that changes
+            whenever the Auto Scaling Group replaces the instance.
         psk : str
             Pre-shared key for IKE authentication. Pass a CDK token from
             ``secret_value_from_json(...).unsafe_unwrap()`` so the value is
@@ -106,9 +111,13 @@ class IalirtVpnConstruct(Construct):
         # Every AWS Site-to-Site VPN connection automatically provisions
         # two auto-assigned tunnel IPs.
         # These LASP IKE Gateways must be given to NOAA.
+        self.vpn_connections: dict[str, ec2.CfnVPNConnection] = {}
         for site, ip in {"WASH": wash_ip, "DENV": denv_ip}.items():
             # AWS needs to know the router's public IP and ASN to establish the tunnel.
-            # bgp_asn=64583 is NOAA's ASN per the ICD.
+            # bgp_asn=64583 is NOAA's ASN per the ICD. This must match the ASN
+            # configured on NOAA's actual router — AWS silently rejects the BGP
+            # session (not the tunnel itself) if the peer AS doesn't match what's
+            # registered here.
             cgw = ec2.CfnCustomerGateway(
                 self,
                 f"NoaaCustomerGateway{site}",
@@ -117,17 +126,18 @@ class IalirtVpnConstruct(Construct):
                 type="ipsec.1",
             )
 
-            # Create the VPN connection between our Virtual Private Gateway (VGW)
-            # and NOAA's customer gateway. Each connection gets two tunnels by default
-            # (AWS requirement for redundancy) — both use the same crypto settings.
-            # BGP is used (static_routes_only=False) so that if one site (WASH or DENV)
-            # goes down, BGP automatically reroutes traffic through the other.
-            # Data flows one way: NOAA sends to us. We do not send to NOAA.
-            ec2.CfnVPNConnection(
+            # Create the VPN connection between our Transit Gateway (TGW) and
+            # NOAA's customer gateway. Each connection gets two tunnels by
+            # default (AWS requirement for redundancy) — both use the same
+            # crypto settings. BGP is used (static_routes_only=False) so that
+            # if one site (WASH or DENV) goes down, BGP automatically reroutes
+            # traffic through the other. Data flows one way: NOAA sends to us.
+            # We do not send to NOAA.
+            self.vpn_connections[site] = ec2.CfnVPNConnection(
                 self,
                 f"NoaaVpnConnection{site}",
                 customer_gateway_id=cgw.ref,
-                vpn_gateway_id=vpn_gateway.ref,
+                transit_gateway_id=transit_gateway_id,
                 type="ipsec.1",
                 static_routes_only=False,
                 vpn_tunnel_options_specifications=[tunnel, tunnel],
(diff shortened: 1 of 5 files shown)


laspsandoval:
I think we won't need this:

d)	The following networks/prefixes requiring access to NOAA resources will be routed to NOAA N-Wave via static routing by LASP:
i)	###.###.##.#/26

if we use BGP

ahotasu:
@laspsandoval This is a great fix for security purposes, but the system (LASP's IMAP? AWS infrastructure) still won't be covered by an SSP. Is Rachel going to add the AWS infrastructure to LASP's institutional SSP?

laspsandoval:
Outdated.

sds_data_manager/constructs/ialirt_vpn_construct.py
```diff
@@ -0,0 +1,144 @@
...
+        """Create NOAA N-Wave customer gateways and VPN connections.
+
+        Parameters
+        ----------
+        scope : Construct
+            Parent construct.
+        construct_id : str
+            A unique string identifier for this construct.
+        transit_gateway_id : str
+            The Transit Gateway to attach the VPN connections to. A Transit
```
lacoak21, line 29: Nice - did you find this out by trail and error?

sds_data_manager/utils/stackbuilder.py
```diff
@@ -434,6 +438,173 @@ def build_sds(
         account_name=account_name,
     )
 
+    # Retrieve the NOAA VPN pre-shared key from Secrets Manager.
```
lacoak21: stackbuilder.py is getting pretty dense. Would it make sense to extract some of this out to another module?

sds_data_manager/utils/stackbuilder.py
```diff
@@ -434,6 +438,173 @@ def build_sds(
...
+    # "/ialirt/noaa/noaa-vpn-psk" before deploying this stack.
+    noaa_vpn_psk = (
+        secretsmanager.Secret.from_secret_name_v2(
+            ialirt_stack, "NoaaVpnPsk", "/ialirt/noaa/noaa-vpn-psk"
+        )
+        .secret_value_from_json("psk")
+        .unsafe_unwrap()
+    )
+
+    # Retrieve NOAA's border router IPs from SSM Parameter Store.
```
lacoak21: If these are not available will the tunnel still be created?

lacoak21 (approved):
LGTM! I tried tracking the flow down and it looks good. This was on a high level though so the testing will be key. Great work!!

sds_data_manager/constructs/ialirt_vpn_construct.py
laspsandoval: Yes. This was a trial and error situation.

sds_data_manager/utils/stackbuilder.py
laspsandoval: No. The tunnel will only be created if those are available.

sds_data_manager/utils/stackbuilder.py
laspsandoval: Done

Issue #1340, opened 2026-05-14
Dependency Hash - sort before serializing.


lacoak21:
Currently there is no sorting done on the ProcessingInputCollection before it gets hashed in batch_starter.py. I think for most cases its fine but if there are any small tweaks or updates to dependencies it may cause a different order in the files and therefore produce a new dependency hash although it may contain the same files.

PR #1353, opened 2026-05-19
Batch starter refactor work

maxineofficial:
# Change Summary
Key changes:

## Overview
- Add types and some batch starter functions
- Add tests

The main focus of this PR is adding some shared types/interfaces that we can use in dagster. I also added some functions from batch starter to ensure that the types/interfaces make sense and work with the functions we will require. 

The dependency updates can be safely ignored, again they are mostly to demonstrate what I am expecting those functions to return, but I am happy to remove them if they conflict with dagster work.


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
maxineofficial, line 1: This is the most critical file to review, as it defines some shared interfaces and types for passing around into the different pieces of information for a processing job determination scheme. I focused on using functions rather than classes for most of the actual algorithms, as I think that will be more helpful with Dagster than a large class containing functions.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
```diff
@@ -13,10 +15,143 @@
 
 
 @dataclass
-class DependencyNode:
-    """Shared between batch starter and dependency lambda.
+class TimeRange:
+    """A date range with optional pointing numbers.
+
+    Stores start and end times as np.datetime64 values and provides
+    conversion to and from the yyyymmdd string format used in filenames.
```
tmplummer: Something to be aware of... `np.datetime64` does not have leapseconds. Each day is considered to be exactly 86400 seconds long. So, round-tripping, from string to np.datetime64 to string, will be ok, but any case like converting from TDB seconds to np.datetime64 to string must be done carefully.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
```diff
@@ -41,25 +176,19 @@ class DependencyNode:
...
+    ProcessingInputCollection for job starting.
+    """
 
-    source: str
-    data_type: str
-    descriptor: str
     required: bool = True
     kickoff_job: bool = True
-    date_range: list = field(default_factory=list)
+    relative_time_range: list = field(default_factory=list)
```
tech3371: I really like the `TimeRange` class. It's clean and easy way to get start date of the filename.

I want to make sure I don't confuse that time with this `relative_time_range`. How are you envisioning this and above to be different? Can you see if my guess is correct?

TimeRange - is the start and end date of a file. Eg. daily files for in-situ, start and end is the day itself. For ENA and glows with repoint ids, start and end date could be day itself or more and it would be more precise in hhmmss.

relative_time_range - is the time to query the upstream for. Eg. look for all dependencies within this date range (last 3 days date to processing date, future 3 days data to processing date).

My main question and concern is how we can make sure that use don't confuse processing date with query date range.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
```diff
@@ -1,190 +1,325 @@
...
+        # Expecting a return of {"status": int, "data": ProcessingInputCollection}
+        # If dependency status is 200, then it means that we have complete set of
+        # dependencies needed.
+        if response["status"] == 200:
+            return collection.deserialize(json.dumps(response["data"]))
+
+    return collection
+
+
+def process_job(potential_job_node: ProcessingJobNode, time_range):
```
tech3371: can we move this function at the top since it calls to all below functions?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
tech3371, line 1: I do like the functional programming here since it seems easier to integrate into dagster!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
```diff
@@ -1,190 +1,325 @@
...
-        str
-            The calculated CRID for the potential job.
-        """
-        # TODO: Update CRID calculation logic or decide if it should be
-        # its own class.
-        # 1. Review and keep logic from current CRID logic
-        # 2. Refactor current CRID logic into this funciton
-        return ""
-
-    def is_duplicate_job(
```
tech3371, line 79: We will need this function

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
```diff
@@ -1,190 +1,325 @@
...
+            dependencies=dependencies,
+            crid=crid,
+            version=version,
+        )
+        # Next: submit the actual job, using submit_job
+
+
+def determine_dependency_node(processing_job: ProcessingJobNode, time_range):
+    """Find the dependency node for a processing job."""
+    return DependencyNode()
```
tech3371: can you expand on what you want to use this function for? Right now, we return all possible jobs and not filtering by `processing_job` or `time_range`.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
```diff
@@ -1,190 +1,325 @@
...
+    because the unique constraint on (dependency_hash, container_image_digest)
+    handles duplicate detection reactively when submitting a new job.
+
+    The duplicate detection works like this: a job is skipped only if BOTH the
+    dependency_hash AND container_image_digest are identical to a previous
+    INPROGRESS or SUCCEEDED job. If either the hash changes or the image digest changes
+    (due to a new software version), a new job submission is allowed with a bumped
+    version.
+
+    Priority order:
```
tech3371, line 96: These comments and docs were so helpful!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/dependency_new.py
tech3371, line 1: I can pick up this one and finish remaining work related to this.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
tech3371: Also, `time_span` in `ProcessingJobNode`. It would be helpful to know how all these are connected and how we can distinguish them

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
```diff
@@ -128,64 +257,60 @@ def _validate_date_range(self, date_range) -> None:
...
-        self._validate_dates(self.start_date, self.end_date)
 
-    def _validate_dates(self, start_date: datetime, end_date: datetime) -> None:
-        """Validate start_date and end_date are datetime instances."""
-        if not isinstance(start_date, datetime):
-            raise ValueError(
-                f"'start_date' must be a datetime instance, got {type(start_date)}"
-            )
-        if not isinstance(end_date, datetime):
+    def convert_to_cli_call(self, version) -> str:
```
tech3371, line 285: like this!

tech3371 (commented):
Thank you Maxine for this work. You cleaned up and connect dots nicely! I had one major question about the date range and time span parameters. Otherwise, changes in this PR looks good!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/dependency_new.py
maxineofficial: It sounds like this will be really related to dagster, so I think it makes sense to delay working on this until we have more information from @bryan-harter about tickets/updates for that system.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
maxineofficial: I'm not sure if it would be useful or not, which is why I didn't fill it in or use it anywhere. It was more to demonstrate that you should be able to go from ProcessingJobNode -> DependencyNode, but I can remove it because it's vestigial.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
maxineofficial: I think the duplicate job checking using the submit job function is sufficient, actually. It uses the processing job table to determine if it's duplicated inside of the not-yet-complete submit_job(). I'm not sure if we will need to have is_duplicate_job() separate from submit_job() yet - curious what @bryan-harter and @lacoak21 think

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
maxineofficial: You're right that TimeRange is a specific time range, with the qualifier that it should ALWAYS include :hhMMss, but sometimes that will just be 00:00:00. start_time should equal the beginning of the repointing in pointing_number_start, and end_time should equal the ending of the repointing in pointing_number_end. 

The dependencynode ONLY has relative time ranges, so +/- 2 days, without specifying WHICH days.

The time range does not specify if it's a processing time range or a dependency query time range. I think it makes the most sense to call get_dependencies with the processing time range and the dependency node relative time range requirements, so inside get_dependencies there will be a step to expand the processing time range to include the relative time range requirements, and then the query returns a processinginputcollection which spans the larger time range. 

ie if MAG needed a 2 day buffer on either side, the processing job might be 2026-02-01, the relative time range would be (-2, 2), and the query range inside get_dependencies would query against 2026-01-30 to 2026-02-03 and return all the files. So, the dependency coverage range is only relevant in that one query and not tracked elsewhere. 

time_span inside ProcessingJobNode is the specific job time range. So, for ENAs, it would be a single repoint, for most in-situ it would be 1 day, for maps it would be 3mo, whatever is needed for one single processing job.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
bryan-harter: I believe try_to_submit_job() should be sufficient in most cases for cancelling duplicate jobs before they start. I believe Luisa basically just used try_to_submit_job() in the IDEX dagster code, rather than do any explicit duplicate checking. She added some code that explains *why* submitting a job failed, and then has some extra code in dagster to handle things from there. So I do think this function can be a bit redundant, though for clarity, we could also move the part of the function that checks for duplicates into this function, and have `try_to_submit_job()` call `is_duplicate_job()`. 

That said, for the dagster work I've done so far, I am actually using `is_duplicate_job()` for a separate reason from the checks in `try_to_submit_job()`, which is to try to determine if the asset was created in the *old* pipeline. The reason why I had to do that is when we first spin Dagster up and materialize thousands of L0 files, Dagster is learning about these files for the first time. Therefore, Dagster will try submitting a boatload of L1a jobs. It will gather up the dependencies for the job, and see that the spacecraft_clock kernel is a newer version in the vast majority of cases, therefore the dependency_hash is going to be different. So try_to_submit_job() will see a different dependency hash, and say it is fine to run. However, presumably we don't want new jobs to run only because the spacecraft clock kernel changed, since that isn't a dependency that usually kicks things off. 

At some point when all the data is in Dagster though, the `is_duplicate_job()` function could go away. That's because new spacecraft clock kernels arriving won't trigger any jobs.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
maxineofficial: Hmm. Annoying. Is there a better datetime tool that you know of that might work better?  Alternatively, we could round up by 1 minute in either direction when querying just to make sure we get all the data - I think that's the biggest issue I can see, the SPICE querying... the rest is _mostly_ full day or full repoints, with more specific time span management in processing with SPICE...

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
maxineofficial: Should hopefully help with future version updates!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
maxineofficial: Could we determine old job systems based on ingestion date or processing job run date instead @bryan-harter? Or maybe we could reverse engineer the dependency hash and compare the spacecraft clock kernel in our first runthough?  Or, could we just ... skip running processing, like temporarily disconnect dagster from the batch starter, let it submit loads of jobs that don't run, and then turn it back on once we have completely switched over? I don't really love having a function hanging around just for the transition period, but it may just be the easiest way regardless.... 

@lacoak21 do you have any thoughts here?

maxineofficial: As a note, I did split `try_to_submit_job()` up a little bit, but the duplicate protection piece you mention is still in the job submission step (renamed to `submit_job()`)

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
```diff
@@ -1,190 +1,325 @@
...
+
+def process_job(potential_job_node: ProcessingJobNode, time_range):
+    """Process the job by resolving dependencies and submitting to batch."""
+    # TODO: potential_job_node represents a range of expected jobs.
+    # Loop here through each 24 hour span or reprocessing.
+
+    dependency_node = determine_dependency_node(potential_job_node, time_range)
+    dependencies = get_dependencies(dependency_node)
+    if dependencies.processing_input:
+        crid = dependency_hash(dependencies.serialize())
```
lacoak21, line 45: Are we switching to using the dependency hash as the CRID? The dependency hash wont have info about second level upstream files meaning we wont be able to tell if we expect a file to be processed. I guess with dagster we really dont need a CRID anyway so maybe this is a moot point.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
```diff
@@ -1,190 +1,325 @@
...
+        The serialized dependencies string OR the ProcessingInputCollection. Either way,
+        the hash will use the output of ProcessingInputCollection.
+
+    Returns
+    -------
+    str
+        The first 8 characters of the SHA-256 hash of the serialized dependencies.
+    """
+    serialized_dependencies = dependencies
+    # TODO: update serialize to sort the files before sending it here, to ensure
```
lacoak21, line 75: I have a ticket for this FYI https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/1340

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
tech3371: Ah ok. This is good context. With this explanation, I like to rename some of these names. Eg. `TimeRange` to `ProcessingTimeRange`. and `relative_time_range` to `dependency_query_time_range`. Finally, may be remove `time_span` and reuse one of `ProcessingTimeRange` or `dependency_query_time_range`. I can take this on if you like!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
tech3371: it's ok. I can do this. it's minor

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
tech3371: remove it :)

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
lacoak21: Hm that is a good point about the spacecraft clock. 

What if we create the dependency hash separately for the "spice basic kernels" (leapsecond and spacraft_clock) and the rest of the deps and concatenate them to make up the dependency hash. That way we dont actually have to reverse engineer the hash and find the kernel we can just subset the first x characters of the hash and know that those are the "spice basic kernels". That way we can have more control over what we consider a duplicate. We should need to then actually implement is_duplicate_job().

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/dependency_new.py
tech3371: can you remove changes in this file besides the s3 upload then? We can add in as needed later. I can do this also.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
tech3371: totally!

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
lacoak21: I agree that disconnecting dagster from batch starter isn't my first choice. It would be nice to just have a toggle on whether we reprocessed if we get a new time kernel. I can see the argument for not adding this work just for this case though.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
```diff
@@ -1,190 +1,325 @@
...
+        repointing=node.time_span.pointing_number_start,
+    )
+    dependency_file_path = dependency_file.construct_path()
+    response = upload_dependency_file(dependency_file_path, dependencies.serialize())
+    if not response:
+        return None
+    return str(dependency_file_path)
+
+
+def submit_job(
```
lacoak21, line 240: In the dagster IDEX implementation I found that it was pretty necessary to add some sort of return info from this function. I basically created a class that contains info about the job submission whether it failed due to a file upload, skipped because it was a duplicate or succeeded. It was also nice to output the ProcessingJob record as a dictionary so dagster could log that for the user. This could be in future work but I just wanted to mention it here in case people had thoughts. 


@dataclass
class BatchJobSubmit:
    """Class to store information about a batch job submission."""
    status: str
    message: str
    # the ProcessingJob information as a dictionary.
    job: dict | None = None

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/dependency_new.py
```diff
@@ -203,20 +211,17 @@ def get_downstream_dependency_nodes(self, input_node: DependencyNode) -> list:
...
         return []
 
-    def get_upstream_dependency(
-        self, session, input_upstream_node: UpstreamDependencyNode
-    ):
+    def get_upstream_dependency(self, session, input_upstream_node: ProcessingJobNode):
         """Get upstream dependencies for a given upstream node.
 
-        UpstreamDependencyNode contains required Inputs:
+        ProcessingJobNode contains required Inputs:
```
lacoak21, line 214: This is a good rename. Definitely clears up the usage.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
```diff
@@ -13,10 +15,143 @@
...
+            Pointing number for the start time.
+        pointing_number_end : int or None, optional
+            Pointing number for the end time.
+
+        Returns
+        -------
+        TimeRange
+            A TimeRange instance with parsed datetime values.
+        """
+        start_time = np.datetime64(
```
lacoak21: Why not use datetime.strptime anymore?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
maxineofficial: I named it `TimeRange` because it can represent a generic time range, not just a processing time range. I can rename relative time range. I don't really understand your last point - time_span is the `TimeRange` associated with that instance of the class, so it already is using `TimeRange`.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
maxineofficial: I think that's a good idea. We can also easily exclude some of the SPICE dependencies from the hash altogether. Is that something you'd like in this PR?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
maxineofficial: I do think we should collapse the CRID and dependency hash into one thing. I am not sure if we need the second level upstream files... is that something that's required for processing still?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
maxineofficial: that is for the datetime library, this uses numpy.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
lacoak21: No I just double checked the dagster run condition checks if any upstream jobs are in progress. So yay we can get rid of the CRID check.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
lacoak21: True I was thinking something like this though 

np.datetime64(datetime.strptime(start_time_string, "%Y%m%d"))

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
lacoak21: Im not sure. Ill bring it up in the optional SDC meeting today. The question that is blocking this is whether in normal situations, do we want a reprocessing to happening with an updated spacecraft kernel? I dont know enough. Maybe @tmplummer or @tenzin knows.

maxineofficial: I think that will be easy to update in the code as-is, so maybe we can merge this and go back if we want to add more dependency hash management. Since this code won't be writing to databases for a while it should be ok to update as-needed

lacoak21: Yeah sounds good. Seems like people were ok with just removing the spacecraft_clock from the dependency hash but Bryan brought up a good point that this still wouldn't fix the dagster init issue because we would have to fix all the dependency hash that currently exist.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
maxineofficial: Sounds good to me - my only correction/question would be if we can use ProcessingJobNode instead of a dict for the job info. We could add a nice print function to the class if it's used in logging.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
maxineofficial: Oh, I understand - I pulled this snippet from the imap-data-access version of this code. I don't really see a specific advantage either way, is there something you're considering here that I am missing? Happy to change it if you prefer.

lacoak21: Ah got it. No strong opinions.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
maxineofficial: Yeah, we would need to find some way to remove the spacecraft clock kernel from all the existing hashes. Seems like the best way would be to just use is_duplicate_job and figure out some way to distinguish the old files. Maybe we could like make the dependency hash one character longer? Then it would be immediately easy to tell if it's in the old system and already run, VS a processing job...

tech3371 (approved):
Approving.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
tech3371: created ticket to revisit this

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/types.py
tech3371: created ticket to revisit this

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/batch_starter_new.py
tech3371: can you create ticket for this if we are not capturing this change in upcoming work of this week? If we are tacking in upcoming work this week, then no need to create ticket.

PR #1380, opened 2026-05-21
updated dependencies for lo/l1b/goodtimes (& bgrates)

### LINKED ISSUE #1379: Update dependencies for lo/l1b/goodtimes and lo/l1b/bgrates
With the pending change in PR #3228 in `imap_processing`, `lo/l1b/goodtimes` and `lo/l1b/bgrates` now depend on `repoint/repoint/historical` to get the repoint id out of which they extract the pointing start time (to put into the `epoch` for cdfs for background rates/sigma).

Dependencies need to be updated to reflect this.

vineetbansal:
Closes #1379 

The `yaml` was missing one product entirely. Added it to make it consistent with the `.csv`.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/dependencies/imap_lo_dependencies.yaml
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/dependencies/imap_lo_dependencies.yaml
@@ -197,6 +197,9 @@ lo_esa_fit_factors: &lo_esa_fit_factors
   - upstream_source: repoint
     upstream_data_type: repoint
     upstream_descriptor: historical
+  - upstream_source: lo
+    upstream_data_type: ancillary
+    upstream_descriptor: bg-rates-anti-ram-overrides
 
 # ======== Level L1C ========
 (l1c, pset):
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/dependencies/imap_lo_dependencies.yaml
```diff
@@ -194,6 +194,18 @@ lo_esa_fit_factors: &lo_esa_fit_factors
   - upstream_source: lo
     upstream_data_type: l1b
     upstream_descriptor: nhk
+  - upstream_source: repoint
+    upstream_data_type: repoint
+    upstream_descriptor: historical
+
+(l1b, bgrates):
```
tmplummer: Wait... It looks to me like the bgrates product is a by-product of the goodtimes job. So, I don't think that we want to have this node defined. From the ongoing infrastructure update, we will likely be adding expected outputs for cases exactly like this, but for now I think we can remove this dependency definition since the batch starter only needs to know about the goodtimes job. This applies to the `bgrates` definition in the `dependency_config.csv` as well.

tmplummer (changes requested):

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/dependency_refactoring/dependencies/imap_lo_dependencies.yaml
vineetbansal: done

vineetbansal: @tmplummer - hold on on this PR. Another small changing coming.. (sorry!)

vineetbansal: @tmplummer - I've added an ancillary dependency that was missing for the product, which I didn't notice till right now. This should be good to merge now.

tmplummer (approved):
LGTM

PR #1383, opened 2026-05-25
Release lambda handler logics

### LINKED ISSUE #1307: Release API Lambda handler logics:
Trigger Source:
`imap-data-access release --instrument --start-date --end-date --release-type --release-number (optional)` -> API endpoint -> Release API Lambda

release types: withhold-data-release, unrelease, early-release

Lambda Handler Logic:
- [x] add dummy file to s3 to populate release table (needs update to indexer) @tech3371 
- [x] look through release table records and find all rows with descriptor `withhold-data-release-<###>` or `unrelease`  that indicates all withhold files  for given instrument and date range and filter latest version.
- [ ] download and read files
- [ ] release everything but withhold files
- [ ] update s3 tag to released=`<released-number>` or `None` if not released.
- [ ] update science table's `released` column table to be associated with release number.

tech3371:
# Change Summary
closes #1307 

## Overview

Release lambda handler logics. Shifted from using manifest files to only through release API/CLI. If manifest file is needed, we can add later but starting with simple approach first.

## File changes

- updated release API to only work through API key endpoint. We only want to make release through `imap-data-access` CLI which will validate if someone as write access through API key validation.
- Added logics to release lambda handler


tech3371:
I tested this in dev and website and it works.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,596 @@
...
+        row.file_path
+        for row in (
+            session.query(science_table.file_path)
+            .join(
+                max_ver_subq,
+                (science_table.instrument == max_ver_subq.c.instrument)
+                & (science_table.descriptor == max_ver_subq.c.descriptor)
+                & (science_table.start_date == max_ver_subq.c.start_date)
+                & (science_table.version == max_ver_subq.c.max_version),
+            )
```
tmplummer, line 189: Does this need to be added?
```suggestion
            .join(
                max_ver_subq,
                (science_table.instrument == max_ver_subq.c.instrument)
                & (science_table.data_level == max_ver_subq.c.data_level)
                & (science_table.descriptor == max_ver_subq.c.descriptor)
                & (science_table.start_date == max_ver_subq.c.start_date)
                & (science_table.version == max_ver_subq.c.max_version),
            )
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,596 @@
...
+        )
+        .group_by(
+            science_table.instrument,
+            science_table.data_level,
+            science_table.descriptor,
+            science_table.start_date,
+        )
+        .subquery()
+    )
+    latest_science_files = [
```
tmplummer: Apparently, you can avoid the list comprehension here by doing:
```
  latest_science_files = (
      session.query(science_table.file_path)
      .join(...)
      .filter(...)
      .scalars()  # Extracts the single column value from each row
      .all()      # Returns list of strings directly
  )```

File: sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py [hunk 3]
```diff
@@ -1,6 +1,596 @@
...
+
+    s3_file_path = (
+        path_obj.construct_path()
+        .relative_to(imap_data_access.config["DATA_DIR"])
+        .as_posix()
+    )
+
+    logger.debug(f"Downloading manifest file from S3 path: {s3_file_path}")
+    download_path = download_from_s3(s3_file_path)
+    logger.debug(f"Download path after download: {download_path}")
```
tmplummer:
```suggestion
    logger.debug(f"Local path after download: {download_path}")
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,596 @@
...
+            release_table.start_date <= end_date,
+            release_table.descriptor == descriptor,
+        )
+        .one_or_none()  # TODO: update this logic if we want to
+        # support more than one withhold file per release batch
+    )
+    return withhold_files
+
+
+def query_latest_science_files(session, instrument, start_date, end_date):
```
tmplummer: I think this will fail to get the correct pointing files for days when there are multiple pointings that start on the same day. e.g. April 7th, Hi has the following files:
```
imap_hi_l1a_45sensor-hk_20260407-repoint00209_v001.cdf
imap_hi_l1a_45sensor-hk_20260407-repoint00210_v002.cdf
imap_hi_l1a_45sensor-hk_20260407-repoint00211_v003.cdf
```

I think this function will only return: `imap_hi_l1a_45sensor-hk_20260407-repoint00211_v003.cdf`
It would be good to add this as a test case.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,596 @@
...
+        "statusCode": 200,
+        "data": {
+            "release_type": release_type,
+            "released_flag": released_flag,
+            "query_params": query_params,
+        },
+    }
+
+
+def query_withhold_files(session, instrument, start_date, end_date, release_number):
```
tmplummer: I think this function also needs to add an optional pointing/repoint input.

See comment below on `query_latest_science_files` function.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,596 @@
...
+    """Handle 'release' type requests."""
+    start_date = datetime.datetime.strptime(query_params["start_date"], "%Y%m%d")
+    end_date = datetime.datetime.strptime(query_params["end_date"], "%Y%m%d")
+
+    with db.Session() as session:
+        # Query for withhold files to exclude from release.
+        science_files_to_exclude = []
+        ancillary_files_to_exclude = []
+
+        withhold_files = query_withhold_files(
```
tmplummer: I'm kind of confused as to why we need to query for a withhold file. Won't the filename be provided as part of the release cli arguments? So you could get it from `query_params["exclude_file"]`

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,596 @@
...
+            query_params["instrument"],
+            start_date,
+            end_date,
+        )
+        if science_files_to_exclude != []:
+            science_files_to_update = [
+                file_path
+                for file_path in science_files_to_update
+                if Path(file_path).name not in science_files_to_exclude
+            ]
```
tmplummer: Would it make more sense to do this filtering as part of the query_latest_science_files query by adding a `not_in()` filter on filename?

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,596 @@
...
+
+        ancillary_files_to_update = get_latest_ancillary_files(
+            session,
+            query_params["instrument"],
+            start_date,
+            end_date,
+        )
+        if ancillary_files_to_exclude != []:
```
tmplummer: Same, I think you could filter in the `get_latest_ancillary_files` function. That would also eliminate the need to split up the whithold files by type. You could just pass in the full list and let the `not_in()` filter do the work.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,596 @@
...
+            ancillary_files_to_update = [
+                file_path
+                for file_path in ancillary_files_to_update
+                if Path(file_path).name not in ancillary_files_to_exclude
+            ]
+
+        # For all science and ancillary files, update released flag for all
+        # applicable files.
+        session.query(models.ScienceFiles).filter(
```
tmplummer: I'm wondering if it would make more sense to return the query results from the `get_latest_` functions instead of extracting the filenames. Then you would just need to perform the `update` here on all the query results instead of needing to re-query to get the same DB rows that you already previously grabbed.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,596 @@
...
+            science_files.append(filename)
+        elif isinstance(file_obj, AncillaryFilePath):
+            ancillary_files.append(filename)
+        else:
+            logger.warning(f"Unrecognized file type in manifest, skipping: {filename}")
+
+    return science_files, ancillary_files
+
+
+def release_type_handler(released_flag, query_params):
```
tmplummer: What is the purpose of the `released_flag`? It seems like maybe this function could also unrelease data by passing in `False`?

Edit: Oh, is that the release number? Flag usually indicates a boolean.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,596 @@
...
+            .all()
+        )
+    ]
+    logger.info(
+        f"Found {len(latest_science_files)} science file(s) for instrument={instrument}"
+    )
+    return latest_science_files
+
+
+def get_latest_ancillary_files(
```
tmplummer, line 205: I had trouble understanding this function so I asked Claude to help review. Here is what they came up with:
```
  def get_latest_ancillary_files(
      session,
      instrument: str,
      start_date: datetime.datetime,
      end_date: datetime.datetime,
  ) -> list:
      """Get latest-version ancillary files for an instrument over a date range."""
      ancillary_table = models.AncillaryFiles

      # Step 1: Get latest version per (descriptor, start_date, end_date)
      # Filter by instrument early to reduce data processed
      row_num_col = (
          func.row_number()
          .over(
              partition_by=[
                  ancillary_table.descriptor,
                  ancillary_table.start_date,
                  ancillary_table.end_date,
              ],
              order_by=ancillary_table.version.desc(),
          )
          .label("row_num")
      )

      latest_versions = (
          session.query(
              ancillary_table.file_path,
              ancillary_table.descriptor,
              ancillary_table.start_date,
              ancillary_table.end_date,
          )
          .filter(ancillary_table.instrument == instrument)
          .add_columns(row_num_col)
          .subquery()
      )

      latest = (
          session.query(
              latest_versions.c.file_path,
              latest_versions.c.descriptor,
              latest_versions.c.start_date,
              latest_versions.c.end_date,
          )
          .filter(latest_versions.c.row_num == 1)
          .subquery()
      )

      # Step 2: Files WITH end_date - simple overlap check
      with_end_date_query = session.query(latest.c.file_path).filter(
          latest.c.end_date.isnot(None),
          latest.c.start_date <= end_date,
          latest.c.end_date >= start_date,
      )

      # Step 3: Files WITHOUT end_date - use LEAD() to find coverage end
      next_start_col = (
          func.lead(latest.c.start_date)
          .over(partition_by=latest.c.descriptor, order_by=latest.c.start_date)
          .label("next_start_date")
      )

      no_end_with_next = (
          session.query(latest.c.file_path, latest.c.start_date, next_start_col)
          .filter(latest.c.end_date.is_(None))
          .subquery()
      )

      no_end_date_query = session.query(no_end_with_next.c.file_path).filter(
          no_end_with_next.c.start_date <= end_date,
          or_(
              no_end_with_next.c.next_start_date.is_(None),
              no_end_with_next.c.next_start_date > start_date,
          ),
      )

      # Combine and return
      combined = union_all(with_end_date_query, no_end_date_query).order_by("file_path")

      file_paths = [row.file_path for row in session.execute(combined).fetchall()]
      logger.info(f"Found {len(file_paths)} ancillary file(s) for {instrument=}")
      return file_paths
```

  Summary of Changes
  
Change:  Benefit
- Filter by instrument in Step 1:                 Window functions process only relevant data
- Remove instrument from later queries:  Less data carried through
- Remove version from later queries:        Not needed after ranking
- Simplify union syntax:                             Cleaner code
- Use add_columns():                                Avoids repeating column list

tmplummer (commented):
I can't comment on the infrastructure changes. Mostly dug into the `release_api.py` module.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
tech3371: `release_flag` is a boolean and what value we are setting to. I am over doing it for that for sure and will simplify.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
tech3371: true..good point

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
tech3371: I will try that

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -1,6 +1,585 @@
...
+    return {
+        "statusCode": 200,
+        "data": {
+            "release_type": release_type,
+            "query_params": query_params,
+        },
+    }
+
+
+def query_latest_science_files(
```
tech3371, line 108: @tmplummer I did a quick hack to get repoint query to work. Now it works. This function logic will change drastic once we implement release version. I think it will be simplified a lot. Right now, our different version for different products and level is making this logic long. But that will be simplified in the future.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
tech3371: Added this change and added test for this.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
tech3371: done

tmplummer (approved):
Thanks for making those changes. I did a quick review and didn't notice any other issues.

tech3371:
> Thanks for making those changes. I did a quick review and didn't notice any other issues.

Thank you for those comments. Looks better and I like those simplification!!

PR #1406, opened 2026-06-01
Cadence sensor

### LINKED ISSUE #1402: Create Partition for cadence jobs
### Summary

Refactor cadence cron job to use dagster sensor now.

### Requirements

The more specific the better.
  - [ ] Use dagster sensor every x time daily and see if it’s time for next map partition time. If it’s time, then create partition for the job. eg. cadence_3mo_2026-07-18T15:00:00_to_2026-10-16T22:30:00 will be found ready by sensor on July 18 at 3pm, and dagster creates partition at that time. Partition creation will mean that we will trigger the map job for that partition. This will fit nicely with future progressive map work.
  - [ ] Do that same for every cadence job, `1mo, 3mo, 6mo, 1yr`

### Related tickets

_No response_

tech3371:
# Change Summary
closes #1402

### DIFF
--- a/sds_data_manager/orchestration/custom_partitions.py
+++ b/sds_data_manager/orchestration/custom_partitions.py
@@ -16,7 +16,7 @@ from dagster import (
 from sds_data_manager.lambda_code.SDSCode.database import database as db
 from sds_data_manager.lambda_code.SDSCode.database import models
 from sds_data_manager.orchestration import config
-from sds_data_manager.orchestration.types import MapJobs
+from sds_data_manager.orchestration.maps_utils import get_map_partition_to_create
 
 IDEX_10_DAY_RANGES_PATH = (
     "sds_data_manager/lambda_code/SDSCode/utils/idex_10_day_CDF_names.csv"
@@ -306,10 +306,10 @@ def add_cadence_map_partitions(context: SensorEvaluationContext):
         # already exist, and trigger runs for the new partitions.
         # TODO: If progressive map, retrigger for all valid progressive map partitions
         # based on current time. Eg. call this instead
-        #   MapJobs().get_progressive_map_partition_names()
+        #   get_progressive_map_partition_names()
         missing_partitions = [
             partition_name
-            for partition_name in MapJobs().get_map_partitions_to_create(cadence_str)
+            for partition_name in get_map_partition_to_create(cadence_str)
             if partition_name not in existing_partitions
         ]
         if not missing_partitions:
--- a/sds_data_manager/orchestration/custom_partitions.py
+++ b/sds_data_manager/orchestration/custom_partitions.py
@@ -1,10 +1,12 @@
-"""This file contains information for working with our custom partitions."""
+"""Contains information for working with our custom partitions."""
 
 import datetime
 
 import pandas as pd
 from dagster import (
+    DefaultSensorStatus,
     DynamicPartitionsDefinition,
+    RunRequest,
     SensorEvaluationContext,
     SensorResult,
     StaticPartitionsDefinition,
@@ -14,6 +16,7 @@ from dagster import (
 from sds_data_manager.lambda_code.SDSCode.database import database as db
 from sds_data_manager.lambda_code.SDSCode.database import models
 from sds_data_manager.orchestration import config
+from sds_data_manager.orchestration.types import MapJobs
 
 IDEX_10_DAY_RANGES_PATH = (
     "sds_data_manager/lambda_code/SDSCode/utils/idex_10_day_CDF_names.csv"
@@ -22,14 +25,29 @@ IDEX_30_DAY_RANGES_PATH = (
     "sds_data_manager/lambda_code/SDSCode/utils/idex_30_day_CDF_names.csv"
 )
 
+MISSION_START_TIME = "2026-04-01T00:00:00"
+CADENCE_TRIGGER_HOUR_UTC = 5
+
+cadence_3mo_partitions = DynamicPartitionsDefinition(name="cadence_3mo_partitions")
+cadence_6mo_partitions = DynamicPartitionsDefinition(name="cadence_6mo_partitions")
+cadence_1yr_partitions = DynamicPartitionsDefinition(name="cadence_1yr_partitions")
+
+CADENCE_PARTITION_DEFS = {
+    "3mo": cadence_3mo_partitions,
+    "6mo": cadence_6mo_partitions,
+    "1yr": cadence_1yr_partitions,
+}
+
+
 ##### THIS TELLS DAGSTER THAT SOME FILES ARE DIVIDED UP BY POINTING NUMBER
 repoint_partitions = DynamicPartitionsDefinition(name="repoint_partitions")
 
 
 @sensor(minimum_interval_seconds=600)
 def add_repoint_partitions(context: SensorEvaluationContext):
-    """Periodically polls the PointingTable, and tells dagster that new
-    repoint numbers exist.
+    """Poll the PointingTable and notify dagster of new repoint numbers.
+
+    This sensor identifies new repointing periods from the database.
     """
     with db.Session() as session:
         pointing_records = session.query(models.PointingTable).all()
@@ -125,8 +143,9 @@ idex10_partitions = DynamicPartitionsDefinition(name="idex_10_day_partitions")
 
 @sensor(minimum_interval_seconds=86400)
 def add_idex_10_day_partitions(context: SensorEvaluationContext):
-    """Periodically polls the PointingTable, and tells dagster that
-    new repoint numbers exist.
+    """Poll the PointingTable and notify dagster of new repoint numbers.
+
+    This sensor creates 10-day IDEX partitions.
     """
     start_date = context.cursor or config.MISSION_START_TIME
     start_dt = datetime.datetime.fromisoformat(start_date).replace(
@@ -188,8 +207,9 @@ idex30_partitions = DynamicPartitionsDefinition(name="idex_30_day_partitions")
 
 @sensor(minimum_interval_seconds=86400)
 def add_idex_30_day_partitions(context: SensorEvaluationContext):
-    """Periodically polls the PointingTable, and tells dagster
-    that new repoint numbers exist.
+    """Poll the PointingTable and notify dagster of new repoint numbers.
+
+    This sensor creates 30-day IDEX partitions.
     """
     start_date = context.cursor or config.MISSION_START_TIME
     start_dt = datetime.datetime.fromisoformat(start_date).replace(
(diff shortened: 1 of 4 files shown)


sds_data_manager/orchestration/custom_partitions.py
```diff
@@ -15,10 +16,27 @@
...
 
+MISSION_START_TIME = "2026-04-01T00:00:00"
+CADENCE_SENSOR_HOUR_UTC = 5
+
+cadence_1mo_partitions = DynamicPartitionsDefinition(name="cadence_1mo_partitions")
+cadence_3mo_partitions = DynamicPartitionsDefinition(name="cadence_3mo_partitions")
+cadence_6mo_partitions = DynamicPartitionsDefinition(name="cadence_6mo_partitions")
+cadence_1yr_partitions = DynamicPartitionsDefinition(name="cadence_1yr_partitions")
+
+CADENCE_PARTITION_DEFS = {
```
lacoak21, line 35: Is this what you were talking about in your comment in slack? 

I would keep this definition here and then in imap_jobs.py you could do something like: 

```
partition_map = {
            "daily":   custom_partitions.daily_partitions,
            "repoint": custom_partitions.repoint_partitions,
            "10d":     custom_partitions.idex10_partitions,
            # NOTE: Right now, IDEX is the only instrument who uses 1mo cadence job that
            # maps to exactly 30 days. If this changes, this logic will need update.
            "30d":     custom_partitions.idex30_partitions,
            "3mo":     custom_partitions.idex30_partitions,
            "6mo":     custom_partitions.idex30_partitions,
            "1yr":     custom_partitions.whole_mission_partition,

        } | CADENCE_PARTITION_DEFS
```

sds_data_manager/orchestration/custom_partitions.py
```diff
@@ -15,10 +16,27 @@
...
 from sds_data_manager.orchestration import config
+from sds_data_manager.orchestration.types import CadenceDays
 
 IDEX_10_DAY_RANGES_PATH = "sds_data_manager/lambda_code/SDSCode/utils/idex_10_day_CDF_names.csv"
 IDEX_30_DAY_RANGES_PATH = "sds_data_manager/lambda_code/SDSCode/utils/idex_30_day_CDF_names.csv"
 
+MISSION_START_TIME = "2026-04-01T00:00:00"
+CADENCE_SENSOR_HOUR_UTC = 5
+
+cadence_1mo_partitions = DynamicPartitionsDefinition(name="cadence_1mo_partitions")
```
lacoak21: We can actually remove the 1mo partition. IDEX l2b was the only one that used that and now we have a custom lookup table for it. Its called 30d now.

sds_data_manager/orchestration/custom_partitions.py
```diff
@@ -193,10 +211,70 @@ def add_idex_30_day_partitions(context: SensorEvaluationContext):
         cursor=end_dt.isoformat()
     )
 
+
+
+@sensor(
+    minimum_interval_seconds=3600,  # Check every hour
```
lacoak21: Should this be once a day?

sds_data_manager/orchestration/types.py
```diff
@@ -538,3 +538,106 @@ class ProcessingJobType:
...
+        num_days = self.days - 1
+        if start_date:
+            end_date = start_date + datetime.timedelta(days=num_days)
+
+        if as_str:
+            start_date = start_date.strftime("%Y-%m-%dT%H:%M:%S")
+            end_date = end_date.strftime("%Y-%m-%dT%H:%M:%S")
+        return start_date, end_date
+
+    def get_cadence_partition_name(
```
lacoak21:
```suggestion
    def get_cadence_partition_names(
```

It would be helpful if this was plural.

sds_data_manager/orchestration/custom_partitions.py
```diff
@@ -193,10 +211,70 @@ def add_idex_30_day_partitions(context: SensorEvaluationContext):
...
+
+    if now_date.hour < CADENCE_SENSOR_HOUR_UTC or context.cursor == today:
+        return SensorResult(cursor=context.cursor)
+
+    partition_requests = []
+    run_requests = []
+
+    for cadence_str, partition_def in CADENCE_PARTITION_DEFS.items():
+        existing_partitions = set(context.instance.get_dynamic_partitions(partition_def.name))
+        missing_partitions = [
```
lacoak21, line 310: Maybe add a comment here about what you are doing. 

something like: 

recalculate partitions from the mission launch through now and compare against existing partitions in dagster.

sds_data_manager/orchestration/custom_partitions.py
tech3371: cool!

sds_data_manager/orchestration/custom_partitions.py
tech3371: It checks every hour but sensor kicks of at 5am UTC time below. I don't know of good cron setting in Dagster to kick of at 5am daily :)

sds_data_manager/orchestration/custom_partitions.py
tech3371: good idea!

sds_data_manager/orchestration/types.py
```diff
@@ -538,3 +581,231 @@ class ProcessingJobType:
     POINTING = "pointing"
     CADENCE = "cadence"
     POINTING_ATTITUDE = "pointing_attitude"
+
+
+class CadenceJob:
```
tech3371: @tmplummer What do you think of this as first draft?

I haven't updated dagster logic yet because I wanted to get this in good state first before I move to dagsters.

tmplummer: I think this captures what we talked about. I am having a hard time understanding the bigger picture of how this fits in.

I think there possibly could be some improvement by defining a cadence job by a start datetime and end datetime rather than a string.

I also think that it might be help to define different CadenceJob types by subclassing?

One other thought is for that you probably only need to generate one partition of each type (3mo, 6mo, 1yr) based on the current date.

This is all really vague feedback because I just don't understand how this interacts with dagster and the bigger picture.

tech3371: I refactor to use each cadence type. what do you think of new approach? I added comment in dagsters of how it will be used. I haven't tested dagster component in dev yet but it has rough code and logic in it.

sds_data_manager/orchestration/types.py
```diff
@@ -494,3 +538,184 @@ class ProcessingJobType:
...
+
+
+print(MapJobs(current_time=datetime.datetime(2026, 7, 18, 0, 0, 0, tzinfo=datetime.timezone.utc)).get_map_partitions_to_create("3mo"))
+print(MapJobs().get_map_partitions_to_create("6mo"))
+print(MapJobs().get_progressive_map_partition_names())
+
+# Output of above print statements:
+# ['cadence_3mo_2026-04-18T00:00:00_to_2026-07-18T00:00:00']
+# []
+# ['cadence_3mo_2026-04-18T00:00:00_to_2026-06-02T23:50:40', 'cadence_6mo_2026-01-17T00:00:00_to_2026-06-02T23:50:40']
```
tmplummer: Remove?

sds_data_manager/orchestration/types.py
```diff
@@ -494,3 +494,194 @@ class ProcessingJobType:
...
+    ).get_map_partitions_to_create("3mo")
+)
+print(MapJobs().get_map_partitions_to_create("6mo"))
+print(MapJobs().get_progressive_map_partition_names())
+
+# Output of above print statements:
+# ['cadence_3mo_2026-04-18T00:00:00_to_2026-07-18T00:00:00']
+# []
+# ['cadence_3mo_2026-04-18T00:00:00_to_2026-06-02T23:50:40',
+#  'cadence_6mo_2026-01-17T00:00:00_to_2026-06-02T23:50:40']
```
tmplummer: Remove?

sds_data_manager/orchestration/types.py
```diff
@@ -494,3 +494,194 @@ class ProcessingJobType:
...
+    end: datetime.datetime
+
+    def to_partition_key(self) -> str:
+        """Convert this window to a Dagster partition key string."""
+        start_str = self.start.strftime("%Y-%m-%dT%H:%M:%S")
+        end_str = self.end.strftime("%Y-%m-%dT%H:%M:%S")
+        return f"cadence_{self.cadence}_{start_str}_to_{end_str}"
+
+
+class BaseMapJob:
```
tmplummer: I think I am having trouble understanding what these classes are used for. Maybe adding docstring descriptions would help?

sds_data_manager/orchestration/types.py
```diff
@@ -494,3 +494,194 @@ class ProcessingJobType:
...
+    def get_current_window(self) -> MapWindow | None:
+        """Return the active map window for current_time, if it exists."""
+        for window in self.get_windows():
+            if window.start <= self.current_time < window.end:
+                return window
+        return None
+
+
+class Map3MoJob(BaseMapJob):
+    """Map definition for 3-month partitions."""
```
tmplummer: This comment says it is defining 3-month partitions. Should the class name reflect that it really has to do with partitions rather than jobs? Jobs have a certain context withing dagster, right?

sds_data_manager/orchestration/types.py
```diff
@@ -494,3 +494,194 @@ class ProcessingJobType:
...
+    """Map definition for 1-year partitions."""
+
+    cadence: ClassVar[str] = "1yr"
+    partitions: ClassVar[tuple[str, ...]] = (
+        # Partition is always 365 days (or 366 on leap year).
+        "cadence_1yr_{year}-01-17T00:00:00_to_{year_plus_1}-01-17T00:00:00",
+    )
+
+
+class MapJobs:
```
tmplummer: It might be instructive to put together a sample of how this class will be used. I see a couple methods that are very close to being static methods (`_get_map_job`, `get_current_window`) and am wondering if there is much benefit in this being a class.

tmplummer (commented):
A few things to consider.

sds_data_manager/orchestration/maps_utils.py
tech3371, line 1: @tmplummer can you see if this makes more sense now? I tried to refactor again.

sds_data_manager/orchestration/types.py
```diff
@@ -494,3 +494,149 @@ class ProcessingJobType:
...
+
+    cadence: ClassVar[str] = "3mo"
+    partitions: ClassVar[tuple[str, ...]] = (
+        # First partition can be 91 days(or 92 days on leap year).
+        "cadence_3mo_{year}-01-17T00:00:00_to_{year}-04-18T00:00:00",
+        # Next two partition are always 91 days.
+        "cadence_3mo_{year}-04-18T00:00:00_to_{year}-07-18T00:00:00",
+        "cadence_3mo_{year}-07-18T00:00:00_to_{year}-10-17T00:00:00",
+        # Last partition is always 92 days.
+        "cadence_3mo_{year}-10-17T00:00:00_to_{year_plus_1}-01-17T00:00:00",
```
tmplummer, line 619: I wonder if it would make things cleaner to define these as tuples of `(month, day)` ranges rather than strings.

I fed that idea into Claude and here was the suggestion:
>  1. Redundancy: The partition string format is defined in two places - in the templates AND in MapWindow.to_partition_name()
>  2. Fragile parsing: get_windows() parses strings to extract dates, which is error-prone
>  3. Hard to read: The long template strings obscure the actual boundaries                                                                                                                                                                                      
>                                                                                                                                                                                                                                                              
>  Try this:
```
  from dataclasses import dataclass
  from typing import ClassVar
  import datetime


  @dataclass(frozen=True)
  class MapWindow:
      """Represent an ENA map cadence window with explicit date range."""

      cadence: str
      start: datetime.datetime
      end: datetime.datetime

      def to_partition_name(self) -> str:
          """Convert this window to a Dagster partition key string."""
          start_str = self.start.strftime("%Y-%m-%dT%H:%M:%S")
          end_str = self.end.strftime("%Y-%m-%dT%H:%M:%S")
          return f"cadence_{self.cadence}_{start_str}_to_{end_str}"


  @dataclass(frozen=True)
  class WindowBoundary:
      """A (month, day) boundary point for a map window."""
      month: int
      day: int

      def to_datetime(self, year: int) -> datetime.datetime:
          """Convert to a datetime for the given year."""
          return datetime.datetime(year, self.month, self.day, tzinfo=datetime.timezone.utc)


  class BaseENAMapPartition:
      """Base class for defining ENA map partition windows."""

      cadence: ClassVar[str]
      # Define boundaries as (month, day) pairs - much clearer than string templates
      boundaries: ClassVar[tuple[WindowBoundary, ...]]

      def __init__(self, current_time: datetime.datetime):
          self.current_time = current_time

      def get_windows(self) -> list[MapWindow]:
          """Build all configured windows for the cadence type."""
          windows: list[MapWindow] = []
          year = self.current_time.year

          for i in range(len(self.boundaries) - 1):
              start_boundary = self.boundaries[i]
              end_boundary = self.boundaries[i + 1]

              # Handle year rollover (e.g., Oct -> Jan means end is next year)
              start_year = year
              end_year = year if end_boundary.month >= start_boundary.month else year + 1

              windows.append(MapWindow(
                  cadence=self.cadence,
                  start=start_boundary.to_datetime(start_year),
                  end=end_boundary.to_datetime(end_year),
              ))
          return windows

      def get_current_window(self) -> MapWindow | None:
          """Return the active map window for current_time."""
          for window in self.get_windows():
              if window.start <= self.current_time < window.end:
                  return window
          return None


  class Map3MoPartition(BaseENAMapPartition):
      """3-month map partitions."""

      cadence: ClassVar[str] = "3mo"
      boundaries: ClassVar[tuple[WindowBoundary, ...]] = (
          WindowBoundary(1, 17),   # Jan 17
          WindowBoundary(4, 18),   # Apr 18
          WindowBoundary(7, 18),   # Jul 18
          WindowBoundary(10, 17),  # Oct 17
          WindowBoundary(1, 17),   # Jan 17 (next year - marks end of last window)
      )


  class Map6MoPartition(BaseENAMapPartition):
      """6-month map partitions."""

      cadence: ClassVar[str] = "6mo"
      boundaries: ClassVar[tuple[WindowBoundary, ...]] = (
          WindowBoundary(1, 17),   # Jan 17
          WindowBoundary(7, 18),   # Jul 18
          WindowBoundary(1, 17),   # Jan 17 (next year)
      )


  class Map1YrPartition(BaseENAMapPartition):
      """1-year map partitions."""

      cadence: ClassVar[str] = "1yr"
      boundaries: ClassVar[tuple[WindowBoundary, ...]] = (
          WindowBoundary(1, 17),   # Jan 17
          WindowBoundary(1, 17),   # Jan 17 (next year)
      )
```

>  Benefits:
>  1. Single source of truth: Dates are defined once as structured data, serialization happens in one place (to_partition_name)
>  2. No parsing: Eliminates fragile string parsing in get_windows()
>  3. Readable: WindowBoundary(1, 17) is clearer than parsing "{year}-01-17T00:00:00"
>  4. Validates naturally: Invalid dates like (2, 30) will fail at to_datetime() time
>  5. Easier to maintain: Adding a new cadence just requires listing boundary points

>  The year-rollover logic in get_windows() handles the case where boundaries cross December→January automatically by comparing months.

tmplummer (commented):
Hey, this looks much better to me. I just had one more idea while reviewing... see the long comment that Claude helped me with.

sds_data_manager/orchestration/types.py
tech3371: I will address this in a new PR. This PR was closed as result of dagster branch merging it dev. I will open PR against dev branch now which is where I like to merge :)

bryan-harter:
Oops sorry, I didn't mean to close this!

PR #1409, opened 2026-06-04
Science db update to add release number

### LINKED ISSUE #1399: Update ScienceFileBase() to support `vMMM.mmmm`
### Summary

Update science table, ScienceFileBase, to support release version

### Requirements

The more specific the better.
  - [ ] Create new column for release number, `release_version` or `release_number` and update `version` column to be `major_version` or `data_version` or `cycle_number` (how many cycles data has been rerun). 
  - [ ] If we want to combine both, DB can combine two columns easily.

### Related tickets

_No response_

tech3371:
# Change Summary
closes #1399

## Overview

Adding `release_number` and `data_version`.

## File changes

- Added release_number table
- Updates to science table to add and update column
- DB Migration code (Tested in dev and works)

## Testing

TODO: fix test once table updates are looking good to review.


sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -331,3 +343,13 @@ class IDEXL0Files(Base):
     start_date = Column(DateTime, nullable=False, primary_key=True)
     version = Column(String(4), nullable=False, primary_key=True)  # vXXX
     ingestion_date = Column(DateTime(timezone=True))
+
+
+class GlobalRelease(Base):
+    """Global Release table that tracks release number."""
+
+    __tablename__ = "global_release"
```
tech3371, line 351: I appreciate new name suggestion

sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -331,3 +343,13 @@ class IDEXL0Files(Base):
...
+    id = Column(Integer, Identity(start=1, increment=1), primary_key=True)
+    release_number = Column(Integer, nullable=False, unique=True)
+    updated_date = Column(DateTime, nullable=False)
+    updated_by = Column(String, nullable=False)
```
tech3371: I know we had comment about safe guard updates to this table. That's why I added these two columns. For these two columns, I think we can retrieve names easily if this is desired to track when last update happened and by who. All API key are stored in dynamodb

maxineofficial: I don't think we need to track "updated by" - but maybe we could track "previously updated" date, so like when the previous version was updated?

pyproject.toml
```diff
@@ -117,7 +117,7 @@ lint.select = ["B", "D", "E", "F", "I", "N", "S", "W", "PL", "PT", "UP", "RUF"]
 # S104: Possible binding to all interfaces
 # PLR0913: Too many arguments in function definition
 # PLR2004: Magic value in comparison
-lint.ignore = ["D104", "D203", "D213", "D300", "D413", "S101", "S104", "PLR0913", "PLR2004"]
+lint.ignore = ["D104", "D203", "D213", "D300", "D413", "S101", "S104", "PLR0913", "PLR2004", "S608"]
```
maxineofficial: What is this ignoring? Probably, alembic files should go in the per-file-ignores.

alembic/versions/1892125c2f95_release_version_and_global_release_table.py
```diff
@@ -0,0 +1,129 @@
...
+
+def upgrade() -> None:
+    """Upgrade schema: convert version string → int + add constraints."""
+    # First, add global_release table
+    op.create_table(
+        "global_release",
+        sa.Column(
+            "id",
+            sa.Integer(),
+            sa.Identity(always=False, start=1, increment=1),
```
maxineofficial, line 33: Start at 1? or 0?

alembic/versions/1892125c2f95_release_version_and_global_release_table.py
```diff
@@ -0,0 +1,129 @@
...
+    for table in TABLES:
+        # 1. Rename version to data_version
+        op.alter_column(
+            table,
+            "version",
+            new_column_name="data_version",
+            existing_type=sa.String(length=4),
+        )
+
+        # 2. Convert data_version from "v001" → 1
```
maxineofficial, line 52: I think we'll want to be careful running this, as it will break existing database readers. I would say we should remove it from this PR. The creation steps, for the new table and the new column, don't affect running code so that seems ok. I don't want to merge DB changes separately from the rest of the code updates.

maxineofficial (changes requested):

sds_data_manager/lambda_code/SDSCode/database/models.py
maxineofficial: I'm also fine with updated by, I'm just not sure it will be that helpful in the long run.

pyproject.toml
tech3371: It was because was doing `f"{table}" in alembic files. yep. I can ignore the whole alembic files.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: sounds good. If you and one more person like to remove `updated_by`, I am happy to!

alembic/versions/1892125c2f95_release_version_and_global_release_table.py
tech3371: good catch!

alembic/versions/1892125c2f95_release_version_and_global_release_table.py
tech3371: I personally feel that doing this transition in alembic has been so simple and helpful. I was able to rename column, apply dtype changes and removing `v` prefix through alembic, and finally converting to int. And revert all those changes back to what we have was so easily too. There is so much benefit doing through this.

If your concerns is with clashing with current code reading it, that's a concern we need to address and updated in following ticket's work. Following ticket's work will update to read this new columns and update code as needed. To address your concerns, we can create working branch similar to what we did with dagster and store these intermediate updates there and apply changes to dev and prod when testing and when ready. Does this plan sound ok?

maxineofficial: Yes, if this isn't merged into dev yet and the changes don't occur against the real prod database, that addresses my concern

tech3371: perfect!

alembic/versions/1892125c2f95_release_version_and_global_release_table.py
tech3371: actually, it's table `id` which can start at 1 but I can update to start at 0 if you like.

maxineofficial: It's fine to start at 1, just wondering

pyproject.toml
```diff
@@ -117,9 +117,11 @@ lint.select = ["B", "D", "E", "F", "I", "N", "S", "W", "PL", "PT", "UP", "RUF"]
...
```
maxineofficial: Can you remove this ignore, if it isn't needed?

maxineofficial:
Why are so many tests getting skipped? What about them is failing?

tech3371:
> Why are so many tests getting skipped? What about them is failing?

@maxinelasp All those test will need to uncommented in future work. If I do it in this PR, I will need to implement mock and so on which I skipped it since those tests will need to be updated and passed in the future.

sds_data_manager/lambda_code/SDSCode/database/models.py
tech3371: since I haven't heard anyone else, I am going to remove that `update_by` column.

maxineofficial (approved):

PR #1421, opened 2026-06-09
ENH: adding support for mission-wide data release reprocessing trigger

### LINKED ISSUE #1412: Update API to take in mission wide data release reprocessing
### Summary

To automate release reprocessing in the future, we are creating a global release table which keeps latest release number. If new release reprocessing is needed, this API will take in new release_number to reprocess. DB update is listened by Dagster and will kickoff release reprocessing jobs when it detects update to DB.

This DB is used by normal pipeline to determine job version too.

### Requirements

The more specific the better.
  - [ ] This new release reprocessing will go to endpoint `/release` and it's target lambda. 
  - [ ] Update lambda handler logic for new `--release-type reprocess`. Relevant code -  https://github.com/IMAP-Science-Operations-Center/sds-data-manager/tree/dev/sds_data_manager/lambda_code/SDSCode/api_lambdas
  - [ ] lambda handler logics should do these. If `release_type == 'reprocess'`, it should expect `release_number` parameter from API call. and if inputs params looks good, update `global_release` table's row with new `release_number` received and `updated_date` to be the date that request came through.
  - [ ] In `/release` lambda handler, API key user is allowed to send request to this API and it makes sure that API key has write access. Follow-up with SDC team about if we should limit this operation to one user or all full access API users to safeguard accident mission wide reprocessing trigger.

### Related tickets

_No response_

tech3371:
# Change Summary
closes #1412 

## Overview

`/release` API takes in mission wide reprocessing and updates global release table. Update science query logic to use new release_number update in science table.

## File changes

- added reprocessing `release-type` handler
- Update science table query logic to query by release number now.
- minor refactor

## Testing

- update tests

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
@@ -238,10 +238,8 @@ def get_latest_ancillary_files(
         ),
     )
 
-    # Combine
-    combined = union_all(with_end_date_query, no_end_date_query).order_by(
-        text("file_path")
-    )
+    # Combine — ORDER BY positional index works across all DB backends for UNION ALL
+    combined = union_all(with_end_date_query, no_end_date_query).order_by(text("1"))
 
     latest_ancillary_files = [row[0] for row in session.execute(combined).fetchall()]
 
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
@@ -106,98 +106,29 @@ def validate_query_params(event):
 
 
 def query_latest_science_files(
-    session, instrument, start_date, end_date, science_files_to_exclude=None
+    session,
+    instrument,
+    start_date,
+    end_date,
+    release_number,
+    science_files_to_exclude=None,
 ):
     """Query for the latest-version science file paths matching given criteria."""
     science_table = models.ScienceFiles
 
-    # Check if any file in the range has a non-null repointing
-    has_repointing = (
-        session.query(science_table)
-        .filter(
-            science_table.instrument == instrument,
-            science_table.start_date >= start_date,
-            science_table.start_date <= end_date,
-            science_table.repointing.isnot(None),
-        )
-        .first()
-        is not None
+    # query for release number for the given instrument and date range
+    latest_science_files = session.query(science_table).filter(
+        science_table.instrument == instrument,
+        science_table.start_date >= start_date,
+        science_table.start_date <= end_date,
+        science_table.release_number == release_number,
     )
 
-    if has_repointing:
-        max_ver_subq = (
-            session.query(
-                science_table.instrument,
-                science_table.data_level,
-                science_table.descriptor,
-                science_table.start_date,
-                science_table.repointing,
-                func.max(science_table.version).label("max_version"),
-            )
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -106,98 +106,29 @@ def validate_query_params(event):
...
-            science_table.repointing.isnot(None),
-        )
-        .first()
-        is not None
+    # query for release number for the given instrument and date range
+    latest_science_files = session.query(science_table).filter(
+        science_table.instrument == instrument,
+        science_table.start_date >= start_date,
+        science_table.start_date <= end_date,
+        science_table.release_number == release_number,
```
tmplummer, line 137: I don't see how this is getting the *latest* science files like the function name and docstring implies. This query will return all files for a given release number. So, if there is a `v001.000` and a `v001.001`, both will get returned.

tmplummer (commented):
Just one question...

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
tech3371: good catch. I should only return rows with `release_number == input_release_number` and `data_version == 0`. I should rename function to `query_release_version()`. What do you think?

tech3371: I updated that function to query to only releaseable science files for given release number. let me know what you think.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -105,99 +105,44 @@ def validate_query_params(event):
...
+        science_table.data_version == 0,
```
tmplummer, line 138: My only concern with requiring `data_version == 0` is that it means the release reprocessing has to make exactly one of each data product. I hope that the dagster refactor helps with this, but for Hi goodtimes, I was seeing an issue where in one reprocessing run, a single goodtimes product could get reprocessed multiple times because of the upstream dependency on current plus 6 nearest repoints. I am not asking for a change here, just something to think about and watch for as we test the release reprocessing.

tmplummer (approved):
Looks good.

sds_data_manager/lambda_code/SDSCode/api_lambdas/release_api.py
```diff
@@ -498,6 +478,68 @@ def unrelease_type_handler(query_params):
...
+        if existing_record:
+            return {
+                "statusCode": 400,
+                "body": json.dumps(
+                    f"Release number {release_number} already exists. "
+                    "Please choose a release number that has not been used before."
+                ),
+            }
+
+        # If new release number is not increment by 1 from the latest release number.
```
lacoak21, line 507: If we are adding this check (which i think is necessary and a good call) do we even need to take in the release number from the cli command?

tech3371: I think so. CLI gives us ability to trigger this anytime. That has been proven to be useful :) since we can make release 'anytime'.

lacoak21: Oh i meant do we need to specifically take in the release number. 

cant we just call 

imap-data-access release reprocess? 
And then increment it by one

PR #1427, opened 2026-06-10
Global release table query 

### LINKED ISSUE #1424: Add a new api endpoint that queries latest major version from table.
imap-data-access webpoda.py needs to be able to query the latest release number for naming l0 files. It should NOT query the db directly because that is handled in sds-data-manager and is out of imap-data-access scope.

lacoak21:
# Change Summary

closes #1424 

## Overview
Read only api query for the latest release number


tech3371 (approved):
looks good!

PR #1428, opened 2026-06-11
Dagster

bryan-harter:
# Change Summary

## Overview
This is... way bigger than should ever be merged. Honestly, I wouldn't even look at it. I'm just making sure that the usual checks pass. 

## File changes
All of them

## Testing
This has been running in prod for a couple weeks. I still really should get some extensive unit and/or integration tests, and that is a separate ticket.


maxineofficial:
You can mark this as a draft if you don't want people to look over it lol

PR #1436, opened 2026-06-15
Stopping jobs from kicking off with no SPICE files.

### LINKED ISSUE #1432: BUG - jobs with missing spice dependenices are kicked off with no spice
### Description of the issue

There are some Ultra processing jobs that are are getting kicked off with no spice dependencies despite having them in the required upstream dependencies.

This function: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/dev/sds_data_manager/orchestration/spice.py#L173
- Returns `None` if the spice files from the metakernel API query don't proved all the required kernels.
- Returns `None` if the metakernel API retruns an error
The outer function: https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/dev/sds_data_manager/orchestration/imap_job.py#L867
- Sees the `None` and just thinks that there are not spice dependencies needed. This causes the batch job to get triggered with no spice files.

Expected behavior would be to not submit these jobs due to missing spice dependencies. Perhaps instead of returning None, the linked function above should raise and error?

```python
Code
```

_No response_

bryan-harter:
# Change Summary

## Overview
Adds a change to have the code raise an error if no SPICE files were found. 

## File changes
Added a check to raise an error if SPICE output was None, and there were SPICE files required. 

## Testing
- Working on testing suite in another ticket

### DIFF
--- a/sds_data_manager/orchestration/imap_job.py
+++ b/sds_data_manager/orchestration/imap_job.py
@@ -781,9 +781,9 @@ class IMAPJobHandler:
         )
         if not spice_files and self.job_config.spice_inputs:
             # If no SPICE files are returned, but there are SPICE inputs, raise failure
-            raise MissingDependenciesError(f"""Missing SPICE files between 
+            raise MissingDependenciesError(f"""Missing SPICE files between
                                            {target_start} and {target_end}""")
-            
+
         return spice_files
 
     def get_science_files_inputs(self, context, target_start, target_end):
--- a/sds_data_manager/orchestration/imap_job.py
+++ b/sds_data_manager/orchestration/imap_job.py
@@ -779,6 +779,11 @@ class IMAPJobHandler:
         spice_files = spice.get_upstream_dependency_inputs_spice(
             self.job_config.spice_types, target_start, target_end
         )
+        if not spice_files and self.job_config.spice_inputs:
+            # If no SPICE files are returned, but there are SPICE inputs, raise failure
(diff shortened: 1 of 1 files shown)


sds_data_manager/orchestration/imap_job.py
```diff
@@ -779,6 +779,11 @@ def get_spice_file_inputs(self, session, target_start, target_end):
         spice_files = spice.get_upstream_dependency_inputs_spice(
             self.job_config.spice_types, target_start, target_end
         )
+        if not spice_files and self.job_config.spice_inputs:
+            # If no SPICE files are returned, but there are SPICE inputs, raise failure
+            raise MissingDependenciesError(f"""Missing SPICE files between
```
tmplummer, line 784: The `get_upstream_dependency_inputs_spice` function will return None if either:
- The files returned by the metakernel query doesn't provide all of the required kernels
- The metakernel API returns error (`metakernel_response["statusCode"] != 200`).

I'm wondering if the `raise` should be done in that function so that the error message can distinguish between those two different causes.

bryan-harter: Yeah that would probably be a better choice in the long run, and we can certainly add it in now if you think it is worth it! 

My thinking was, I *think* we're going to revamp the SPICE handling code at some point hopefully soon so we can better handle things like ensuring full coverage (and ignoring the gaps caused by repointing) and whatnot, so maybe we might as well wait and add in better error descriptions when that code is implemented.

tmplummer (approved):

PR #1438, opened 2026-06-16
Dagster check for running upstream jobs

### LINKED ISSUE #1410: Dagster check for upstream jobs running.
1. First step is to determine whether we need a CRID replacement in dagster. Do we need to check if any upstream jobs (not just direct upstream) are running before triggering a job or will the time buffers in the sensors solve this. 

2. if needed, expand on Bryans commented out code that checks if upstream jobs are running to include the whole upstream tree.

lacoak21:
# Change Summary

closes #1410
## Overview
Uncomment Bryans code that checks for upstream jobs. Make a slight update to the function so that it checks all upstream jobs not just parent jobs. 

## File changes
  - uncomment code, remove todo, and update logic to check all upstream jobs. 

## Testing
Add a small test for this function. This should be merged into the larger test suite once we have that but I wanted to write one to help me check my logic.


bryan-harter (approved):
Nice, looks good! Glad there was a built-in function for finding all upstream files!

bryan-harter:
Still looks good, and thanks for adding to the unit tests!

PR #1442, opened 2026-06-17
Fix Dagster backfill run id collisions

### LINKED ISSUE #1441: Dagster reprocessing sensor run collisions
Hi SDC,

We attempted to reprocess a set of 40 files (4 CoDICE Lo 3d distribution descriptors x 10 days) about an hour ago and have seen only 6 of them run in the Processing Jobs UI. In the AWS console, it looks like the other 34 requests failed to be handled from the main ReprocessQueue and were dropped into a dead-letter queue "reprocessDQL".

Is this expected behavior? Will these remaining requests be automatically retried eventually, or should we resubmit them manually?

Nicholas and Peter

This was an email sent from menlo. I investigated it and turns out there were run ID collisions happening within the reprocessing sensor.

lacoak21:
# Change Summary

closes #1441

## Overview
The reprocessing sensor was stamping backfill run IDs with the date. I think this was causing run collisions if you were to submit a bunch of requests at once the example in the ticket. 

I also bumped up the max retries for the DQL to 2. 

Another change was to process each message separately and if one errors out, log a message and continue instead of causing every other message to fail as well.

### DIFF
--- a/sds_data_manager/constructs/instrument_lambdas.py
+++ b/sds_data_manager/constructs/instrument_lambdas.py
@@ -88,9 +88,10 @@ class ReprocessingTools(Construct):
             fifo=True,
             # Removes messages with identical content.
             content_based_deduplication=True,
-            # The dead letter queue will take messages that failed retry.
+            # The dead letter queue will take messages that failed retry. Allow 2
+            # retries before the message gets sent to the DQL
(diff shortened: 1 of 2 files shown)


bryan-harter (approved):

pleasant-menlo:
Looks reasonable, thanks!

PR #1446, opened 2026-06-18
Adding major version to dep config yaml files

tech3371:
# Change Summary


tech3371:
Luisa, I am merging this to unblock the upcoming work as we discussed.

PR #1450, opened 2026-06-18
Unit Tests for Dagster

### LINKED ISSUE #1417: Unit Testing Dagster
### Summary

We should create unit tests for the current processing infrastructure. Right now, it is unclear to me what the best method for unit testing Dagster is. 

There are many functions we can (and probably have) performed unit tests for for determining spin, ancillary, spice, and repoint dependencies. These should now point to the appropriate functions within the new "orchestration" folder. 

However, determining which science files to use now comes from materializations in Dagster. How would we test out the full dependency pipeline here? 

### Requirements

The more specific the better.
  - [ A complete unit test suite that hits the most important pieces of the new "orchestration" folder]

### Related tickets

_No response_

bryan-harter:
# Change Summary

## Overview
This adds a single (rather large) test to pytest, ensuring that a dagster sensor will appropriately pick up a new materialization of a glows_l0_raw file, kick off processing, and then eventually materialize the L1a files. 

Note that this is not a complete list of things that need to be unit tested. For example, the hi-goodtimes logic still remains untouched. However, this can be implemented ASAP, which should help ensure nothing breaks when we re-implement the spacecraft jobs. 

## File changes
The two primary file changes are

tests/orchestration/conftest.py
- Set up the environment by overriding AWS commands and the RDS

tests/orchestration/conftest.py 
- Runs an "end to end" test of a L0 file arriving, and Dagster taking things through to materializing the L1a assets that are generated.  

But these files were also touched

tests/orchestration/imap_job.py
- These *look* like big changes, but really all I did was encapsulate almost everything in "build_asset()" into another function, which makes it easier for testing purposes. 

pyproject.toml
- I added a dependency to add "testcontainers" for pytest. This ensures that we are testing with a similar postgresql instance like we are on production. 

## Testing
- This is all tests


sds_data_manager/orchestration/imap_job.py
```diff
@@ -128,7 +128,10 @@ def __init__(self, job: ProcessingJobNode):
         job : ProcessingJobNode
             The job node to process.
         """
-        self.BATCH_JOB_TIMEOUT_SECONDS = 3600  # 1 hour, can be adjusted as needed
+        self.BATCH_JOB_TIMEOUT_SECONDS = 3600  # 1 hour, can be adjusted as needed.
+        self.WAIT_TIME_AFTER_BATCH = (
```
tmplummer, line 132: Is this the wait time in seconds?

tmplummer (commented):
From a very quick review, this looks reasonable. Are we tracking any tickets to bulk up the unit testing for all this code?

sds_data_manager/orchestration/imap_job.py
bryan-harter: Yup! Yeah good call, it should also say "seconds" in the name.

bryan-harter:
Yep I will make a couple of tickets! 

I'm working on the spacecraft l1a job now, and that'll also include tests which will hit more areas. 

The main thing I can think that should really be tested after this is the hi-goodtimes code, and that code is something we should revisit as a whole. I just did a lot of copy+pasting of the code that was in dependency.py (so I knew it was exactly the same), without a lot of refactoring to rely on things that Dagster might know. So revisiting the hi-goodtimes needs a ticket in general. 

And then the other one is that Tenzin had made a ton of awesome unit tests for calculating the dependencies, but I just had to comment out for now. But I would like them re-worked! So I'm planning on making a ticket for that.

PR #1451, opened 2026-06-18
Update query API to handle new major/minor versions, while preserving…

### LINKED ISSUE #1413: Update query API to take in major and minor version
### Summary

imap-data-access query --release-number

If nothing provided, then default to current release number.

### Requirements

The more specific the better.
  - [ ] Update query lambda handler to take in release_number parameter now
  - [ ] If nothing provided, then default to release number from global_release table.

### Related tickets

_No response_

maxineofficial:
## Summary

Adds major-version querying to the science query API and moves "latest" resolution server-side, following the science version-scheme rework that split `version` into integer `major_version` / `minor_version` columns.

Science queries now default to the latest major version when no major version is specified — a behavior change from the previous "return all versions unless filtered" default.

## Changes

- **`version` backwards-compatibility alias**: the science `version` query param is still accepted and maps to the new columns. `v001` -> `minor_version` (deprecated minor-only form); `v001.0002` -> `major_version` + `minor_version` (full `vMMM.mmmm` form). Malformed values return 400.
- **Server-side "latest"** (science only):
  - No `major_version` supplied (the new default) -> latest major version with  ALL of its minor versions, per file series.
  - `latest=true` -> the single newest file (latest major, then latest minor) per series, mirroring the existing `/spice-query` `latest=true` convention.
  - A concrete `major_version` disables the latest restriction and filters to that major. Precedence: concrete major_version > latest=true > default.
- **Released-aware latest**: for unauthenticated requests, the max-version computation runs inside the correlated subqueries with a `released` filter, so an unreleased newer version cannot hide the latest *released* version.
- A "file series" is grouped by instrument / data_level / descriptor / start_date / repointing (NULL-safe on repointing).
- Other tables (ancillary, spice) and the quicklook table are unchanged.
- Refactored the handler into helpers (`_parse_version_alias`, `_resolve_science_version_mode`, `_latest_version_filters`, `_format_search_results`) to keep it readable and within lint limits.

## Testing

- `tests/lambda_endpoints/test_query_api.py`: un-skipped, fixtures updated to the new integer version columns, and new cases added for the latest-major default, `latest=true`, concrete major, the `version` alias (both forms), unreleased-version exclusion for unauthenticated users, legacy `vXXX` filenames, and invalid-version 400s.

## Notes
- Response keys are now `major_version` / `minor_version` (no combined `version` key). 
- Depends on the DB migration that introduced `major_version` / `minor_version` (already on the branch).

🤖 Co-authored with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/query_api.py
@@ -5,7 +5,7 @@ import json
 import logging
 from collections import namedtuple
 
-from sqlalchemy import func, select
+from sqlalchemy import and_, func, or_, select
 
 from ..api_lambdas.utils import is_authenticated_user
 from ..database import database as db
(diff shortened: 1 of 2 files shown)


lacoak21 (approved):
LGTM

tech3371 (approved):
Looks good to me too. Really thought out work. Thanks Maxine!

We will test it with other components and will update if we find anything.

PR #1457, opened 2026-06-23
Api updates

### LINKED ISSUE #1412: Update API to take in mission wide data release reprocessing

### LINKED ISSUE #1423: Update all API to use new update in science file updates
Update upload or download or query to use
* update to use new version science file validator object from `imap-data-access`
* update DB query to use new science table

tech3371:
# Change Summary
Closes #1412

## Overview

Updated API and pipeline lambdas to use new major and minor of science files. Query API updates are done by Maxine is other PR.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/idex_l0_indexer.py
+++ b/sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/idex_l0_indexer.py
@@ -263,7 +263,8 @@ def lambda_handler(event, context):
     )
     idex_params = {
         "ingestion_date": params["ingestion_date"],
-        "version": params["version"],
+        "major_version": params["major_version"],
+        "minor_version": params["minor_version"],
         "file_path": params["file_path"],
     }
     logger.info(f"Writing metadata to idex-l0-files db table: {idex_params}")
--- a/tests/lambda_endpoints/conftest.py
+++ b/tests/lambda_endpoints/conftest.py
@@ -72,7 +72,7 @@ def ancillary_file():
 @pytest.fixture(scope="module")
 def science_file():
     """Path to a valid science file."""
-    return "imap/swe/l1a/2010/01/imap_swe_l1a_test-description_20100101_v000.0001.cdf"
+    return "imap/swe/l1a/2010/01/imap_swe_l1a_test-description_20100101_v001.0001.cdf"
 
 
 @pytest.fixture(scope="module")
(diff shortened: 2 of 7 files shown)


sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/utils.py
```diff
@@ -0,0 +1,38 @@
...
+        Modified params dictionary with "version" removed and
+        "major_version" and "minor_version" converted to integers.
+
+    Examples
+    --------
+    >>> params = {
+    ...     "instrument": "hi",
+    ...     "data_level": "l1a",
+    ...     "version": "v001.0000",
+    ...     "major_version": "001",
```
lacoak21: Why are these strings here?

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/utils.py
```diff
@@ -0,0 +1,38 @@
+"""Utility functions for pipeline lambdas."""
+
+
+def extract_major_minor_version(params: dict) -> dict:
```
lacoak21: With this [PR done](https://github.com/IMAP-Science-Operations-Center/imap-data-access/pull/332), I think we can remove this function entirely, major_version and minor_version are already integers.

lacoak21 (approved):
LGTM just had a few comments.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/utils.py
tech3371: totally! I will wait for that PR and pre-release and remove this function.

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/utils.py
tech3371: I think our file params returns them as string. If they don't, then I got this wrong.

PR #1458, opened 2026-06-23
Modifying how pointing_attitude is kicked off

bryan-harter:
# Change Summary

## Overview
Pointing attitude was kicked off daily with the most recent changes. However, it makes more sense to kick it off for every new repoint. 

In addition, I made a change to the sensor so that things are kicked off on every new repoint partition that has been added. I think this makes a little more sense than triggering on a new repoint file itself, since it ensures that we've had time for the actual partition to get created. 

## File changes
sds_data_manager/orchestration/custom_behavior/spacecraft.py
- The sensor will trigger on new repoint partitions. The previous behavior was that it would trigger every day. 

sds_data_manager/orchestration/dependencies/imap_spacecraft_dependencies.yaml
- Set to be repoint partitions instead of daily. This is only partially accurate, since technically it creates a file that could span many repoints. However, we don't currently keep track of the output of this job in Dagster, and "repoint" reflects the frequency that we want to kick off this job. Tim's PR is working on adding a new custom partition to handle the output, if we decide we want to keep track of the output.  

## Testing
tests/orchestration/test_spacecraft.py
- Change to reflect the new sensor output. We'd get a run for each repoint, rather than each day, and the test only adds 10 repoints.

### DIFF
--- a/sds_data_manager/orchestration/dependencies/imap_spacecraft_dependencies.yaml
+++ b/sds_data_manager/orchestration/dependencies/imap_spacecraft_dependencies.yaml
@@ -5,7 +5,7 @@
 # and how to properly configure them for your pipeline lambdas.
 
 (l1a, pointing-attitude):
-  partition: daily
+  partition: repoint
   inputs:
     - source: repoint
       data_type: repoint
--- a/tests/orchestration/test_spacecraft.py
+++ b/tests/orchestration/test_spacecraft.py
@@ -70,7 +70,7 @@ def test_spacecraft_l1a_sensor(mock_db_session, ephemeral_instance):
     run_requests = list(sensor_result)
 
     # Verify things were kicked off
-    assert len(run_requests) > 270, "Expected numerous kickoffs to be attempted."
+    assert len(run_requests) == 10, "Expected a run for each repoint partition."
 
 
 def test_spacecraft_l1a_no_repoint(
--- a/sds_data_manager/orchestration/custom_behavior/spacecraft.py
+++ b/sds_data_manager/orchestration/custom_behavior/spacecraft.py
@@ -1,6 +1,17 @@
 """Override behavior for Spacecraft processing."""
 
-from sds_data_manager.orchestration import imap_job
+import datetime
+import json
+
+from dagster import (
+    RunRequest,
+    SensorEvaluationContext,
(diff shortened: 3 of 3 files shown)


lacoak21 (approved):
LGTM!

tests/orchestration/test_spacecraft.py
```diff
@@ -70,7 +70,7 @@ def test_spacecraft_l1a_sensor(mock_db_session, ephemeral_instance):
     run_requests = list(sensor_result)
 
     # Verify things were kicked off
-    assert len(run_requests) > 270, "Expected numerous kickoffs to be attempted."
+    assert len(run_requests) == 10, "Expected a run for each repoint partition."
```
tech3371, line 73: Nice!

tech3371 (approved):
LGTM! nice

sds_data_manager/orchestration/dependencies/imap_spacecraft_dependencies.yaml
```diff
@@ -5,7 +5,7 @@
 # and how to properly configure them for your pipeline lambdas.
 
 (l1a, pointing-attitude):
-  partition: daily
+  partition: repoint
```
tmplummer, line 8: Question:
Since this is set to a repoint partition, if we were to reprocess for some reason (like an updated attitude history kernel), will this trigger a job for each repoint? That is not the behavior that we want. Ultimately, we would want to run only a single job for the time covered by the updated ah kernel. That is why in the WIP PR I put up, I went to the lengths of creating a custom partition.

bryan-harter: Yes that makes complete sense! We'll still need that PR, I just put something in place to get things up and running last week. 

If we did need to reprocess the pointing_attitude in the current state of the system, I *think* we'd just need to do something like set the start_date and end_date equal only to the final date of 90 day window. Then it would only trigger the creation of that full 90 day file instead of all other repoints. 

Right now if we set the start_date to the actual start date of the file we wish to generate, that would be no bueno and it would make 90 files. So your changes are definitely much better. I thought about merging them into this PR, but I didn't *quite* follow the logic of the partition stuff and I didn't want to break anything.

Issue #1414, opened 2026-06-05
Downtime and transition science file version from old to new


tech3371:
### Summary

Come up with plan on how to transition existing record db or s3 to use new science version. 

### Requirements

The more specific the better.
  - [x] DB transition will be done through alembic migration script work in #1399 
  - [x] How to rename s3 files to use new version. Write lambda script to rename all existing s3 files to `v001.mmmm` where `mmmm` is current version all files have. Update all files to have `v001.mmmm`

### Process for second round of file renaming:

- [x] Validate files for renaming (@vineetbansal  and @lacoak21):
   - [x] Restore RDS backup to new RDS
   - [x] Query for all rows that have old versioning convention
   - [x] Compare to entries in `deprecated-data-archive` bucket
      - [x] Need to move deprecated IDEX files that already existed. Descriptor for these files were:
         - l1a: Sci-1week, evt, catlst, msg
         - L1b: Sci-1week, evt, catlst, msg
         - l2a: Sci-1week
         - These were moved to s3://deprecated-data-archive/deprecated_idex_products/
- [x] Run rename script (@vineetbansal & @bryan-harter)
- [x] Validating output from rename script:
   - [x] Check that for every source file, we have exactly one destination file (@bryan-harter )
   - [x] Verify that each destination file can be opened (@bryan-harter )
   - [x] Spot check a subset by doing some sort of diff of the source file and destination file (@bryan-harter & @tmplummer)
- [x] Dry run DB sync pointing at the new bucket to verify that no new files exist (@tech3371  & @bryan-harter)
   - [ ] If there are any new files, we need to turn off Dagster triggering before moving files
   - [ ] Summary from DB sync script - 
      {
         "db_files": 311967,
         "temp_bucket_files": 297973,
         "missing_from_temp": 14013,
         "found_in_prod": 14013,
         "missing_everywhere": 0
      }

- [ ] Move new files from their location to the production bucket making sure to not include the deprecated IDEX files (@tech3371  & @bryan-harter)

bryan-harter:
Sounds like renaming the files is the best way to go, but it will require careful timing with the imap-data-access library to ensure things don't stay broken for too long

vineetbansal:
[...]

PR #1461, opened 2026-06-25
S3 transition code foundation

tech3371:
# Change Summary
Related to #1414

### DIFF
--- a/sds_data_manager/constructs/instrument_lambdas.py
+++ b/sds_data_manager/constructs/instrument_lambdas.py
@@ -2,13 +2,19 @@
 
 import datetime
 
-from aws_cdk import Duration
+from aws_cdk import Duration, Environment
 from aws_cdk import aws_ec2 as ec2
+from aws_cdk import (
+    aws_iam as iam,
+)
 from aws_cdk import aws_lambda as lambda_
+from aws_cdk import aws_s3 as s3
+from aws_cdk import aws_secretsmanager as secrets
 from aws_cdk import aws_sqs as sqs
 from constructs import Construct
 
 from sds_data_manager.constructs.api_gateway_construct import ApiGateway
+from sds_data_manager.constructs.database_construct import SdpDatabase
 
 
 class ReprocessingTools(Construct):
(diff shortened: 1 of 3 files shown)


sds_data_manager/lambda_code/SDSCode/database/s3_file_transition.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    # First take on steps of transitioning S3 files to new version. TBD
+    # 1. Able to connect to s3 bucket and list files.
+    # 2. Able to download files from s3 bucket, either using s3 API or imap-data-access.
+    # 3. Able to rename files to new version and validate the new version
+    #    using imap-data-access.
+    # 4. Store in tmp location. TBD on details
+    # 5. Able to make query to DB and science table.
+    # 6. Verify the new version files record matches with what's in science table.
+    # 7. Once verified, switch to the new version files.
```
tech3371, line 25: step 8, rename DB column `file_path`'s version to new version. I noticed that DB migration script didn't do that. Should I do that through DB Alembic script? 

Image

sds_data_manager/constructs/instrument_lambdas.py
```diff
@@ -236,3 +242,96 @@ def calculate_next_run(first_job, today, interval_minutes):
             minutes=(events_passed + 1) * interval_minutes
         )
         return next_run
+
```
lacoak21, line 245: What if this was just a script in a /utils folder or something. I only hesitate because we getting close to our 500 cloud formation resource max (I think we are around 430 or something). It will only be used once correct?

sds_data_manager/lambda_code/SDSCode/database/s3_file_transition.py
```diff
@@ -0,0 +1,26 @@
...
+    Lambda code defined here will help with making sure
+    dev and prod has same setup when transitioning.
+    """
+    # First take on steps of transitioning S3 files to new version. TBD
+    # 1. Able to connect to s3 bucket and list files.
+    # 2. Able to download files from s3 bucket, either using s3 API or imap-data-access.
+    # 3. Able to rename files to new version and validate the new version
+    #    using imap-data-access.
+    # 4. Store in tmp location. TBD on details
+    # 5. Able to make query to DB and science table.
```
lacoak21, line 23: Can you add why we need to query? Is it to check if the file already exists with the new version?

sds_data_manager/lambda_code/SDSCode/database/s3_file_transition.py
lacoak21: Oh good catch. Hmm I vote we add it to the lambda or the script. I think using alembic for this may be overkill because we are not modifying anything about the schema (models.py) of the database. It probably is easier to just use sqalchemy to do this in the python script.

lacoak21 (approved):
Looks good so far! are you going to finish the lambda in this PR?

tech3371:
> Looks good so far! are you going to finish the lambda in this PR?

I am only creating lambda and adding pseudocode notes for Vineet and us to think about implementation details. I wanted this PR to setup Lambda that Vineet can try with.

Lambda implementation will come later.

sds_data_manager/constructs/instrument_lambdas.py
tech3371: I plan to take this lambda out after transition is done. Does that help?

It's nice to have lambda that we can run through console with same setting as in dev and prod and use just for this transition task since it's pretty involved.

sds_data_manager/lambda_code/SDSCode/database/s3_file_transition.py
tech3371: This is details that I don't know about yet and would be helpful to brainstorm ideas. I added DB query because during this transition, I wanted to make sure that we still stay synced between DB and new s3 files with new names. I don't know how to do that yet but I knew that we need to make query to DB at least. That's why I added that at least for now.

tech3371:
I am going to merge this and bring some of discussion to slack next week.

PR #1463, opened 2026-06-25
Add major version and output products to dependency hash

### LINKED ISSUE #1462: Ability to reprocess jobs with new version json
Right now our duplicate job checks only the dependencies. 

we need to be able to update a dependency file major version and reprocess a job. 

e.g. if we bumped a major version output in this file 

sds_data_manager/orchestration/dependencies/imap_mag_dependencies.yaml we need to be able to reprocess even if the dependencies are the same.

lacoak21:
# Change Summary

closes #1462 
## Overview

We need to add the output products and major versions to the dependency hash. This will give us the ability to reprocess files when the dependencies have not changed but we have bumped a major version number. The minor version should NOT be included in the dependency hash because the minor version should only get updated when we have a dependency update, a major version bump, or a code change. The major can change when there hasnt been a code change or a dependency update.

This PR also adds some small fixes I found during testing today. 

## File changes
- sds_data_manager/orchestration/custom_behavior/idex.py
   - in the idex.py code, we were deduping files to only materialize the latest ones. I updated this code to use the new version format
- sds_data_manager/orchestration/dependencies/imap_mag_dependencies.yaml
   - During one of my first tests, the mag l1a job failed because the version dict was missing the output for burst-raw. ***NOTE*** we should double check that all outputs are being accounted for because it will error out if they are not in the dependency yaml. This is new behavior after Vineets PR. 
      - dependency hash updates.

### DIFF
--- a/sds_data_manager/orchestration/dependencies/imap_mag_dependencies.yaml
+++ b/sds_data_manager/orchestration/dependencies/imap_mag_dependencies.yaml
@@ -36,6 +36,11 @@
       data_type: l1a
       descriptor: burst-mago
       major_version: 1
+    - source: mag
+      data_type: l1a
+      descriptor: burst-raw
+      major_version: 1
+
 
 (l1b, burst-magi):
   partition: daily
--- a/sds_data_manager/orchestration/imap_job.py
+++ b/sds_data_manager/orchestration/imap_job.py
@@ -1070,7 +1070,7 @@ class IMAPJobHandler:
         dependency_strings.append(self._get_container_image_digest())
         # Append a string of each output descriptor and its major version, sorted
         # alphabetically by descriptor
-        # e.g. 'burst-magi:1,burst-mago:1,norm-magi:1,norm-mago:1'
+        # e.g. 'burst-magi:1,burst-mago:1,burst-raw:1,norm-magi:1,norm-mago:1'
         version_string = ",".join(
             sorted(
                 [
--- a/sds_data_manager/orchestration/imap_job.py
+++ b/sds_data_manager/orchestration/imap_job.py
@@ -1213,7 +1213,7 @@ class IMAPJobHandler:
         # Add the dependency hash and version hash to the dependency file name to ensure
         # uniqueness. The dependency hash is based on the upstream dependencies and the
         # container image digest. The version hash is based on the output versions.
-        dep_hash = self._dependency_hash(serialized_dependencies)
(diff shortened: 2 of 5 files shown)


bryan-harter (approved):
Looks good to me!

sds_data_manager/orchestration/dependencies/imap_mag_dependencies.yaml
```diff
@@ -36,6 +36,11 @@
       data_type: l1a
       descriptor: burst-mago
       major_version: 1
+    - source: mag
```
tech3371, line 39: thanks for catching it! I am guess you caught that when you didn't see the file produced when matching with expected output list?

sds_data_manager/orchestration/imap_job.py
```diff
@@ -1031,21 +1031,29 @@ def filter_conditions(table):
...
+    def _dependency_hash(
+        self, serialized_dependencies: str, output_versions: dict[str, dict[str, int]]
+    ) -> str:
         """Generate a hash for the serialized dependencies.
 
         This is a unique ID for a particular run. Dagster will refuse to run a job with
-        the same dependency_hash.
+        the same dependency_hash. It is created from the upstream dependencies,
+        the container image hash, and the output products and their major version
+        numbers.
```
tech3371:
```suggestion
        the container image hash, and the output products and their major and
        minor version numbers.
```

tech3371: On second note, ignore my code suggestion. I found out why you updated and your PR description very helpful. Can you add that in this doc as reason to not add minor to hash? I think it will help us in the future for reference and understanding.
`
We need to add the output products and major versions to the dependency hash. This will give us the ability to reprocess files when the dependencies have not changed but we have bumped a major version number. The minor version should NOT be included in the dependency hash because the minor version should only get updated when we have a dependency update, a major version bump, or a code change. The major can change when there hasnt been a code change or a dependency update.
`

tech3371 (approved):
LGTM

sds_data_manager/orchestration/dependencies/imap_mag_dependencies.yaml
lacoak21: No - processing actually errored out. I added this as a note in the PR descriptor because we need to double check that all job outputs are accounted for.

PR #1466, opened 2026-06-30
cadence sensor

tech3371:
# Change Summary

### DIFF
--- a/sds_data_manager/orchestration/custom_partitions.py
+++ b/sds_data_manager/orchestration/custom_partitions.py
@@ -262,7 +262,7 @@ def add_idex_30_day_partitions(context: SensorEvaluationContext):
 # Run daily (24 hours = 86400 seconds)
 # TODO: update to run daily or weekly or at specified time
 # based on progressive discussion in the future.
-@sensor(minimum_interval_seconds=86400)
+@sensor(minimum_interval_seconds=86)
 def add_cadence_map_partitions(context: SensorEvaluationContext):
     """Create missing cadence partitions daily.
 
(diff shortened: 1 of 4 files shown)


sds_data_manager/orchestration/custom_partitions.py
```diff
@@ -242,9 +259,60 @@ def add_idex_30_day_partitions(context: SensorEvaluationContext):
...
+def add_cadence_map_partitions(context: SensorEvaluationContext):
+    """Create missing cadence partitions daily.
+
+    This sensor checks for new cadence partitions that need to be created
+    and adds them to Dagster.
+    """
+    added_any = False
+
+    # Calculate the currently active windows for all cadences.
+    current_time = datetime.datetime.now(datetime.timezone.utc)
```
lacoak21, line 275: @tech3371 I updated this function call to 

1. take in the datetime.now instead of the cadence string 
2. Call this only once outside of the loop because it will return current windows for all candences

sds_data_manager/orchestration/custom_partitions.py
```diff
@@ -242,9 +259,60 @@ def add_idex_30_day_partitions(context: SensorEvaluationContext):
...
+    # or manual triggers to run jobs on these partitions.
+    return SensorResult()
+
+
 sensors = [
     add_repoint_partitions,
     add_daily_partitions,
     add_idex_10_day_partitions,
     add_idex_30_day_partitions,
+    add_cadence_map_partitions,
```
lacoak21, line 317: This was the key missing part that actually lets dagster know to create this sensor in our project.

PR #1477, opened 2026-07-08
Fix Hi l1b goodtimes custom dependency handling

### LINKED ISSUE #1473: BUG - Hi goodtimes jobs are failing and not getting retriggered
### Description of the issue

Looking at the Hi Goodtimes jobs, the batch job errors out with this traceback:
```
  File "/usr/local/bin/imap_cli", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/lib/python3.12/site-packages/imap_processing/cli.py", line 1849, in main
    instrument.process()
  File "/usr/local/lib/python3.12/site-packages/imap_processing/cli.py", line 453, in process
    products = self.do_processing(dependencies)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/imap_processing/cli.py", line 865, in do_processing
    datasets = hi_goodtimes.hi_goodtimes(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/imap_processing/hi/hi_goodtimes.py", line 131, in hi_goodtimes
    raise ValueError(
ValueError: Goodtimes cannot yet be processed for repoint00301: repoint00305 has not yet been completed according to the repoint table.
```

In the Hi code, I check that the goodtimes product is not trying to get generated before the future DE data needed for the statistical checks has had a chance to be downlinked and processed. I do this by making sure that the repoint table has N+4 repoint in it. So, this seems like a bug in triggering of the goodtimes job. I previously had it set up to trigger across pointing (e.g. l1b de for pointing N would trigger potential jobs for the range [N-4:n+4]).

_No response_

_No response_

```python
Code
```

_No response_

tmplummer:
Fix Hi L1B Goodtimes jobs kicking off too early / not retriggering after failure

Problem

Hi Goodtimes needs L1B DE data from neighboring repoints (not just its own) to compute quality flags, but the custom HiGoodtimesJob handler meant to enforce that was never actually running. A typo in its JobBuilderRegistry registration keys (45-sensorgoodtimes/90-sensorgoodtimes instead of 45sensor-goodtimes/90sensor-goodtimes) meant the lookup never matched, so Goodtimes silently fell back to the generic IMAPJobHandler, which only requires the current repoint's own DE file.

This caused two symptoms:
- Goodtimes jobs kicked off as soon as their own repoint's DE existed, before neighboring/future DE data had been downlinked and processed.
- If the resulting job failed, the duplicate-detection dependency_hash never changed (it only reflected the current repoint's own files), so the job was never retried even once the missing neighbor data arrived.

Changes

sds_data_manager/orchestration/custom_behavior/hi.py:
- Fixed the JobBuilderRegistry.register keys so HiGoodtimesJob actually activates.
- Fixed a bug where a neighboring repoint's INPROGRESS DE job was supposed to raise MissingDependenciesError and wait, but the None check was on the wrong variable (metadata_list, which is never None) instead of repoint_list. Passing None into get_all_files_by_repoint_numbers crashed with a TypeError instead of skipping.
- Moved the future-pointing/PointingTable check inside the with db.Session() block instead of running it after the session had already closed.
- Clarified the comment/error message around the future-coverage check: PointingTable records that a pointing's data has been downlinked, not merely scheduled, so its absence means more data may still be coming (wait) while its presence without a matching DE file means Hi simply has no data for that pointing (safe to proceed).

tests/orchestration/test_hi.py (new):
- test_hi_goodtimes_registered: regression test asserting job_handlers actually resolves to HiGoodtimesJob for both goodtimes descriptors — this is the test that would have caught the registry typo.
- test_hi_goodtimes_waits_when_future_pointing_unknown: verifies the job skips/waits when it doesn't have enough future DE neighbors and the pointing at the edge of the window hasn't been downlinked yet.
- test_hi_goodtimes_proceeds_when_no_more_data_expected: verifies the job proceeds (and includes the neighbor DE files in its dependencies) once PointingTable confirms there's no more data to wait for.
- test_hi_goodtimes_waits_for_inprogress_neighbor: regression test for the misplaced None-check bug — verifies an in-progress neighbor causes a graceful MissingDependenciesError, not a crash.

Closes: #1473


sds_data_manager/orchestration/custom_behavior/hi.py
```diff
@@ -12,69 +12,88 @@
...
+                    # at the edge of our window already exists in
+                    # PointingTable - its existence there without a
+                    # corresponding Hi L1B DE means Hi simply has no data for
+                    # it, not that we haven't waited long enough, so our
+                    # nearest set won't change no matter how long we wait.
+                    # But if that pointing doesn't exist yet, the picture is
+                    # still incomplete: more pointings - and likely more Hi
+                    # L1B DE files - will show up later and change our
+                    # nearest set. Hold off and let the sensor retrigger this
+                    # job once that data has filled in.
```
tech3371, line 62: Thank you for these detail notes!

tech3371 (approved):
LGTM

tmplummer:
pre-commit.ci autofix

PR #1491, opened 2026-07-15
Adding 'version' back to query result.

tech3371:
# Change Summary

## Overview

On the website, I noticed that 'Version' column is showing up as empty because query result doesn't include 'version'. It only has major and minor version. Adding 'version' back into query return.


lacoak21 (approved):
LGTM

PR #1494, opened 2026-07-15
Update l1a output products

lacoak21:
More updates to output products


tech3371 (approved):
wow...that's a lot more missing. Thank you for going through those lists!

PR #1495, opened 2026-07-15
Increase Query Lambda function timeout from 1 minute to 3 minutes

tech3371:
# Change Summary

## Overview

Increasing query lambda timeout to be 3min in case query takes bit longer at times.


tech3371:
closing this. updating lambda timeout won't help here. API gateway has timeout of 29sec. If we don't return by then, the response won't make it to user in time.

Issue #1493, opened 2026-07-15
Science data querying times out and fails for relatively small queries when not specifying a major version


pleasant-menlo:
Since the recent change to vMMM.mmmm versioning, science data queries have timed out for relatively small numbers of files.

For example:

```imap-data-access query --instrument glows --data-level l3a --start-date 20260501 --end-date 20260701```

times out after 30 seconds and returns an error IMAPDataAccessError('503 Service Unavailable: {"message":"Service Unavailable"}')

If we specify a fixed major version with `--major-version 1`, the query completes in under 1 second and returns 705 results.

Examples of queries that previously would run successfully but are now failing due to timeout:
* imap-data-access query --instrument glows --data-level l3e --descriptor survival-probability-lo --version latest
* imap-data-access query --instrument glows --data-level l3a --descriptor hist --version latest
* imap-data-access query --instrument lo --data-level l1b --descriptor nhk --version latest

hafarooki:
As discussed under #1256, this is probably due to the join in the new version subquery. There is no index on this table, so it is O(N^2). Working on a PR to add an index, that might fix this issue (but not fix #1256)

tech3371:
This has been fixed now. after PR is deployed, I verified that query is working now by running previously failed query. 

@pleasant-menlo can you check if your issue was fixed now.

pleasant-menlo:
We are still seeing timeout errors for all of the queries above.

alastairtree:
I am also seeing timeouts at 30s and slow queries suggesting you do probably need more/better indexes. For example this API call times out and is already quite specific

https://api.imap-mission.com/api-key/query?table=science&instrument=mag&data_level=l1d&descriptor=burst-rtn&ingestion_start_date=20260608&ingestion_end_date=20260716&extension=cdf

I used to do a bit DBA work, and knowing the types of queries you will be getting are almost always going to be instrument + date I would add 2 indexes:
- instrument + start_date + end date, 
- instrument + ingest_date. 

Those two would allow a query optimiser to quickly cut out most of the table's data for the majority of queries for 2 pretty small index datasets.

hafarooki:
Copying this from my post on #1256 . Despite the index being added, the first command is still taking >30s. I'm confused why they would exhibit such difference despite the addition of the index. I'll have to think more carefully about why those subqueries are so slow.

---

[...]

Issue #1256, opened 2026-04-17
BUG - imap-data-access fails for hi


tech3371:
### Description of the issue

When we query for `hi` instrument, we get internal server error. But for other instrument, it works.

### Steps to reproduce the issue

```
IMAP_API_KEY=$IMAP_PROD imap-data-access query --instrument hi
```

### Expected vs Actual behavior

Return list of files

### Code Snippet (If applicable)

```python
Code
```

tech3371:
<img width="1555" height="130" alt="Image" src="https://github.com/user-attachments/assets/7f2ed70b-d073-4122-bb7a-5c8db2c9a520" />

tech3371:
We can't fix this bug because the reason command failed was because AWS API call has 1min or less restriction. 

So, if we call like this, `imap-data-access query --instrument hi`, there are so many files to query and structure before returning it, AWS API returns error before backend can return responses.

Some suggestions to get around it:
* We need to ask for date range or some other combination. Eg.
`imap-data-access query --instrument hi --start-date 20260204`. Once I passed in date range, it worked. 
* Even specifing data level worked too for all date range, `imap-data-access query --instrument hi --data-level l1c`. It may not work as project goes on. So, you may need to call specific level or date range or other combination.

lacoak21:
The ULTRA team found out that they are now getting errors with the following command. 

imap-data-access  query --instrument ultra --data-level l1b

It may be nice to implement pagination or something for this. Or nice error messages.

lacoak21:
Bob also requested a way to query all of the descriptors

bryan-harter:
On EMM, we initially solved this problem by uploading the query result as a JSON file to a bucket, creating a public link to that JSON file, and then having the API return a 304 redirect to that file. Then from the user's perspective, it looks like the API has returned a query larger than 6MB. (And the bucket automatically clears files after 24 hours so we aren't full of random JSON files). 

But recently the limits with lambdas and payloads was enough that we just started using a different API service altogether, and just have a FastAPI server running on Fargate. Probably overkill for this project though lol.

hafarooki:
I was curious about this, since i was having a similar issue yesterday.

First of all, it looks like the error code has changed. In @tech3371 's original post, the error code was 500. But I am getting error code 503:
```bash
>>> imap-data-access query --instrument mag  
IMAPDataAccessError('503 Service Unavailable: {"message":"Service Unavailable"}')
```

[...]

PR #1496, opened 2026-07-15
add index to science the two science file tables

hafarooki:
# Change Summary

## Overview
As pointed out in #1493, the recent versioning change made the queries really slow. See comments in #1256 for how we figured out the problem. It boils down to lacking an index on the table. So in this PR I added an index.

## File changes
- **`models.py`** — Added a composite `Index` on the shared `ScienceFileBase`
  mixin covering `instrument`, `data_level`, `descriptor`, `start_date`,
  `repointing`, `major_version`, and `minor_version`. Because both
  `ScienceFiles` and `QuicklookFiles` inherit from this mixin, `__table_args__`
  is built via `@declared_attr` so each subclass gets its own distinctly-named
  index (`idx_<tablename>_version`) rather than sharing a single Index object
  (which SQLAlchemy rejects). The existing `CheckConstraint`s are preserved.
- **`alembic/versions/00bb63a40cf1_science_files_version_index.py`** — New
  migration creating the two indexes on `science_files` and `quicklook_files`,
  with a matching `downgrade()` that drops them.

**Please review the alembic script in particular, I'm not sure I did that right.**

## Testing
There were no unit tests for `models.py` so I did not modify any tests.

### DIFF
--- a/alembic/versions/00bb63a40cf1_science_files_version_index.py
+++ b/alembic/versions/00bb63a40cf1_science_files_version_index.py
@@ -1,0 +1,47 @@
+"""science_files_version_index.
+
+Revision ID: 00bb63a40cf1
+Revises: e5dee361947f
+Create Date: 2026-07-15 18:05:00.000000
+
+"""
+
+from collections.abc import Sequence
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "00bb63a40cf1"
+down_revision: str | Sequence[str] | None = "e5dee361947f"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+# Speeds up the correlated "latest version" subqueries in query_api.py, which
+# scan for MAX(major_version)/MAX(minor_version) grouped by these columns.
+# See GitHub issue #1256.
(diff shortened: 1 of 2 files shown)


sds_data_manager/lambda_code/SDSCode/database/models.py
```diff
@@ -174,10 +174,24 @@ class ScienceFileBase:
...
+                # separate name for index for each subclass
+                f"idx_{cls.__tablename__}_version",
+                "instrument",
+                "data_level",
+                "descriptor",
+                "start_date",
+                "repointing",
+                "major_version",
+                "minor_version",
+            ),
```
tech3371, line 191: nice find!!

alembic/versions/00bb63a40cf1_science_files_version_index.py
tech3371, line 1: I will try this against dev DB.

tech3371, line 1: I tested in dev and can confirm that this alembic migration script works and new index shows up in science_files table. tested upgrade and downgrade too.

tech3371 (approved):

lacoak21 (approved):

PR #1515, opened 2026-07-21
Fix issue truncating spice coverage start/end date to day boundary

### LINKED ISSUE #1513: BUG - start and end date used for metakernel api are truncated to day boundary
### Description of the issue

## Summary 
get_upstream_dependency_inputs_spice in sds_data_manager/orchestration/spice.py rounds the start and end of the requested SPICE coverage window down to midnight UTC before querying the metakernel API, discarding the hour/minute/second precision that the caller actually provided. For jobs partitioned on repoint boundaries (which are not day-aligned), this means the wrong time window — sometimes a zero-width one — gets sent to spice_metakernel_api, so the SPICE files returned may not actually cover the time range being processed.

##  Impact                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                
  For a repoint-partitioned job (e.g. a Hi L1B DE job keyed on a repoint window), two concrete failure modes:                                                                                                                                                   
                                                                                                                                                                                                                                                                
  1. Repoint window entirely within one calendar day (e.g. 14:32:10 → 21:47:33 on the same date): both start_date and end_date round down to the same midnight. The add_24_hrs fallback (if end_date == start_date) does not trigger, because it compares the   
  original, untruncated datetimes, which are not equal. The result is start_time == end_time — a zero-width query sent to the metakernel API instead of the real ~7 hour window.                                                                                
  2. Repoint window crossing midnight (e.g. 22:00 day N → 03:00 day N+1): end_time rounds down to midnight of day N+1, which is earlier than the true end time (03:00). The query window is truncated short on the end side and can miss SPICE kernel coverage  
  needed for the final hours of the repoint.                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                                
  In both cases, get_spice_file_inputs (imap_job.py:836) may fail to find the correct SPICE files for the run — either raising MissingDependenciesError spuriously, or (worse) silently returning kernels that don't actually cover the full requested time     
  range.

_No response_

_No response_

```python
Code
```

_No response_

tmplummer:
## Fix SPICE metakernel query truncating start/end times to day boundaries                                                                                                                                                                                       
                                                                                                                                                                                                                                                                
  get_upstream_dependency_inputs_spice rounded start_date/end_date to day resolution before querying SPICE kernel coverage, via a %Y%m%d string round-trip. For non-day-aligned windows (e.g. repoints), this could produce a zero-width or truncated query,    
  missing kernel coverage.                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                
  Fix: convert directly via datetime subtraction against a hardcoded J2000 epoch constant (2000-01-01T11:58:55.816 UTC, the true TT epoch), preserving full precision. Avoided spiceypy.datetime2et() on purpose — this function runs in a fresh ECS Fargate task per Dagster run, so furnishing a leapseconds kernel there would add an S3/DB round trip to every job run. The hardcoded constant is accurate to a few seconds, plenty for kernel-coverage matching.                                                      
                                                                                                                                                                                                                                                                
  Tests: tests/orchestration/test_spice.py, covering same-day and cross-midnight precision, accuracy vs. spiceypy.datetime2et, and the existing 24-hour-window behavior.

Closes: #1513


lacoak21 (approved):
LGTM

PR #1526, opened 2026-07-22
1524 bug   fix lo dependency yaml

tmplummer:
Try agian to fix the Lo dependency definition. This time add spice and repoint table.


lacoak21 (approved):
LGTM - I'll redeploy when you merge this.

PR #1527, opened 2026-07-22
BUG: Fix IDEX add_l1a_partition sensor 

### LINKED ISSUE #1522: BUG -IDEX l0 add_10_day_partition sensor closing window
### Description of the issue

add_idex_10_day_partitions and add_idex_30_day_partitions are both no longer working properly. 

Suppose the sensor runs on July 22.
It does this:
start_dt = context.cursor
end_dt = now + 40 days → about Sept 1
On the first run, context.cursor is usually mission start, so it sees a big range and adds partitions.
Then it returns:
cursor = end_dt
So now the cursor becomes Sept 1.
Next day
On July 23, it runs again:
start_dt = cursor → Sept 1
end_dt = now + 40 days → about Sept 2
So the scan window is only 1 day wide.
But IDEX partitions are 10-day windows, so that 1-day slice usually contains:
maybe 0 CSV rows, or
only 1 start date
And the code needs 2 adjacent start dates to build one partition:
2026-07-19 → 2026-07-29
2026-07-29 → 2026-08-08

This means no new partitions are being added. We now have l0 data gaps. 

_No response_

_No response_

```python
Code
```

_No response_

lacoak21:
# Change Summary

closes #1522
## Overview
Fix IDEX 10 day partition sensor. As mentioned in the issue, the window that dagster was checking each day was 1day. This meant that basically no IDEX 10day partitions were being added. Since these windows are static, I decided to just use the mission start as the start date and each day check for new partitions in a 40 day window. The extra time to do this should be negligible and ensure that we always have the right partitions for the current time and near future. 

## File changes
- sds_data_manager/orchestration/custom_partitions.py
- Simplify logic


tests/orchestration/test_custom_partitions.py
```diff
@@ -0,0 +1,52 @@
...
+@patch.object(custom_partitions.datetime, "datetime", _FrozenDatetime)
+def test_add_idex_10_day_partitions():
+    """Check that add_idex_10_day_partitions adds the correct partitions."""
+    with instance_for_test() as instance:
+        instance.add_dynamic_partitions(
+            "idex_10_day_partitions",
+            ["idex10_2025-09-27T00:00:00_to_2025-10-07T00:00:00"],
+        )
+        context = build_sensor_context(instance=instance)
+        sensor_result = custom_partitions.add_idex_10_day_partitions(context)
```
tech3371, line 29: Adding this Sept 27 - Oct 7 is initiating other `new_partitions` in Dynamic pool, it seems. Why is other partitions are added and needed?

tests/orchestration/test_custom_partitions.py
```diff
@@ -0,0 +1,52 @@
...
+            "idex_30_day_partitions",
+            ["idex30_2025-09-27T00:00:00_to_2025-10-07T00:00:00"],
+        )
+        context = build_sensor_context(instance=instance)
+        sensor_result = custom_partitions.add_idex_30_day_partitions(context)
+
+    new_partitions = sensor_result.dynamic_partitions_requests[0].partition_keys
+    assert new_partitions == [
+        "idex30_2025-09-24T00:00:00_to_2025-09-27T00:00:00",
+        "idex30_2025-09-27T00:00:00_to_2025-10-27T00:00:00",
```
tech3371, line 55: why is this now going back in time? Eg. Sept 24 to Sept 27?

tech3371 (commented):
I had couple questions understanding the test and intention behind the change but the code changes itself didn't look big and looks good.

tests/orchestration/test_custom_partitions.py
lacoak21: So this part instance.add_dynamic_partitions(
            "idex_10_day_partitions",
            ["idex10_2025-09-27T00:00:00_to_2025-10-07T00:00:00"],
        ) is to mock what we already have. And this step custom_partitions.add_idex_10_day_partitions(context) is to trigger the sensor to add more partitions.

bryan-harter (approved):
This looks good! If something like this happens again, you should be able to manually change the cursor of the sensor in the dagster console to whatever you want. I just tested it by setting the cursor to nothing, then running "Preview Tick Result". It went ahead and created the mission partitions for you. I did it for the 10 day, if you want to try it out for the 30 day.

PR #1531, opened 2026-07-24
Increase mag job timeout

lacoak21:
# Change Summary

closes #3346
## Overview
Mag l1d jobs were hitting the batch job timeout. The simplest solution is to increase the timeout for all Mag jobs. 

## File changes


tmplummer (approved):
It would be interesting to get a list of nominal run-times for all our different jobs. I suspect that the 3-hour timeout was already tailored to the MAG L1D job but this change looks fine to me.

PR #1532, opened 2026-07-24
Make reconstructed ephemeris optional for SWE L3 and SWAPI L3a

### LINKED ISSUE #1523: BUG - Dependencies missing optional flag
### Description of the issue

When the dependency configuration was translated from CSV to YAML, `SOFT_NO_TRIGGER` dependencies did not get the appropriate `required: false` flag.

Most of these have been handled by now, but a couple remain:

- SWAPI L3a requires ephemeris_reconstructed
- SWE L3 requires ephemeris_reconstructed
  - However, SWE L3 also requires MAG L1d which in turn requires ephemeris_reconstructed, so the direct requirement is not responsible for delaying processing.
- ~~Ultra L1b `{45,90}sensor-priority-{1,2,3,4}-de` used to be `SOFT_NO_TRIGGER` inputs to pointing sets, but are now required~~
  - This change may be intentional; check with @lacoak21 

SWAPI L3a and SWE L3 dependencies on ephemeris_reconstructed should have the `required: false` flag so the jobs can run if reconstructed ephemeris is not available.

pleasant-menlo:
# Change Summary

## Overview
When dependency configuration was translated from CSV into YAML, reconstructed ephemeris was mistakenly made required for SWE L3 and SWAPI L3a. This makes it optional again by adding the flag `required: false`, allowing these products to be triggered with only predicted ephemeris.

Fixes #1523

### DIFF
--- a/sds_data_manager/orchestration/dependencies/imap_swapi_dependencies.yaml
+++ b/sds_data_manager/orchestration/dependencies/imap_swapi_dependencies.yaml
@@ -19,6 +19,7 @@ l3a_spice_common: &l3a_spice_common
   - source: ephemeris_reconstructed
     data_type: spice
     descriptor: historical
+    required: false
     trigger_job: false
   - source: ephemeris_predicted
     data_type: spice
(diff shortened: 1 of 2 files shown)


tmplummer (approved):
🚀

Issue #1521, opened 2026-07-22
BUG - Spectral index maps not triggering


pleasant-menlo:
### Description of the issue

The ENA spectral index "spx" map products are intended to be produced in response to a new ENA intensity file being uploaded. However, they do not seem to be triggering currently.

### Expected vs Actual behavior

Each spectral index job has a single ENA intensity map as input. When any new file with that descriptor is uploaded, we would like the spectral index job to run with the new file as input. Ideally this would happen quickly, just like ordinary daily files, but a delay of a few hours would be acceptable.

For example, see this dependency configuration:
https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/b6ba9287f866b01e693bfd36be3f40589fb901ec/sds_data_manager/orchestration/dependencies/imap_hi_dependencies.yaml#L942-L952
A new file `imap_hi_l3_h45-ena-h-hf-sp-full-hae-4deg-6mo_20260117_v001.0007.cdf` was uploaded at 20260729 20:24:19, which should have triggered a Hi L3 `h45-spx-h-hf-sp-full-hae-4deg-6mo` job. No such job has been seen.

### Product list

(hi, l3, h45-spx-h-hf-sp-anti-hae-4deg-1yr)
(hi, l3, h45-spx-h-hf-sp-anti-hae-6deg-1yr)
(hi, l3, h45-spx-h-hf-sp-full-hae-4deg-6mo)
(hi, l3, h45-spx-h-hf-sp-full-hae-6deg-6mo)
(hi, l3, h45-spx-h-hf-sp-ram-hae-4deg-1yr)
(hi, l3, h45-spx-h-hf-sp-ram-hae-6deg-1yr)
(hi, l3, h90-spx-h-hf-sp-anti-hae-4deg-1yr)
(hi, l3, h90-spx-h-hf-sp-anti-hae-6deg-1yr)
(hi, l3, h90-spx-h-hf-sp-full-hae-4deg-6mo)
(hi, l3, h90-spx-h-hf-sp-full-hae-6deg-6mo)
(hi, l3, h90-spx-h-hf-sp-ram-hae-4deg-1yr)
(hi, l3, h90-spx-h-hf-sp-ram-hae-6deg-1yr)
(hi, l3, hic-spx-h-hf-sp-full-hae-4deg-1yr)
(hi, l3, hic-spx-h-hf-sp-full-hae-6deg-1yr)
[...]

PR #1534, opened 2026-07-24
Make all L3 map jobs use the daily partition

pleasant-menlo:
# Change Summary

## Overview
L3 map jobs fall in two categories:
- Jobs which are triggered every day and analyze the available data to determine if anything should be processed, e.g. Hi `sp-maps`
- Jobs which should be triggered by the upload of a single new file, e.g. spectral index jobs.

The second category are [not triggering as expected](https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/1521). We would like to see if changing their `partition` configuration to `daily` helps.

## File changes
Changes all L3 map jobs to `partition: daily`

## Testing
N/A

### DIFF
--- a/sds_data_manager/orchestration/dependencies/imap_hi_dependencies.yaml
+++ b/sds_data_manager/orchestration/dependencies/imap_hi_dependencies.yaml
@@ -908,16 +908,15 @@ sensor90_l1c_pset: &sensor90_l1c_pset
     - *spice_l2
     - *ancillary_90sensor_l2
     - *sensor90_l1c_pset
-
-# L3 Products - 45sensor
   outputs:
     - source: hi
       data_type: l2
       descriptor: h90-ena-h-sf-nsp-ram-hae-6deg-6mo
       major_version: 1
 
+# L3 Products - 45sensor
 (l3, h45-spx-h-hf-sp-anti-hae-4deg-1yr):
-  partition: 1yr
+  partition: daily
   inputs:
     - source: hi
       data_type: l3
@@ -929,7 +928,7 @@ sensor90_l1c_pset: &sensor90_l1c_pset
       major_version: 1
 
 (l3, h45-spx-h-hf-sp-anti-hae-6deg-1yr):
-  partition: 1yr
+  partition: daily
   inputs:
     - source: hi
       data_type: l3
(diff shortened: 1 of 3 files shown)


tmplummer (commented):
Hmmm, I suspect that this change will make it look like there are many missing products in dagster. Maybe @bryan-harter has a suggestion since Luisa is currently out.

tech3371 (changes requested):
mismatching `1yr` cadence descriptor with `partition` is going to cause some mismatch of logic in dependency lookup and more tech debt in the future. I recommend not doing this update.

`partition` is used to determine date time range to look up dependency. If it's updated to daily, it will only look for one day range of dependency. If your desired outcome is to trigger job daily, a dagster sensor logic can help with that. @bryan-harter or @lacoak21 will be able to help with.

pleasant-menlo:
Thanks Tenzin, that makes sense. We will investigate the issue further.

tech3371:
@pleasant-menlo If you can open a ticket with details, I can try to help with little knowledge I have of dagster. I need to learn and dive more in that part of pipeline anyway :).

Details that would be helpful in that ticket are:
1. which product do you want this behaviour to apply. eg. (hi, l3a, <descriptor). If more than one descriptor, list of descriptor in that format would be helpful.
2. Dependency query logic. will that need to change for this case or not. Eg. I saw it's for `1yr` cadence job. will dependency query still lookup for 1 yr or do you want us to by pass dependency look and kickoff job at daily cadence. **Note** that `partitions` will still look like same as other instrument's cadence partition, eg. `cadence_1yr_{year}-01-17T00:00:00_to_{year+1}-01-17T00:00:00` format. Also, if you want to bypass dependency logic, that can show your job as succeeded on dagster and more implecation that I don't know until I dive into it. These are the initial ones I can think of.
3. at what cadence do you want this unique case to run at. Eg. daily or every x hours or every x weeks etc. This helps us know what frequency to set for this case's sensor.
4. Anything else we should know about.

pleasant-menlo:
We updated the ticket at https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/1521 with some more details, and Bryan is looking into some things. Cheers!

PR #1536, opened 2026-07-24
6mo map job for imap-lo

### LINKED ISSUE #1535: Imap-Lo - 6mo quickmaps in dagster

vineetbansal:
Closes #1535 

We need a placeholder job that can be triggered, even if it does nothing useful just yet. The product added here is exactly the one we're hoping to generate by next week, as per @mitchellshen


tmplummer (approved):
LGTM

PR #1537, opened 2026-07-27
Correct deps for 6mo lo maps

### LINKED ISSUE #1535: Imap-Lo - 6mo quickmaps in dagster

vineetbansal:
Closes #1535 

An earlier PR on this had incorrect dependencies. These will work with the new map algorithm for Lo.
Also moved an errant comment to its correct place in the list.

All other maps' dependencies will likely need to be updated to match this one too, but that's another PR once this one is validated by the Lo team.


tmplummer (approved):
Looks good to me.

PR #1539, opened 2026-07-28
Removed trigger_jobs for pointing_attitude and attitude_history (ie set to True)

### LINKED ISSUE #1476: Turn on kickoff=True for data products depending on the attitude_history and the DPS (pointing_attitude) kernel
### Summary

Right now, job kickoff is false for just about every product that depends on the pointing_attitude and DPS kernels. This is because these kernels are appended to rather than overwritten, stopping after 90 days. If kickoff=True, then we could trigger a file at the beginning of the window up to 90 times, even though the data at the beginning of that file never changed.  

AFTER the ticket https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/596 is complete, we need to turn on the ability for these new kernels to kick off processing once more. 

Additionally, once this ticket is done, we should coordinate with Margaret/MOC/POC to receive brand new pointing_attitude files. These files will essentially kick off a large reprocessing effort, as they will be fixing an issue with the coordinate system used in these files. 

### Requirements

The more specific the better.
  - [ Set kickoff=True in the YAML files where data products use these SPICE kernels as input ]
  - [ Once implemented, coordinate with Margaret/ops to handle the delivery of the new pointing_attitude files ]

### Related tickets

_No response_

daralynnrhode:
# Updated products that depend on the pointing_attitude and DPS kernels to be True 

## Overview

Made changes to the yaml files to remove all `trigger_jobs: false` found under `pointing_attitude` and `attitude_history` depend-ees.
closes #1476

The ticket also mentions coordinating with Margaret and the MOC/POC to get brand new pointing_attitude files, so just putting that here so it doesn't slip through. 

## File changes

- sds_data_manager/orchestration/dependencies/imap_glows_dependencies.yaml
- sds_data_manager/orchestration/dependencies/imap_hi_dependencies.yaml
- sds_data_manager/orchestration/dependencies/imap_lo_dependencies.yaml
- sds_data_manager/orchestration/dependencies/imap_spacecraft_dependencies.yaml
- sds_data_manager/orchestration/dependencies/imap_swe_dependencies.yaml
- sds_data_manager/orchestration/dependencies/imap_ultra_dependencies.yaml

### DIFF
--- a/sds_data_manager/orchestration/dependencies/imap_glows_dependencies.yaml
+++ b/sds_data_manager/orchestration/dependencies/imap_glows_dependencies.yaml
@@ -32,7 +32,6 @@ hist_spice: &hist_spice
   - source: pointing_attitude
     data_type: spice
     descriptor: historical
-    trigger_job: false
 
 # ======== Product-specific dependencies ===========
 (l1a, all):
--- a/sds_data_manager/orchestration/dependencies/imap_hi_dependencies.yaml
+++ b/sds_data_manager/orchestration/dependencies/imap_hi_dependencies.yaml
@@ -29,7 +29,6 @@ spice_common: &spice_common
   - source: attitude_history
     data_type: spice
     descriptor: historical
-    trigger_job: false
   - source: ephemeris_reconstructed
     data_type: spice
     descriptor: historical
@@ -50,7 +49,6 @@ spice_l1b: &spice_l1b
   - source: spin
     data_type: spin
     descriptor: historical
-    trigger_job: false
 
 spice_l1c: &spice_l1c
   - *spice_common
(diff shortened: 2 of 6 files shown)


tech3371 (approved):
@daralynnrhode and @tmplummer Are we triggering now on these kernel arrival because of recent update you made with reprocessing only for updated time coverage? If so, remove `trigger_job` looks good to me. Wanted to double check.

tmplummer:
@tech3371 
> @daralynnrhode and @tmplummer Are we triggering now on these kernel arrival because of recent update you made with reprocessing only for updated time coverage? If so, remove `trigger_job` looks good to me. Wanted to double check.

I didn't write the issue, but yes, that is the idea from the issue. I think this is good to merge. 🤞 that this will just work now.

tmplummer (approved):

PR #1541, opened 2026-07-29
All 6mo imap-lo descriptors

### LINKED ISSUE #1535: Imap-Lo - 6mo quickmaps in dagster

vineetbansal:
Closes #1535 (for sure this time).

Since the process of map entries showing up in dagster is not 100% automatic, I thought I might as well add all entries that we're going to need in the next few weeks for the Lo team to validate. This is the official list from the Lo team spreadsheet.

Some of these jobs will fail for sure since the code is not hooked up to work with all descriptors, but this will allow us to gradually push code to make all of them work.

I've collapsed the 1yr maps into a single section as well, with the same input dependencies as the 6mo maps. The 1yr entries may need revisiting once the 6mo maps are done.

### DIFF
--- a/sds_data_manager/orchestration/dependencies/imap_lo_dependencies.yaml
+++ b/sds_data_manager/orchestration/dependencies/imap_lo_dependencies.yaml
@@ -51,6 +51,25 @@ lo_esa_fit_factors: &lo_esa_fit_factors
     data_type: ancillary
     descriptor: esa-eta-fit-factors
 
+lo_l2_map_inputs: &lo_l2_map_inputs
+  - *spice_common
+  - *lo_esa_fit_factors
+  - source: spin
+    data_type: spin
+    descriptor: historical
+  - source: repoint
+    data_type: repoint
+    descriptor: historical
+  - source: lo
+    data_type: l1b
+    descriptor: histrates
+  - source: lo
+    data_type: l1b
+    descriptor: goodtimes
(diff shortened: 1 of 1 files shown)


sds_data_manager/orchestration/dependencies/imap_lo_dependencies.yaml
```diff
@@ -332,10 +302,11 @@ lo_esa_fit_factors: &lo_esa_fit_factors
 
 # ======== Level L2 ========
 
-(l2, l090-enansnbs-h-sf-nsp-ram-hae-6deg-6mo):
+(l2, all-maps-6mo):
```
tmplummer: Was this modeled after L3 maps? I don't know exactly how L3 works, but I believe they are just manually running through a list of hard coded descriptors in a single "all-maps" job. This loses the ability to reprocess just select maps. For example, if the CG correction gets updated, you have to reprocess all spacecraft frame maps even though those wont change. It seems like putting it into one batch job could be problematic.

vineetbansal: ah - yes I looked at the l3 section and thought it was a clever way for our infrastructure to reduce lines in these yamls. Yes - we'll definitely need to reprocess selected maps. Expanding it and updating..

tmplummer (approved):
Looks reasonable.

PR #1544, opened 2026-07-31
A few improvements to dagster

bryan-harter:
# Change Summary

## Overview
These are a few small unrelated changes, but related to the discussions we had about dagster the other day:

1) The default sensor time is now 5 minutes instead of 15. It was largely 15 just to stop the system from being overwhelmed by the initial processing efforts when dagster turned on for the first time, but 5 minutes should be plenty now 

2) Removed the code the code that prevented failed jobs from running again with no changes. This again was something to help prevent a boatload of processing runs from occurring during the early deployment steps, but now this is causing a hassle and should just be better to re-run. 

3) Increased the allowed batch job timeout to 5 hours. Leaving it at 1 hour caused an issues where the job completed successfully, but Dagster didn't know about it. 

4) Perhaps the nicest one - instead of skipping an asset when we don't have the dependencies, instead we return an "AssetObservation". This keeps the little icon "grey", but it actually returns metadata about the asset at that partition. Now you can click on a grey partition for an asset, and the metadata tells you exactly what dependency it failed to retrieve.  

## File changes
sds_data_manager/orchestration/config.py
- Setting a 5 minute sensor timer instead of 15

sds_data_manager/orchestration/imap_job.py
- 5 hour timeout 
- Added AssetObservation
- Removed "Already Failed" batch job checking 

## Testing
tests/orchestration/test_spacecraft.py
- Fixed a unit test to verify that an AssetObservation is returned instead of nothing if we don't meet the dependencies.

### DIFF
--- a/sds_data_manager/orchestration/config.py
+++ b/sds_data_manager/orchestration/config.py
@@ -11,22 +11,22 @@ VALID_CADENCE_STRS = ["1mo", "3mo", "6mo", "1yr"]
 FIRST_MAP_START_DATE = datetime.datetime(2026, 1, 17, tzinfo=datetime.timezone.utc)
 
 sensor_schedules = {
-    "l0": 900,
-    "l1": 900,
-    "l1a": 900,
-    "l1b": 900,
-    "l1c": 900,
-    "l1d": 900,
-    "l2": 900,
-    "l2a": 900,
-    "l2b": 900,
-    "l2c": 900,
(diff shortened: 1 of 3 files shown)


tmplummer (approved):
Nice improvements! Interested in responses to the Copilot comments but otherwise, looks good to me.

tech3371 (approved):
Thank you for quick updates!

PR #1545, opened 2026-07-31
Adding a read only web server for Dagster

### LINKED ISSUE #1416: Create a public-facing read-only version of Dagster
### Summary

Right now, the Dagster instance is full access, though only accessible via the VPN. We can create a read-only version that other members of the team can see. 

As a part of this ticket, it would be great if we could discuss options for authentication with the web team. We *could* make it completely public, but ideally it would be hidden behind a login similar to the Batch job status page on the imap-mission.com. 

As an alternative, we could grant access only to specific IP addresses (such as different institute's VPNs). 

### Requirements

The more specific the better.
  - [ Create a new read-only dagster webserver ECS task ]
  - [ Investigate methods to lock down access to it ]
  - [ Implement any necessary security features ]

### Related tickets

_No response_

bryan-harter:
# Change Summary

## Overview
Creates a read only version of Dagster that Princeton can see . Closes https://github.com/IMAP-Science-Operations-Center/sds-data-manager/issues/1416

## File changes
- sds_data_manager/constructs/dagster_construct.py
Basically a copy+paste of the regular web server. However, there is the --read-only flag now, and we allow more IP addresses.

### DIFF
--- a/sds_data_manager/constructs/dagster_construct.py
+++ b/sds_data_manager/constructs/dagster_construct.py
@@ -302,6 +302,7 @@ class DagsterEcsConstruct(Construct):
             run_task_def.task_definition_arn
         )
 
+        # Dagster Webserver
         webserver_service = ecs_patterns.ApplicationLoadBalancedFargateService(
             self,
             "DagsterWebserver",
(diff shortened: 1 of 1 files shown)


tech3371 (approved):
Nice! I am sure Princeton folks will appreciate this. This makes it easy to add IP address in the future.

PR #1546, opened 2026-07-31
Better Batch Job monitoring in Dagster

bryan-harter:
# Change Summary

## Overview
This code started as a way for to print out some logs from the Batch jobs, to enable better debugging from within the Dagster interface. However, I realized we could just move the entire Batch job monitoring out of the file indexer and into Dagster. Since a dagster container is already sitting there waiting for changes in job status, it can also just be the one that inserts status into the DB. 

## File changes
sds_data_manager/constructs/indexer_lambda_construct.py
- Removing AWS Batch job status triggers from the indexer

sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/indexer.py
- Removing Batch job status handling code

sds_data_manager/orchestration/imap_job.py
- Moved Batch job status handling into Dagster. 
- During the Asset creation, we'll now print out the last 100 lines from the cloudwatch logs. This will cause those logs to show up in Dagster. 

## Testing
Several of the unit tests had to be modified to accompany these changes. But they do confirm that the batch job submission and subsequent modification work as intended.

### DIFF
--- a/sds_data_manager/orchestration/imap_job.py
+++ b/sds_data_manager/orchestration/imap_job.py
@@ -65,6 +65,9 @@ BATCH_JOB_RETRY_STRATEGY = {
 }
 # Create an sqs client
 SQS_CLIENT = boto3.client("sqs", region_name="us-west-2")
+# Create a client to access Cloudwatch
+LOGS_CLIENT = boto3.client("logs", region_name="us-west-2")
+LOG_GROUP_NAME = "/aws/batch/job"
 
 # Logger setup
 logger = logging.getLogger(__name__)
@@ -113,6 +116,7 @@ class BatchJobSubmit:
     message: str
     # the ProcessingJob information as a dictionary.
     job: dict | None = None
+    response: dict | None = None
 
 
 class IMAPJobHandler:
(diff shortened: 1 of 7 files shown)


sds_data_manager/orchestration/imap_job.py
```diff
@@ -304,39 +303,99 @@ def run_job(
...
+            elif job_info["status"] == "FAILED":
+                break
+            else:
+                time.sleep(wait_time)
+                continue
+
+        # Print out some logs
+        container_info = job_info.get("container", {})
+        log_stream_name = container_info.get("logStreamName")
+        if log_stream_name:
```
bryan-harter, line 336: This was the original intent behind the the PR. I wanted to get the log stream from the submitted processing job and then print out the most recent 100 lines of Batch job into Dagster. Then it would be easier to debug things without going into Cloudwatch.

But I realized I'd be doing some duplicate work with the file-indexer, so I ultimately just moved file-indexer's ProcessingJob table duties into here to keep things (hopefully) simpler.

lacoak21 (approved):
this looks good to me! Good idea. The only thing I would check is that the timeout in wait_for_batch_job is greater than the largest batch timeout we have

PR #1549, opened 2026-08-04
Adding Dagster Deploy to GitHub Actions

### LINKED ISSUE #1418: Add a step to deploy the latest Dagster Code
### Summary

When we update Dagster logic and push a new version of the CDK, a new Docker image will be created in the "dagster-image" repo. However, the ECS will continue to happily run the previous version of the Dagster daemon and webserver, until there is some sort of error with Fargate. 

To get around this for now, to immediately deploy the items we have been using the commands `aws ecs update-service --cluster {cluster-name} --service {service-tash} --force-new-deployment --region us-west-2`. 

The simplest thing to do for now would be to add a github action that performs this task for us after a CDK deploy. One issue with this is that running sensors will immediately stop and throw an error. In theory this should not be an issue, since the sensor will pick up again next time. However, I just have not thoroughly thought through the consequences of having the daemon interrupted by force. I'm 99% sure that this is not an issue, but I wanted to remind whoever picks up this ticket of this concern. 

As a part of this ticket, it would be nice to ensure that there is no way in AWS to specify that ECS should redeploy itself on a new image (without doing something crazy like having a lambda watching the ECR and having it force the deployment). 

### Requirements

The more specific the better.
  - [ The GitHub Actions CDK deploy will deploy the newest version of the dagster daemon and dagster webserver]

### Related tickets

_No response_

daralynnrhode:
# Adding Dagster Deploy to GitHub Actions

## Overview

closes #1418 

Added a dagster deploy step to the GitHub Actions. I opted to keep it as a separate step instead of combing with the CDK deploy step, but can combine the two if it's needed/wanted. 
I did do some research on GitHub Actions and workflows and this is what I was able to come up with. 

## File changes

- .github/workflows/deploy.yml

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -125,7 +125,7 @@ jobs:
       # Force ECS redeploy of Dagster daemon + webserver (prod only)
       # ----------------------------------------------------
       - name: Force new ECS deployment for Dagster
-        if: env.should_tag == 'true'
+        if: env.is_prod == 'true'
         run: |
           CLUSTER=$(aws ecs list-clusters --region us-west-2 \
             --query "clusterArns[?contains(@, 'DagsterCluster')]" --output text)
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -129,10 +129,10 @@ jobs:
         run: |
           CLUSTER=$(aws ecs list-clusters --region us-west-2 \
             --query "clusterArns[?contains(@, 'DagsterCluster')]" --output text)
- 
+
           SERVICES=$(aws ecs list-services --cluster "$CLUSTER" --region us-west-2 \
             --query "serviceArns[?contains(@, 'Dagster')]" --output text)
- 
+
           for svc in $SERVICES; do
             echo "Forcing redeploy of $svc"
             aws ecs update-service --cluster "$CLUSTER" --service "$svc" --force-new-deployment --region us-west-2
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -122,6 +122,23 @@ jobs:
           poetry run cdk deploy $STACKS --context account_name=$ACCOUNT_NAME --require-approval never
 
       # ----------------------------------------------------
+      # Force ECS redeploy of Dagster daemon + webserver (prod only)
+      # ----------------------------------------------------
+      - name: Force new ECS deployment for Dagster
+        if: env.should_tag == 'true'
+        run: |
+          CLUSTER=$(aws ecs list-clusters --region us-west-2 \
+            --query "clusterArns[?contains(@, 'DagsterCluster')]" --output text)
+ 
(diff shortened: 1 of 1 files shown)


bryan-harter:
Looks great to me!

bryan-harter (approved):

lacoak21 (approved):
LGTM! Nice job! Was this tested with dev?

daralynnrhode:
@lacoak21 no, because I am unsure how to go about doing that

.github/workflows/deploy.yml
```diff
@@ -121,6 +121,23 @@ jobs:
           # poetry run to get the environment we installed everything into
           poetry run cdk deploy $STACKS --context account_name=$ACCOUNT_NAME --require-approval never
 
+      # ----------------------------------------------------
+      # Force ECS redeploy of Dagster daemon + webserver (prod only)
+      # ----------------------------------------------------
+      - name: Force new ECS deployment for Dagster
+        if: env.should_tag == 'true'
```
lacoak21: Maybe we should change this boolean name e.g.
```suggestion
        if: env.is_prod == 'true'
```

lacoak21:
> @lacoak21 no, because I am unsure how to go about doing that

Ok yeah this is tricky. Do you have AWS credentials yet? 

I dont think there is an easy way to test the full thing out but as an example I just ran these commands for the dev account and it looks good: 

CLUSTER=$(aws ecs list-clusters --region us-west-2             --query "clusterArns[?contains(@, 'DagsterCluster')]" --output text --profile imap-dev)

services=$(aws ecs list-services --cluster "$CLUSTER" --region us-west-2 --query "serviceArns[?contains(@, 'Dagster')]" --output text --profile imap-dev)

for svc in "${services[@]}"; do echo "Forcing redeploy of $svc"; aws ecs update-service --cluster "$CLUSTER" --service "$svc" --force-new-deployment --region us-west-2 --profile imap-dev; done

daralynnrhode:
> Ok yeah this is tricky. Do you have AWS credentials yet?
> 
> I dont think there is an easy way to test the full thing out but as an example I just ran these commands for the dev account and it looks good:

I do not have ASW credentials yet, last I heard was when I first mentioned it to you. Am I good to merge in then?

lacoak21:
> I do not have ASW credentials yet, last I heard was when I first mentioned it to you. Am I good to merge in then?

yes good to merge after you update the boolean var name above! Unless you have a reason to keep it as is.

tech3371:
you will not need credentials to test GitHub action after you merge into dev because GitHub action was suppose to be independent of individual credentials. GitHub action shouldn't take in personal credentials. It's enabled through some IAM policy and role instead.

Can we add that steps to dev action too? Eg. Add it after CDK deploy step on merge to `dev` branch. And if it works in dev, we can expect it to work in prod. That would be a good verification step and steps we would need in general workflow as well.

tech3371:
Eg. we can can enable 'Force new ECS deployment for Dagster' step in this image.
Screenshot 2026-08-05 at 10 26 16 AM

PR #1550, opened 2026-08-05
Update batch job retry strategy

### LINKED ISSUE #1472: BUG - Unable to reprocess jobs that have a 503 or empty processing log error.
### Description of the issue

After triggering reprocessing for L3 in situ products, we saw a handful of failing jobs: some due to a 503 error and others with "(Empty processing log)". We are unable to trigger reprocessing via the command line because Dagster recognizes these as previously failed jobs.

We need dagster to automatically retry these jobs OR have someway of making sure they dont get counted as dupicates

_No response_

_No response_

```python
Code
```

_No response_

lacoak21:
# Change Summary

closes #1472

## Overview

We no longer need the "spot task was interrupted" retry because we are using on demand instances now. I also added a retry when the exit code is 75. This is the exit code that specifies this job should be retried. For example, when there was an imap-data-access 503 error we should retry the job. We should also retry the job when the job could not pull the container. E.g the status reason was "CannotPullContainerError".

## File changes
   - Update retry strategy

### DIFF
--- a/sds_data_manager/orchestration/imap_job.py
+++ b/sds_data_manager/orchestration/imap_job.py
@@ -57,9 +57,10 @@ BATCH_JOB_RETRY_STRATEGY = {
     "attempts": 10,
     "evaluateOnExit": [
         {
-            "onStatusReason": "Your Spot Task was interrupted.",
+            "onExitCode": "75",
             "action": "RETRY",
-        },
+        },  # retry jobs that failed with the rerunable exit code.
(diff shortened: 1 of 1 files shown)


bryan-harter (approved):
Seems good to me, as long as the AI comment turns out to be incorrect/not relevant

lacoak21:
> Seems good to me, as long as the AI comment turns out to be incorrect/not relevant

I think it is relevant but I did change it to onStatusReason

PR #1551, opened 2026-08-05
Deploying Dagster for both Prod and Dev

daralynnrhode:
# Changing From Prod only to both Prod and Dev Deployment

## Overview

Removing the conditional prod only statement so dagster also deploys for dev 

Addresses feedback on #1549 pertaining to #1418 ticket.

## File changes

.github/workflows/deploy.yml


bryan-harter (approved):
Looks good! Somehow I missed in the first review that it was only for prod.

tech3371 (approved):
Thank you for this update! I am interested in how this goes because I tried duplicating this effort in my forked repo and was much more effort than I was prepared for at the time :). Thank you for quick help on this!!!

PR #1552, opened 2026-08-05
Support for type=spin/repoint/thruster/kernels for the spice-query api

leowerneck:
This is a slight rewrite of #1250 by @vineetbansal, which splits some of the new tests for better granularity and introduces a helper function to ensure consistent formatting of the response output.

---

# Change Summary

Fixes issue #911 

Added support to query non-SPICE tables (spin/repoint/thruster) using the `/spice-query` endpoint.

The `type` parameter decides which table to query - `kernels` searches through the spice table. Other possible values are `spin`, `thruster`, or `repoint`, which delegate to the existing lambda in `non_spice_table_api.py`.

When the `type` parameter is not set, it is assumed to be `kernels`, preserving legacy behavior. A `type` parameter set to a specific kernel type like `spacecraft_clock` still works as before.

Parameters like `file_name`, `start_time` and `end_time` that the SPICE api already supports are now mapped to the non-SPICE API, so the caller has to go through minimal changes to start using these.

## File changes

- Main functionality added to `spice_query_api.py` that internally calls `non_spice_tables_api`.
- I had to move an import in `spice_metakernel_api.py` to where it is used to get around a circular import issue (`spice_utilities.py` imports `spice_metakernel_api` and `spice_metakernel_api.py` imports `spice_utilities`).

## Testing

- Added a test to `test_spice_query_api.py` to ensure that `type` parameter is interpreted correctly for SPICE tables.
- Added tests to `test_spin_api.py` to ensure that calls made through `spice_query.api` return expected responses. Existing tests left intact, so the old API endpoints work as well.

### DIFF
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/non_spice_table_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/non_spice_table_api.py
@@ -122,7 +122,7 @@ def lambda_handler(event, context):  # noqa: PLR0912
                 logger.debug(err_msg)
                 return response
 
-        # Reset statusCode to 200 if we got this far
+        # Reset status_code to 200 if we got this far
         status_code = 200
         search_results = session.execute(query).scalars().all()
         # format the search results into a list of dictionaries
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/non_spice_table_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/non_spice_table_api.py
@@ -13,6 +13,16 @@ logger = logging.getLogger(__name__)
 logger.setLevel(logging.INFO)
 
 
+def get_json_response(
+    statusCode, body_str, headers={"Content-Type": "application/json"}
+):
+    return {
+        "statusCode": statusCode,
+        "headers": headers,
+        "body": json.dumps(body_str),
+    }
+
+
 def lambda_handler(event, context):  # noqa: PLR0912
     """Handle API requests for the non-SPICE data.
 
@@ -22,6 +32,8 @@ def lambda_handler(event, context):  # noqa: PLR0912
         "Spin/Repoint/small-forces Query Event: " + json.dumps(event, indent=2)
     )
 
+    # Initialize statusCode to 400
+    statusCode = 400
     raw_path = event.get("rawPath", "")
     if "spin" in raw_path:
         table = SpinFiles
@@ -30,13 +42,9 @@ def lambda_handler(event, context):  # noqa: PLR0912
     elif "small-forces" in raw_path:
         table = SmallForcesFile
     else:
-        response = {
-            "statusCode": 400,
-            "body": json.dumps(
-                "Invalid path. Path must contain either 'spin' or 'repoint'."
-            ),
-        }
-        logger.debug("Invalid path, must contain either 'spin' or 'repoint'.")
+        errorMessage = "Invalid path, path must contain either 'spin' or 'repoint'."
+        response = get_json_response(statusCode, errorMessage)
+        logger.debug(errorMessage)
         return response
 
     # add session, pick model like in indexer and add query to filter_as
@@ -59,17 +67,9 @@ def lambda_handler(event, context):  # noqa: PLR0912
         for param, value in query_params.items():
             # confirm that the query parameter is valid
             if param not in valid_parameters:
-                response = {
-                    "statusCode": 400,
-                    "body": json.dumps(
-                        f"{param} is not a valid query parameter. "
-                        + f"Valid query parameters are: {valid_parameters}"
-                    ),
-                }
-                logger.debug(
-                    f"Received an invalid query parameter [{param}],"
-                    " valid options are: {valid_parameters}"
-                )
+                errorMessage = f"{param} is not a valid query parameter. Valid query parameters are: {valid_parameters}"
+                response = get_json_response(statusCode, errorMessage)
+                logger.debug(errorMessage)
                 return response
             try:
                 if param == "start_date" and table != RepointFiles:
@@ -113,13 +113,13 @@ def lambda_handler(event, context):  # noqa: PLR0912
                     parsed_date = datetime.datetime.strptime(value, "%Y%m%d")
                     query = query.where(table.ingestion_date <= parsed_date)
             except ValueError:
-                response = {
-                    "statusCode": 400,
-                    "body": json.dumps(f"Invalid value for {param}: {value}"),
-                }
-                logger.debug(f"Invalid value for {param}: {value}")
+                errorMessage = f"Invalid value for {param}: {value}"
+                response = get_json_response(statusCode, errorMessage)
+                logger.debug(errorMessage)
                 return response
 
+        # Reset statusCode to 200 if we got this far
+        statusCode = 200
         search_results = session.execute(query).scalars().all()
         # format the search results into a list of dictionaries
         if table == RepointFiles:
@@ -135,7 +135,7 @@ def lambda_handler(event, context):  # noqa: PLR0912
                 }
                 for result in search_results
             ]
-            return {"statusCode": 200, "body": json.dumps(search_results)}
+            return get_json_response(statusCode, search_results)
 
         # Spin or small-forces files have a start_date field
         search_results = [
@@ -148,4 +148,4 @@ def lambda_handler(event, context):  # noqa: PLR0912
             }
             for result in search_results
         ]
-        return {"statusCode": 200, "body": json.dumps(search_results)}
+        return get_json_response(statusCode, search_results)
--- a/sds_data_manager/lambda_code/SDSCode/api_lambdas/non_spice_table_api.py
+++ b/sds_data_manager/lambda_code/SDSCode/api_lambdas/non_spice_table_api.py
@@ -24,7 +24,52 @@ def get_json_response(status_code, body_str, headers=None):
     }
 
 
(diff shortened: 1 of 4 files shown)


sds_data_manager/lambda_code/SDSCode/api_lambdas/non_spice_table_api.py
tech3371, line 1: looks like this code I have wrote a while back is hard to read and follow. Sorry for this and appreciate your understanding.

I am wondering if we should start fresh with logic something like this and reuse other code we have written since then or in parallel.

```
If table == 'spin':
    copy code from this https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/dev/sds_data_manager/orchestration/spin.py::get_spin_files()
elif table == 'repoint':
     copy code from this https://github.com/IMAP-Science-Operations-Center/sds-data-manager/blob/dev/sds_data_manager/orchestration/repoint_file.py::get_latest_repoint_file()
elif table == 'small-forces':
     create new function that queries latest small forces file.
else:
     return error json response of unrecognize table input.

Comment (tech3371) [reply sds_data_manager/lambda_code/SDSCode/api_lambdas/non_spice_table_api.py:1]: As you work on this, I appreciate if you can think about first task in this ticket, https://github.com/IMAP-Science-Operations-Center/imap-data-access/issues/346. This backend work you are doing here will support interface for that work.

Eg. that ticket will need to add CLI handler for command like this. These are examples. These are not implemented yet.
```
imap-data-access query --table spice --type repoint --version latest
imap-data-access query --table spice --type spin --start-date <> --end-date <> --version latest
imap-data-access query --table spice --type thruster --start-date <> --end-date <> --version latest
```

File: sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_query_api.py [hunk 1]
```diff
@@ -73,7 +123,7 @@ def lambda_handler(event, context):
                     query = query.where(
                         models.SPICEFiles.min_date_j2000 <= float(value)
                     )
-                elif param == "type":
+                elif param == "type" and value != "kernels":
```
tech3371, line 126: popping 'type' and generalizing to 'kernel' is making me concern because type may be used to filter kernels elsewhere and many user interfaces such as website support query by specific kernel type and return result of that kernel type. I have a feeling that they use this filter logic to return requested kernel. Eg. `type=attitude_history` will filter and return just attitude history kernel.

In general, I would be careful changing logic in here file since it's heavily used by all pipeline and user interface. This is where I appreciate double check form @bryan-harter .

tech3371: ok..I looked again more carefully and looks like you get queryStringParameters again at line 90 and there you are not popping `type`. good.

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_query_api.py
```diff
@@ -28,9 +39,53 @@ def lambda_handler(event, context):
...
+
+    Passing a kernel-type value such as ``attitude_history`` filters the SPICE table
+    by ``kernel_type``. The remaining SPICE-specific parameters
+    (``file_name``, ``start_time``, ``end_time``, ``latest``,
+    ``start_ingest_date``, ``end_ingest_date``) apply.
+
+    When ``type`` is one of the non-SPICE values, the ``file_name`` parameter is
+    renamed to ``file_path``, ``start_time`` is mapped to ``start_date``, ``end_time``
+    is mapped to ``end_date``, and the remaining parameters are passed to the
+    non-SPICE API.
```
tech3371, line 60: I would consolidate to something like this:

```suggestion
    * ``<other kernel type>`` - Use metakernel API to query and return requested kernels.
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_query_api.py
```diff
@@ -28,9 +39,53 @@ def lambda_handler(event, context):
...
     """
     logger.debug("SPICE Query Event: " + json.dumps(event, indent=2))
 
+    # Initialize status_code to 400
+    status_code = 400
+
+    query_params = event.get("queryStringParameters", {})
+    table_type = query_params.get("type", "kernels")
```
tech3371, line 68: Consolidating to reduce confusion with query_params at line 90.
```suggestion
    table_type = event.get("queryStringParameters", {}).get("type", "")
```

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_query_api.py
```diff
@@ -28,9 +39,53 @@ def lambda_handler(event, context):
...
+
+    if table_type in _NON_SPICE_RAW_PATHS:
+        # Remove `type`, since it is not a valid non-SPICE query parameter.
+        query_params.pop("type")
```
tech3371, line 72: I would avoid this line or move it into `_remap_to_non_spice_params()` to reduce confusion.
```suggestion

```

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_query_api.py
```diff
@@ -126,14 +177,33 @@ def lambda_handler(event, context):
...
-            "statusCode": 200,
-            "headers": {"Content-Type": "application/json"},
-            "body": json.dumps(search_results),  # returns a list of tuples
-        }
-
-        return response
+        return non_spice_table_api.get_json_response(status_code, search_results)
+
+
+def _remap_to_non_spice_params(query_params: dict) -> dict:
```
tech3371, line 183: I don't think we need this function. We can expect user to pass human readable time, `yyyymmdd`. We normally convert str to datetime to be able to query DB. @tmplummer do we need to support `hhmmss` precision for these kinds of query?

Also, I don't think lambda layer has `spiceypy` library unless it has changed recently that I don't know of.

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_query_api.py
```diff
@@ -28,9 +39,53 @@ def lambda_handler(event, context):
...
+    if table_type in _NON_SPICE_RAW_PATHS:
+        # Remove `type`, since it is not a valid non-SPICE query parameter.
+        query_params.pop("type")
+
+        # Remap incoming parameters to ones supported by the non-SPICE tables API.
+        try:
+            query_params = _remap_to_non_spice_params(query_params)
+        except ValueError:
+            err_msg = "Expected start/end times in ET."
+            return non_spice_table_api.get_json_response(status_code, err_msg)
```
tech3371, line 79: See my comment on the `_remap_to_non_spice_params()`

sds_data_manager/lambda_code/SDSCode/api_lambdas/spice_query_api.py
tmplummer: This is just for `imap-data-access query --table spice` right? I don't think that we need to support higher precision than day if that is the case. It is needed in the spice and metakernel APIs. I also don't see any need to support queries where the times are ET times instead of `yyyymmdd` strings.

tmplummer:
Sorry for commenting before I had the full context of this PR and the issues driving it. It seems to me there are some inconsistencies in our current approach to handling spin, repoint, and sff files. They are being handled separately from spice files and ingested into their own tables when I think that they can in fact be handled almost exactly like the other spice kernels. When ingesting those file types, we should be able to extract all the same details that are extracted when any other spice kernel is ingested, namely: 
- file_name
- file_root
- kernel_type
- version
- min_date_j2000
- max_date_j2000
- file_intervals_j2000
- min_date_datetime
- max_date_datetime
- file_intervals_datetime
- min_date_sclk
- max_date_sclk
- file_intervals_sclk
- sclk_kernel
- lsk_kernel
- ingestion_date
- timestamp

If we migrated the DB to include these files in the spice table, then the spice-api as is would support querying these file types and the metakernel api would also correctly prioritize the correct file for a given time. Then, there would just remain the issue of supporting querying the spice table via the query API.

@tech3371, @bryan-harter what do you think about that approach?

tech3371:
> Sorry for commenting before I had the full context of this PR and the issues driving it. It seems to me there are some inconsistencies in our current approach to handling spin, repoint, and sff files. They are being handled separately from spice files and ingested into their own tables when I think that they can in fact be handled almost exactly like the other spice kernels. When ingesting those file types, we should be able to extract all the same details that are extracted when any other spice kernel is ingested, namely:
> 
> * file_name
> * file_root
> * kernel_type
> * version
> * min_date_j2000
> * max_date_j2000
> * file_intervals_j2000
> * min_date_datetime
> * max_date_datetime
> * file_intervals_datetime
> * min_date_sclk
> * max_date_sclk
> * file_intervals_sclk
> * sclk_kernel
> * lsk_kernel
> * ingestion_date
> * timestamp
> 
> If we migrated the DB to include these files in the spice table, then the spice-api as is would support querying these file types and the metakernel api would also correctly prioritize the correct file for a given time. Then, there would just remain the issue of supporting querying the spice table via the query API.
> 
> @tech3371, @bryan-harter what do you think about that approach?

I think the reason we had to set up different tables were because those files don't have as much details as SPICE kernels. Most of those other API's files are CSV or ASCII file. I think that was the reason for setting up different table. Also, I was lack SPICE knowledge and didn't want to mess up SPICE API in case those files are included in metakernel file query and user face kernel loading issues because they were expecting only Kernel types in their metakernel files. @bryan-harter Do you see Tim's suggestion as possibility? If so, that would simplify things for sure!

PR #1555, opened 2026-08-06
Read IDEX partition CSV from imap_processing

### LINKED ISSUE #1327: Remove the idex 10day file csv from sds-data-manager
import it directly from imap-processing so they dont diverge.

lacoak21:
# Change Summary

closes #1327

## Overview
Use the static idex partition info from imap_processing. Unfortunately I still need copies of "idex_10_day_CDF_names.csv" in imap_processing AND sds-data-manager because imap_processing is not a requirement for dagster right now and I would prefer to keep it that way. Dagster needs those files in custom_paritions.py. 

## File changes
- sds_data_manager/lambda_code/SDSCode/pipeline_lambdas/idex_l0_indexer.py
   - read static file from imap_processing


bryan-harter (approved):
Looks good!

